PR #255, opened 2017-07-17
Update dependencies without breaking changes

unitymind:
* Trying to update `mio` introduce breaking changes due using `tokio` project - still on `0.5` version
* `params` crate still rely on `body-parser 0.5.0` and `serde-*-0.8.x` family due more than half year not released `iron`

In general, it's good practice keep dependencies up to date :)


unitymind:
Exciting about blockchain on Rust with well documentation. Great technical direction! :)

defuz:
Thanks! 👍

Issue #190, opened 2017-06-26
Call update_config only when it's really needed


defuz:

DarkEld3r:
Closed by #256.

Issue #137, opened 2017-05-30
Different timeout adjuster implementations should be configured through ConsensusConfig


DarkEld3r:
Technically this is a breaking change because `ConsensusConfig` should be modified.

DarkEld3r:
Closed by #256.

PR #256, opened 2017-07-17
Timeout adjusters refactoring

DarkEld3r:
Several tasks are handled in this pull request:
1. Additional `Dynamic` timeout adjuster is added.
2. Timeout adjusters should be configured dynamically (#137).
3. Config should be updated only when it is needed (#190).

(#250)

### DIFF
--- a/exonum/src/storage/proof_map_index/mod.rs
+++ b/exonum/src/storage/proof_map_index/mod.rs
@@ -106,7 +106,7 @@ impl<T, K, V> ProofMapIndex<T, K, V> {
     /// immutable methods are available. In the second case both immutable and mutable methods are
     /// available.
     /// [`&Snapshot`]: ../trait.Snapshot.html
-    /// [`&mut Fork`]: ../struct.Fork.html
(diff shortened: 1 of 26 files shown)


exonum/src/storage/proof_map_index/mod.rs
```diff
@@ -106,7 +106,7 @@ impl<T, K, V> ProofMapIndex<T, K, V> {
     /// immutable methods are available. In the second case both immutable and mutable methods are
     /// available.
     /// [`&Snapshot`]: ../trait.Snapshot.html
-    /// [`&mut Fork`]: ../struct.Fork.html
+    /// [`&mut Fork`]: ../struct.Fork.h tml
```
vldm: remove space

vldm:
@DarkEld3r can you give me a line
>Config should be updated only when it is needed (#190).
where this handled?

DarkEld3r:
@vldm 
Actually `update_config` still called, but inside I'm checking configurations for equality: https://github.com/exonum/exonum/pull/256/files#diff-ca8cd50c9741eb27c277c423d41066b4R486

Issue #232, opened 2017-07-12
Panics during command line arguments parsing


DarkEld3r:
Currently our code panics if there are invalid command line arguments (`exonum/helpers/fabric/mod.rs`). It is not very user-friendly because error looks like "thread 'main' panicked at...".

PR #257, opened 2017-07-17
Make panic message cleaner.

vldm:
Closes https://github.com/exonum/exonum/issues/232

I am using `set_hook` here, instead of `catch_unwind` because I need to prevent printing default panic message in `std::*::default_hook`.

The message will look like
```
➜  exonum-configuration git:(cargo-toml) ✗ ./target/debug/examples/configuration 
error: Subcommand not found

```
instead of 
```
➜  exonum-configuration git:(cargo-toml) ✗ ./target/debug/examples/configuration 
thread 'main' panicked at 'Subcommand not found', /Users/vladimirmotylenko/bitfury/exonum/exonum/src/helpers/fabric/clap_backend.rs:84
```

Cons: 
 1. User can't print backtrace for cmd parsing now. (but he shouldn't want it)
 2. We can add color, or other marks, to increase usability
 3. It only catch and prints `String` represented exceptions.

### DIFF
--- a/exonum/src/helpers/fabric/builder.rs
+++ b/exonum/src/helpers/fabric/builder.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::fmt;
+use std::panic::{self, PanicInfo};
 use std::ffi::OsString;
 
 use blockchain::{Service, Blockchain};
@@ -87,11 +88,31 @@ impl NodeBuilder {
         }
     }
 
+    // handle error, and print it.
+    fn panic_hook(info: &PanicInfo) {
+        let msg = match info.payload().downcast_ref::<&'static str>() {
+            Some(s) => *s,
+            None => {
+                match info.payload().downcast_ref::<String>() {
+                    Some(s) => &s[..],
+                    None => "Box<Any>",
+                }
+            }
+        };
+        println!("error: {}", msg);
+    }
+
     /// Runs application.
     pub fn run(self) {
-        if let Some(mut node) = self.parse_cmd() {
+        let old_hook = panic::take_hook();
+        panic::set_hook(Box::new(Self::panic_hook));
+        let feedback = self.parse_cmd();
+        panic::set_hook(old_hook);
+
+        if let Some(mut node) = feedback {
             node.run().expect("Node return error")
         }
+
     }
 }
 


vldm:
@DarkEld3r we can, but does we need this? 

1. This panics could happen only at program start.
2. I can't imagine case, where user really needs to handle this kinds of errors.

DarkEld3r:
> This panics could happen only at program start.

I don't see how it makes them special. 😄 
Also can you guarantee that there will be no other panics (unrelated to the command line arguments)? Especially after some time and when our code-base grows bigger.

> I can't imagine case, where user really needs to handle this kinds of errors.

Perhaps, you are right and user don't need to handle such errors, but panics are intended for unrecoverable failures and command line parsing is definitely expected to fail. I don't like to move towards "try/catch" approach.

vldm:
@DarkEld3r 
> I don't see how it makes them special.
There is no valid way to recover if initialization fails.
> Also can you guarantee that there will be no other panics (unrelated to the command line arguments)? Especially after some time and when our code-base grows bigger.

I recover old hook, just after cmd parsing. So there should be no other panics in `parse_cmd` unrelated to the command line arguments. :)


```
         let old_hook = panic::take_hook();
         panic::set_hook(Box::new(Self::panic_hook));
         let feedback = self.parse_cmd();
         panic::set_hook(old_hook);
```

> but panics are intended for unrecoverable failures
Panic is for errors, that:
a) Can't be handled in place.
b) There is commonly no difference what kind of errors cause panic.

And errors during application initialization is definitely this kind of errors. 

BTW: Only one thing in this hook differ from original one, is that it skip backtrace, and print more cleaner output.

DarkEld3r:
> So there should be no other panics in parse_cmd unrelated to the command line arguments. :)

Arbitrary code can be called inside `parse_cmd`. :smile: For some cases backtrace can be useful. 

>  a) Can't be handled in place.

Is it your opinion or common practice for the rust code? Could you please provide some link to the documentation/guide if it is the latter?

As far as I know, `panic` is for [unrecoverable errors](https://doc.rust-lang.org/book/second-edition/ch09-01-unrecoverable-errors-with-panic.html#unrecoverable-errors-with-panic) and for some "special cases". 

Additionally, (library) user can change panics handling from the unwinding to the abort. Generally libraries must not rely on any specific option. I think that we already have some problems with that, but I don't like one more such case.

vldm:
> As far as I know, panic is for unrecoverable errors and for some "special cases".

"Unrecoverable errors" in terms of rust Is an error that can't be handled in place (first paragraph of in your link). 

"Unrecoverable" doesn't mean that this error shouldn't be handled in whole application. Only end user should know how to deal with "panics". If he want to catch this panic, he will be able to do it with `catch_unwind`.

>Additionally, (library) user can change panics handling from the unwinding to the abort. Generally libraries must not rely on any specific option. I think that we already have some problems with that, but I don't like one more such case.

The hook will called even in abort.
```
The panic hook is invoked when a thread panics, but before the panic runtime is invoked. As such, the hook will run with both the aborting and unwinding runtimes.
```
https://doc.rust-lang.org/std/panic/fn.set_hook.html

DarkEld3r:
> The hook will called even in abort.

:+1: Nice to know.

> "Unrecoverable errors" in terms of rust Is an error that can't be handled in place

I cannot agree with that. :smile: Anyway, I am OK with proposed change, because it makes the error message better. I just don't like the idea of messing with panics, but if no one else has objections, then it is OK for me too.

Issue #203, opened 2017-06-30
Expose common typedefs from the core


DarkEld3r:
Currently some useful typedef are "private" for the core. For example, `Milliseconds` typedef is located in the `events` module. Also there are `Round`, `Height` and `ValidatorId` in the `/exonum/src/node/state.rs`.

Additional discussion: https://github.com/exonum/exonum-btc-anchoring/pull/58#discussion_r124996422.

#79 somehow relates to this.

DarkEld3r:
Closed by #262.

PR #262, opened 2017-07-21
Use tuple structs for typedefs

### LINKED ISSUE #79: Use tuple struct instead of simple `type`
Currently we have "typedefs" for some things like height, round, etc.:
```rust
pub type Round = u32;
pub type Height = u64;
pub type ValidatorId = u32;
```
Instead they can be made into tuple struct.

Advantages:
- Prevents possible (probably, unlikely) errors, for example: `fn foo(Round, ValidatorId)`.
- Force consistency: currently we have many places where row types are used instead of our typedefs.
- "Cool typesafety". :laughing: 

Disadvantages: 
- Additional "boilerplate": `round.0` instead of `round` if we need underlying value.
- ?

I can make such refactoring if we decide that we need it.

DarkEld3r:
This is mostly an experiment, so sandbox compilation is currently broken. Generally I like the idea, but it still has some rough edges. For example, `std::iter::Step` is still unstable, so round-loops look a little ugly. There is still room for additional improvements. 

Closes #79 and #203.


exonum/src/node/consensus.rs
```diff
@@ -31,9 +32,9 @@ where
     #[cfg_attr(feature = "flame_profile", flame)]
     pub fn handle_consensus(&mut self, msg: ConsensusMessage) {
         // Ignore messages from previous and future height
-        if msg.height() < self.state.height() || msg.height() > self.state.height() + 1 {
+        if msg.height() < self.state.height() || msg.height() > self.state.height().next() {
             warn!(
-                "Received consensus message from other height: msg.height={}, self.height={}",
+                "Received consensus message from other height: msg.height={:?}, self.height={:?}",
```
vldm: can you implement `Display` for new types, and print it as number, there no reason to debug print in log

exonum/src/node/state.rs
```diff
@@ -954,10 +948,10 @@ impl State {
 
     /// Returns `true` if the node doesn't have proposes different from the locked one.
     pub fn have_incompatible_prevotes(&self) -> bool {
-        for round in self.locked_round + 1..self.round + 1 {
+        for round in self.locked_round.0 + 1..self.round.0 + 1 {
```
vldm: You should iterate over `Round`s

exonum/src/node/consensus.rs
```diff
@@ -250,7 +251,7 @@ where
     /// Executes and commits block. This function is called when node has full propose information.
     pub fn has_full_propose(&mut self, hash: Hash, propose_round: Round) {
         // Send prevote
-        if self.state.locked_round() == 0 {
+        if self.state.locked_round().0 == 0 {
```
vldm: maybe as method `next` in `Height`, you could implement static method `Round::zero()`, and keep internal value private?

DarkEld3r:
@vldm I like your suggestions, but I'm not sure about making internal values private. I think that it is just some protection from accidental mistakes, so it should not be too restrictive. Additionally, `ValidatorId` and `Height` are used as indexes (`usize`), so we still need to perform such conversion somehow.

exonum/src/blockchain/schema.rs
```diff
@@ -92,8 +93,8 @@ where
     }
 
     /// Returns table that keeps a list of transactions for the each block.
-    pub fn block_txs(&self, height: u64) -> ProofListIndex<&T, Hash> {
-        ProofListIndex::new(gen_prefix(CONSENSUS, 4, &height), &self.view)
+    pub fn block_txs(&self, height: Height) -> ProofListIndex<&T, Hash> {
+        ProofListIndex::new(gen_prefix(CONSENSUS, 4, &height.0), &self.view)
```
alekseysidorov: Can you avoid of usage `height.0` in code?

exonum/src/blockchain/schema.rs
```diff
@@ -163,17 +164,17 @@ where
...
 
     /// Returns the current height of the blockchain. Its value is equal to `last_height + 1`.
-    pub fn current_height(&self) -> u64 {
+    pub fn current_height(&self) -> Height {
         let last_height = self.last_height();
         match last_height {
-            Some(last_height) => last_height + 1,
-            None => 0,
+            Some(last_height) => last_height.next(),
+            None => Height(0),
```
alekseysidorov: In my opinion `Height::zero()` is more clear.

exonum/src/blockchain/schema.rs
```diff
@@ -273,15 +274,15 @@ where
...
         let actual_from = self.configs_actual_from();
         for i in (0..actual_from.len()).rev() {
-            if actual_from.get(i).unwrap().actual_from() <= height {
+            if actual_from.get(i).unwrap().actual_from() <= height.0 {
                 return i as u64;
             }
         }
         panic!(
-            "Couldn't not find any config for height {}, \
+            "Couldn't not find any config for height {:?}, \
```
alekseysidorov: Why do you not implement `Display` for height?

exonum/src/node/state.rs
```diff
@@ -220,7 +214,7 @@ where
 
     /// Inserts a new message if it hasn't been inserted yet.
     pub fn insert(&mut self, message: &T) {
-        let voter = message.validator() as usize;
+        let voter = message.validator().0 as usize;
```
alekseysidorov: I think that `message.validator().as_usize()` is better.

exonum/src/node/state.rs
```diff
@@ -647,16 +641,16 @@ impl State {
...
 
     /// Increments the node height by one and resets previous height data.
     // FIXME use block_hash
     pub fn new_height(&mut self, block_hash: &Hash, height_start_time: SystemTime) {
-        self.height += 1;
+        self.height.increment();
         self.height_start_time = height_start_time;
-        self.round = 1;
-        self.locked_round = 0;
+        self.round = Round(1);
```
alekseysidorov: What do you think about `Round::first()` name?

exonum/src/blockchain/schema.rs
DarkEld3r: Unfortunately, no: `gen_prefix` is a generic function, that works with different "suffixes".

exonum/src/blockchain/schema.rs
DarkEld3r: Actually, `Display` is already implemented for `Height`, I just forgot to update usage.

exonum/src/node/state.rs
DarkEld3r: Just for one call? :smile: Additionally, in sandbox tests different rounds are used.

DarkEld3r:
Review status: 0 of 27 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[exonum/src/encoding/tests.rs, line 247 at r8](https://reviewable.io:443/reviews/exonum/exonum/262#-Kqj1PdHPA8UamJKTOmk:-Kqj1PdHPA8UamJKTOml:b-omu225) ([raw file](https://github.com/exonum/exonum/blob/84bd2cb6a55683f813fc3f2ebda55ad094da9b4f/exonum/src/encoding/tests.rs#L247)):*
> ```Rust
>     );
>     // read
>     assert_eq!(prevote.validator(), VALIDATOR);
> ```

Reviewable test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/exonum/exonum/262)*

Issue #186, opened 2017-06-23
Node panic with `panicked at 'no entry found for key'`


vldm:
In benchmark there was error: `panicked at 'no entry found for key'`.
The error was only once, so I can reproduce it, with `RUST_BACKTRACE=1`

Probably as part of https://github.com/exonum/exonum-core/issues/159
we need to checks if we got index lookup `map[key]` without check of map\arrays boundaries.

Update 1:
The backtrace looks like:
```
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:355
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:371
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:549
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:511
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:495
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:471
   8: core::panicking::panic_fmt
There are 97500 transactions in the pool, speed = 500 / Duration { secs: 0, nanos: 200000000 }
             at /checkout/src/libcore/panicking.rs:69
   9: core::option::expect_failed
             at /checkout/src/libcore/option.rs:794
  10: exonum::blockchain::Blockchain::create_patch
  11: exonum::node::basic::<impl exonum::node::NodeHandler<S>>::handle_message
  12: <exonum::node::NodeHandler<S> as exonum::events::EventHandler>::handle_event
  13: <mio::event_loop::EventLoop<H>>::run_once
  14: exonum::node::Node::run_handler
  15: exonum::node::Node::run
  16: validator_generator::main
  17: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:98
  18: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:433
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:57
  19: __libc_start_main
  20: _start
[...]

PR #264, opened 2017-07-24
Fix panic, in handle_block

vldm:
Closes https://github.com/exonum/exonum/issues/186

### DIFF
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -221,7 +221,9 @@ impl Blockchain {
             let last_hash = self.last_hash();
             // Save & execute transactions
             for (index, hash) in tx_hashes.iter().enumerate() {
-                let tx = &pool[hash];
+                let tx = pool.get(hash).expect(
+                    "BUG: Cannot find transaction in pool.",
+                );
 
                 fork.checkpoint();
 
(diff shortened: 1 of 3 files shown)


exonum/src/node/state.rs
```diff
@@ -703,10 +703,16 @@ impl State {
     /// Transaction is ignored if the following criteria are fulfilled:
     /// - transactions pool size is exceeded
     /// - transaction isn't contained in unknown transaction list of any propose
-    pub fn add_transaction(&mut self, tx_hash: Hash, msg: Box<Transaction>) -> Vec<(Hash, Round)> {
+    /// - transaction isn't a part of block
+    pub fn add_transaction(
+        &mut self,
+        tx_hash: Hash,
+        msg: Box<Transaction>,
+        // if tx is in some of propose or in a block,
```
DarkEld3r: Maybe move that comment to the function description?

vldm: For me  it's more clear, with rustfmt `block` mode, when the comment is as close as possible with the place that it comments.

exonum/src/blockchain/mod.rs
```diff
@@ -221,7 +221,9 @@ impl Blockchain {
             let last_hash = self.last_hash();
             // Save & execute transactions
             for (index, hash) in tx_hashes.iter().enumerate() {
-                let tx = &pool[hash];
+                let tx = pool.get(hash).expect(
+                    "BUG: Cannot found transaction in pool.",
```
gisochre: find

exonum/src/node/state.rs
```diff
@@ -703,10 +703,16 @@ impl State {
...
+        // we should add it, or we can stuck on some state
```
gisochre: could become

gisochre: in

PR #270, opened 2017-07-26
Export proof key size constant

DarkEld3r:


defuz:
@DarkEld3r Perhaps the name `PROOF_MAP_KEY_SIZE` would be more suitable?

PR #273, opened 2017-07-31
Added possibility to choose an external storage at compile time

aleksuss:


exonum/src/storage/tests.rs
```diff
@@ -12,17 +12,35 @@
...
 // limitations under the License.
 
+#[cfg(not(feature = "memorydb"))]
 use tempdir::TempDir;
 
-use super::{Database, LevelDB, LevelDBOptions, MemoryDB, Snapshot, Fork};
+#[cfg(not(feature = "memorydb"))]
+use std::path::Path;
 
+use super::{Database, Snapshot, Fork};
```
alekseysidorov: You can create local `mod` for each db and select them by feature.

aleksuss:
By default, tests run for every type of database. With feature param you can explicitly define what type of database you want to use. But MemoryDB is present always because too many doctests depend on it.

PR #274, opened 2017-07-31
Store snapshot as raw pointer

DarkEld3r:
Just an experiment. :smile:


DarkEld3r:
Discussed: `Snapshot` should store `Arc<Db>` inside.

PR #282, opened 2017-08-02
Update Travis script (sudo false)

DarkEld3r:


DarkEld3r:
Doc-tests fails randomly without `sudo: required`. :cry:

Issue #279, opened 2017-08-02
Fields in information types are private.


alekseysidorov:
In some types like `TxLocation` or `TxInfo` fields are private. That does not allow to create own api on top on them.

PR #278, opened 2017-08-01
Fix deserialization panic.

vldm:
Closes #277

PR #283, opened 2017-08-02
Make fields public

vldm:
Based on https://github.com/exonum/exonum/pull/278
Closes https://github.com/exonum/exonum/issues/279

### DIFF
--- a/exonum/src/blockchain/tests.rs
+++ b/exonum/src/blockchain/tests.rs
@@ -14,6 +14,30 @@
 
 #![allow(dead_code)]
 
+use crypto::Hash;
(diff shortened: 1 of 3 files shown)


alekseysidorov:
The `TxLocation` contains the height that is needed to find the nearest anchoring transaction.

gisochre:
@alekseysidorov review plz. `TxLocation` fields are pub as components of `encoding_struct!`.

alekseysidorov:
But location field it `TxInfo` is private. If I want to add information about anchoring to `TxInfo` I need to get tx_location again.

PR #286, opened 2017-08-03
Fix misleading variable names in the map proof tree nodes.

dmitry-timofeev:


alekseysidorov:
Please add "BREAKING CHANGE" label. This change will break api.

gisochre:
@dmitry-timofeev why is this needed? for clarity? 
@boguslavsky whatcha think?

dmitry-timofeev:
> for clarity?

Exactly. The variables in those structures are clearly **not** hashes, therefore, the naming is misleading.

PR #287, opened 2017-08-03
Rename messages::Block to BlockResponse

### LINKED ISSUE #244: Different names for `messages::Block` and `blockchain::Block`
These types have different semantic, so naming should reflect that.

DarkEld3r:
Firstly I though about using `BlockInfo` name, but it is already used in the `explorer` module. Any ideas of a better naming?

Closes #244.

### DIFF
--- a/exonum/src/messages/protocol.rs
+++ b/exonum/src/messages/protocol.rs
@@ -47,8 +47,8 @@ pub const PROPOSE_MESSAGE_ID: u16 = 2;
 pub const PREVOTE_MESSAGE_ID: u16 = 3;
 /// `Precommit` message id.
 pub const PRECOMMIT_MESSAGE_ID: u16 = 4;
-/// `Block` message id.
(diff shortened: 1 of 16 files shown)


exonum/src/messages/protocol.rs
```diff
@@ -244,15 +244,15 @@ message! {
...
         /// The sender's public key.
-        field from:           &PublicKey          [00 => 32]
+        field from:           &PublicKey                [00 => 32]
         /// Public key of the recipient.
-        field to:             &PublicKey          [32 => 64]
-        /// Block.
-        field block:          blockchain::Block   [64 => 72]
+        field to:             &PublicKey                [32 => 64]
+        /// Block header.
+        field block:          blockchain::BlockHeader   [64 => 72]
```
alekseysidorov: Why not header?

defuz:
@DarkEld3r I suggest also renaming all `RequestFoo` into `FooRequest`.

Issue #289, opened 2017-08-05
Can't share blockchain state using MemoryDB 


vldm:
Cloning blockchain with internal memorydb
`blockchain().clone()`
Will clone current state, instead of sharing reference to blockchain.

This lead to broken api endpoints.

defuz:
Closed by #292

PR #292, opened 2017-08-08
Share blockchain state using MemoryDB

aleksuss:
#289


aleksuss:
@alekseysidorov If this test add into the code from master branch it will be fail

Issue #296, opened 2017-08-11
Panic in deserialization


vldm:
Panic when trying to decode malicious buffer.

1st Test: (Buffer.len < expected size)
```
encoding_struct! {
    #[allow(dead_code)]
    struct Parent {
        const SIZE = 8;
        field child: Child [0 => 8]
    }
}

encoding_struct! {
    #[allow(dead_code)]
    struct Child {
        const SIZE = 32;
        field child: &Hash [0 => 32]
    }
}

#[test]
fn test_zero_size_segment() {
    let buf = vec![8,0,0,0, // not overlap
                   0,0,0,0,0]; // but with zero size

    <Parent as Field>::check(&buf, 0.into(), 8.into(), 8.into()).expect("Found error in check");
}
```

2nd Test ( if segment_size == 0 we ignore all kind of errors)
```
#[test]
fn test_incorrect_pointer() {
    let buf = vec![222,0,0,0, // incorrect pointer
                   0,0,0,0,0]; // but with zero size

    <Parent as Field>::check(&buf, 0.into(), 8.into(), 8.into()).expect("Found error in check");
}
```

3rd Test (if message.len >= body.len && message.len < message_header + signature_size) :
`attempt to subtract with overflow`

```

#[test]
#[allow(dead_code)]
#[should_panic(expected = "Found error in from_raw: UnexpectedlyShortPayload")]
fn test_message_with_small_size() {
    message! {
        struct SmallField {
            const TYPE = 0;
            const ID = 0;
            const SIZE = 1;
            field test: bool [0 => 1]
        }
    }

    let buff = vec![1;1];
    let raw = Arc::new(MessageBuffer::from_vec(buff));
    let _message = <SmallField as FromRaw>::from_raw(raw).expect("Found error in from_raw");
}
```

PR #297, opened 2017-08-11
Fix malicious buffer deserialization panics.

vldm:
Closes https://github.com/exonum/exonum/issues/296

After some fuzz tests, and additional review I detected new panics in deserialization.

All test cases are filed in issue.

### DIFF
--- a/exonum/src/encoding/tests.rs
+++ b/exonum/src/encoding/tests.rs
@@ -24,22 +24,54 @@ use messages::{RawMessage, Message, FromRaw, Connect, Propose, Prevote, Precommi
 
 use super::{Field, Offset};
 
+#[allow(dead_code)]
+// This structures used to test deserialize,
+// so we should ingore unused `new` method.
+mod ignore_new {
+    use crypto::Hash;
+    encoding_struct! {
+        struct Parent {
+            const SIZE = 8;
+            field child: Child [0 => 8]
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


exonum/src/encoding/tests.rs
```diff
@@ -24,22 +24,51 @@ use messages::{RawMessage, Message, FromRaw, Connect, Propose, Prevote, Precommi
 
 use super::{Field, Offset};
 
+#[allow(dead_code)]
+mod ignore_new {
```
alekseysidorov: What is purpose of this module?

exonum/src/messages/tests.rs
```diff
@@ -25,3 +27,21 @@ fn test_message_without_fields() {
     }
     drop(NoFields::new(&SecretKey::new([1; 64])));
 }
+
+#[test]
+#[allow(dead_code)]
```
alekseysidorov: Why you use this attribute? Can you explain me?

exonum/src/encoding/tests.rs
vldm: I use `encoding_struct` to set buffer schema. So I need to ignore `dead_code` warning about `new` method. I add some comment.

exonum/src/messages/tests.rs
vldm: the same as above.

Issue #73, opened 2017-04-27
Network discovery failure via RequestPeers


gisochre:
This was observed to sometimes result in
1. network discovery problems: not all validators getting updates for transactions which resulted in them having empty txpools and broadcasting empty proposals.
2. network partitioning (stop/start all validators, watch them not continuing progress on blocks). 

```
    pub fn handle_request_peers(&mut self, msg: RequestPeers) {
        let peers: Vec<Connect> = self.state.peers().iter().map(|(_, b)| b.clone()).collect();
        for peer in peers {
            self.send_to_peer(*msg.from(), peer.raw());
        }
    }
------
    pub fn send_to_peer(&mut self, public_key: PublicKey, message: &RawMessage) {
        if let Some(conn) = self.state.peers().get(&public_key) {
            trace!("Send to addr: {}", conn.addr());
            self.channel.send_to(&conn.addr(), message.clone());
        } else {
            warn!("Hasn't connection with peer {:?}", public_key);
        }
    }
```
If node A missed node B's `Connect`, node A won't send its peers to B upon being requested.

Proposed fix: add `addr` and `time` fields to `RequestPeers`, effectively combining `Connect` and `RequestPeers`. (and combining handling logic too). 
```
        addr:           SocketAddr  [32 => 38]
        time:           SystemTime  [38 => 50]
```

PR #299, opened 2017-08-14
Fix Network discover failure

vldm:
Closes https://github.com/exonum/exonum/issues/73

With `PollOpt::edge()` mio could ignore connection events, if they appear in one short period.

https://carllerche.github.io/mio/mio/struct.Poll.html#edge-triggered-and-level-triggered


vldm:
@gisochre can you confirm, that bug is fixed?

PR #302, opened 2017-08-16
Update clippy to 0.0.152

aleksuss:


exonum/src/storage/proof_list_index/key.rs
```diff
@@ -15,7 +15,7 @@
 use super::super::StorageKey;
 
 const HEIGHT_SHIFT: u64 = 56;
-const MAX_INDEX: u64 = 72057594037927935; // 2 ** 56 - 1
+const MAX_INDEX: u64 = 72_057_594_037_927_935; // 2 ** 56 - 1
```
vldm: may be it's a better way to save this index in hex?
`0xFFFFFFFFFFFFFF`

.travis.yml
```diff
@@ -33,8 +33,8 @@ sudo: required
 
 env:
   global:
-  - NIGHTLY_RUST=nightly-2017-07-05
-  - CLIPPY_VERS=0.0.142
+  - NIGHTLY_RUST=nightly-2017-08-15
```
vldm: Does this `env` still used somewhere?

PR #306, opened 2017-08-18
Make DBKey public

DarkEld3r:
Currently `DBKey` is already exposed through `MapProof` enum:
```rust
pub enum MapProof<V> {
    /// A boundary case with a single element tree and a matching key.
    LeafRootInclusive(DBKey, V),
    /// A boundary case with a single element tree and a non-matching key
    LeafRootExclusive(DBKey, Hash),
    /// A boundary case with empty tree.
    Empty,
    /// A root branch of the tree.
    Branch(BranchProofNode<V>),
}
```

### DIFF
--- a/exonum/src/storage/proof_map_index/key.rs
+++ b/exonum/src/storage/proof_map_index/key.rs
@@ -91,25 +91,28 @@ impl DBKey {
         }
     }
 
+    #[doc(hidden)]
     pub fn from(&self) -> u16 {
         self.from
     }
 
     // TODO: terrible hack, try to remove this
+    #[doc(hidden)]
     pub fn set_from(&mut self, from: u16) {
         self.from = from
     }
 
+    #[doc(hidden)]
     pub fn to(&self) -> u16 {
         self.to
     }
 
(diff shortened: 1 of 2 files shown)


exonum/src/storage/proof_map_index/key.rs
```diff
@@ -91,25 +91,28 @@ impl DBKey {
...
         self.from = from
     }
 
+    #[doc(hidden)]
     pub fn to(&self) -> u16 {
         self.to
     }
 
-    /// Length of the `DBKey`
+    /// Returns length of the `DBKey`.
```
dmitry-timofeev: This comment explains nothing, because the _length_ of the DBKey is 34 bytes :-)

It seems to return the number of significant **bits** in the 'data' field. Must be from 0 for root node of the MPT and up to 256 for a leaf node. 

On top of that value of 'to' in DBKey is not the same as the value of the last byte of the key in the underlying storage (which is from 0 to 255 for branch nodes, and always zero for leaves). See `impl StorageKey for DBKey` for more.

Issue #276, opened 2017-08-01
Use stream hashing/signing where possible


defuz:
Review all code base and find places where the use of streaming hashing/signing will get rid of extra allocations.

aleksuss:
I couldn't find any place where extra allocation in `hash` and `sign/verify` used. If you could show me a place like that I would really appreciate it.

DarkEld3r:
Closed by #307.

PR #307, opened 2017-08-21
Using hash streaming for getting rid of reallocation

aleksuss:
#276


exonum/src/storage/proof_map_index/key.rs
```diff
@@ -191,10 +191,10 @@ impl DBKey {
         self.to == (KEY_SIZE * 8) as u16
     }
 
-    // TODO: terrible hack, try to remove this
-    pub fn to_vec(&self) -> Vec<u8> {
-        let mut buffer = vec![0u8; DB_KEY_SIZE as usize];
-        self.write(&mut buffer);
+    /// Represents `DBKey` as byte array and returns it
+    pub fn as_bytes(&self) -> Box<[u8]> {
```
alekseysidorov: Why `Box<[u8]>` is better than `Vec<u8>`?

DarkEld3r: As for me, that isn't very important, but `Box<[u8]>` is slightly "stricter": it has no dynamic size/capacity. :smile:

alekseysidorov:
Can you add entry to CHANGELOG.md file?

Issue #127, opened 2017-05-24
Errors in passed test output


vldm:
in log we have some messages like 
```
test handle_tx_ignore_existing_tx_in_blockchain ... ok
[1495659937 : 786] - [ ERROR ] - exonum::node::consensus - Wrong propose leader detected: actual=3, expected=2
test ignore_propose_from_non_leader ... ok
test has_full_propose_wrong_state_hash ... ok
[1495659937 : 800] - [ ERROR ] - exonum::node::consensus - Received propose with wrong last_block_hash msg=Propose { validator: 2, height: 1, round: 1, prev_hash: "Hash(00000000...)", transactions: [] }
test ignore_propose_with_incorrect_prev_hash ... ok
[1495659937 : 808] - [ ERROR ] - exonum::node::consensus - Received propose with already commited transaction, msg=Propose { validator: 3, height: 2, round: 1, prev_hash: "Hash(daa1e36f...)", transactions: ["Hash(28776065...)"] }
```
watch:
https://travis-ci.com/exonum/exonum-core/builds/46748879#L564-L569

It's because of logger initialization on some tests.
Logger has external state, that has side effects in other tests.

To fix it we can init logger, to output in stdout, or didn't init it in tests at all.
https://doc.rust-lang.org/log/env_logger/struct.LogBuilder.html#method.target

PR #308, opened 2017-08-22
Remove error from test log

vldm:
Closes https://github.com/exonum/exonum/issues/127

This pr just remove most of `logger::init()` in tests.
Every test should be self-reproducible, so we don't need debug output's in master for tests.

There still some logs in `events/tests.rs` left, waiting for https://github.com/exonum/exonum/pull/300

Additionally removes old profiler tests code.


gisochre:
@vldm, you made it! You finally killed all test output) Pity, it's not 
> To fix it we can init logger, to output in stdout

PR #312, opened 2017-08-30
Sparse list index - a new data structure

alekseysidorov:
`SparseListIndex` is similar to `ListIndex` but it can contains null elements in the middle.


exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,714 @@
...
+            self.set_size(size);
+        }
+        self.base.put(&index, value)
+    }
+
+    /// Clears the list, removing all values.
+    ///
+    /// # Notes
+    ///
+    /// Currently this method is not optimized to delete large set of data. During the execution of
```
DarkEld3r: ...to delete **a** large set of data...

exonum/src/storage/mod.rs
```diff
@@ -66,6 +66,8 @@
 //! configuration. Similar to combination of [`Box`] and [`Option`] from standard library.
 //! - [`ListIndex`] is a list of items stored in sequential order. Similar to [`Vec`] collection
 //! from standard library.
+//! - [`SparseListIndex`] is a list of items stored in sequential order. Similar to `ListIndex`
+//! but it may contains null items.
```
DarkEld3r: ...but may contain...

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,714 @@
...
+        let total_length = BigEndian::read_u64(&buf[0..8]);
+        let items_count = BigEndian::read_u64(&buf[8..16]);
+        SparseListSize {
+            total_length,
+            items_count,
+        }
+    }
+}
+
+/// The list of items is similar to the [`ListIndex`], but it may contains null items.
```
DarkEld3r: ...may contain null...

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,735 @@
...
+    /// let mut fork = db.fork();
+    /// let mut index = SparseListIndex::new(vec![1, 2, 3], &mut fork);
+    ///
+    /// index.extend([1, 2, 3, 4, 5].iter().cloned());
+    ///
+    /// for val in index.iter() {
+    ///     println!("{}", val);
+    /// }
+    /// ```
+    pub fn iter(&self) -> SparseListIndexIter<V> {
```
defuz: Since sparse list is less trivial than a list, I think that its iterators should be more like the iterators of map:

indices() -> iterator by indeces (like keys in maps)
values() -> iterator by values
iter() -> iterator by pairs (index, value)

exonum/src/storage/mod.rs
```diff
@@ -66,6 +66,8 @@
 //! configuration. Similar to combination of [`Box`] and [`Option`] from standard library.
 //! - [`ListIndex`] is a list of items stored in sequential order. Similar to [`Vec`] collection
 //! from standard library.
+//! - [`SparseListIndex`] is a list of items stored in sequential order. Similar to `ListIndex`
+//! but it may contain null items.
```
defuz: "may contain null items" -> "may contain indexes without values"

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,838 @@
...
+        let mut buf = [0; 16];
+        BigEndian::write_u64(&mut buf[0..8], self.capacity);
+        BigEndian::write_u64(&mut buf[8..16], self.length);
+        buf
+    }
+}
+
+impl StorageValue for SparseListSize {
+    fn hash(&self) -> Hash {
+        hash(&self.to_array())
```
defuz: Using `HashStream` from `crypto.rs` should be more efficient here.

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,838 @@
...
+    }
+
+    fn into_bytes(self) -> Vec<u8> {
+        self.to_array().to_vec()
```
defuz: ...So you do not need `to_array` method and you can write bytes directly into a `Vec` here.

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,838 @@
...
+    ///
+    /// index.push(42);
+    /// assert!(!index.is_empty());
+    /// ```
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+
+    /// Returns the total amount of elements (including "empty" elements) in the list. In fact
```
defuz: > In fact capacity reflects total amount of elements that have ever been put to the index and never gets reduced.

It's not entirely true because `set` method with index argument that is outside of capacity can break this statement.

I would say something like:

> The value of capacity is determined by the maximum index of the element ever inserted into the index.

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,838 @@
...
+    ///
+    /// let db = MemoryDB::new();
+    /// let mut fork = db.fork();
+    /// let mut index = SparseListIndex::new(vec![1, 2, 3], &mut fork);
+    /// assert_eq!(None, index.pop());
+    ///
+    /// index.push(1);
+    /// assert_eq!(Some(1), index.pop());
+    /// ```
+    pub fn pop(&mut self) -> Option<V> {
```
defuz: This method is not very obvious since it uses the number of non-empty items to index in a sparse list. In the first place, this means that it extracts the element from the middle of the list, and secondly, it can return `None` in a situation where the index is not empty yet.

I think that we should leave only 2 methods:

- `pop` which works exactly like `pop_front` (and of course its behavior should be specified in the docs)
- `push` which works like `push_back`

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,838 @@
...
+    /// assert_eq!(2, index.capacity());
+    /// assert_eq!(1, index.len());
+
+    /// assert_eq!(Some(12), index.remove(1));
+    /// assert_eq!(2, index.capacity());
+    /// ```
+    pub fn remove(&mut self, index: u64) -> Option<V> {
+        let mut size = self.size();
+        if index >= size.capacity {
+            warn!(
```
defuz: I think the `warning` is unnecessary here. In some situations, an attempt to remove outside the capacity can be expected behavior. Moreover, in this case, it would be necessary to add a similar warning for `get` method for consistency.

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,838 @@
...
+    /// let mut fork = db.fork();
+    /// let mut index = SparseListIndex::new(vec![1, 2, 3], &mut fork);
+    ///
+    /// index.push(1);
+    /// assert_eq!(Some(1), index.get(0));
+    ///
+    /// index.set(0, 10);
+    /// assert_eq!(Some(10), index.get(0));
+    /// ```
+    pub fn set(&mut self, index: u64, value: V) {
```
defuz: Since you still need to get the old value by the index, it may be useful to return it from the method as `-> Option<V>`

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,838 @@
...
+    /// let db = MemoryDB::new();
+    /// let mut fork = db.fork();
+    /// let mut index = SparseListIndex::new(vec![1, 2, 3], &mut fork);
+    /// assert_eq!(None, index.pop());
+    ///
+    /// index.push_back(1);
+    /// assert_eq!(Some(1), index.pop_front());
+    /// ```
+    pub fn pop_front(&mut self) -> Option<V> {
+        let _opt = {
```
defuz: Strange name for the variable: why `_opt`, not just `first_index` or something like that.

exonum/src/storage/sparse_list_index.rs
```diff
@@ -0,0 +1,838 @@
...
+            self.indices().next()
+        };
+        if let Some(min_index) = _opt {
+            return self.remove(min_index);
```
defuz: Сall `remove` method is formally correct here, but it is not effective because `remove` does additional checks that are not necessary in this case:

- We are guaranteed to know that the index is less than capacity
- We are guaranteed to know that an element with such an index exists

And since we still need get the value from the index, you may use `self.iter().next()` instead of `self.indices().next()`

defuz:
@skletsun And please do not forget to add the mention in `CHANGELOG.md`.

exonum/src/storage/sparse_list_index.rs
```diff
@@ -562,15 +511,21 @@ where
...
-            return self.remove(min_index);
+
+        if let Some((first_index, first_elem)) = first_item {
+            let mut size = self.size();
+            self.base.remove(&first_index);
+            size.length -= 1;
+            self.set_size(size);
+            return Some(first_elem);
         }
         None
```
aleksuss: I suppose more rusty is:
```
 if let Some((first_index, first_elem)) = first_item {
             let mut size = self.size();
             self.base.remove(&first_index);
             size.length -= 1;
             self.set_size(size);
             Some(first_elem)
} else {
          None
}
```

aleksuss: I suppose more rusty is:
```rust
if let Some((first_index, first_elem)) = first_item {
    let mut size = self.size();
    self.base.remove(&first_index);
    size.length -= 1;
    self.set_size(size);
    Some(first_elem)
} else {
    None
}
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Implement `FromStr` and `ToString` traits for public sodium types (#318)
+- Sparse list index - a new data structure (#312)
```
DarkEld3r: A thing to consider: do we want a more detailed changelog? Maybe "...a new data structure similar to `ListIndex`, but it may contain indexes without elements".

PR #314, opened 2017-09-04
Clippy update to 0.0.156

aleksuss:
Rust nightly-2017-09-03

### DIFF
--- a/sandbox/tests/requests.rs
+++ b/sandbox/tests/requests.rs
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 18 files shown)


sandbox/tests/requests.rs
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-extern crate exonum;
-extern crate sandbox;
+//extern crate exonum;
```
DarkEld3r: I suppose this file is a stub for additional tests. Personally I prefer to remove it completely and add an issue.

PR #318, opened 2017-09-05
Implement `FromStr` for public sodium types.

### LINKED ISSUE #317: Missing `FromStr` impelementations for types that can be converted from hex.
There are no `FromStr` implementation for types like `PublicKey`, so `Context` in clap fabric module can't get these types from arguments directly.

alekseysidorov:
Closes #317

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- Implement `FromStr` and `ToString` traits for public sodium types (#318)
+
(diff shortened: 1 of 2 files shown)


vldm:
Can you implement `https://doc.rust-lang.org/alloc/string/trait.ToString.html` for public types as well.
(there should be only one convenient way to represent public types)
And may be implement `FromStr` for sodium private types?

alekseysidorov:
In my opinion, for private types it is necessary for a developer to know exactly what he is doing.

ivanovaleksey:
Hello everyone, why `ToString` was implemented?
Wouldn't it be better to implement `Display`?

ivanovaleksey:
Thanks @vitvakatu, will make a PR soon.

Issue #75, opened 2017-04-28
Persistent storing of sent consensus messages


defuz:
We need this to fulfill the requirements of the consensus algorithm.

vldm:
@skletsun it would be good, if you could test this issue (before and after merging pr)
https://github.com/exonum/exonum/issues/218

Probably testcase looks like:
1. validator sent `Propose` `X`.
2. validator reboot
3. validator sent another `Propose` `Y`.

PR #322, opened 2017-09-11
Consensus messages cache [ECR-205]

skletsun:
Introducing a cache for consensus messages that stores all required information to database until new height achieved.

Closes: https://github.com/exonum/exonum/issues/75


exonum/src/blockchain/mod.rs
```diff
@@ -343,6 +343,9 @@ impl Blockchain {
                     schema.precommits_mut(&block_hash).push(precommit.clone());
                 }
 
+                // cleanup the consensus messages cache
+                schema.consensus_messages_cache_mut().clear();
```
DarkEld3r: I'm not sure that we need such comments. :smile: The code is pretty self-explanatory:  `....consensus_messages_cache...().clear();`.

In my opinion comments are needed in the following cases:
1. Code is non-trivial.
2. Intent isn't obvious (for example, you can want to explain why we need to clear the cache).

exonum/src/blockchain/mod.rs
```diff
@@ -381,6 +384,41 @@ impl Blockchain {
...
+    pub fn save_message(&mut self, raw: &RawMessage) {
+        let mut fork = self.fork();
+
+        {
+            let mut schema = Schema::new(&mut fork);
+            schema.consensus_messages_cache_mut().push(raw.clone());
+        }
+
+        // apply changes
+        self.merge(fork.into_patch()).unwrap();
```
DarkEld3r: Perhaps it will be better to return `Result<(), Error>` instead of calling unwrap? In that case high level code can decide what to do with that error.

exonum/src/blockchain/mod.rs
```diff
@@ -381,6 +384,41 @@ impl Blockchain {
...
+        {
+            let mut schema = Schema::new(&mut fork);
+            let mut index = schema.consensus_messages_cache_mut();
+            for msg in iter {
+                index.push(msg.clone());
+            }
+        }
+
+        // apply changes
+        self.merge(fork.into_patch()).unwrap();
```
DarkEld3r: Same here.

exonum/src/blockchain/schema.rs
```diff
@@ -62,6 +62,9 @@ pub struct Schema<T> {
     view: T,
 }
 
+/// Constants for the 'ord' values used in prefix generation
+const CONSENSUS_MESSAGES_CACHE_ORD: u8 = 9;
```
DarkEld3r: :+1: 

How about adding other constants?

exonum/src/node/mod.rs
```diff
@@ -83,7 +83,7 @@ where
     inner: S,
 }
 
-/// Handler that that performs consensus algorithm.
+/// Handler that performs consensus algorithm.
```
DarkEld3r: :smile: :+1:

exonum/src/blockchain/mod.rs
skletsun: Good point, thanks

exonum/src/blockchain/schema.rs
skletsun: It was my initial intention ;)

exonum/src/blockchain/mod.rs
```diff
@@ -386,23 +386,23 @@ impl Blockchain {
...
 
         {
             let mut schema = Schema::new(&mut fork);
             schema.consensus_messages_cache_mut().push(raw.clone());
         }
 
-        // apply changes
-        self.merge(fork.into_patch()).unwrap();
+        self.merge(fork.into_patch())?;
+        Ok(())
```
DarkEld3r: This can be rewritten the following way:
```rust
self.merge(fork.into_patch())
```

exonum/src/blockchain/mod.rs
```diff
@@ -412,12 +412,12 @@ impl Blockchain {
...
             for msg in iter {
-                index.push(msg.clone());
+                index.push(msg);
             }
         }
 
-        // apply changes
-        self.merge(fork.into_patch()).unwrap();
+        self.merge(fork.into_patch())?;
+        Ok(())
```
DarkEld3r: Same here.

exonum/src/blockchain/mod.rs
skletsun: Indeed. Somehow I mixed it with a different method and thought self.merge() returns Result<T, Error> instead of Result<(), Error>.

exonum/src/node/consensus.rs
```diff
@@ -384,6 +384,14 @@ where
 
             // Change lock
             if self.state.has_majority_prevotes(round, propose_hash) {
+
+                // Put our consensus messages to the cache. This is for validators only.
+                if self.state.is_validator() {
+                    self.blockchain
+                        .save_messages(self.state.prepare_consensus_messages().into_iter())
```
alekseysidorov: You can use `IntoIterator` instead of `Iterator` in `save_messages` to avoid `into_iter()` call.

skletsun: Done

skletsun:
Closes: https://github.com/exonum/exonum/issues/75

skletsun:
Closes: https://github.com/exonum/exonum/issues/75

exonum/src/blockchain/schema.rs
```diff
@@ -362,6 +431,23 @@ impl<'a> Schema<&'a mut Fork> {
         ProofMapIndex::new(gen_prefix(CONSENSUS, 8, &()), &mut self.view)
```
vldm: missing `STATE_HASH_AGGREGATOR_ORD`

sandbox/tests/consensus.rs
```diff
@@ -2808,6 +3177,16 @@ fn test_schema_config_changes() {
     );
 }
 
+// helper method to check for messages in consensus cache
```
DarkEld3r: Please remove the commented code or add an explanation.

vldm: remove this comments?

DarkEld3r:
Please add changelog entry ("Internal improvements" section).

DarkEld3r:
I cannot believe I have done it. Please review very thoroughly. At the first glance, everything should be OK: after restart the node processes all messages, therefore it broadcasts all responses one more time. But I want to be 100% sure that I'm not missing something.

exonum/CHANGELOG.md
```diff
@@ -88,6 +88,11 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `ExonumJsonDeserialize` trait is implemented for `F32` and `F64`. (#461)
 
+### Internal improvements
+
+- Consensus messages are stored persistently (in the database), so restart will
+  not affect the node's behaviour. (#322)
```
slowli: Nitpick: "behaviour" is the British spelling, the US one is "behavior".

exonum/src/blockchain/schema.rs
```diff
@@ -373,42 +388,55 @@ impl<'a> Schema<&'a mut Fork> {
...
     /// [1]: struct.Schema.html#method.peers_cache
     pub fn peers_cache_mut(&mut self) -> MapIndex<&mut Fork, PublicKey, Connect> {
-        MapIndex::new("core.peers_cache", &mut self.view)
+        MapIndex::new("core.peers_cache", self.view)
+    }
+
+    /// Mutable reference to the [`consensus_messages_cache`][1] index.
+    ///
+    /// [1]: struct.Schema.html#method.consensus_messages
+    pub fn consensus_messages_cache_mut(&mut self) -> ListIndex<&mut Fork, RawMessage> {
```
slowli: I'd say *all* mutating methods of the schema should have `pub(crate)` or lesser visibility, especially those dealing with indices directly. In this particular case, I would argue even read-only indices should be `pub(crate)` (OTOH, it may adversely impact testing, but in this case I'd say conditional compilation may still be preferable to exposing a potentially very dangerous interface).

exonum/src/blockchain/mod.rs
```diff
@@ -439,6 +443,47 @@ impl Blockchain {
...
+    pub fn save_messages<I>(&mut self, round: Round, iter: I)
+    where
+        I: IntoIterator<Item = RawMessage>,
+    {
+        let mut fork = self.fork();
+
+        {
+            let mut schema = Schema::new(&mut fork);
+            let mut index = schema.consensus_messages_cache_mut();
+            for msg in iter {
```
slowli: Could `extend()` method work here?

exonum/src/blockchain/mod.rs
```diff
@@ -439,6 +443,47 @@ impl Blockchain {
         let it = peers_cache.iter().map(|(k, v)| (k, v.clone()));
         it.collect()
     }
+
+    /// Saves the given raw message to the consensus messages cache.
+    pub fn save_message(&mut self, round: Round, raw: &RawMessage) {
```
slowli: Would it make sense to reduce this method to `save_messages(round, ::std::iter::once(raw))` to reduce code duplication?

exonum/src/blockchain/mod.rs
```diff
@@ -349,6 +349,10 @@ impl Blockchain {
                 for precommit in precommits {
                     schema.precommits_mut(&block_hash).push(precommit.clone());
                 }
+
+                // Consensus messages cache is useful only during one height, so it should be
+                // cleared when a new height is achieved.
+                schema.consensus_messages_cache_mut().clear();
```
slowli: Is the message cache only cleared here? If so, does it mean that a node may process the same message several times if it crashes repeatedly as per [`NodeHandler::initialize()`](https://github.com/skletsun/exonum/blob/49b05f37dced3341a4f709c7602c09f9d3bbb4e9/exonum/src/node/mod.rs#L505)?

exonum/src/node/mod.rs
```diff
@@ -487,14 +487,25 @@ impl NodeHandler {
             info!("Trying to connect with peer {}", address);
         }
 
-        let round = Round::first();
+        let snapshot = self.blockchain.snapshot();
+        let schema = Schema::new(&snapshot);
+
+        // Recover previous saved round if any
+        let round = schema.consensus_round();
```
slowli: Could there be any adverse effects, if say, the remembered round is 1000, and the other nodes are now at greater height with "normal" values of rounds?

exonum/CHANGELOG.md
vitvakatu: Yes, we've decided to use American english only

exonum/src/node/mod.rs
DarkEld3r: This isn't a problem: other nodes will ignore messages from the "previous" height. The restarted node firstly processes saved messages, which is not optimal, but makes code more maintainable because there is only one point for messages processing.

exonum/src/blockchain/mod.rs
DarkEld3r: Right. Why this is a problem?

DarkEld3r: I mean how can it be avoided? On the one  hand, node can become byzantine if it doesn't store its messages. On the other hand, it can keep crashing forever.

Probably we can provide some interface to explicitly clear cached messages during initialization. What do you think?

PR #329, opened 2017-09-19
Add (very) initial support for the metrics infrastructure

### LINKED ISSUE #320: Add macro for collecting metrics
Macro similar to the logging macros (`error!`, `trace!`, etc.) should be added for collecting different metrics.

DarkEld3r:
Closes #320.


DarkEld3r:
@alekseysidorov Nice idea, but it will require a more elaborate implementation that should be flexible enough to use with different backends. I'm not sure that we ready for that (right now). 😄

exonum/src/helpers/metrics.rs
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+//! Utilities for collecting metrics.
+
+use std::time::SystemTime;
+use std::fmt::Arguments;
+
+/// Adds given metric with given value.
+///
+/// Metric name should be in the following format: "module_name.metric_name", where "module_name"
```
aleksuss: Clippy says you should replace double quotes for ` :-)

exonum/src/helpers/metrics.rs
```diff
@@ -0,0 +1,66 @@
...
+    if cfg!(feature = "metrics-hadoop") {
+        write_to_hadoop(metric_name, value, &time);
+    }
+
+    if cfg!(feature = "metrics-log") {
+        trace!("{} {} {}", metric_name, value, time);
+    }
+}
+
+fn write_to_hadoop(_metric_name: &str, _value: &Arguments, _time: &str) {
```
aleksuss: What do you think about adding this attribute `#[cfg(feature = "metrics-hadoop")]` ?

DarkEld3r: Unfortunately, `cfg!(feature = "...")` (macro, not an attribute) works not the same way as C defines, so code under that `cfg!` will still be compiled and cause a compilation error if a function is marked with `#[cfg(...)]` attribute.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Implement `FromStr` and `ToString` traits for public sodium types (#318)
+- Base support for metrics infrastructure.
```
defuz: Please be as specific as possible in a changelog. Something like: "Added new macro `metric!` to collect statistical information."

exonum/src/helpers/metrics.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    #[cfg(feature = "metrics-log")]
+    {
+        trace!("{} {} {}", metric_name, value, time);
+    }
+}
+
+#[cfg(feature = "metrics-hadoop")]
+fn write_to_hadoop(_metric_name: &str, _value: &Arguments, _time: &str) {
```
defuz: Please comment or completely delete this function entirely.

exonum/src/node/consensus.rs
```diff
@@ -461,6 +461,13 @@ where
 
         let height = self.state.height();
 
+        let mempool_size = self.state
+            .transactions()
+            .read()
+            .expect("Expected read lock")
+            .len();
+        metric!("node.mempool", "{}", mempool_size);
```
defuz: Why do we need to format here? Any metric is always just a number. So I prefer to see: 

```rust
metric!("node.mempool", mempool_size);
```

PR #331, opened 2017-09-27
WIP: Added template for pull requests

aleksuss:


aleksuss:
Of course we should. But later :)

aleksuss:
Any suggestions are welcome

defuz:
1. Why do we need "Status" section? As for me it is quite useless and takes up a lot of space at the very beginning. Instead, it would be great to add a reminder, something like `<!--- Add "WIP:" if PR is not ready to merge yet --->`

2. Please delete the line `# Description`, it's just garbage.

3. "short description of the pull request". What is a "description"? I think we need to motivate the author to use the right words here. Looking for PR I would like to know the answer to two questions: 

   - What is the motivation to make these changes?
   - Why is this implemented in this way? (techinal decisions)

4. "Related issues" Here it is good to recall the use of magic construction "Fixed #000".

5. I think "to-do list" should be more verbose to be unambiguous. Example: "Changelog updated", "Added necessary unit-tests" etc

6. "In which projects should related issues be created?" The most important thing: **why** should they be open? I suggest "What another projects are affected by this change?"

7. "suggest other related project names" ...or delete unnecessary

aleksuss:
Added some changes considering suggestions by @defuz

DarkEld3r:
#400.

PR #335, opened 2017-10-04
Float field

Ralvke:
In [cryptocurrency](https://github.com/exonum/cryptocurrency) works fine (with disabled signature verify), but need testing with [exonum-client](https://github.com/exonum/exonum-client)


vldm:
Thank you, for your pull request.

Can you explain why you need float numbers in blockchain? 😄 
There some problems with them, that stopped us to implement them. 

Before we merge it, we need to made another changes:

1. In your pr `f32` implementation is missing, so you need to add it.
2. In json we can't serialize `nan` and `inf`, so in our binary representation, we should validate that `f32` does not contain this values. For reference how to validate value, you can check `Field` implementation for `bool`.


Some other changes that should be mentioned before code landing:
- [ ] We should update docs for new types.
- [ ] We should implement this types in exonum client.

PR #348, opened 2017-10-24
WIP: Add allow_origin option

### LINKED ISSUE #263: Cross-Origin headers for public API
I cannot use Exonum REST API directly if static assets (html, css, js) loaded from other server:
`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource`.

How about to add `Access-Control-Allow-Origin` header and make it configurable through `config.toml`?
```
[api]
domain = "<valid domain or *>"
```

Or will we recommend to use reverse-proxy (nginx) with CORS headers to access Exonum node?

DenisKolodin:
Fixes #263 

I convert `Option<String>` (from `api.allow_origin` field of the config) to `AccessControlAllowOrigin` header:

- `None` to `Null`
- `Some("*")` to `Any`
- `Some(_)` to `Value(_)`

It also adds middleware to Iron (after the public api's chain) and set `AccessControl` header to every response. This middleware adds `Null` value to header if option not set. Do I keep this behavior or I should drop this header if `allow_origin` value had not set? Ideas?

### DIFF
--- a/exonum/src/node/mod.rs
+++ b/exonum/src/node/mod.rs
@@ -18,7 +18,7 @@
 
 use router::Router;
 use mount::Mount;
-use iron::{Chain, Iron};
+use iron::{Chain, Iron, AfterMiddleware, Request, Response, IronResult};
 use iron::headers::AccessControlAllowOrigin;
 use toml::Value;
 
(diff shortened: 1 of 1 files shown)


slowli:
@DenisKolodin Does it work with `POST` requests (i.e., transaction submission)? It seems these requests will trigger [preflight](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests), because `Content-Type` is not among allowed for simple requests.

DenisKolodin:
@slowli It returns `Access-Control-Allow-Origin` for every type of request (including POST). But it doesn't handle preflights, because it needs extra exploration about supported methods of services. But I can send some predefined values for `Access-Control-Allow-Methods`, `Access-Control-Allow-Headers` and `Access-Control-Max-Age` for `OPTIONS` requests. Does it make sense?

slowli:
@DenisKolodin Did you take a look at [`iron-cors`](https://docs.rs/iron-cors/0.6.0-rc.1/iron_cors/)? The package seems to support several allowed hosts (I'm not sure your implementation does), and in the latest RC, preflight requests, too.

vldm:
@DenisKolodin Can we close it in favor of ﻿﻿https://github.com/exonum/exonum/pull/406 ?

Issue #346, opened 2017-10-24
Explorer API range endpoint


quicky84:
Current design of the [blocks-in-range API](https://exonum.com/doc/advanced/node-management/#public-endpoints) does not allow for easy retrieval of the blocks in the chain.
For example, you would like to retrieve only non-empty blocks from the tail of the chain, thus `GET {explorer_base_path}/blocks?count=k&skip_empty_blocks=true`. Imagine that the last k+1 blocks are empty, then the result will be `[]`. Client-side it's not possible to conclude whether you exhausted all blocks (there no transactions in the chain) or just hit a range of (at least k+1) empty blocks. There is no way to continue exploring the chain because you don't know the heights of the involved blocks or, at least, the height of the latest block. See also exonum/exonum-doc#132

For now, the workaround is to retrieve all blocks starting from the tail until you hit the first block and then filter on `tx_count`.

tipoc123:
vote up!

DarkEld3r:
Closed with #509.

PR #351, opened 2017-10-27
API: Update Blocks-in-range endpoint

quicky84:
This is regarding the issue exonum/exonum#346

The PR modifies the behavior of the endpoint in the following ways.
Given `latest` (default: `max_height`), `count`, and `skip_empty_blocks` (default: false)
at most `max(count, max_height-1)` blocks satisfying the criterion bundled with the corresponding range will be returned.
Response has the following form: 
```
{
  range: {
    from: number
    to: number
  },
  blocks: BlockHeader[]
}
```
where `BlockHeader` is described in the [doc](https://exonum.com/doc/advanced/node-management/#response_6)

For example, if `latest=100`, `count=10`, and `skip_empty_blocks=true` then it will fetch at most 10 non-empty blocks. This is different from the current behavior which will explore blocks in range `90..100` and return non-empty blocks in this range.
The proposed behavior is an improvement because in this way response contains exactly what is requested: `count` blocks satisfying the `skip_empty_blocks` condition ending at `latest`.
On the other hand, returning explored ranges will allow user to position themself in the chain. Please refer to the issue exonum/exonum#346

### DIFF
--- a/exonum/src/explorer/mod.rs
+++ b/exonum/src/explorer/mod.rs
@@ -176,7 +176,7 @@ impl<'a> BlockchainExplorer<'a> {
         BlocksRange {
             range: Range {
                 from: height + 1,
-                to: upper - 1
+                to: upper - 1,
             },
             blocks: v,
         }


quicky84:
If someone care to elaborate the error, I will gladly try to fix it.

DarkEld3r:
It seems that you found a heisenbug not related to your changes. :sweat_smile:

quicky84:
All right, glad to hear that I didn't break stuff :)

slowli:
On an unrelated note: *ideally*, there should be test coverage for the explorer, but seemingly there is none right now (otherwise, the build would break 😄 ). Guys, could you please advise whether it's possible to provide this coverage right now, whether it is reasonable to do it within this PR, and what needs to be accomplished to make it happen.

exonum/src/explorer/mod.rs
```diff
@@ -55,6 +55,25 @@ pub struct TxInfo {
     pub proof_to_block_merkle_root: ListProof<Hash>,
 }
 
+/// Range Information
```
DarkEld3r: Nitpick: we (usually) ending comments by dot. :smile: Additionally, "information" should be in lower case.

exonum/src/explorer/mod.rs
```diff
@@ -55,6 +55,25 @@ pub struct TxInfo {
...
+#[derive(Debug, Serialize)]
+pub struct Range {
+    /// Left border of the range, >=0
+    pub from: u64,
+    /// Right border of the range
+    pub to: u64,
+}
+
+/// Information on blocks coupled with the corresponding
+/// range in the clockchain
```
DarkEld3r: Typo: **b**lockchain.

quicky84: All right, I was not aware of the full-stop thing. 
Which release will these changes be included in?

exonum/src/explorer/mod.rs
quicky84: All right, I was not aware of the full-stop thing.

quicky84:
Which release will these changes be included in?

exonum/src/explorer/mod.rs
alekseysidorov: Can you describe more detail? It is necessary to clarify that this is the range of blocks and describe whether the bound values are included in the range.

DarkEld3r:
@quicky84 
> Which release will these changes be included in?

`0.4` if nothing goes wrong. :smile:

slowli:
@quicky84 It seems that now it is possible to test the proposed changes with the help of [`exonum-testkit`](http://github.com/exonum/exonum-testkit/):
```rust
let mut testkit = ...;
let api = testkit.api();
create_some_blocks(&mut testkit); // Put some blocks into the testkit blockchain...
let range: BlocksRange = api.get(ApiKind::Explorer, "v1/blocks");
// Assert something about `range`
```

`ApiKind` is currently undocumented; this is a bug, which I hope we will fix pretty soon.

DarkEld3r:
@quicky84 Would you be interested in finishing this pull request? I will help you with any questions.

quicky84:
Honestly by the time of approval I hand an impression that everything was OK.
What else has to be done to get it merged?

DarkEld3r:
Closed in favor of #502.

PR #352, opened 2017-10-29
Fix request messages id description

DarkEld3r:


slowli:
Should other comments be fixed as well? I grep'd the codebase, and it seems that there are other mentions of old message names, e.g., in [`node/state.rs`](https://github.com/exonum/exonum/blob/00cf5720aca6b438612f952e531370419202659c/exonum/src/node/state.rs#L37), [`node/basic.rs`](https://github.com/exonum/exonum/blob/00cf5720aca6b438612f952e531370419202659c/exonum/src/node/state.rs#L37) and [`node/requests.rs`](https://github.com/exonum/exonum/blob/00cf5720aca6b438612f952e531370419202659c/exonum/src/node/requests.rs#L53) (the last one with a typo, too).

Issue #354, opened 2017-10-30
test_network_reconnect test (sometimes) fails


slowli:
The [`test_network_reconnect` test](https://github.com/exonum/exonum/blob/00cf5720aca6b438612f952e531370419202659c/exonum/src/events/tests.rs#L262) sometimes fails (subjectively, 20% of the time), but (strangely enough) seemingly only when run together with other tests from the `events::tests` module.

Relevant backtrace:

```none
  7: exonum::events::tests::TestHandler::wait_for_message
            at src/events/tests.rs:104
  8: exonum::events::tests::test_network_reconnect
            at src/events/tests.rs:293
  9: <F as test::FnBox<T>>::call_box
            at /checkout/src/libtest/lib.rs:1480
            at /checkout/src/libcore/ops/function.rs:223
            at /checkout/src/libtest/lib.rs:141
 10: __rust_maybe_catch_panic
            at /checkout/src/libpanic_unwind/lib.rs:99
```

Log output if `init_logger()` is placed on the top of the test:

![test-failure-log](https://user-images.githubusercontent.com/9612896/32171770-f7fce9d0-bd81-11e7-8efb-d2881e9a2fba.png)

Steps to reproduce:
- Check out the newest `master` (commit 00cf5720aca6b438612f952e531370419202659c)
- Run `cargo test --lib events::tests` 5-10 times

I did not manage to reproduce the problem when only `test_network_reconnect` is run. I run tests on the VirtualBox 5.0.22 with the Ubuntu 14.04 Server image and the Win 8.1 host. It seems that the problem may depend on the OS and/or other external config, and it may be related to concurrency.

Marginally related to #351, where this problem manifested itself on one of Travis CI builds.

PR #355, opened 2017-10-31
Fix network reconnect test

vldm:
Fixed a problem in `test_network_reconnect` 
Our assumption was that `DisconnectWithPeer` event is processed synchronized.
But if reconnect time is small enough, it's possible that we already disconnect and reconnect to peer, and only after this starting to process `DisconnectWithPeer`.
This cause to disconnect just after reconnect. And double "`Connect` message" processing.

In real code, the time between disconnect and reconnect is significant, and this case could be ignored.
Another reason to ignore this, is that it is a valid case to process double "`Connect` message".

Closes: https://github.com/exonum/exonum/issues/354

### DIFF
--- a/exonum/src/events/tests.rs
+++ b/exonum/src/events/tests.rs
@@ -278,10 +278,8 @@ fn test_network_reconnect() {
     t1.send_to(second, msg.clone());
     assert_eq!(t2.wait_for_message(), msg);
 
-    t1.disconnect_with(second);
-    assert_eq!(t1.wait_for_disconnect(), second);
-
     drop(t2);
+    assert_eq!(t1.wait_for_disconnect(), second);
 
     // Handle second attempt.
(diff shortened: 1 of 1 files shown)


exonum/src/events/tests.rs
```diff
@@ -292,9 +290,9 @@ fn test_network_reconnect() {
     t1.send_to(second, msg.clone());
     assert_eq!(t2.wait_for_message(), msg);
 
-    drop(t2);
-
+    t1.disconnect_with(second);
     assert_eq!(t1.wait_for_disconnect(), second);
+    drop(t2);
```
slowli: Does it make sense to call `drop()` just before the variable goes out of scope? [The Rust docs](https://doc.rust-lang.org/std/mem/fn.drop.html) seem to tell that it is redundant in this case.

vldm: Fxd.

PR #357, opened 2017-11-02
Use 'ServiceFactory' as trait object

DarkEld3r:


defuz:
@DenisKolodin Motivation?

DarkEld3r:
@defuz It was my idea. :smile: I think that approach has two advantages:
1. It works better if we want the framework to provide some `start`/`main` function that takes additional service factories. Currently it cannot be done by function, only by macro because service factory is used as type, not as object.
2. I suppose it is better if we want to change interface into that crates service with given version.

defuz:
@DarkEld3r, BTW is it breaking change? Do we need to update CHANGELOG?

vldm:
Please update changelog before merge.

alekseysidorov:
Can you also update services code?

PR #367, opened 2017-11-07
Move panic after index name validation into the separate function

DarkEld3r:
Motivation: don't repeat `if`/`else` with panic. :smile:

### DIFF
--- a/exonum/src/storage/base_index.rs
+++ b/exonum/src/storage/base_index.rs
@@ -61,14 +61,12 @@ impl<T> BaseIndex<T> {
     /// [`&Snapshot`]: ../trait.Snapshot.html
     /// [`&mut Fork`]: ../struct.Fork.html
     pub fn new<S: AsRef<str>>(name: S, view: T) -> Self {
-        if is_valid_name(&name) {
-            BaseIndex {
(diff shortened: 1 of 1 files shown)


exonum/src/storage/base_index.rs
```diff
@@ -252,19 +248,42 @@ fn is_valid_name<S: AsRef<str>>(name: S) -> bool {
...
-#[test]
-fn test_index_name_validator() {
-    assert!(is_valid_name("index_name"));
-    assert!(is_valid_name("_index_name"));
-    assert!(is_valid_name("AinDex_name_"));
-    assert!(is_valid_name("core.index_name1Z"));
-    assert!(is_valid_name("configuration.indeX_1namE"));
-    assert!(is_valid_name("1index_Namez"));
+/// Calls the `is_valid_name` function with the given name and panics if it returns `false`.
+fn check_name<S: AsRef<str>>(name: S) {
```
vldm: nitpic: maybe it will be clearer if a function with a `panic` inside will have the prefix `assert_`? `assert_valid_name` for example.

DarkEld3r: I like the idea. Fixed.

PR #369, opened 2017-11-08
Updated clippy up to 0.0.169

aleksuss:


exonum/src/blockchain/schema.rs
```diff
@@ -356,7 +356,7 @@ impl<'a> Schema<&'a mut Fork> {
 
     /// Adds a new configuration to the blockchain, which will become an actual at
     /// the `actual_from` height in `config_data`.
-    pub fn commit_configuration(&mut self, config_data: StoredConfiguration) {
+    pub fn commit_configuration(&mut self, config_data: &StoredConfiguration) {
```
vldm: it is not clear for me why we need reference there?
May be, it will be better to remove clone() at line :383

exonum/src/events/network.rs
```diff
@@ -224,7 +224,7 @@ impl NetworkPart {
             network_config,
             self.listen_address,
             handle.clone(),
-            self.network_tx.clone(),
+            &self.network_tx.clone(),
```
vldm: Why we need clone there?

exonum/src/blockchain/schema.rs
aleksuss: `put` in indexes wait a value by value, not by reference

exonum/src/events/network.rs
aleksuss: Yes. Clone here is redundant

exonum/src/blockchain/schema.rs
vldm: yes, and we already got this value in `pub fn commit_configuration(&mut self, config_data: StoredConfiguration)`

aleksuss: But we use `config_data` after putting it into an index in `info!` macros. Therefor we should to call `clone()` anyway. No?

vldm: We can move this info before x)

aleksuss: Yes we can. But we're talking about what we have now :-) 
BY THE WAY: I guess that if `info` would be above put, clippy would had skipped func

PR #370, opened 2017-11-13
Fix crate_authors! macro usage

vldm:
In new version rustc/clap `crate_authors!` can not return static string.

### DIFF
--- a/exonum/src/helpers/fabric/clap_backend.rs
+++ b/exonum/src/helpers/fabric/clap_backend.rs
@@ -29,19 +29,17 @@ impl ClapBackend {
         I: IntoIterator<Item = T>,
         T: Into<OsString> + Clone,
     {
-        let app = clap::App::new("Exonum application based on fabric configuration.")
-            .version(crate_version!())
-            .author(crate_authors!("\n"))
-            .about(
-                "It contain basic set of command, to deploy network on exonum.",
-            );
-
         let subcommands: Vec<_> = commands
             .iter()
             .map(|command| ClapBackend::command_into_subcommand(command))
             .collect();
-
-        let matches = app.subcommands(subcommands.into_iter())
+        let matches = clap::App::new("Exonum application based on fabric configuration.")
+            .version(crate_version!())
+            .author(crate_authors!("\n"))
+            .about(
+                "It contain basic set of command, to deploy network on exonum.",
+            )
+            .subcommands(subcommands.into_iter())
             .get_matches_from_safe(line)
             .unwrap();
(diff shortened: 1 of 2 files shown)


exonum/src/helpers/fabric/clap_backend.rs
```diff
@@ -29,9 +29,10 @@ impl ClapBackend {
         I: IntoIterator<Item = T>,
         T: Into<OsString> + Clone,
     {
+        let authors = crate_authors!("\n"); // crate_authors! broken in tx_sender
```
DarkEld3r: I did understand the problem, but I didn't get this comment. Is it useful at all?

divortep: That might be fixed in a bit different way, just by a minor shift.
It seems that the func chain works if it is closed by get_matches(), like:
```
let matches = clap::App::new("Exonum").author(crate_authors!("\n")).get_matches();
```

So we can change the existing code so:
```
pub fn execute(commands: &[CollectedCommand]) -> Feedback {
    let subcommands: Vec<_> = commands
        .iter()
        .map(|command| ClapBackend::command_into_subcommand(command))
        .collect();

    let matches = clap::App::new("Exonum application based on fabric configuration.")
        .version(crate_version!())
        .author(crate_authors!("\n"))
        .about("Exonum application based on fabric configuration.")
        .subcommands(subcommands.into_iter())
        .get_matches();

    let subcommand = matches.subcommand();
    for command in commands {
        if command.name() == subcommand.0 {
            return command.execute(Context::new_from_args(
                command.args(),
                subcommand.1.expect("Arguments not found."),
            ));
        }
    }

    panic!("Subcommand not found");
}
```

From my perspective that is bug on clap-rs side, since even their example for the macros doesn't work.

vldm: I like this solution more than creating temporary variable.

exonum/src/helpers/fabric/clap_backend.rs
```diff
@@ -29,19 +29,17 @@ impl ClapBackend {
         I: IntoIterator<Item = T>,
         T: Into<OsString> + Clone,
     {
+        let subcommands: Vec<_> = commands
+            .iter()
+            .map(|command| ClapBackend::command_into_subcommand(command))
+            .collect();
         let app = clap::App::new("Exonum application based on fabric configuration.")
```
divortep: It appears to me you forgot to rename 'app' variable to 'matches' here.

PR #371, opened 2017-11-16
Change `Patch` into struct instead of typedef

DarkEld3r:
Motivation: in some cases it is useful to access `Patch` without destroying `Fork`. 

Notes/alternatives: 
1. Instead of using `patch.changes` in `db.rs` corresponding methods can be added to the `Patch` itself. 
2. `changes` field can be "private" (not `pub(crate)`), but that requires an `Iterator` implementation.

Technically this is a breaking change: `Patch` was public, but I hope no one uses it. :smile:


exonum/src/storage/db.rs
```diff
@@ -23,14 +23,29 @@ use super::Result;
...
+    pub(crate) changes: HashMap<String, BTreeMap<Vec<u8>, Change>>,
+}
+
+impl Patch {
+    /// Creates a new empty `Patch` instance.
+    fn new() -> Self {
+        Self { changes: HashMap::new() }
+    }
+
+    /// Returns amount of changes.
```
dmitry-timofeev: the number of changes (if changes is countable in this context)?

exonum/src/storage/db.rs
```diff
@@ -23,14 +23,29 @@ use super::Result;
 use self::NextIterValue::*;
 
 /// A set of serial changes that should be applied to a storage atomically.
-pub type Patch = HashMap<String, BTreeMap<Vec<u8>, Change>>;
+#[derive(Debug, Clone)]
+pub struct Patch {
+    pub(crate) changes: HashMap<String, BTreeMap<Vec<u8>, Change>>,
```
dmitry-timofeev: Shall we make `changes` field crate-private and allow any code to use the underlying map, or add an API to perform operations on a _patch_? In the latter case you will have 

```
self.patch.get_changes(name)  // and
self.patch.contains_changes(name) // (or even a better name of methods)

// instead of

self.patch.changes.get(name)
self.patch.changes.contains_key(name)
```

defuz: I expect to see here:

```rust
pub fn changes(&self, name: &str) -> Option<Changes>
```

where `Changes` is our type that implement `Iterator` over `Change`.

exonum/src/storage/db.rs
```diff
@@ -34,9 +38,43 @@ impl Patch {
...
+        self.changes.get(name)
+    }
+
+    /// Returns a mutable reference to the changes corresponding to the `name`.
+    fn get_changes_mut(&mut self, name: &str) -> Option<&mut Changes> {
+        self.changes.get_mut(name)
+    }
+
+    /// Creates a changes map for the given name if doesn't exist already.
+    fn create_changes_if_needed(&mut self, name: &str) {
```
aleksuss: @DarkEld3r Can you explain pls, why do we need this method?

exonum/src/storage/db.rs
```diff
@@ -34,9 +38,43 @@ impl Patch {
...
+    }
+
+    /// Returns the number of changes.
     pub fn len(&self) -> usize {
-        self.changes.len()
+        let mut count = 0;
+        for cf in self.iter() {
+            count += cf.1.len();
+        }
+        count
```
aleksuss: May be something like this will be better ?
`self.iter().fold(0, |acc, &cf| acc + cf.1.len())`

defuz: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.sum

exonum/src/storage/db.rs
DarkEld3r: There are three places with the following code:
```rust
if !self.patch.contains_key(name) {
    self.patch.insert(name.to_string(), BTreeMap::new());
}
```
I suppose it is enough to move it into a separate function. :smile:

defuz: Instead of call `create_changes_if_needed`  and than `get_changes_mut(...).unwrap()` you may want:

```rust
fn get_changes_mut(...) {
     self.changes.entry(name).or_insert_with(BTreeMap::new)...
}
```

exonum/src/storage/db.rs
```diff
@@ -22,8 +23,78 @@ use super::Result;
 
 use self::NextIterValue::*;
 
+/// Map containing changes with corresponding key.
+/// Changes
```
dmitry-timofeev: Is this line needed (you will see in the API docs in an IDE a sentence followed by a single word)?

exonum/src/storage/db.rs
DarkEld3r: `sum` sums the *elements* of an iterator, so it cannot be used here. :disappointed:

CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Changed a signature of `open` function in a `rocksdb` module. `RocksDBOptions` should pass by the reference. (#369)
+- Removed `round` method from the `ServiceContext`. (#366)
+- `ValidatorState` in the `ServiceContext` replaced by the `ValidatorId`. (#366)
+- `add_transaction` in the `ServiceContext` replaced by the `transaction_sender` which implements the `TransactionSend` trait. (#366)
+- The `Node` constructor now requires `db` and `services` variables instead of `blockchain` instance. (#366)
+- The `Blockchain` constructor now requires services keypair and an `ApiSender` instance. (#366)
+- `mount_*_api` methods in `Blockchain` instance now do not require `ApiContext`. (#366)
```
dmitry-timofeev: Off-topic: sounds interesting :-)

exonum/src/storage/db.rs
```diff
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 use std::collections::HashMap;
-use std::collections::btree_map::{BTreeMap, Range};
+use std::collections::btree_map::{BTreeMap, Range, Iter as BTMIter, IntoIter as BTMIntoIter};
+use std::collections::hash_map::{Iter as HMIter, IntoIter as HMIntoIter, Entry as HMEntry};
```
skletsun: I feel like these names are more close to the Rust naming conventions (upper camelcase for types, https://doc.rust-lang.org/1.0.0/style/style/naming/README.html): BtmIter, BtmIntoIter, HmIter, HmIntoIter, HmEntry. What do you think?

PR #372, opened 2017-11-16
Remove redundant `HexValue` trait.

alekseysidorov:
Within the changes you can use `AsRef<RawMessage>` in `where` conditions.
Also this PR removes `HexValue` trait.


exonum/src/messages/raw.rs
```diff
@@ -280,3 +281,24 @@ impl Message for RawMessage {
...
+        let vec = Vec::<u8>::from_hex(v.as_ref())?;
+        Ok(RawMessage::new(MessageBuffer::from_vec(vec)))
+    }
+}
+
+impl<M: FromRaw> HexValue for M {
+    fn to_hex(&self) -> String {
+        self.raw().to_hex()
+    }
+    fn from_hex<T: AsRef<str>>(v: T) -> Result<Self, FromHexError> {
```
DarkEld3r: Nitpick: an empty line between methods? 😄

exonum/src/messages/raw.rs
```diff
@@ -280,3 +281,25 @@ impl Message for RawMessage {
...
+}
+
+impl<M: FromRaw> HexValue for M {
+    fn to_hex(&self) -> String {
+        self.raw().to_hex()
+    }
+
+    fn from_hex<T: AsRef<str>>(v: T) -> Result<Self, FromHexError> {
+        let raw = RawMessage::from_hex(v)?;
+        M::from_raw(raw).map_err(|_| FromHexError::InvalidHexLength)
```
slowli: Nitpick: I wonder if this error mapping is the best we can do. It seems that `from_raw()` may fail for reasons unrelated to message length (e.g., invalid service / message ID, overlapping segments, etc.). Related: is service / message ID supposed to be checked for validity in `from_hex()` (I can't find corresponding code in [`from_raw()`](https://github.com/exonum/exonum/blob/f198c93a86deb3650db8284d1c6d869113c17e2d/exonum/src/messages/spec.rs#L133)), or is it supposed to operated only on trusted inputs?

defuz: This implementation is not possible if the service ID is not a constant (will happen in near future). I propose to abandon this.

PR #375, opened 2017-11-22
Move exonum to context logger.

vldm:
# Exonum now using slog infrastucture to print log

For core developer, this change lead to slightly modification calls of logging macro. For now this calls should contain additional variables: logging context. 

For service developer, this lead to add additional logging methods `warn` \ `error` \ `info` to `ServiceContext`.

# Details and motivations:
Logging is a powerful mechanism that allows not only displaying information for the user, but later using this information to monitor exonum health, and process it by other third-party utilities.
WIthout structural logging, it's important to write regular expressions for each type of message.

## Working with context
Logging events in exonum could be splitted into 4 main parts:
1. network module - tokio thread, that provides abstraction layer of communication between nodes. Source of logging events on relatively low network level.
2. consensus module - is a main algorithmyc part of exonum. Source of all necessery algorythm related events.
3. service - source of service related events
4. global - other stuf from common utils

Each of this part work in some sort of "context", that somehow printed in logs.
Additionally, inside exonum `handle_messages`, exist another "sub-context" specific for message.

Usually in exonum this context scattered in different logs lines, and it is hard to understand full context by looking into log on some moment of time.

To somehow fix this situation, i decided to keep logging context into specific variables:
`State->consensus_logger` for algorithm specific events;
`NodeHandler->network_logger` for network specific events;
`ServiceContext->logger` for service specific events;
and global logger for global events, that can be used throught slog_scope

To take this context from message, i propose to introduce additional method `logger` for each message type. (By implementing private trait)

## Configuration
Now, exonum support multiple target outputs, additional info in `logger/config.rs`


CHANGELOG.md
```diff
@@ -12,6 +12,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added a new function `merge_sync`. In this function a write will be flushed from the operating system buffer cache before the write is considered complete. (#368)
 - Added conversion into boxed values for values which implement `Service` or `Transaction` traits. (#366)
 - Added constructor for the `ServiceContext` which can be useful for the alternative node implementations. (#366)
+- Added new specialised constructor for `Node` `with_logger` which creates `Node`, with user context. (#375)
+- Added new logging methods to `ServiceContext` (`logger`, `warn`, `error`, `info`). (#375)
+- Added new module `logger`, for logging initialisation. (#375)
+- Added `StubLogger` structure in `logger` module - empty logger for testing purposes. (#375)
+- Added `init_logger` method in `logger` module, which perform global logger initialisation, and returns new logger instance. (#375)
```
DarkEld3r: ...which perform**s** global...

exonum/Cargo.toml
```diff
@@ -15,7 +15,13 @@ description = "An extensible framework for blockchain software projects."
...
 
 [dependencies]
-log = "0.3.8"
+slog = {version="2.0.12" }
+slog-scope = "4.0.0"
+slog-json = "2.0.2"
+slog-term = "2.3.0"
+slog-async = "2.1.0"
+slog-envlogger = "2.0.0"
+lazy_static = "0.2.9"
```
DarkEld3r: Can we use latest (`0.2.11`) `lazy_static` version?

exonum/src/blockchain/service.rs
```diff
@@ -187,9 +191,41 @@ impl ServiceContext {
             fork,
             stored_configuration,
             height,
+            logger,
         }
     }
 
+    /// Returns current logger context, usable when important extended logging system.
```
DarkEld3r: I didn't get the second part of the comment. :smile:

exonum/src/blockchain/service.rs
```diff
@@ -187,9 +191,41 @@ impl ServiceContext {
...
+    /// If one needs more logging feature, one can use `slog` ecosystem to print log message.
+    ///
+    /// Example:
+    /// ```rust, no_build
+    /// #[macro_use] extern crate slog;
+    ///
+    /// // context: ServiceContext
```
DarkEld3r: I propose to use `let context: ServiceContext = ...` instead of comment, because this isn't going to be compiled anyway.

exonum/src/blockchain/service.rs
```diff
@@ -187,9 +191,41 @@ impl ServiceContext {
             fork,
             stored_configuration,
             height,
+            logger,
         }
     }
 
+    /// Returns current logger context, usable when important extended logging system.
+    /// If one needs more logging feature, one can use `slog` ecosystem to print log message.
```
DarkEld3r: I suppose "features" should be in plural here.

exonum/src/blockchain/service.rs
```diff
@@ -187,9 +191,41 @@ impl ServiceContext {
...
+    /// // context: ServiceContext
+    ///
+    /// debug!(context.logger(), "My debug message");
+    /// trace!(context.logger(), "My trace message with context"; "context" => "my_value");
+    /// ```
+    pub fn logger(&self) -> &Logger<ExonumLogger> {
+        &self.logger
+    }
+
+    /// Print message into log with error level.
```
DarkEld3r: "Prints" here and below in order to be consistent with other documentation.

exonum/src/logger/builder.rs
```diff
@@ -0,0 +1,147 @@
+use slog_term::{TermDecorator, FullFormat, TermDecoratorBuilder, CompactFormat};
+use slog::{Never, Drain, FnValue, PushFnValue, Record};
+use slog_json::Json;
+use std::time::SystemTime;
```
DarkEld3r: Please separate thirdparty and std usings by an empty line.

exonum/src/logger/builder.rs
```diff
@@ -0,0 +1,147 @@
...
+    let logger = format.build();
+    match opts.level {
+        LoggingLevel::Env => Box::new(::slog_envlogger::new(logger).fuse()),
+        LoggingLevel::Off => Box::new(StubLogger::new().fuse()),
+        v => Box::new(logger.filter_level(v.into_slog_level().unwrap()).fuse()),
+    }
+}
+
+fn make_json<W: io::Write + Send + 'static>(
+    w: W,
```
DarkEld3r: What is `w` here? Can it have a more meaningful name?

exonum/src/logger/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+    GLOBAL_LOGGER.with(|v| {
+        v.set(Some(set_global_logger(
+            Logger::root_typed(logger_cloned, o!("module" => "global")),
+        )))
+    });
+    ret
+}
+
+impl StubLogger {
+    /// Creates new instance of testing logger.
```
DarkEld3r: ...a new instance...

exonum/src/events/network.rs
```diff
@@ -375,7 +394,7 @@ impl Listener {
                     None => Err(other_error("Incoming socket closed")),
                 })
                 .and_then(move |(connect, stream)| {
-                    trace!("Received handshake message={:?}", connect);
+                    trace!(logger, "Received handshake message={:?}", connect);
```
alekseysidorov: Why not 
```rust
trace!(logger, "Received handshake"; "message" => %connect);
```

exonum/src/logger/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+/// Exonum logger drain
+pub type ExonumLogger = Arc<SendSyncRefUnwindSafeDrain<Ok = (), Err = Never>>;
+
+/// Stub logger for tests
+#[derive(Debug)]
+pub struct StubLogger;
+
+thread_local!(static GLOBAL_LOGGER: Cell<Option<GlobalLoggerGuard>> = Cell::new(None););
+
+/// Performs the logger initialization.
```
alekseysidorov: Is the "global logger" more clear description?

exonum/src/node/basic.rs
```diff
@@ -42,22 +43,26 @@ impl NodeHandler {
...
-        info!("Received Connect message from peer: {}", addr);
+        info!(self.network_logger(), "Received Connect message";  "peer_address" => %addr);
         self.handle_connect(connect);
     }
 
     /// Handles the `Disconnected` event. Node will try to connect to that address again if it was
     /// in the validators list.
     pub fn handle_disconnected(&mut self, addr: SocketAddr) {
-        info!("Disconnected from: {}", addr);
+        info!(self.network_logger(), "Disconnected message";  "peer_address" => %addr);
```
alekseysidorov: I wonder that we have a some architecture issues if we need a `network_logger` here.

exonum/src/node/mod.rs
```diff
@@ -79,6 +81,19 @@ pub enum NodeTimeout {
     PeerExchange,
 }
 
+impl NodeTimeout {
```
alekseysidorov: Why not `Display`?

exonum/src/node/state.rs
```diff
@@ -94,6 +97,10 @@ pub struct State {
 
     timeout_adjuster: Box<TimeoutAdjuster>,
     propose_timeout: Milliseconds,
+
+    // We should keep copy of root logger, to update consensus_logger in future
```
alekseysidorov: Does this mean that there are some problems in the code structure?

matklad:
My two cents (haven't really used slog before, and relatively unfamiliar with exonum):

Look like there are two improvements over the status quo here:

1) Logging is "structured", that is, instead of just strings, one may log arbitrary key-value maps
2) Logging is contextual, that is, we thread a particular logger instance through the code, which adds additional information to each log message. 

The first one is a clear win, while the second one does increase code complexity. The bit about [cloning loggers](https://github.com/exonum/exonum/pull/375/files#diff-7b9d4914caf2fba71b38f92c00c611eaR138) looks especially worrisome. 


Structured (not contextual) logging is coming (slowly 🐌 ) to `log` crate([issue](https://github.com/rust-lang-nursery/log/issues/149)), and *may* be implemented in the current [0.4 cycle](https://users.rust-lang.org/t/log-0-4-0-released/14608).

It's interesting that you can implement poor man's contextual logger on top of structured logging by rewriting `log!(context.logger(), "message")` as `log!("message"; context => context)`.

So an interesting approach would be to try only "structured" part, but still employ only a single global logger.

PR #378, opened 2017-11-24
Cache peers between node restarts

alexander-irbis:
Restore State::peers after restart of process


exonum/src/blockchain/mod.rs
```diff
@@ -28,6 +28,7 @@
 //!
 //! [1]: https://github.com/exonum/exonum-doc/blob/master/src/get-started/create-service.md
 
+use std::collections::HashMap;
```
DarkEld3r: Please move `std` using to the other ones.

exonum/src/blockchain/schema.rs
```diff
@@ -351,6 +357,13 @@ impl<'a> Schema<&'a mut Fork> {
         ProofMapIndex::new("core.state_hash_aggregator", &mut self.view)
     }
 
+    /// Mutable reference to the [`peers_cache`][1] index.
+    ///
+    /// [1]: struct.Schema.html#method.peers
```
DarkEld3r: #method.peers_cache

exonum/src/node/mod.rs
```diff
@@ -292,6 +292,8 @@ impl NodeHandler {
             &config.listener.consensus_secret_key,
         );
 
+        let peers = blockchain.get_saved_peers();
```
DarkEld3r: Not an issue, but I prefer not to introduce additional variable here. :smile:

exonum/src/blockchain/mod.rs
```diff
@@ -400,6 +402,53 @@ impl Blockchain {
...
+            let mut peers = schema.peers_cache_mut();
+            let pubkey = peers.iter()
+                .filter_map(|(k, v)| match v.addr() == *addr {
+                    true => Some(k),
+                    false => None,
+                })
+                .next();
+            if let Some(pubkey) = pubkey {
+                peers.remove(&pubkey);
+            }
```
DarkEld3r: Filtering can be replaced by the following:
```rust
if let Some(pubkey) = peers.iter().find(...) {
    ...
}
```

alexander-irbis:
Lol, something went wrong

alexander-irbis:
Ok, all fixed

exonum/src/node/mod.rs
```diff
@@ -354,10 +355,18 @@ impl NodeHandler {
...
+        for address in &peers {
+            if address == &listen_address {
+                continue;
+            }
+            self.connect(address);
+            info!("Trying to connect with peer {}", address);
+        }
         for address in &self.peer_discovery.clone() {
-            if address == &self.system_state.listen_address() {
+            if address == &listen_address || peers.contains(address) {
```
vldm: This loop copy-paste looks strange for me, can we just `.chain(` iters, and collect it into `HashSet`?

exonum/src/blockchain/mod.rs
```diff
@@ -98,6 +100,14 @@ impl Blockchain {
...
 
+    /// Recreates the blockchain to reuse with a sandbox.
+    pub fn clone_with_api_sender(&self, api_sender: ApiSender) -> Blockchain {
+        Blockchain {
+            api_sender,
+            ..self.clone()
+        }
+    }
+
     /// Returnts service `VecMap` for all our services.
```
alekseysidorov: Oh, can you also fix this typo?

exonum/src/blockchain/mod.rs
```diff
@@ -98,6 +100,14 @@ impl Blockchain {
         }
     }
 
+    /// Recreates the blockchain to reuse with a sandbox.
```
alekseysidorov: And what will happen with the `Mount`s which were created by `mount_*_api`. It seems that they will continue to use an old `api_sender` instance.
If this method is needs only for sandbox testing it should be marked by `#[doc(hidden)]` attribute.

alexander-irbis:
I expect this is the final version if no issues.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `mount_*_api` methods in `Blockchain` instance now do not require `ApiContext`. (#366)
 - Removed redundant `current_height` method in `Schema` and rename `last_height` to `height`. (#379)
 - `last_block` now returns `Block` instead of `Option<Block>`. (#379)
+- The list of peer connections is now restored to the last state after the process is restarted. (#378)
```
DarkEld3r: Perhaps this should be in the "added" section. :smile:

alexander-irbis: I guess this not only adds new functionality, but more about changing behavior.

alexander-irbis:
Hm, I feel I resolved the same conflict in CHANGELOG.md twice

PR #380, opened 2017-11-27
Alternative `MapProof` implementation

slowli:
This PR changes the format of `MapProof`s and principles of their generation.

More detailed description is coming.

See the description on the wiki: https://wiki.bf.local/display/EXN/Flat+proofs+for+ProofMapIndex


DarkEld3r:
Ready for review. :smile: 

@slowli If you don't want, then someone else (me, for example) can update this pull request according to the latest changes.

slowli:
@DarkEld3r I'll try to update the PR in the next 2-3 days and will post here if I encounter any problems.

exonum/src/storage/proof_map_index/mod.rs
```diff
@@ -1057,13 +1057,75 @@ where
     }
 }
 
-impl<T, K, V> ::std::fmt::Debug for ProofMapIndex<T, K, V>
+impl<T, K, V> fmt::Debug for ProofMapIndex<T, K, V>
```
slowli: I've refactored the `Debug` implementation for `ProofMapIndex` in order not to use `Box`es. IMO, this one is more understandable as well.

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -653,13 +642,48 @@ where
...
+                entries: entries.into_iter().map(OptionalEntry::into).collect(),
+                hash: h,
+            }
         })
     }
 }
+
+impl<K, V> CheckedMapProof<K, V> {
+    /// Retrieves references to keys that the proof shows as missing from the map.
+    pub fn missing_keys(&self) -> Vec<&K> {
```
slowli: I would like to want to return `impl Iterator<Item = &K>` here, but `impl Traits` are not stable yet.

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -653,13 +642,48 @@ where
...
-            )
+            CheckedMapProof {
+                entries: entries.into_iter().map(OptionalEntry::into).collect(),
+                hash: h,
+            }
         })
     }
 }
+
+impl<K, V> CheckedMapProof<K, V> {
```
slowli: Would it make sense to add a consuming method returning `self.entries` and/or implement `IntoIterator<Item = (K, Option<V>)`?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -465,17 +465,38 @@ impl StorageKey for ProofPath {
...
-            None
-        } else {
-            for i in self.start()..min(self.end(), other.end()) {
-                let cmp = self.bit(i).cmp(&other.bit(i));
-                if cmp != Ordering::Equal {
-                    return Some(cmp);
-                }
+            return None;
+        }
+        // NB: This check can be moved to "real" code; the method below does not work
```
slowli: The new implementation improves performance of the proof verification (measured by the corresponding bench) by ~35%, making it almost precisely equal to the one before changing ordering of bits in `ProofPath`. As a downside it's less understandable and I'm not completely sure about its correctness (it *is* tested, incl. some fuzzing).

matklad:
@slowli is there any high-level description of how the proof likes? Like, some ascii-art which shows how contours looks like? If there are, I think it might be a good idea to include then into the code (perhaps even crate a `README.md` file in the `proof_map_index/` dir?).

slowli:
@matklad Right now, there is some description and discussion of the idea on [the internal wiki](https://wiki.bf.local/display/EXN/Flat+proofs+for+ProofMapIndex). I understand that it's not in the best shape, but I suppose it's better than nothing.

As a side note: it is possible to verify proofs in a traditional manner, by building a tree and calculating its hash; and, just like now, proofs can be visualized as a tree, just without intermediate nodes mentioned in the proof JSON presentation.

slowli:
@matklad Here is the [explanation](https://github.com/slowli/exonum/blob/map-view-explanation/exonum/src/storage/proof_map_index/README.md) (for a single-entry proofs for now, but the multi-proof case in quite similar). I created a fork of this branch because I'm not sure that source code is the best place for such explanations.

matklad:
@slowli thanks a lot! 

>I'm not sure that source code is the best place for such explanations.

I find that dumping such explanations in README.md file alongside the code works best. Surely, they can become outdated, but that's still better than no explanation at all :) See, for example, how rustc does it for (a few) of it's internal crates: https://github.com/rust-lang/rust/tree/master/src/librustc_typeck

matklad:
So, I definitely like the flat proof format: it feels very elegant indeed!

A couple of questions:


The first is why do we need `CountorNode`s for verification? They are hard to think about, because they are half-nodes, and because we need to traverse pairs of entries (`for w in entries.windows(2)`) to construct them. 

Could we do verification using a stack of full nodes? That is, we traverse entries, push them onto the stack and then try to combine the two topmost entries, if they were neighrours in the tree

Pseudorust:

```Rust
let mut stack = vec![]

for entry in entries {
    stack.push(entry);
    while stack.len() > 2 && siblings(&stack[stack.len() - 1], &stack[stack.len() - 2]) {
        let (right, left) = (stack.pop(), stack.pop());
        stack.push(combine(left, right));
    }
}
assert!(stack.len() == 1);
return stack.pop().unwrap();
```

For the example in readme, it would yield the following sequence of actions

```
> []
push H[1:2]
> [H[1:2]]
push V3
> [H[1:2], V3]
push V4
> [H[1:2], V3, V4]
pop V3, V4; push H[3:4]
> [H[1:2], H[3:4]]
pop H[1:2], H[3:4]; push H[1:4]
> [H[1:4]]
push H[5:7]
> [H[1:4], H[5:7]]
pop H[1:4], H[5:7]; push R
```


The second concerns the representation of `MapProof`, in particular, that fact that we need separate `entries` and `proof` array, and then need merge them together: https://github.com/slowli/exonum/blob/ba7a0e508b204329c3eb08dd61a2d1bc5cf560d2/exonum/src/storage/proof_map_index/proof.rs#L639

Does it make sense to store **all** hashes in the `proof` array, and represent entries as `Entry { idx: usize, value: V }`, such that `proof[idx].hash == value.hash()`? I don't see any major benefits in such representation, but it seems a little more direct perhaps?

slowli:
@matklad Re `ContourNode`s: They are used mainly for performance reasons. I suppose they can be replaced with "ordinary" nodes from the `node` module at the cost of additional memory consumption and additional plumbing (e.g., nodes do not store their keys). As for your proposition, I don't quite get how you determine that 2 nodes are siblings. Is it based on the same algorithm that I use (in which case the `siblings` function in your pseudocode is not pure)?

Re: hashes in proof array: Here are some considerations for the current design.

- I would like for a client to be able to ignore the `proof` part completely if he chooses to do so. **This is a controversial point and IMO it should be discussed further.** Using less controversial phrasing, I would want for the proof without a `proof` part to look like an ordinary data collection. 
- I would want to have data as normalized as possible.

An alternative to your idea would be to have a proof contain references to entries in the simplest possible form: a constant marker indicating to put the next entry from the entry list. For example:

```json
{
  "proof": [
    { "path": "001...", "hash": "deadbeef..." },
    "entry",
    { "path": "101", "hash": "c0ffee..." }
  ],
  "entries": [
    { "key": "SomeKey", "value": "SomeValue" }
  ]
}
```

This would allow to eliminate the sorting step of the verification, so I think it could be worth the effort.

slowli:
Also, I would like to point out that while I find reworking `ProofMapKey` useful, the corresponding functionality could be removed from this PR and dealt with separately. To explain: I don't think we should allow to use anything other than hashes for keys. Even if the other key type follows a distribution close to uniform (as is the case with `PublicKey`), proofs will leak information about other entries in the index, which could be avoided with hashes. Moreover, using hashes could allow to key an index by any hashable type. As a downside, using `k.hash()` as a key raises a question how to restore `k` in the general case.

matklad:
>I don't quite get how you determine that 2 nodes are siblings. Is it based on the same algorithm that I use (in which case the siblings function in your pseudocode is not pure)?

🤔 looks I'll need to reinvent `ContourNode` for that :) Or perhaps the following would work? Let's use `Vec<MapProofEntry>` for the stack. To determine if `stack[-2]` and `stack[-1]` are siblings, we check if `common_prefix_len(stack[-2].path, stack[-1].path) > common_prefix_len(stack[-1].path, entry.path)`. To merge two top stack entries, we set `path` to their common prefix, and calculate hash using the two hashes and suffixes after the common prefix.

matklad:
Could we implement `get_proof` as 

```
pub fn get_proof(&self, key: K) -> MapProof<K, V> {
    self.get_multiproof(::std::iter::once(key))
}
```

? This also allows to kill `add_proof_entries`.

slowli:
You can do that, but it will negatively impact performance (measured by `build_proofs` benches). On my PC, the old `get_proof` implementation is ~10% faster than the one using `get_multiproof` (~15.5 ms vs ~17.5 ms/iter on any of `build_proofs` benches).

matklad:
Hm, so it was possible to get rid of `ContourNode` during proof verification? I wonder if a similar trick could be applied to the proof construction algorithm? 

Am I correct that, in order to create a multiproof, we (conceptually) construct a subtree which is the union of paths from the root node to the leaves from the keyset, and then add all hashes along the border of this subtree? Would it be possible then to reconstruct the proof by walking this border directly and collecting the hashes from exterior of the proof-subtree along the way? Something like this:


```Rust
enum Direction {
    DownLeft,
    DownRight,
    UpRight,
    UpLeft,
    DontMove,
}

fn direction_from_to(a: ProofPath, b: ProofPath) -> Direction {
    if is_ancestor(a, b) {
        // return DownLeft or DownRight
    } else {
        // return UpLeft or UpRight
    }
}

fn proof(keys: Iterator<Item = ProofPath>) -> Vec<(ProofPath, Hash)> {
    let mut curr = root;
    let mut proof: Vec<(ProofPath, Hash)> = vec![];
    for next_key in keys.chain(::std::iter::once(root)) {
        loop {
            match direction_from_to(curr, next_key) {
                Direction::DownLeft => {
                    curr = curr.left_child;
                },
                Direction::DownRight => {
                    proof.push(curr.left_child);
                    curr = curr.right_child
                },
                Direction::UpRight => {
                    proof.push(curr.right_child);
                    curr = curr.parent;
                },
                Direction::UpLeft => {
                    curr = curr.parent
                },
                Direction::DontMove => {
                    proof.push(next_key);
                    break;
                },
            }
        }
    }
}
```

slowli:
@matklad 

> Am I correct that, in order to create a multiproof, we (conceptually) construct a subtree which is the union of paths from the root node to the leaves from the keyset, and then add all hashes along the border of this subtree?

Yes, that's precisely right.

> Would it be possible then to reconstruct the proof by walking this border directly and collecting the hashes from exterior of the proof-subtree along the way?

Maybe, but there are several questions regarding your proposal / code:

- I'm not sure it's possible to add proof entries into the builder just by observing movement directions, i.e., without remembering whether we have visited the children of the node. 
- I would like to minimize retrievals from the database, as they may be slow (not necessarily, but still). The current procedure has a property that it reads nodes from the index no more than once. It seems that your proposal would not have this property.

matklad:
>I'm not sure it's possible to add proof entries into the builder just by observing movement directions, i.e., without remembering whether we have visited the children of the node.

Hm, agree. I think we need to understand the direction of the turn...

```
      B
      |
      | # if we go back from A to B, we need to remember if 
      A # arrived from X or Y
     / \
    /   \
   X     Y
```

Perhaps remembering the last edge would work out? 

>It seems that your proposal would not have this property.

I think it is possible to maintain the path (as a stack) from the root to the curr. When you go down the stack, you push, when you go up, you pop. On the first glance it seems that such stack would be enough to avoid double loads.

slowli:
@matklad I'd say that what you're describing now is *exactly* how the current code works 😉 There is a stack (aka contour), and its elements remember which their children were visited.

matklad:
Agree that that’s basically the same. I just wonder if we indeed need another  data structure (ContourNode) in addition to the prior entry that we already have. It seems like we don’t need to actually *store* visited_left and visited_right, because we can *probably* figure this out based on the local info about traversal(the previos node on he tour). But this all just vague intuition, I have zero confidence in any of this :)

slowli:
@DarkEld3r Should `ProofMapKey` interface stay as-is (i.e., independent from `StorageKey`), or should I revert it to the state presently used in core? As I've explained previously, I'm OK with any of these options.

Also, I can't help but notice that the Travis build of this PR *stably* exits abnormally with a `pthread` failure. IMO, this is a sign that our concurrent code may have some problems (the PR introduces some long-running fuzz tests, but they themselves do not have overt concurrency, so the problem may be elsewhere).

dmitry-timofeev:
Just wondering, do you run fuzz tests as part of the standard build on CI servers?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -188,26 +281,37 @@ pub(crate) trait BitsRange {
...
+                let tail = x.trailing_zeros() as u16;
+                return min(i * 8 + tail - self.start(), max_len);
+            }
+        }
+
+        max_len
+    }
+
+    #[doc(hidden)]
+    fn matches_from(&self, other: &Self, from: u16) -> bool {
```
DarkEld3r: This method isn't public, should it be `doc(hidden)` additionally?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -311,6 +415,111 @@ impl StorageKey for ProofPath {
...
 }
 
+impl PartialOrd for ProofPath {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        if self.start() != other.start() {
+            return None;
+        }
+        // NB: This check can be moved to "real" code; the code below does not work
+        // if `self.start() % 8 != 0` without additional modifications.
+        debug_assert_eq!(self.start(), 0);
```
DarkEld3r: Personally, I'm against debug assertions for such cases. I prefer to replace it with `assert_eq`. As for me, debug assertions are needed when such checks are costly in terms of performance.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -311,6 +415,111 @@ impl StorageKey for ProofPath {
...
+    }
+
+    #[test]
+    fn test_proof_path_serialization() {
+        let path = ProofPath::new(&[1; 32]).prefix(3);
+        assert_eq!(serde_json::to_value(&path).unwrap(), json!("100"));
+        let path: ProofPath = serde_json::from_value(json!("101001")).unwrap();
+        assert_eq!(path, ProofPath::new(&[0b00_100101; 32]).prefix(6));
+
+        // Fuzz tests for roundtrip
```
DarkEld3r: Please add a dot at the end of the comment. :upside_down_face:

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -12,16 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// spell-checker:ignore lhash, rhash, lkey, rkey, bpath, repr
+// spell-checker:ignore bpath, deser, precheck, repr
```
DarkEld3r: We already have "precheck" and "repr" in the "dictionary". Plus I prefer to replace `deser` by the full form.

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -44,538 +42,848 @@ impl Serialize for ProofPath {
...
+    /// Non-terminal node for a map consisting of a single node.
+    #[fail(display = "non-terminal node as a single key in proof")]
+    NonTerminalNode(ProofPath),
+
+    /// One path in the proof is a prefix of another path.
+    #[fail(display = "embedded paths in proof")]
+    EmbeddedPaths {
+        /// Prefix key
+        prefix: ProofPath,
+        /// Key containing the prefix
```
DarkEld3r: Generally, we add dots at the end for even such comments. :upside_down_face:

exonum/src/storage/proof_map_index/tests.rs
```diff
@@ -597,30 +1147,133 @@ fn iter(db: Box<Database>) {
...
+    assert_eq!(keys, vec![other_key.hash()]);
+    assert_eq!(table.get(&key), None);
+    assert_eq!(table.get(&other_key), Some(vec![1, 2, 3]));
+    assert_eq!(
+        table.merkle_root(),
+        hash_isolated_node(&ProofPath::new(&other_key.hash()), &hash(&vec![1, 2, 3]))
+    );
+}
+
+// // // // Macros // // // //
```
DarkEld3r: Personally, I'm against such "fancy" comments. :upside_down_face: Probably, except for the tutorials.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -32,30 +32,118 @@ pub const PROOF_PATH_LEN_POS: usize = KEY_SIZE + 1;
 /// A trait that defines a subset of storage key types which are suitable for use with
 /// `ProofMapIndex`.
 ///
-/// The size of the keys must be exactly 32 bytes and the keys must have a uniform distribution.
-pub trait ProofMapKey: StorageKey {}
+/// The size of the keys must be exactly [`PROOF_MAP_KEY_SIZE`] bytes and the keys must have
+/// a uniform distribution.
+///
+/// [`PROOF_MAP_KEY_SIZE`]: constant.PROOF_MAP_KEY_SIZE.html
```
vitvakatu: Here and everywhere else. Try to open `PROOF_MAP_KEY_SIZE` in docs - it turns out it's impossible to learn actual value of this constant from docs.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -188,26 +281,37 @@ pub(crate) trait BitsRange {
     /// Returns the raw bytes of the key.
     fn raw_key(&self) -> &[u8];
 
+    /// Returns the number of matching bits with `other` starting from position `from`.
+    fn match_len(&self, other: &Self, from: u16) -> u16 {
```
vitvakatu: What about testing this method?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -311,6 +415,111 @@ impl StorageKey for ProofPath {
     }
 }
 
+impl PartialOrd for ProofPath {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        if self.start() != other.start() {
+            return None;
+        }
+        // NB: This check can be moved to "real" code; the code below does not work
```
vitvakatu: Does this comment need to be here?

exonum/src/storage/proof_map_index/mod.rs
```diff
@@ -352,84 +292,47 @@ where
...
+    ///
+    /// ```
+    /// use exonum::storage::{MemoryDB, Database, ProofMapIndex};
+    ///
+    /// let db = MemoryDB::new();
+    /// let snapshot = db.snapshot();
+    /// let index: ProofMapIndex<_, [u8; 32], u8> = ProofMapIndex::new("index", &snapshot);
+    ///
+    /// let proof = index.get_multiproof(vec![[0; 32], [1; 32]]);
+    /// # drop(proof);
```
vitvakatu: For what?

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -12,26 +12,24 @@
...
-use std::fmt;
-
-use crypto::{Hash, HashStream};
-use super::super::{Error, StorageValue};
+use crypto::{CryptoHash, Hash, HashStream};
+use storage::StorageValue;
 use super::key::{BitsRange, ChildKind, ProofMapKey, ProofPath, KEY_SIZE};
+use super::node::{Node, BranchNode};
+
+// Expected size of the proof, in a number of hashed entries.
```
vitvakatu: `in a number` -> `in number`?

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -44,538 +42,841 @@ impl Serialize for ProofPath {
...
+    path: ProofPath,
+    hash: Hash,
+}
+
+// Used instead of `(K, Option<V>)` only for the purpose of clearer (de)serialization.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+enum OptionalEntry<K, V> {
+    Missing { missing: K },
+    KV { key: K, value: V },
```
vitvakatu: It looks like huge over-engineering to me. Why not `struct OptionalEntry { K, Option<V> }`?

slowli: The major reason is that if we ever support storing `Option<V>` as `StorageValue`, it would be impossible to tell `(key, None)` and `(key, Some(None))` in JSON serialization. I agree this concern is somewhat artificial, but nevertheless I'd like to split missing and existing entries, so they cannot be mixed up. Another way would be to avoid overt serialization of missing entries completely, but this makes proofs less "transferable" in a sense that it would become impossible to tell what exactly the proof proves.

exonum/src/storage/proof_map_index/mod.rs
slowli: `drop` is used in some existing `ProofMapIndex` code samples, so I just copied it from there. Do you propose to remove it?

exonum/src/storage/indexes_metadata.rs
```diff
@@ -172,7 +172,23 @@ pub fn set_index_type(name: &str, index_type: IndexType, is_family: bool, view:
 #[cfg(test)]
 mod tests {
     use super::{IndexType, IndexMetadata, INDEXES_METADATA_TABLE_NAME};
-    use storage::{MemoryDB, Database, MapIndex, ProofMapIndex};
+    use storage::{MemoryDB, Database, MapIndex, ProofMapIndex, StorageKey};
+
+    impl StorageKey for [u8; 32] {
```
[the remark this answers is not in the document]
slowli: Because without it, the metadata tests (and only metadata tests) don't work :upside_down_face: In the present implementation, `ProofMapKey` is completely delineated from `StorageKey` (imo, quite deservedly so - the user-supplied key types are *not* stored as keys, `ProofPath`s are), so, as a consequence, there is no more `StorageKey` implementation for `[u8; 32]` (it was previously defined [in `proof_map_index/keys.rs`](https://github.com/exonum/exonum/blob/8e01e2ec04ec92bea8012a4f4da1e389d8d7708c/exonum/src/storage/proof_map_index/key.rs#L42)). Obviously, there may be other solutions:

- Defining `StorageKey` for `[u8; 32]` in `storage/keys.rs`
- Changing `[u8; 32]` to some other type in metadata tests

I'm inclined to go with the second route. Would you be satisfied with it?

exonum/src/storage/proof_map_index/mod.rs
DarkEld3r: We have such `drop`s in other tests. As far as I can remember, they were added to suppress the compiler warnings in one specific case: https://github.com/intellij-rust/intellij-rust/issues/1554
I suppose it can be removed.

exonum/src/storage/proof_map_index/proof.rs
vitvakatu: Thanks, I've been thinking about this today, now I'm convinced.

exonum/src/storage/indexes_metadata.rs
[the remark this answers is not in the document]
vitvakatu: Well, I think second option would be okay. I don't think we should define `StorageKey` for some types only for test purposes, it seems weird.

slowli:
@DarkEld3r How about `pthread lock` errors on Travis? I'm afraid it will stay. I've tried to not use `thread_rng` repeatedly for tests, but so far no luck. Should I try something else (e.g., reduce running time of tests)?

PR #381, opened 2017-11-27
Fix `actual_configuration` behaviour.

alekseysidorov:


exonum/src/blockchain/schema.rs
```diff
@@ -146,30 +146,44 @@ where
     }
 
     /// Returns latest committed block.
+    ///
+    /// # Panics
+    /// - If the "genesis block" was not created.
```
DarkEld3r: Nitpick: I propose to use the same formatting for doc-comments as in standard library. Example:
```
    /// Discription...
    ///
    /// # Panics
    ///
    /// Panics if ...
    ///
    /// # Examples
```
https://doc.rust-lang.org/src/alloc/vec.rs.html#448

exonum/src/blockchain/schema.rs
```diff
@@ -263,10 +277,14 @@ where
         }
         panic!(
             "Couldn't not find any config for height {}, \
-                that means that genesis block was created incorrectly.",
+             that means that genesis block was created incorrectly.",
             height
         )
     }
+
+    fn next_height(&self) -> Height {
```
DarkEld3r: This function isn't public, but I propose to add a doc-comment too. 
There is a possibility to generate documentation for a "private" api by the way.

PR #384, opened 2017-11-29
Add serialization support for f32/f64

DarkEld3r:


aleksuss:
@DarkEld3r I guess it will be great to add some unit tests

exonum/src/encoding/serialize/json.rs
```diff
@@ -129,27 +129,29 @@ macro_rules! impl_deserialize_bigint {
...
-                         .is_some()
+    (@impl $typename:ty) => {
+        impl ExonumJson for $typename {
+            fn deserialize_field<B: WriteBufferWrapper>(value: &Value,
+                                                         buffer: &mut B,
+                                                         from: Offset,
+                                                         to: Offset )
+                -> Result<(), Box<Error>>
+            {
+                let number = value.as_f64().ok_or("Can't cast json as float")?;
```
alekseysidorov: I wonder if an serde deserializer can get here not a finite value?

exonum/src/encoding/fields.rs
```diff
@@ -249,6 +249,75 @@ impl<'a> Field<'a> for i8 {
...
+
+    unsafe fn read(buffer: &'a [u8], from: Offset, to: Offset) -> Self {
+        LittleEndian::read_f32(&buffer[from as usize..to as usize])
+    }
+
+    fn write(&self, buffer: &mut Vec<u8>, from: Offset, to: Offset) {
+        LittleEndian::write_f32(&mut buffer[from as usize..to as usize], *self);
+    }
+
+    fn check(
```
slowli: Is this method covered by tests? Can't find it.

exonum/src/encoding/serialize/json.rs
```diff
@@ -129,27 +129,29 @@ macro_rules! impl_deserialize_bigint {
...
-        impl<'a> ExonumJson for $typename {
-            fn deserialize(value: &Value, buffer: &'a mut Vec<u8>,
-                            from: usize, to: usize ) -> bool {
-                    value.as_f64()
-                         .map(|v| v as $typename)
-                         .map(|val| val.write(buffer, from, to))
-                         .is_some()
+    (@impl $typename:ty) => {
+        impl ExonumJson for $typename {
+            fn deserialize_field<B: WriteBufferWrapper>(value: &Value,
```
slowli: Could you address somewhere how conversion to / from JSON works for `f32` / `f64`? There are values that have a finite decimal representation, which do not exactly map to a `f32` / `f64` value (e.g., `0.1`, because it is an infinite fraction in the binary base), and some can lead to ambiguous results when converting from JSON and back (e.g., `9876543211234567` for `f64` - it has 16 significant digits, which is more than `floor(53 * log_2(10)) = 15` precision provided by the `f64` type).

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added a new function `merge_sync`. In this function a write will be flushed from the operating system buffer cache before the write is considered complete. (#368)
 - Added conversion into boxed values for values which implement `Service` or `Transaction` traits. (#366)
 - Added constructor for the `ServiceContext` which can be useful for the alternative node implementations. (#366)
+- Added serialization support floating point types (`f32` and `f64`). Note: special values (Infinity and NaN) aren't supported. (#384)
```
slowli: I don't quite get what "unsupported" means here. Take the following code:

```rust
use std::f64;
use messages::Message;

message! {
    struct F64Message {
        const TYPE = 0;
        const ID = 0;
        const SIZE = 8;
        field foo: f64 [0 => 8]
    }
}

let sk = SecretKey::new([1; 64]);
let x = F64Message::new(f64::INFINITY, &sk);
let buffer = x.raw().clone();
let y = F64Message::from_raw(buffer).unwrap();
assert_eq!(x, y);
```

It fails on `unwrap()`, which means that it's now possible to create and serialize unserializable messages. Honestly doesn't look like a good design decision to me. Is non-support of infinities and NaNs temporary? Otherwise, maybe it could make sense to provide `Field` implementation for *wrappers* of `f32` / `f64`, which are guaranteed not to equal infinities or NaNs?

DarkEld3r:
I'm going to address the comments more thoroughly, for now just a quick note: Infinity and NaN aren't supported by JSON format, therefore the restriction.

slowli:
@DarkEld3r JSON doesn't indeed support Infinity and NaNs (although some extensions, say, [JSON5](http://json5.org/), do; and I suppose it could be possible to create a crutch like string-encoded `"Infinity"`, `"-Infinity"`, and `"NaN"` values as a JSON-compliant alternative). But, strictly speaking, [pure JSON](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) does not have restrictions on the length of numbers, either; which brings the question why `u64` / `i64` are serialized as strings 😄 (I understand that the practical answer is that the built-in JavaScript JSON parser would otherwise corrupt `u64` / `i64` parsing in some cases.)

vldm:
Is there a reason why we need float in our type system? May be as good alternative would be fixed pointe arythmetic?

DarkEld3r:
I like the idea of using fixed point arithmetic, but I still think that flexibility is good for generic (library/framework) solution, so ideally I prefer to provide both solution and highlight floats disadvantages and complications in the documentation.

dmitry-timofeev:
@DenisKolodin , @Tsiba  I don't think any updates to the serialization format shall go into a release until the specification has been updated. And normally, it shall be updated _first_. Is there a task for that?

DarkEld3r:
https://github.com/DarkEld3r/exonum/pull/1

DarkEld3r:
These changes require more tests:
- Roundtrip of `Float::MIN,` `Float::MAX`, `Float::Zero`, `Float::One`, `Float::MINUS_ONE`.
- Denormalized floats.
- `NaN` and `Infinity`.

DarkEld3r:
https://github.com/DarkEld3r/exonum/pull/2

PR #385, opened 2017-11-29
Add transaction execution status [ECR-332]

DarkEld3r:
Motivation: sometimes it is convenient to know transaction execution status. I suppose core can provide this functionality, so services shouldn't reinvent the wheel for themselves.

Further improvements: execution status can be returned as part of `v1/transactions` request.


exonum/src/blockchain/schema.rs
```diff
@@ -20,6 +20,20 @@ use helpers::Height;
 use super::{Block, BlockProof, Blockchain};
 use super::config::StoredConfiguration;
 
+// TODO: Replace by `const` when/if `contact!` accepts it.
+macro_rules! core_prefix { () => ( "core." ) }
+
+const TRANSACTIONS: &str = concat!(core_prefix!(), "transactions");
```
alekseysidorov: Looks very ugly. Can you try to rewrite macro?

```rust
core_index_name!(TRANSACTIONS_STATUS => "transactions_status");
```

Or like this!

```rust
define_indexes! {
      const NAME = "core";
      index TRANSACTIONS_STATUS => "transactions_status";
}
```

Or any another option.

exonum/src/storage/values.rs
```diff
@@ -232,6 +232,27 @@ impl StorageValue for i64 {
...
+    }
+
+    fn from_bytes(value: Cow<[u8]>) -> Self {
+        debug_assert_eq!(value.len(), 1);
+        debug_assert!(value[0] == 0 || value[0] == 1);
+
+        match value[0] {
+            0 => true,
+            1 => false,
+            _ => unreachable!(),
```
aleksuss: 0 - `true` and 1 - `false` ? Can you please explain why ?

DarkEld3r: Terrible mistake. :sweat_smile:

slowli:
I think `Result<(), E>` where `E` is transaction-specific (an associated type?) could be *even more* descriptive, because there may be various ways in which a transaction may fail. For example, `TxTransfer` in the cryptocurrency tutorial could have `E` defined like this:

```rust
enum TxTransferError {
    MissingSender,
    MissingReceiver,
    InsufficientFunds,
}
```

(There is at least one edge case that is missing -`u64` overflow on the receiver's balance; seems like it's missing from the tutorial code 🤔 ) But this would require to (de)serialize `enum`s.

DarkEld3r:
@slowli It can be expanded even more by adding associated `Ok` type, so it would be possible to have different kinds of successful execution. :smile: I'm not sure if this has some sense though.

Anyway, we had a little discussion and decided to store integer value (I suppose `u8` is enough) instead of `bool`. That allows us to have different results (I have in mind "succeeded", "failed" and "panic") and to simplify further transition to `enums`.
Services can use their custom return codes, but I see no easy way to prevent conflicts in their values. I suppose it is OK for the first implementation.

slowli:
Yeah, regarding the execution status that should be stored in the blockchain, I was thinking of something like

```rust
enum Status<E> {
    /// Everything went fine
    Ok, // or make `Ok(T)` and `Status<T, E>`
    /// User-generated error returned by the transaction
    Error(E),
    /// Panic during error execution
    Panic(String),
}
```

Not sure `String` (i.e., the error description) would be optimal and possible to retrieve at all times.

> Services can use their custom return codes, but I see no easy way to prevent conflicts in their values.

Why is possible conflict a concern? It seems to me that it would be fine that codes would overlap for different transactions; it could be seen as an equivalent for "overlapping" tags in the enum serialization - a non-issue if the type being deserialized is known.

DarkEld3r:
> Why is possible conflict a concern? It seems to me that it would be fine that codes would overlap for different transactions

Overlapping is OK for different transactions, but it isn't OK when "predefined/reserved" values from core overlap with values from transaction. For example, we decide to use `2` value for panic and service author can use the same value for his own specific result. (For now) we only can mention "reserved" values in the documentation.

DarkEld3r:
:warning: :exclamation: 
This implementation has one potential issue with `TRANSACTION_STATUS_PANIC`. If such status is going to be stored in the blockchain, then fix for a problematic transaction must require a new service version: the same transaction will give different status, otherwise.

Technically, it is possible to get the same problem in the current implementation if a transaction panics before fix and successfully modify storage after. :thinking:

CHANGELOG.md
```diff
@@ -9,10 +9,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Added `patch` method to the `Fork` structure. (#393)
+- Added `StorageValue` implementation for `bool`. (#385)
+- Added `StorageValue` implementation for `bool`. (#385)
```
slowli: Duplicate as a result of merge?

exonum/src/blockchain/schema.rs
```diff
@@ -19,6 +19,31 @@ use storage::{Fork, ListIndex, MapIndex, MapProof, ProofListIndex, ProofMapIndex
 use helpers::Height;
 use super::{Block, BlockProof, Blockchain};
 use super::config::StoredConfiguration;
+use super::transaction::ExecutionStatus;
+
+/// Defines `&str` constants with given name and value.
```
slowli: Nitpick: This comment is technically wrong; the name is prefixed with the `"core."` string.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,245 @@
...
+            status: None,
+        }
+    }
+
+    /// Returns execution status consuming `ExecutionContext`.
+    pub fn into_status(self) -> ExecutionStatus {
+        match self.status {
+            Some(val) => val,
+            None => {
+                if self.fork.patch().len() > self.changes_number {
```
slowli: Do I understand it correctly that a transaction is considered failed if it does not commit anything to the storage? If so, I'm against this; there are some cases where not committing anything is a perfectly reasonable behavior.

exonum/src/blockchain/mod.rs
```diff
@@ -239,22 +242,35 @@ impl Blockchain {
...
+                    let r =
+                        panic::catch_unwind(panic::AssertUnwindSafe(|| tx.execute(&mut context)));
+                    (r, context.into_status())
+                };
 
-                match r {
-                    Ok(..) => fork.commit(),
+                let transaction_status = match r {
+                    Ok(()) => {
+                        fork.commit();
```
slowli: Do I understand it correctly that a transaction with an explicitly set `Failed` status may persist changes to the storage anyway? A developer may expect for these changes to be discarded, so if this behavior is intended, it should be described *very* carefully.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+
+    /// Returns reference to the `Fork`.
+    pub fn fork(&mut self) -> &mut Fork {
+        self.fork
+    }
+
+    /// Sets `Transaction`'s execution status. See `ExecutionStatus` for the details.
+    pub fn set_execution_status(&mut self, status: ExecutionStatus) {
```
slowli: Is it possible to set `Panic` status? If yes, I'd say this is a suboptimal design and may warrant separate types for:

- Execution status that can be set by the developer, and
- Execution status that is persisted to the storage

DarkEld3r:
@slowli I suppose the whole idea should be slightly redesigned. Ideally, I prefer to discuss it with you @defuz, @vldm and @alekseysidorov.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,203 @@
...
+    fn execute(&self, fork: &mut Fork) -> TransactionStatus;
+}
+
+/// Execution status of the transaction.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TransactionStatus {
+    /// Successful transaction execution.
+    Succeeded,
+    /// Panic occurred during transaction execution. This status should not be used explicitly.
+    Panic,
```
matklad: Naming seems slightly inconsistent. Should it be `Paniced` to match with `Failed`? If we are going to change this, then perhaps go with all nouns?

```
pub enum TransactionStatus {
   Success,
   Panic,
   Failure,
   Custom(u8),
}
```

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,203 @@
...
+    ///     // ...
+    /// #   fn verify(&self) -> bool { true }
+    /// }
+    /// # fn main() {}
+    fn execute(&self, fork: &mut Fork) -> TransactionStatus;
+}
+
+/// Execution status of the transaction.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TransactionStatus {
```
matklad: An alternative would be to use 

```Rust
type TransactionStatus = Result<(), TransactionError>;

enum TransactionError {
    Panic,
    UnknownFailure,
    Code(u8),
}
```

matklad: Do we want to use `?` when working with transaction? Do we want clients to use `?` when defining transactions? Do we want to allow clients to define `Into<TransactionError> for MyCustomClientError` ?

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,203 @@
...
+        0 => TransactionStatus::Succeeded,
+        1 => TransactionStatus::Panic,
+        2 => TransactionStatus::Failed,
+        val => {
+            assert!(
+                val > MAX_RESERVED_VALUE,
+                "Invalid ExecutionStatus value: {}",
+                val
+            );
+            TransactionStatus::Custom((val - MAX_RESERVED_VALUE - 1) as u8)
```
matklad: Let's add an assertion that `(val - MAX_RESERVED_VALUE - 1) <= u8::MAX_VALUE`?

exonum/src/blockchain/transaction.rs
DarkEld3r: `Result` is nice, but I don't like `()` as `Ok` value: someone can want to have different statuses for the successful execution. What do you think about `Result<u8, TransactionError>`? Not sure about "raw" `u8` though.

matklad: Ah, so `Custom` in `TransactionStatus` can work for both errors and successes?

DarkEld3r: Yes, I had that idea. It isn't ideal, but it should be flexible enough for different potential use-cases.

matklad: That means, that we can't really say if the transaction was successful or not, at the core level?

So, the fully-general(but still code-based) type would look like this?

```Rust
type TransactionResult = Result<TransactionValue, TransactionError>;

struct TransactionValue(u8);
const Success: TransactionVaue = TransactionValue(0);

enum TransactionError {
    Panic,
    UnknownFailure,
    Code(u8)
}
```

I don't really know the full set of tradeoffs here, so I don't know which solution is better. Allowing to use `?` when defining transactions seems somewhat nice though?

DarkEld3r: I agree that `?` is really convenient, but I am not sure about such `TransactionValue` implementation. Maybe it makes sense to use `enum` too? Something like that:
```rust
type TransactionResult = Result<TransactionValue, TransactionError>;

enum TransactionValue {
    Success,
    Code(u8),
}

enum TransactionError {
    Panic,
    UnknownFailure,
    Code(u8)
}
```

matklad: @DarkEld3r yep, using `enum` for TransactionValue looks great! We might want to add `#[doc(hidden)] __non_exhaustive` if we think that we might want to extend `u8` to something else in the future.

vitvakatu:
I'm voting for `transaction::Error` naming strategy. As far as I can see it have got some popularity among Rust community last time.

DarkEld3r:
I also prefer to use namespaces instead of more verbose (type) names, but in our case there can be several errors/results/contexts in the same scope, so they should have different names anyway. This way, I propose to stick with current naming for now and improve ergonomics of names and imports in a separate pull request.

matklad:
>I also prefer to use namespaces instead of more verbose (type) names, but in our case there can be several errors/results/contexts in the same scope, so they should have different names anyway. 

The nice thing about `transaction::Result` is that the user gets to decide if they

```
use transaction::Result;
fn foo() -> Result { }
```

or 

```
use transaction::{self};
fn foo() -> transaction::Result { }
```

>This way, I propose to stick with current naming for now and improve ergonomics of names and imports in a separate pull request.

➕ 1️⃣

DarkEld3r:
OK, now it is ready for review (again), please take a look. 😅

Quick summary:
- `Transaction::execute` returns `TransactionResult`: alias for `Result<TransactionValue, TransactionError>`.
- `TransactionValue` is `Success` or user defined code (`u8`).
- `TransactionError` is `Panic`, `UnknownFailure` or user defined code (`u8`).

This is relatively simple, but leaves enough flexibility for Exonum users to build business logic on top of that.

Possible improvements:
- Hide `TransactionError::Panic` from the user. This can be done by introducing two types of errors: one for result of `Transaction::execute` ("user error") and one for `Schema::transaction_results`. As for me, it isn't worth the effort because it complicates implementation without introducing significant advantages: even if user returns TransactionError::Panic from the transaction it will not break logical correctness (we don't and probably shouldn't rely on such statuses in the core).
- Replace `TransactionValue::Code` and `TransactionError::Code` by `StorageValue`. Although it looks like a good idea, it cannot be done more or less easily. This requires some kind of "id" (similar to the transaction `ID`) which is controversial and leaves space to mistakes. 

According to the above, I propose to stay with the current implementation at least for now. 😫

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,235 @@
...
+use encoding::serialize::json::ExonumJson;
+
+// There are three `0...255` ranges for each category:
+// - `0...255` - special values such as `TransactionValue::Success` or `TransactionError::Panic`.
+// - `256...511` - `TransactionValue::Code`.
+// - `512...767` - `TransactionError::Code`.
+const MIN_SUCCESS: u16 = 256;
+const MAX_SUCCESS: u16 = 511;
+const MIN_ERROR: u16 = 512;
+const MAX_ERROR: u16 = 767;
```
matklad: An alternative would be to write:

```
const MIN_SUCCESS: u16 = 256;
const MAX_SUCCESS: u16 = MIN_SUCCESS + 255;
const MIN_ERROR: u16 = MAX_SUCCESS + 1;
const MAX_ERROR: u16 = MIN_ERROR + 255;
```

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// # fn main() {}
+    fn execute(&self, fork: &mut Fork) -> TransactionResult;
+}
+
+/// Result of successful transaction execution.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TransactionValue {
+    /// Successful transaction execution.
+    Success,
+    /// User defined code. Can have different meanings for different transactions and services.
```
matklad: Nit: should it be `User-defined`?

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,235 @@
...
+// - `512...767` - `TransactionError::Code`.
+const MIN_SUCCESS: u16 = 256;
+const MAX_SUCCESS: u16 = 511;
+const MIN_ERROR: u16 = 512;
+const MAX_ERROR: u16 = 767;
+
+// Special reserved values.
+const TRANSACTION_SUCCESS: u16 = 0;
+const TRANSACTION_PANIC: u16 = 1;
+const TRANSACTION_FAILURE: u16 = 2;
```
matklad: Nice encoding to prevent users from returning build-in statuses! 👍

exonum/src/blockchain/transaction.rs
slowli: Could you please specify an example of several successful statuses (probably, as an example section in the `TransactionValue` docs)? If I understand correctly,they could arise if the transaction logic contains branching execution paths; however, I'm having trouble understanding what `Success` means in this case.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,235 @@
...
+    Code(u8),
+}
+
+/// Result of unsuccessful transaction execution.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TransactionError {
+    /// Panic occurred during transaction execution. This status should not be used explicitly.
+    Panic,
+    /// General failure (unspecified reason).
+    UnknownFailure,
```
slowli: Could you please provide an example where this status would be used? I'm having trouble understanding its use cases.

exonum/src/blockchain/schema.rs
```diff
@@ -66,46 +90,51 @@ where
 
     /// Returns table that represents a map from transaction hash into raw transaction message.
     pub fn transactions(&self) -> MapIndex<&T, Hash, RawMessage> {
-        MapIndex::new("core.transactions", &self.view)
+        MapIndex::new(TRANSACTIONS, &self.view)
+    }
+
+    /// Returns table that represents a map from transaction hash into execution result.
+    pub fn transaction_results(&self) -> MapIndex<&T, Hash, TransactionResult> {
```
slowli: Do I understand correctly that transaction results are not committed to the blockchain state? Would be great if they were - would make lying to lightweight clients about a status of a particular transaction more difficult.

DarkEld3r:
We had one more discussion today. I'm going to create a confluence page for further discussion, because we still can get to consensus. 😢 Lets move the discussion there.

I didn't pay enough attention to the documentation and there is a (good) reason for that. This implementation has already been changed several times and there will be more changes. I don't want to rewrite the documentation each time. 😄 

Unfortunately, associated types [will not work](https://play.rust-lang.org/?gist=6419100483726c4a19a6bc5618a667d0&version=stable): we are storing transactions as trait-objects (`Box<Transaction>`), so associated types must be specified and should be the same for every transaction.

exonum/src/blockchain/transaction.rs
DarkEld3r: It is just a predefined status for "general failure" that can be used when user doesn't care about more detailed errors.

exonum/src/blockchain/transaction.rs
DarkEld3r: Right now I don't have a good example of several different successful statuses, but I suppose that they are needed because we are making a general-purpose (blockchain) framework.

As for `Success` - it is a simple predefined success status, same as `UnknownFailure`.

exonum/src/blockchain/tests.rs
```diff
@@ -19,9 +19,9 @@ use std::collections::BTreeMap;
 use rand::{thread_rng, Rng};
 use serde_json;
 
-use blockchain::{Blockchain, Schema, Transaction};
+use blockchain::{Blockchain, Schema, Transaction, TransactionResult, TransactionValue};
 use crypto::{gen_keypair, Hash};
-use storage::{Database, Error, Fork, ListIndex};
+use storage::{Database, Fork, Error, ListIndex};
```
alexander-irbis: Why? 🙃

alekseysidorov: `rls` sorts entries in alphabetical order.

vitvakatu: It's not alphabetical order!

DarkEld3r: Honestly, I have no idea. :smile:

aleksuss: Rustfmt ?

slowli:
@DarkEld3r Just to clarify: Associated types and/or constants may work if we will use a special datatype for boxed transactions instead of `Box<Transaction>`, which would be returned from `Service::tx_from_raw`. See [this snippet](https://play.rust-lang.org/?gist=8d5bde71a5edf67a79e5af9fee2e83a0&version=undefined) for a (very rough and incomplete) proof of concept. I'm *very* unsure if it would be worth the trouble though, and whether it would be a good architectural decision; `Arc`s and boxed `Fn`s probably create a significant overhead, and the whole construction seems fragile relative to the changes in the `Transaction` interface.

DarkEld3r:
https://wiki.bf.local/display/EXN/Transaction+status

matklad:
@slowli there's a simpler and more efficient way to box transaction with associated types: https://play.rust-lang.org/?gist=39eb6e89127012399af737ace10d73b0&version=stable

slowli:
@matklad Yeah, I thought there ought to be a better way than reinventing dynamic dispatch 🙃 Do you think something like that could be viable, or is this the waste of time?

matklad:
Who needs objects and virtual tables when one has closures 🙃 ?

I don't think I have an opinion, informed by experience, whether associated types here are worth the effort. 

However, I do think that it's generally better to solve the problem only *after* all the requirements and constraints are known, so, if we don't have an immediate use-case for rich errors or rich results, I would suggest to go with the minimal implementation: `-> Result<TxOk, TxErr>`, where `TxOk` and `TxErr` are ZSTs which don't expose internal representation at all and only construable with `new`.

Here is another, even more philosophical idea :) `Transaction` is a part of the boundary between framework's code and user's code, so it has two sides: user defines methods, framework calls them. And this creates the tension between the needs of the implementers of the framework and the needs of the users, because the same methods have to cater for both use cases. A nice solution to this problem is the NVI pattern, which is essentially what both our playgrounds do:

```
pub trait Transaction { // pub, user implements this
    fn verify(&self) -> bool;
    fn execute(&self, fork: &mut Fork);
}

pub(crate) Tx { // pub(crate), this is what we use everywhere
    inner: Box<Transaction>, // Note: there's one place to switch `Box` to `Arc`
}

impl Tx {
    // Note: the interface does not need to be exactly the same
    pub fn verify(&self) -> bool { self.inner.verify() }
    pub fn execute(&self, fork: &mut Fork) { self.inner.execute() }
}
```

TLDR: I have **no idea** what it the best solution for results, however, I think it might be a generally good idea to hide all dynamic dispatch behind a single concrete type internally, and have more implementation flexibility down the line.

exonum/benches/block.rs
```diff
@@ -122,12 +124,13 @@ mod tests {
...
 
-            fn execute(&self, view: &mut Fork) {
-                let mut index = ProofMapIndex::new("balances_txs", view);
+            fn execute(&self, fork: &mut Fork) -> TransactionResult {
+                let mut index = ProofMapIndex::new("balances_txs", fork);
                 let from_balance = index.get(self.from()).unwrap_or(0u64);
                 let to_balance = index.get(self.to()).unwrap_or(0u64);
                 index.put(self.from(), from_balance - 1);
                 index.put(self.to(), to_balance + 1);
+                Ok(TransactionValue::Success)
```
alekseysidorov: In my opinion `TransactionValue` looks somewhat incomprehensible for service developers.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Takes the current blockchain state via `fork` and can modify it if certain conditions
+    /// are met.
+    ///
+    /// # Notes
+    ///
+    /// - When programming `execute`, you should perform state-related checks before any changes
+    /// to the state and return early if these checks fail.
+    /// - If the execute method of a transaction raises a `panic`, the changes made by the
+    /// transactions are discarded, but the transaction itself is still considered committed.
+    /// - A transaction execution status (see `ExecutionStatus` for the details) is stored in the
```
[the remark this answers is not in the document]
alekseysidorov: in my opinion `ExecutionStatus` is more clear than `TransactionResult`.

DarkEld3r:
Please review one more time. :sweat_smile: Quick summary:
- Separate types for `Transaction::execute()` and `Schema::transaction_statuses` (user cannot set `Panic` status).
- `Ok(())` as result of successful execution.
- String error description can be used.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,398 @@
...
+// `Err(TransactionError::Panic)`.
+const TRANSACTION_STATUS_PANIC: u16 = TRANSACTION_STATUS_OK + 1;
+// `Err(TransactionError::Failure)`.
+const TRANSACTION_STATUS_FAILURE: u16 = TRANSACTION_STATUS_PANIC + 1;
+// `Err(TransactionError::Description)`.
+const TRANSACTION_STATUS_DESCRIPTION: u16 = TRANSACTION_STATUS_FAILURE + 1;
+
+/// Return value of the `Transaction`'s `execute' method. Changes made by the transaction are
+/// discarded if `Err` is returned, see `Transaction` documentation for the details.
+pub type ExecutionStatus = Result<(), ExecutionError>;
```
matklad: It's a bit strange that `Result` type does not have "Result" in the name.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,398 @@
...
+/// Result of unsuccessful transaction execution.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ExecutionError {
+    /// General failure (unspecified reason).
+    Failure,
+    /// User-defined error-code. Can have different meanings for different transactions and
+    /// services.
+    Code(u8),
+    /// User-defined string error description.
+    Description(String),
```
matklad: Having both `Code` and `Description` variants seems simultaneously overly restrictive and overly general. Like, you can't do much with code, but you can store arbitrary JSON in description if you want too...

vitvakatu: I suggest `Description(Box<Error>)`

exonum/src/blockchain/transaction.rs
DarkEld3r: I'm not so sure myself. On the one hand, I simply don't like such typedefs: I prefer to have `Result` in the appropriate namespace. On the other hand, forcing user to write something like `use exonum::transaction::Result as TransactionResult` isn't good either.

We need to seriously revise our imports anyway, but it can happen not so soon.

Lets use your comment as a little poll: :+1: for `Result` and :-1: for `Status`.

exonum/src/blockchain/transaction.rs
DarkEld3r: @matklad 
> Like, you can't do much with code

Your can store (almost) anything in the `MapIndex`. :wink: 

The original idea was the following: `Code` gives you enough flexibility to organize the error handling as complex as you want. Additionally, it is easier to deal with codes having versioning in mind. String descriptions are quite the opposite: they are nice when you only need to report error to user, but are harder to maintain. I think that both variants can be useful.

exonum/CHANGELOG.md
```diff
@@ -18,9 +18,13 @@ All notable changes to this project will be documented in this file. The project
   around a type (e.g., `struct QuirkyI32Key(i32)`) and implement `StorageKey` for it
   using big endian encoding. Then, use the wrapper instead of the int type in indices.
   See the unit tests for `StorageKey` for an example. (#443)
+- `Transaction` `execute` method now returns  `TransactionStatus` that is stored in the blockchain and can be accessed
```
slowli: `ExecutionResult` instead of `TransactionStatus`? Also, there is an obsolete space before `TransactionStatus`; and I would use `Transaction::execute` instead of 2 separate words.

exonum/CHANGELOG.md
```diff
@@ -18,9 +18,13 @@ All notable changes to this project will be documented in this file. The project
...
+  through api. The changes made by transactions that return `Err` are discarded. To migrate, add `Ok(())` as the last
+  line to the `execute` method. (#385)
```
slowli: Nitpick: I'd add
> ...to the `execute` method. More generally, make sure that the method returns `Ok(())` on successful execution.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,398 @@
...
+}
+
+/// Converts from `ExecutionResult` into `TransactionResult`.
+pub fn convert_result(result: ExecutionResult) -> TransactionResult {
```
slowli: Why isn't this method implemented as `impl From<ExecutionResult> for TransactionResult`?

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,398 @@
...
+                Err(TransactionError::Description(
+                    String::from_bytes(Cow::Borrowed(&bytes[2..])),
+                ))
+            }
+            value => panic!("Invalid TransactionResult value: {}", value),
+        }
+    }
+}
+
+fn status_as_u16(status: &TransactionResult) -> u16 {
```
slowli: Why isn't this a private method in `TransactionResult`?

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,398 @@
...
+            let (_, patch) = blockchain.create_patch(
+                ValidatorId::zero(),
+                Height(index),
+                &[transaction.hash()],
+                &pool,
+            );
+
+            db.merge(patch).unwrap();
+
+            let mut fork = db.fork();
```
slowli: Nitpick: Using a `Fork` here looks a bit sloppy, but I suppose it's OK for tests.

testkit/tests/currency/cryptocurrency.rs
```diff
@@ -126,12 +126,13 @@ impl Transaction for TxCreateWallet {
...
 
     /// Apply logic to the storage when executing the transaction.
-    fn execute(&self, view: &mut Fork) {
+    fn execute(&self, view: &mut Fork) -> ExecutionResult {
         let mut schema = CurrencySchema { view };
         if schema.wallet(self.pub_key()).is_none() {
             let wallet = Wallet::new(self.pub_key(), self.name(), INIT_BALANCE);
-            schema.wallets_mut().put(self.pub_key(), wallet)
+            schema.wallets_mut().put(self.pub_key(), wallet);
         }
```
slowli: You could add an erroneous status signaling "wallet already exists" here. OTOH, when/if the `cryptocurrency` repo is moved into this repo, this can be done right there 😄

testkit/tests/currency/cryptocurrency.rs
```diff
@@ -144,7 +145,7 @@ impl Transaction for TxTransfer {
 
     /// Retrieve two wallets to apply the transfer. Check the sender's
     /// balance and apply changes to the balances of the wallets.
-    fn execute(&self, view: &mut Fork) {
+    fn execute(&self, view: &mut Fork) -> ExecutionResult {
```
slowli: Similarly, transfers is a good place to showcase multiple errors:
- Unknown sender
- Unknown receiver
- Insufficient funds
- Arithmetic overflow for the receiver's balance (not tested now, btw)

exonum/src/blockchain/transaction.rs
DarkEld3r: Two reasons:
1. I don't think it should be part of the public API: user should use both types separately without converting one to another.
2. Rust [doesn't allow this](https://play.rust-lang.org/?gist=3c7bdc53ebd157379316c546b677fd1d&version=stable). :smile:

exonum/src/blockchain/transaction.rs
DarkEld3r: Yeap, it is only needed to reuse the `create_entry` function.

exonum/src/blockchain/transaction.rs
DarkEld3r: I don't know. :upside_down_face: In C++ it is often considered that free functions should be preferred if possible. I'm not sure about the best practices for Rust in this case.

@aleksuss @matklad @vitvakatu @ozkriff Opinions?

matklad: - I am not sure that there are best practices for Rust for this particular case
- In C++, the rational behind the rule is that free functions do not have access to private members, and this is not applicable to Rust, which uses modules for visibility. 
- I personally generally prefer free-standing functions to methods and traits most of the times, because it the simplest tool of abstraction. That is, I prefer C-style Rust to C++-style Rust for implementation parts of code.
- However, in this particular case I would make it a method: this will allow to get rid of the `status` word from the function name :)

vitvakatu: I don't have strong opinion, but I'd prefer to see this function as method.

slowli: I've used [this](https://doc.rust-lang.org/1.0.0/style/features/functions-and-methods/README.html) as a reference.

exonum/src/blockchain/transaction.rs
slowli: Terribly sorry; I meant using `impl From<ExecutionError> for TransactionError` + `map_err(TransactionError::from)`.

exonum/src/blockchain/transaction.rs
DarkEld3r: OK, now I'm convinced, but Rust complains again:
> cannot define inherent `impl` for a type outside of the crate where the type is defined

I don't think that creating a method for `TransactionError` and reusing it in a (still needed) free function makes code better.

exonum/src/blockchain/transaction.rs
DarkEld3r: Still I don't like extending public API with such conversion: will it be useful for exonum users?.. On the one hand, it seems harmless, on the other hand, it provides more room for breaking changes.

testkit/tests/currency/cryptocurrency.rs
DarkEld3r: I have created the issue for that: https://jira.bf.local/browse/ECR-682

exonum/src/blockchain/transaction.rs
slowli: Well, it could be possible to implement a private trait to convert `TransactionResult` as needed, something like this:

```rust
trait AsCode {
    type Code: StorageValue + CryptoHash;
    fn as_code(&self) -> Self::Code;
}

impl AsCode for TransactionResult {
    type Code = u16;

    fn as_code(&self) -> u16 {
        match *self {
            Ok(()) => TRANSACTION_STATUS_OK,
            Err(TransactionError::Panic) => TRANSACTION_STATUS_PANIC,
            Err(TransactionError::Failure) => TRANSACTION_STATUS_FAILURE,
            Err(TransactionError::Code(c)) => u16::from(c),
            Err(TransactionError::Description(_)) => TRANSACTION_STATUS_DESCRIPTION,
        }
    }
}

// usage
impl CryptoHash for TransactionResult {
    fn hash(&self) -> Hash {
        self.as_code().hash()
    }
}
```

Not sure whether it would be any better than the current implementation, though.

exonum/src/blockchain/transaction.rs
slowli: I think it is an intended addition to the `ExecutionError` / `TransactionError` interface: it seems an invariant w.r.t. future changes that `ExecutionError`s are a subset of `TransactionError`s. But I don't insist on this change.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,398 @@
...
+        match *self {
+            TransactionError::Panic => write!(f, "Panic during execution"),
+            TransactionError::Failure => write!(f, "Unspecified failure"),
+            TransactionError::Code(c) => write!(f, "Error code: {}", c),
+            TransactionError::Description(ref s) => write!(f, "{}", s),
+        }
+    }
+}
+
+// String content (`TransactionError::Description`) is intentionally excluded from the hash
```
slowli: I don't find this reasoning particularly compelling. Consider that excluding a message from the hash calculation means that it could be faked for lightweight clients. Moreover, I'd say that doing this *encourages* users to use error descriptions as they won't break the blockchain.

Maybe, do the following: make a string description an optional field in `ExecutionError` (remaking it into a struct with status and this message fields). Then the message would not be committed into the hash by design, but a service dev would need to specify a status together with it to indicate what's going on and to commit *something* to the blockchain state.

exonum/src/blockchain/schema.rs
```diff
@@ -64,46 +88,51 @@ where
 
     /// Returns table that represents a map from transaction hash into raw transaction message.
     pub fn transactions(&self) -> MapIndex<&T, Hash, RawMessage> {
-        MapIndex::new("core.transactions", &self.view)
+        MapIndex::new(TRANSACTIONS, &self.view)
+    }
+
+    /// Returns table that represents a map from transaction hash into execution status.
```
slowli: Nitpick: using a consistent terminology (i.e., s/status/result/) could be better. This concerns naming of variables / methods as well.

exonum/src/blockchain/transaction.rs
DarkEld3r: Makes sense.

exonum/src/blockchain/transaction.rs
DarkEld3r: I understand your concern, but I'm not sure that the proposed solution will make a big difference: user can still use different descriptions with the same code. Am I missing something?

exonum/src/blockchain/schema.rs
DarkEld3r: Actually, it isn't a nitpick: it is the result of several renamings and it should be fixed. :+1:

exonum/src/blockchain/transaction.rs
DarkEld3r: Discussed: I'm convinced now.

services/time/src/lib.rs
```diff
@@ -115,22 +116,22 @@ impl Transaction for TxTime {
...
+    fn execute(&self, view: &mut Fork) -> ExecutionResult {
         let validator_keys = Schema::new(&view).actual_configuration().validator_keys;
 
         // The transaction must be signed by the validator.
         let signed = validator_keys.iter().any(|&validator| {
             validator.service_key == *self.pub_key()
         });
         if !signed {
-            return;
+            return Ok(());
```
slowli: This calls for an erroneous return ("signed by an invalid author"). Could be dealt with later.

DarkEld3r:
Optional string description has been redesigned slightly, please review.

services/time/src/lib.rs
```diff
@@ -115,22 +116,22 @@ impl Transaction for TxTime {
...
         }
 
         let mut schema = TimeSchema::new(view);
         match schema.validators_times().get(self.pub_key()) {
             // The validator time in the storage should be less than in the transaction.
             Some(storage_time) if storage_time >= self.time() => {
-                return;
+                return Ok(());
```
slowli: Now, this case is interesting. On the one hand, it is sort of unexpected that the storage time of a validator decreases. OTOH, it's not obvious that such a transaction should not put changes into the storage.

services/time/src/lib.rs
```diff
@@ -158,21 +159,23 @@ impl Transaction for TxTime {
...
+            return Ok(());
         }
         // Ordering time from highest to lowest.
         validator_times.sort_by(|a, b| b.cmp(a));
 
         match schema.time().get() {
             // Selected time should be longer than the time in the storage.
             Some(current_time) if current_time >= validator_times[max_byzantine_nodes] => {
-                return;
+                return Ok(());
```
slowli: Nitpick: it seems this `return` could be safely changed to no-op.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,490 @@
...
+/// # use exonum::blockchain::{ExecutionError, TransactionError};
+///
+/// # let transaction_error: TransactionError = ExecutionError::new(1).into();
+/// // Prints user friendly error description.
+/// println!("Transaction error: {}", transaction_error);
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TransactionError {
+    /// Error type, see `TransactionErrorType` for the details.
+    error_type: TransactionErrorType,
```
slowli: IMO, naming this field `type` could be better; it avoids stuttering.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,490 @@
...
+        Self::panic(None)
+    }
+
+    /// Returns error type of this `TransactionError` instance.
+    pub fn error_type(&self) -> TransactionErrorType {
+        self.error_type
+    }
+
+    /// Returns an optional error description.
+    pub fn description(&self) -> &Option<String> {
```
slowli: Nitpick: maybe, it makes sense to convert this to `Option<&str>` by using `Option::as_ref()` method?

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,490 @@
...
+///   implementation for the details).
+/// - `TransactionErrorType::Panic` is set by the framework if panic is raised during transaction
+///   execution.
+/// - `TransactionError` implements `Display` which can be used for obtaining a simple error
+///   description.
+///
+/// # Examples
+///
+/// ```
+/// # use exonum::blockchain::{ExecutionError, TransactionError};
```
slowli: You sure this example is appropriate? Unless I'm missing something, it would be rendered as a single line of code telling very little how `TransactionError` can be used.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,490 @@
...
+    }
+}
+
+// `TransactionResult` is stored as `u16` plus `bool` (`true` means that optional part is present)
+// with optional string part needed only for string error description.
+impl StorageValue for TransactionResult {
+    fn into_bytes(self) -> Vec<u8> {
+        let mut res = u16::into_bytes(status_as_u16(&self));
+        if let Some(description) = self.err().and_then(|e| e.description) {
+            res.extend(bool::into_bytes(true));
```
slowli: This seems a bit overboard, but it seems that the presence / absence of the description could be packed into the first 2 bytes as well. Is a separate byte used for future compatibility (e.g., the desired serialization of `Option<T>`)?

exonum/src/blockchain/mod.rs
```diff
@@ -254,22 +257,35 @@ impl Blockchain {
...
 
-                let r = panic::catch_unwind(panic::AssertUnwindSafe(|| { tx.execute(&mut fork); }));
+                let catch_result =
+                    panic::catch_unwind(panic::AssertUnwindSafe(|| tx.execute(&mut fork)));
 
-                match r {
-                    Ok(..) => fork.commit(),
+                let transaction_result = match catch_result {
+                    Ok(execution_result) => {
+                        if execution_result.is_err() {
```
aleksuss: I propose to log an error from `execution_result`

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,490 @@
...
+    fn transaction_results_round_trip() {
+        let results = [
+            Ok(()),
+            Err(TransactionError::panic(None)),
+            Err(TransactionError::panic(Some("".to_owned()))),
+            Err(TransactionError::panic(
+                Some("Panic error description".to_owned()),
+            )),
+            Err(TransactionError::code(0, None)),
+            Err(TransactionError::code(
```
slowli: Nitpick: Would it make sense to introduce an auxiliary function returning `Err(TransactionError::code(x, Some(s.to_owned())))` based on `x` and `s`? All these parentheses make me a bit uncomfortable.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,490 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TransactionErrorType {
+    /// Panic occurred during transaction execution.
+    Panic,
+    /// User-defined error-code. Can have different meanings for different transactions and
+    /// services.
+    Code(u8),
+}
+
+/// Extended by the framework result of unsuccessful transaction execution.
```
slowli: I'd suggest to rearrange 2 parts of this sentence:

> Result of unsuccessful transaction execution encompassing both service and framework-wide error handling.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,490 @@
...
+        }
+    }
+}
+
+/// Type of the transaction error.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TransactionErrorType {
+    /// Panic occurred during transaction execution.
+    Panic,
+    /// User-defined error-code. Can have different meanings for different transactions and
```
slowli: Nitpick: s/error-code/error code/.

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,490 @@
...
+        Self {
+            code,
+            description: Some(description),
+        }
+    }
+}
+
+/// Type of the transaction error.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TransactionErrorType {
```
slowli: I don't quite like such a long name for a type, but can't think of a better alternative rn. Probably, something like
```rust
pub enum ErrorCode {
    Panic,
    User(u8),
}
```

but the `User` variant name is surely not the best one.

services/time/src/lib.rs
DarkEld3r: There already is an issue for that: https://jira.bf.local/browse/ECR-682
:smile:

exonum/src/blockchain/transaction.rs
DarkEld3r: `type` is a keyword and I really don't like things like `type_`.

slowli: Jeez, I sure didn't think that through... OTOH, imo `error_type` is close in its semantic to a *code* (see a comment about `TransactionErrorType` naming), so maybe that could be used?

exonum/src/blockchain/transaction.rs
DarkEld3r: I'm not sure, do you have better ideas?

exonum/src/blockchain/transaction.rs
DarkEld3r: Packing this flag was my first intention, but I don't like code complication. I prefer to wait for support of enums in serialization and then update this implementation.

exonum/src/blockchain/transaction.rs
DarkEld3r: I don't like that name, either, but it should not be often used directly, especially in the user code.

exonum/src/blockchain/transaction.rs
slowli: Well, the most complete example I can think of is demonstrating how an `ExecutionError` returned by the service can then be read from the core schema; something like

```rust
message! {
    struct ErrTransaction { }
}

impl Transaction for ErrTransaction {
    // ...
    fn execute(&self, _: &mut Fork) -> ExecutionResult {
         Err(ExecutionError::with_description(1, "Gosh!"))
    }
}

let mut blockchain = ...;
let patch = blockchain.create_patch(...);
blockchain.merge(patch);

let schema = Schema::new(blockchain.snapshot());
let tx_result = schema.transaction_results().get(tx.hash()).unwrap();
let stored_err = tx_result.unwrap_err();
assert_eq!(stored_err.error_type(), TransactionErrorType::Code(1));
assert_eq!(stored_err.description(), Some("Gosh"));
```

Obviously, to make this example actually compile, some hidden code is needed.

exonum/src/blockchain/transaction.rs
DarkEld3r: Personally, I like `error_type` more, lets see other comments if there are any.

vitvakatu: I don't think naming of a single private field can really be a problem.

slowli: There is also a public getter for the field and the type of the field, but I agree that it's not that important.

exonum/src/blockchain/tests.rs
```diff
@@ -19,9 +19,9 @@ use std::collections::BTreeMap;
 use rand::{thread_rng, Rng};
 use serde_json;
 
-use blockchain::{Blockchain, Schema, Transaction};
+use blockchain::{Blockchain, Schema, Transaction, ExecutionResult};
 use crypto::{gen_keypair, CryptoHash, Hash};
-use storage::{Database, Error, Fork, ListIndex};
+use storage::{Database, Fork, Error, ListIndex};
```
skletsun: That strange sorting order again...

exonum/src/blockchain/transaction.rs
```diff
@@ -238,8 +252,8 @@ impl TransactionError {
     }
 
     /// Returns an optional error description.
-    pub fn description(&self) -> &Option<String> {
-        &self.description
+    pub fn description(&self) -> Option<&str> {
+        self.description.as_ref().map(|d| d.as_ref())
```
slowli: Would it make sense to use `.map(String::as_ref)` instead of `.map(|d| d.as_ref())`? I think clippy had suggested the first version previously, but it seems it has stopped to.

exonum/src/blockchain/transaction.rs
```diff
@@ -343,6 +357,54 @@ mod tests {
...
+            (TransactionErrorType::Code(0), None),
+            (TransactionErrorType::Code(1), Some("".to_owned())),
+            (TransactionErrorType::Code(100), None),
+            (
+                TransactionErrorType::Code(255),
+                Some("error description".to_owned()),
+            ),
+        ];
+
+        for value in &values {
```
slowli: Would it make sense to override the test as follows?

```rust
let values = [
    (TransactionErrorType::Panic, None),
    (TransactionErrorType::Panic, Some("panic")),
    (TransactionErrorType::Code(0), None),
    (TransactionErrorType::Code(1), Some("")),
    (TransactionErrorType::Code(100), None),
    (
        TransactionErrorType::Code(255),
        Some("error description"),
    ),
];

for value in values.into_iter() {
    let error = TransactionError::new(value.0, value.1.clone().map(str::to_owned));
    assert_eq!(value.0, error.error_type());
    assert_eq!(value.1, error.description());
}
```

I think it's marginally more readable, but I'm not insisting. Other tests could aget rid of `.to_owned()` in a similar way as well.

exonum/src/blockchain/transaction.rs
```diff
@@ -343,6 +357,54 @@ mod tests {
...
+            assert_eq!(None, error.description);
+        }
+    }
+
+    #[test]
+    fn execution_error_with_description() {
+        let values = [
+            (0, "".to_owned()),
+            (1, "test".to_owned()),
+            (100, "errro".to_owned()),
```
slowli: Misspelling? (Already pointed out by CI.)

exonum/src/blockchain/transaction.rs
```diff
@@ -343,6 +357,54 @@ mod tests {
         static ref EXECUTION_STATUS: Mutex<ExecutionResult> = Mutex::new(Ok(()));
     }
 
+    #[test]
+    fn execution_error_new() {
+        let codes = [0, 1, 100, 255];
+
+        for code in &codes {
```
slowli: Is there a reason to not use `for code in codes.into_iter()` or `for &code in &codes`?

exonum/src/blockchain/transaction.rs
```diff
@@ -0,0 +1,552 @@
...
+/// # use exonum::storage::{MemoryDB, Database};
+/// # use exonum::crypto::Hash;
+/// use exonum::blockchain::Schema;
+///
+/// # let db = MemoryDB::new();
+/// # let snapshot = db.snapshot();
+/// # let transaction_hash = Hash::zero();
+/// let schema = Schema::new(&snapshot);
+///
+/// if let Some(result) = schema.transaction_results().get(&transaction_hash) {
```
slowli: This example is awesome 👍

Issue #15, opened 2017-04-04
Consensus on the threshold of 1/3 sleeping validators


DarkEld3r:
After merging #6 request (#2 issue), we are going to have a problem with consensus on the threshold of 1/3 sleeping validators. Proposed solution: if 1/3 validators send messages for round R or higher, then validators from lower rounds should jump to round R.

PR #388, opened 2017-12-04
Fixed problem with consensus on 1/3 validators sleeps.

vldm:
Closes https://github.com/exonum/exonum/issues/15


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fixed `crate_authors!` macro usage, this macro can't return static string in new clap version. (#370)
 - Fixed mistake in description of the height getter in the `ServiceContext`. (#366)
+- Fixed #15 consensus on the threshold of 1/3 sleeping validators. (#388)
```
DarkEld3r: I like the idea of having reference to extended description (such as issue), but I am not sure it will work with Jira.

exonum/src/node/consensus.rs
```diff
@@ -47,7 +47,15 @@ impl NodeHandler {
                 self.state.height(),
                 self.state.round()
             );
+            let validator = msg.validator();
+            let round = msg.round();
             self.state.add_queued(msg);
+            trace!("Trying reach actual round.");
```
DarkEld3r: ...**to** reach...

exonum/src/node/consensus.rs
```diff
@@ -47,7 +47,15 @@ impl NodeHandler {
...
+            if let Some(r) = self.state.reach_actual_round(validator, round) {
+                info!("Jump to new round = {}", round);
```
DarkEld3r: ...to **a** new...

exonum/src/node/consensus.rs
```diff
@@ -581,7 +589,26 @@ impl NodeHandler {
             self.has_full_propose(hash, round);
         }
     }
+    /// Handles round update
```
DarkEld3r: I would love to see an empty line between methods. :smile:

exonum/src/node/state.rs
```diff
@@ -564,6 +567,34 @@ impl State {
         ValidatorId(((height + round) % (self.validators().len() as u64)) as u16)
     }
 
+    /// Update known round for a validator, try reach actual round
+    /// returns new actual round
+    pub fn reach_actual_round(&mut self, id: ValidatorId, round: Round) -> Option<Round> {
```
DarkEld3r: As far as I can see, this function doesn't actually "try to reach actual round", so I propose to rewrite the comment the following way:

Updates known round for a validator and returns a new actual round if byzantine majority of validators are on a higher round. Otherwise returns `None`.

Perhaps rename function too? Something like `get_actual_round`.

CHANGELOG.md
vldm: This is github issue

DarkEld3r: I know, this isn't a problem right now, but perhaps we need some solution for future.

vldm: I prefer to link only github issues, that would be for outside users.

exonum/src/node/state.rs
```diff
@@ -575,6 +575,13 @@ impl State {
             let known_round = self.validators_rounds.entry(id).or_insert_with(Round::zero);
             if round <= *known_round {
                 // keep only maximum round
+                trace!(
+                    "Received message with round older that we know,\
```
DarkEld3r: I propose to rephrase as "Received a message from a higher round than we know".

exonum/src/node/consensus.rs
```diff
@@ -590,8 +591,24 @@ impl NodeHandler {
...
+        if round <= self.state.round() {
+            return;
+        }
+
+        info!("Jump to a new round = {}", round);
+        self.state.jump_round(round);
+        self.process_new_round();
+    }
+
+    // Try to process consensus messages from future round.
```
DarkEld3r: ...from **the** future round.

exonum/src/node/mod.rs
```diff
@@ -175,6 +177,7 @@ impl Default for EventsPoolCapacity {
             network_events_capacity: 512,
             timeout_requests_capacity: 128,
             api_requests_capacity: 1024,
+            internal_events_capacity: 128,
```
alekseysidorov: Please add `BREAKING CHANGE` badge. And add comment about these parameter to changelog.

exonum/src/events/mod.rs
```diff
@@ -17,7 +17,7 @@ pub mod tests;
 pub mod codec;
 pub mod error;
 pub mod network;
-pub mod timeouts;
+pub mod intrenal;
```
DarkEld3r: Typo in the file name. :smile:

defuz: in**tre**nal

PR #392, opened 2017-12-11
Allow to use borrowed types for some indexes operations [ECR-700]

DarkEld3r:


exonum/src/storage/base_index.rs
```diff
@@ -112,7 +112,7 @@ where
     /// Returns a value of *any* type corresponding to the key of *any* type.
     pub fn get<K, V>(&self, key: &K) -> Option<V>
     where
-        K: StorageKey,
+        K: StorageKey + ?Sized,
```
defuz: If you define `StorageKey` as 

```
pub trait StorageKey: ToOwned + ?Sized
```

so you don't need to add `+ ?Sized` every time

exonum/src/storage/keys.rs
```diff
@@ -70,131 +74,131 @@ pub trait StorageKey {
...
     /// Deserialize a key from the specified buffer of bytes.
     // TODO: should be unsafe (ECR-174)?
-    fn read(buffer: &[u8]) -> Self;
+    fn read(buffer: &[u8]) -> Self::Owned;
 }
 
 impl StorageKey for () {
     fn size(&self) -> usize {
-        0
+        mem::size_of::<Self>()
```
defuz: It looks like the right change (we get rid of the magic constants), but in fact it is logically incorrect. The function `mem::size_of` is depends on architecture. For example, you can imagine some strange architecture on which u16 always takes 3 bytes (why not?).  Hovewer, `BigEndian::write_u16(buffer, *self)` always writes 2 bytes, regardless of architecture.

Since the amount of memory used by different types during serialization is part of the *agreement of the protocol*, it is correct to use constants rather than information about how many bytes these types occupy in memory on a given architecture.

exonum/src/storage/base_index.rs
DarkEld3r: Unfortunately:
```rust
error: `?Trait` is not permitted in supertraits
  --> exonum/src/storage/keys.rs:62:34
   |
62 | pub trait StorageKey: ToOwned + ?Sized {
   |                                  ^^^^^
   |
   = note: traits are `?Sized` by default
```

exonum/src/storage/keys.rs
vldm: @defuz https://doc.rust-lang.org/std/mem/fn.size_of.html
>In general, the size of a type is not stable across compilations, but specific types such as primitives are.

DarkEld3r: @vldm Interesting. 👍  There is even a nice table by the link. You think that `size_of` usage in this case is valid, right?

vldm: @DarkEld3r nope :) This link was just for info about rust.
I am pretty sure, that size of value is highly related with `write` \ `read` implementation, so it should be known and probably finite.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Changed iterators over `Patch` and `Changes` data into custom types instead of standard collection iterators. (#393)
 - Fixed typo in `SparceListIndexKeys` and `SparceListIndexValues` (#398)
+- Changes key-indexes interface allowing to use borrowed types for the search operations. (#392)
```
vitvakatu: Changed? It's not obvious for me what tense should we use in changelog.

DarkEld3r: Sorry, I added the reviewers too soon, but now it should be ready. :sweat_smile:

exonum/src/storage/base_index.rs
```diff
@@ -261,6 +261,8 @@ fn assert_valid_name<S: AsRef<str>>(name: S) {
 mod tests {
     use super::*;
 
+    static INDEX_NAME: &str = "test_index_name";
```
slowli: FYI: this string is seemingly unused. I wonder if there is a way to make the CI builds fail because of this.

PR #393, opened 2017-12-11
Add 'patch' method to the 'Fork' structure

DarkEld3r:
Funny thing: in the #371 pull request I actually forgot to add getter for `Patch`. 😅

### DIFF
--- a/exonum/src/storage/db.rs
+++ b/exonum/src/storage/db.rs
@@ -17,7 +17,7 @@ use std::collections::btree_map::{BTreeMap, Range, Iter as BtmIter, IntoIter as 
 use std::collections::hash_map::{Iter as HmIter, IntoIter as HmIntoIter, Entry as HmEntry};
 use std::collections::Bound::*;
 use std::cmp::Ordering::*;
-use std::iter::Peekable;
+use std::iter::{Peekable, Iterator as StdIterator};
 
 use super::Result;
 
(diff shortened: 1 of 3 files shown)


exonum/src/storage/db.rs
```diff
@@ -431,6 +431,11 @@ impl Fork {
         self.patch
     }
 
+    /// Returns reference to the inner `Patch`.
+    pub fn path(&self) -> &Patch {
```
dmitry-timofeev: It's supposed to be pat**c**h`, isn't it?

DarkEld3r:
@defuz Probably, I should add wrappers for iterators, right?

DarkEld3r:
One more thing to discuss: we have `Iterator` trait in the storage, so its name is the same as for the standard one (`std::iter::Iterator`). Is that a problem?

On the one hand, it is a little confusing and not very convenient to use something like `use std::iter::Iterator as StdIterator;`.
On the other hand, many crates have their own `Error` or `Result` definitions.

Opinions?

ozkriff:
> it is a little confusing and not very convenient to use something like use `std::iter::Iterator as StdIterator;`.

And it may cause some strange error messages I think.

> On the other hand, many crates have their own `Error` or `Result` definitions.

`std::error::Error` trait is not in the prelude and non-std `Result`s and `Error`s are usually just structures or aliases, not traits.

Issue #109, opened 2017-05-17
Update CONTRIBUTING.md with code conventions


DarkEld3r:
Although we use rust-fmt, nevertheless there are conventions that cannot be applied automatically, so we need to document them. Starting points:
- http://aturon.github.io/style/naming/README.html
- https://github.com/exonum/exonum-core/pull/30#issuecomment-293910368
- https://github.com/exonum/exonum-core/pull/84#issuecomment-298358528
- Don't use `_` in public APIs, instead use full variable names and `#[allow(unused_variables, unused_mut)]`

DarkEld3r:
This issue includes more tasks:
- Add a file with developer notes, similar to https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md (#8).
- Fix link and TODO in CONTRIBUTING.md (#8).
- Add information about Clippy and rusfmt (#8).

PR #397, opened 2017-12-12
Reorder imports (treat all 'extern create' as thirdparty)

DarkEld3r:
Motivation: we had a (not described formally) convention to separate usings into three groups: first thirdparty, then `std` and finally the internal ones. I propose to treat all `external crate` usings as thirdparty just because it is a formal criteria.

The order is discussable too.

### DIFF
--- a/exonum/benches/block.rs
+++ b/exonum/benches/block.rs
@@ -23,6 +23,8 @@ extern crate test;
 
 #[cfg(test)]
 mod tests {
+    use std::collections::BTreeMap;
+
     use tempdir::TempDir;
     use futures::sync::mpsc;
     use test::Bencher;
@@ -33,8 +35,6 @@ mod tests {
     use exonum::messages::Message;
     use exonum::helpers::{Height, ValidatorId};
     use exonum::node::ApiSender;
-
-    use std::collections::BTreeMap;
 
     fn create_blockchain(db: Box<Database>) -> Blockchain {
         let dummy_channel = mpsc::channel(1);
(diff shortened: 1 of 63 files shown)


alekseysidorov:
What do you think about order such this?
```rust
use std;
use foreign_crates;
use exonum_crates;
use current_crate;
```

aleksuss:
Let's be closer to community convention ?
```rust
use third_party_crates;

use std;

use current_crate;

pub use current_crate;
```
P.S. third_party_crates include exonum's crates as external dependencies

DarkEld3r:
I like this idea, especially because it seems that `rust-fmt` [probably](https://github.com/rust-lang-nursery/fmt-rfcs/issues/24) will probably enforce a similar order.

DarkEld3r:
Discussed, the following order should be used:
```rust
use std;
...

use extern_crate;
...

use inner;
...

pub use something;
```

Motivation: it is a popular ordering in other crates and in the [rust-fmt RFC](https://github.com/rust-lang-nursery/fmt-rfcs/issues/24).
It can be convenient to separate our (exonum) extern crates from third-party crates, but such rule cannot be (easily) automated.

aleksuss:
It'd be grate if you save this somewhere in a public place

DarkEld3r:
@aleksuss We have an issue (https://github.com/exonum/exonum/issues/109 and in JIRA too) for `CONTRIBUTING.md`/`developer-notes.md`. :trollface:

PR #399, opened 2017-12-15
Remove default 'state_hash' implementation

DarkEld3r:
Motivation: default implementation can confuse users. Implementing `Service` trait it is easily to forget about `state_hash` method because it is already implemented. Without the default implementation user will get a compilation error and and this will force him look at the documentation.

Disadvantages: minimal service code becomes a little more "bloated".


slowli:
I think this is an improvement. But frankly speaking, I find it a bit confusing that tables in the storage are identified by string keys, but are expected to be linearized into a vector for `state_hash()`. May just be me, though.

PR #401, opened 2017-12-21
WIP: health check

ozkriff:


exonum/src/api/public/system.rs
```diff
@@ -12,15 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::time::SystemTime;
+use std::collections::HashMap;
+
+use crypto::PublicKey;
```
DarkEld3r: Module `crypto` is part of Exonum, so it should be in the next section. :wink:

ozkriff: Done

vldm:
Can you provide some info about pull request:
(description, examples, and how we could test it will be enought).

exonum/src/blockchain/service.rs
```diff
@@ -304,24 +310,32 @@ impl SharedNodeState {
...
+
     /// Update internal state, from `Node` State`
     pub fn update_node_state(&self, state: &State) {
-        for (p, c) in state.peers().iter() {
-            self.state
-                .write()
-                .expect("Expected write lock.")
-                .peers_info
-                .insert(c.addr(), *p);
+        for (&public_key, c) in state.peers().iter() {
```
alekseysidorov: Is not the `state.peers().values()` working here?

exonum/src/blockchain/mod.rs
```diff
@@ -392,7 +392,9 @@ impl Blockchain {
         mount
     }
 
-    fn api_context(&self) -> ApiContext {
+    // TODO: I need the public key - how can i get it?
+    // is it ok to make this public?
+    pub(crate) fn api_context(&self) -> ApiContext {
```
alekseysidorov: It seems that the method should be public and documented.
And later we may be add possibility to add the custom handlers to the `NodeBulder`.

DarkEld3r:
Superseded by #405.

PR #402, opened 2017-12-24
Refactor `Transaction::info` method

vldm:
### Description

Removes `Transaction::info`.
Makes `Transaction` inherit `ExonumJson`. 

- [ ] Bug fix (non-breaking change which fixes an issue)

- [x] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:
- [X] All commits are named based on our guideline (read CONTRIBUTING.md)

- [ ] Tests for the changes have been added

We can test api. For example: result of `curl http://127.0.0.1:8000/api/system/v1/transactions/44c6c2c58eaab71f8d627d75ca72f244289bc84586a7fb42186a676b2ec4626b`
looks unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Added `patch` method to the `Fork` structure. (#393)
-- Replaced config param `timeout_events_capacity` with `internal_events_capacity`. (#388)
 
 ### Changed
 - Changed iterators over `Patch` and `Changes` data into custom types instead of standard collection iterators. (#393)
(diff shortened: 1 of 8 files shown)


slowli:
- So why isn't `info()` removed? 🤔 It seems like you've removed both of its uses in the core; is it used somewhere else?
- Maybe, a stupid idea, but: use `Serialize` instead of `ExonumJson` as a supertrait for `Transaction`? It seems more transparent to me, but I'm not sure whether it would be functionally the same.

vldm:
@slowli it was experimental, thats why i didn't remember `info()`
Yes, Serialize is better, but for now, serdes serialize is not object-safe, so it couldn't be used within `Transaction`.

PR #404, opened 2017-12-25
Removed HashMap wrapper in MemoryDB

aleksuss:
### Description
Improve MemoryDB functionality 

Removed `HashMap` wrapper

### Checklist:

Unit-tests

### DIFF
--- a/exonum/src/storage/memorydb.rs
+++ b/exonum/src/storage/memorydb.rs
@@ -13,16 +13,14 @@
 // limitations under the License.
 
 //! An implementation of `MemoryDB` database.
-
 use std::sync::{Arc, RwLock};
 use std::clone::Clone;
 use std::collections::btree_map::BTreeMap;
-use std::collections::HashMap;
 
 use super::{Database, Snapshot, Patch, Iterator, Iter, Result};
 use super::db::Change;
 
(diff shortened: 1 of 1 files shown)


ozkriff:
> - [ ] CHANGELOG.md have been updated

Shouldn't this change be mentioned in the changelog?

aleksuss:
Actually I don't know. There're unimportant changes.

DarkEld3r:
We need to develop a convention for such things. On the one hand, it is an implementation detail. On the other hand, it (probably) improves performance.

slowli:
> Improve MemoryDB functionality

Do you mean performance? (As far as I see, the functionality of `MemoryDB` isn't changed by the PR.) If so, could you please write a benchmark to compare?

Also, I may be missing something, but I'm having trouble understanding why your implementation works correctly. Wouldn't it be the case that if one name (say, `foo`) is a prefix of another name (say, `foobar`) and we call `db.iter("foo", &[])`, it would incorrectly include `foobar`'s entries into the iterator? Could you please write the corresponding test?

aleksuss:
@slowli Yes. You're right about collision with nearest prefixes. I've missed this case. We should write more unit tests 😄

PR #406, opened 2017-12-27
Support for CORS

matklad:
Reincarnation of https://github.com/exonum/exonum/pull/348.

Supporting CORS is not as simple as adding `Allow-Origin: *` to the headers: you also have to deal with preflight protocol.

<details>
<summary> preflight example</summary>

```
[1514359409 : 813] - [ DEBUG ] - hyper::server::request - Request Line: Options AbsolutePath("/api/services/cryptocurrency/v1/wallets") Http11
[1514359409 : 817] - [ DEBUG ] - hyper::server::request - Headers { Host: 127.0.0.1:8000
, Connection: keep-alive
, Access-Control-Request-Method: POST
, Origin: http://example.com
, User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.90 Safari/537.36
, Access-Control-Request-Headers: content-type
, Accept: */*
, Referer: http://example.com/
, Accept-Encoding: gzip, deflate, br
, Accept-Language: en-US,en;q=0.8
, }
[1514359409 : 823] - [ DEBUG ] - hyper::server::response - writing head: Http11 Ok
[1514359409 : 825] - [ DEBUG ] - hyper::server::response - headers [
Headers { Content-Length: 0
, Access-Control-Allow-Origin: *
, Access-Control-Allow-Methods: POST
, Access-Control-Allow-Headers: content-type
, Content-Type: text/plain
, Date: Wed, 27 Dec 2017 07:23:29 GMT
, }]
```

</details>

So it seems like it's better to use some off-the-shelf solution. There's `iron-cors`, but it is not super actively maintained and is not compatible with our version of iron. Currently, some random fork is used, but I suggest that we fork it as `exonum-iron-cors` until the upstream is updated.

This was tested manually, will look into automated tests now.


exonum/src/node/mod.rs
```diff
@@ -151,10 +155,112 @@ impl Default for NodeApiConfig {
...
+        match *self {
+            AllowOrigin::Any => CorsMiddleware::with_allow_any(),
+            AllowOrigin::Whitelist(ref hosts) => CorsMiddleware::with_whitelist(
+                hosts.iter().cloned().collect(),
+            ),
+        }
+    }
+}
+
+impl ser::Serialize for AllowOrigin {
```
matklad: These serde impls are to allow the user specify any of

```
allow_origin = "*"
allow_origin = "http://example.com"
allow_origin = ["http://a.org", "http://b.org"]
```

exonum/src/node/mod.rs
```diff
@@ -151,10 +155,112 @@ impl Default for NodeApiConfig {
...
+    check(r#""*""#, AllowOrigin::Any);
+    check(
+        r#""http://example.com""#,
+        AllowOrigin::Whitelist(vec!["http://example.com".to_string()]),
+    );
+    check(
+        r#"["http://a.org", "http://b.org"]"#,
+        AllowOrigin::Whitelist(vec!["http://a.org".to_string(), "http://b.org".to_string()]),
+    );
+}
```
matklad: Not sure where's the proper place for this test. Also, there's quite a lot of boring serializable stuff in this file, perhaps it's time to extract `node/config.rs` module?

matklad:
Not sure what's up with CI `pthread lock: invalid argument` seems unrelated to the changes: https://travis-ci.org/exonum/exonum/jobs/322095684#L1128.

Question: presumably, this new option should be documented somewhere?

matklad:
I've managed to write a test for this, but it required some changes to both exonum and testkit. 

The problem is that testkit and exonum both have (different) code for configuring endpoints:

exonum: https://github.com/exonum/exonum/blob/8f020f712cb3e171a98f46d2cbc1bf221dd5e00f/exonum/src/node/mod.rs#L753-L776
testkit: https://github.com/exonum/exonum-testkit/blob/eca67627ee5e9afc986cbd6f47ddce4589fd3d69/src/lib.rs#L1158-L1174

And of course this is exactly the place where CORS middleware is registered :) 

I suggest to extract this logic into `#[doc(hidden)] pub fn` and reusing it in test kit. The prototype implementation is in these two commits:

https://github.com/matklad/exonum/commit/6d5d21002657a83966c0da66de09ef9bf74c244d
https://github.com/exonum/exonum-testkit/commit/899de32b61a49dc22ffbed9f6d5c19cce1e5cde2

If this plan sounds right, I'll go ahead and make necessary changes properly (including private API endpoints).

Alternatively, is there perhaps some really-really integrated tests, which I can use to test this functionality?

matklad:
Status: PR is now using master branch of `iron-cors`, which is compatible with iron 0.6. Waiting for a new release of iron-cors.

exonum/src/node/mod.rs
```diff
@@ -767,10 +873,15 @@ impl Node {
                     mount.mount("api/explorer", router);
                 }
 
+                let cors_middlware = self.api_options.allow_origin.as_ref().map(
```
slowli: Nitpick: "e" missing from `cors_middlware`.

exonum/src/node/mod.rs
```diff
@@ -767,10 +873,15 @@ impl Node {
...
+                    AllowOrigin::as_middleware,
+                );
                 let thread = thread::spawn(move || {
                     info!("Public exonum api started on {}", listen_address);
-
-                    let chain = Chain::new(mount);
+                    let mut chain = Chain::new(mount);
+                    if let Some(middlware) = cors_middlware {
```
slowli: Nitpick: "e" is missing from `middlware`, too.

exonum/src/node/mod.rs
```diff
@@ -142,6 +144,8 @@ pub struct NodeApiConfig {
     pub public_api_address: Option<SocketAddr>,
     /// Listen address for private api endpoints.
     pub private_api_address: Option<SocketAddr>,
+    /// Set CORS part of response header for browser-based clients.
```
slowli: Nitpick: is "set" a verb here? It looks out of place, imo. Also, not sure why "browser-based" is explicitly mentioned here. Isn't it the case that CORS headers would be returned for *all* HTTP requests? I'd suggest to use something like

```rust
/// Cross-origin resource sharing ([CORS][cors]) options for responses returned
/// by API handlers.
///
/// [cors]: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
```

exonum/src/node/mod.rs
```diff
@@ -151,10 +155,112 @@ impl Default for NodeApiConfig {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AllowOrigin {
+    /// Allow access from any host.
+    Any,
+    /// Allow access only from the following hosts.
+    Whitelist(Vec<String>),
+}
+
+impl AllowOrigin {
+    fn as_middleware(&self) -> CorsMiddleware {
```
slowli: Since `AllowOrigin` implements `Clone` and this method clones the whitelist anyway, maybe it could make sense to instead implement `From<AllowOrigin>` for `CorsMiddleware` and use it lower as

```rust
let cors_middleware = self.api_options.allow_origin.clone().map(CorsMiddleware::from);
```

matklad:
No longer WIP: this now uses freshly released iron-cors 0.7.

Testing this is blocked on https://github.com/exonum/exonum/pull/409

PR #409, opened 2017-12-28
Expose api handler creation fucntions for testing

matklad:
While trying to come up with a test for CORS functionality, I've found out that testkit and exonum duplicate code for creating andpoints, and that testkit does not use CORS middleware because of it. As a solution to this problem, this PR suggest to extract this functionality into `pub` but `doc(hidden)` functions. So now we can be sure that testkit provides the same API structure as Exonum. On the other hand, now it would be harder to tweak API for testkit, but hopefully we won't need such tweaks?

I've mechanically extracted all dependencies to function parameters, so someone should look at this as a whole and check that API makes sense. 

The testkit side of things is https://github.com/exonum/exonum-testkit/pull/23

### DIFF
--- a/exonum/src/node/mod.rs
+++ b/exonum/src/node/mod.rs
@@ -841,25 +841,14 @@ impl Node {
 
         let private_config_api_thread = match self.api_options.private_api_address {
             Some(listen_address) => {
-                let mut mount = Mount::new();
-                mount.mount("api/services", blockchain.mount_private_api());
-                let shared_api_state = self.handler().api_state().clone();
-                let mut router = Router::new();
-                let node_info =
-                    private::NodeInfo::new(blockchain.service_map().iter().map(|(_, s)| s));
-                let system_api = private::SystemApi::new(
-                    node_info,
+                let handler = create_private_api_handler(
                     blockchain.clone(),
-                    shared_api_state,
+                    self.handler().api_state().clone(),
                     api_sender,
                 );
-                system_api.wire(&mut router);
-                mount.mount("api/system", router);
-
                 let thread = thread::spawn(move || {
                     info!("Private exonum api started on {}", listen_address);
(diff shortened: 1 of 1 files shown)


exonum/src/node/mod.rs
```diff
@@ -837,3 +814,48 @@ impl Node {
...
+        explorer_api.wire(&mut router);
+        mount.mount("api/explorer", router);
+    }
+
+    let mut router = Router::new();
+    let system_api = public::SystemApi::new(pool, blockchain, shared_api_state);
+    system_api.wire(&mut router);
+    mount.mount("api/system", router);
+
+    Chain::new(mount)
```
slowli: Maybe, a stupid question, but: what purpose does wrapping a `Mount` into a `Chain` serve? It seems that the code would work without it (including a potential change of the return type into `impl Handler` in the future).

matklad: The real reason is that I've just copied what was originally written. However, I do think that returning a `Chain` here is the right thing to do: this allows us to add middlewares without changing the signature. This, for example, will happen with CORS: https://github.com/exonum/exonum/pull/406.

slowli: Makes sense, thanks! As far as I get from the Iron docs, `Chain` is kind of like `Box<Handler>` in that it could be wrapped around any handler hiding its implementation details (and allowing to add middleware for the handler).

slowli:
Just noticed that there is a typo in the 6e76763e2d429eb5bbdc89ef8f71e829775685ae commit message: "fucntions" instead of "functions".

alekseysidorov:
`Conflicting files exonum/src/node/mod.rs`

Can you resolve?

matklad:
@alekseysidorov sure, but I was under the impression that @slowli had some questions here.

PR #413, opened 2018-01-04
Automatically calculate field offsets

matklad:
This PR uses macro magic to automatically calculate fields offset. Only implementation of macro is changed at the moment: that is, the user of `message!` still has to specify offsets, but they are ignored. In other words, the change is backwards compatible (if the user does not import exonum macros selectively). 

Currently, only `message!` macro implementation is updated, `encode_struct!` will be next, but I've decided I'd send a PR to gauge feedback.


matklad:
Should be ready for review now. Some things that are not yet done:

1) change the DSL itself so that the user does not need `[from => to]`. I am not sure what is backwards compatibility policy here
2) refactor the `Field` trait so that it does not use unnecessary `to` argument for read and write methods
3) remove `check_offsets` macro.

slowli:
I wonder whether it could be possible to remove `const SIZE = ...` spec, too. As far as I understand from the code, it continues to be used in this PR.

Also, a minor nitpick: typo in ff3cd7939adcbb669ca5d41539ddf22d4dce3c21 commit message: "endocing" instead of "encoding".

matklad:
I wonder if macro based approach is the best here. I can imagine a more type-driven one, based on the trait like

```Rust
trait FieldVisitor {
    fn visit<F: Field>(name: &'str, from: Offset, to: Offset);
}
```


My understanding is that the long-term plan about current encoding implementation is unclear (i.e, we'd want other languages to be able to consume and produce encoded messages, which can't work with Rust-specific macro-based implementation). So I would say quick&dirty temporary macro solution is preferable to some type-level hackery.

matklad:
@alekseysidorov the problem is that Rust macros have some problems with hygene and visibility, so one has to use hand-crafted name-manging to simulate privacy. Here's an example from `lazy_static`: https://github.com/rust-lang-nursery/lazy-static.rs/blob/c0fbc23aa9924eba2c134745f0a353bb97c08d23/src/lib.rs#L128

However, I see that my naming scheme is internally inconsistent, and that I missed a couple of `doc(hidden)` attributes. I'll change all prefixes to `__ex_`.

matklad:
Interesting idea: I think we can make `field_size()` an associated constant, and this will guarantee zero runtime cost. However, one can't call `std::mem::size_of` as a const fn at the moment, so we'll have to hard-code sizes for primitive types, which seems really unfortunate (we will be able to macro-genera sum for composite types though). 

Besides, I am 90% sure that LLVM will inline and constant-fold these functions in release mode anyway. It is [well known](https://twitter.com/ManishEarth/status/936084757212946432) that LLVM inlines all things by default :)

matklad:
Just as a sanity check, not performance differences are observed for `cargo +nightly bench -- tests::bench_execute_block_timestamping_rocksdb` (of course, encoding is probably not a bottleneck for this benchmark).

matklad:
About private macros:

The trick I've mentioned with `use foo::*` actually importing macros as well works only with crates, not with modules. And is not "officially" a feature yet, so it's better not to try to use it. 

About the idea of using a single helper macro with "keyword" based dispatch, I am not sure it's worth doing it. Looks like from the user perspective a single private macro and several private macros are equivalent? In any case, they'll have to blanket `#[macro_use] extern crate exonum`, instead of `#[macro_use(message, encoding_struct)] extern crate exonum`. Blanket `#[macro_use]` also seems to be an established convention, because a lot of crates are using "__" named helper macros.

matklad:
This commit actually allows to omit sizes and offsets from the macro: https://github.com/exonum/exonum/pull/413/commits/f429385c4244a4a755414df1f54a39454e1fdfd2

Currently, this is done in a backwards-compatible way: it was trivial to do and allows to migrate to the new macro form message-by-message, instead of fixing huge number of marco-invocation errors at once. 

If we really want to force users to switch to the new form, it would be better just to completely remove the old one.

Which repositories can be updated together with this PR? Anything else bedides testkit.

slowli:
@matklad If we're talking the macro without header size and offsets, the current form seems not the most intuitive. For one, fields could be declared in the ordinary Rust manner:
```rust
message! {
    struct Tx {
        // ...
        foo: u64,
        bar: &PublicKey,
    }
}
```

I'm not sure about remaining "associated constants" `ID` and `TYPE`. (Btw, could they be actually made into associated constants of the `Message` or `Transaction` trait? Or will the resulting need for complicated transaction boxing outweigh any benefits? It could be related to #385 - if we want strong typing in transaction execution status, that could require similar transaction boxing logic.) In case they could be `Transaction`'s associated constants, it could look pretty neat, as the user needs to implement `Transaction` for the message anyway:

```rust
impl Transaction for Tx {
    const SERVICE_ID = 1000;
    const ID = 1;

    fn verify(&self) -> bool { /* ... */ }
    fn execute(&self, _: &mut Fork) { /* ... */ }
}
```

matklad:
@slowli yeah, I think that the whole encoding story could have an overhaul, especially if we want this storage format to be accessible from other languages. Like, there's also useless `to` argument in `Field`'s methods and such, and we might want to create a custom derive macro, or even a code generator. 

However, I think it makes sense to postpone this major overhaul (because it needs some serious design work), and instead make the minimal changes just to fix this particular footgun. So, I see this similarity with the previous form as an advantage, because it's easier to upgrade existing messages :)

slowli:
You're probably right; however, I can't help thinking that (leaving "associated constants" alone) the `$field:ident: $type:ty,` scheme is more "Rustic" than `field $field:ident: $type:ty`, and does not require *that* much work to derive from the current macro invocations. (Although the "Rust-ness" is incomplete; e.g., ref'd field types such as `&PublicKey` lack a lifetime specifier, which works against my proposal.)

matklad:
Hm, yep, convinced that `$field: $ty,` would be better :) 

Ok, so the final plan is

* use `$field: $ty,` syntax
* don't provide backwards comparability (i.e, support only the new syntax) (discussed during the daily meeting)

Which repositories should I update after this change? Probably `testkit` and `anchoring`? Anything else?

matklad:
Hm, the test failure is puzzling:

```
pthread lock: Invalid argument
error: process didn't exit successfully: `/home/travis/build/exonum/exonum/exonum/target/debug/deps/exonum-620417560db9bdb3` (signal: 6, SIGABRT: process abort signal)
```

https://travis-ci.org/exonum/exonum/jobs/326830141#L1148

And of course everything works locally :( Has anybody seen a similar error?

vldm:
>Has anybody seen a similar error?

It is a floating heisenbug, that appears rarely in travis-CI. Restarting job usually should help.

alekseysidorov:
In my opinion, first of all you need to update the repositories `testkit`,` configuration` and `btc-anchoring`.

matklad:
> It is a floating heisenbug, that appears rarely in travis-CI. Restarting job usually should help.

Heh, that's scary :) There's a similar old bug report in MySql from facebook, which uses rocksdb somehow. I wonder if the version of rocks we use is significant...

Anyway, this is no longer WIP and should be ready to merge. I'd prefer to update other repoes after this is merged.

PR #415, opened 2018-01-08
Encoding: Properly handling float serialization of zero values

vitvakatu:
### Description
Both positive and negative values has the same `classify` representation (`FpCategory::Zero`), which can make possible problems while encoding/decoding floats.

This change fixes behavior by forbidding negative zeros in wrappers' constructors.

### Checklist:

cc @matklad @vldm @DarkEld3r

### DIFF
--- a/exonum/src/encoding/float.rs
+++ b/exonum/src/encoding/float.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use std::mem;
-use std::num::FpCategory;
 use std::error::Error;
 
 use byteorder::{ByteOrder, LittleEndian};
@@ -25,7 +24,8 @@ use encoding::{CheckedOffset, Field, Offset};
 use encoding::serialize::WriteBufferWrapper;
 use encoding::serialize::json::ExonumJson;
 
-/// Wrapper for the `f32` type that restricts non-finite (NaN and Infinity) values.
+/// Wrapper for the `f32` type that restricts non-finite
+/// (NaN, Infinity, negative zero and subnormal) values.
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
 pub struct F32 {
     value: f32,
@@ -36,7 +36,7 @@ impl F32 {
     ///
     /// # Panics
     ///
-    /// Panics if given value isn't normal (either `NaN`, `Infinity` or `SubNormal`).
+    /// Panics if given value isn't normal (either `NaN`, `Infinity`, negative zero or `SubNormal`).
     ///
     /// # Examples
     ///
@@ -51,7 +51,7 @@ impl F32 {
     }
 
     /// Creates a new `F32` instance with the given `value`. Returns `None` if the given value
-    /// isn't normal.
+    /// isn't normal (either `NaN`, `Infinity`, negative zero or `SubNormal`).
     ///
     /// # Examples
     ///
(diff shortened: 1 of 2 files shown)


exonum/src/encoding/float.rs
```diff
@@ -66,9 +66,10 @@ impl F32 {
     /// assert!(val.is_none());
     /// ```
     pub fn try_from(value: f32) -> Option<Self> {
-        match value.classify() {
-            FpCategory::Normal | FpCategory::Zero => Some(Self { value }),
-            _ => None,
+        if value.is_normal() || value == 0.0 {
```
matklad: Hm, at least according to my understanding of IEEE754, `+0.0 == -0.0`, though their bit-patterns are different. So it seems to me that with this code will allow both `-0.0` and `+0.0` as an internal representation of F32, potentially causing non-canonical serialization.

That is, I think the following version should be more correct:

```
if value == 0.0 { 
  Some(Self { value: 0.0 } // normalize -0.0
else if value.is_normal() {
  Some(Self { value })
else {
  None
}
```

It would be great to have a test that checks that +0.0 and -0.0 are both accepted and serialized the same in both binary and JSON format.

vitvakatu:
Okay, now it works as desired (so doesn't work at all). What should I do with negative zeros?
cc @DenisKolodin @vldm @matklad

matklad:
> What should I do with negative zeros?

I propose *normalization*: `F32/64` are guaranteed not to contain negative zero. However, it is possible to construct `F32/64` from `-0.0`, which would be converted to `0.0`.


My concern that JSON serialization is inherently lossy because it represents floats in decimal system is unfounded. That is, it's true that for all non-NaN values `x`, `fromJsonString(toJsonString(x)) == x`. This is so because string representation uses enough decimal digits to restore the original binary number precisely: although the decimal number is different from the original binary number, converting it back to binary gives the original number.

vitvakatu:
I intend to agree with @matklad, but I'm not sure whether it can produce some unpredictable results during serialization/deserialization.

matklad:
>I'm not sure whether it can produce some unpredictable results during serialization/deserialization.

I think we can just check (once) **all** `f32` numbers: https://randomascii.wordpress.com/2014/01/27/theres-only-four-billion-floatsso-test-them-all/

matklad:
An excellent observation @vldm! The crux of the problem is that we can potentially deserialize a message which could have never been produced by the serializer! One possible fix for this is, during deserialization, validate that incoming bytes indeed could have been produced by serializer. I propose the following API for this:

```Rust
fn normalize(x: f32) -> f32 {
    if x == 0.0 { 0.0 } else { x }
}

fn is_representable(x: f32) -> bool {
    x.is_normal()
}

impl F32 {
    fn try_from(x: f32) -> Option<F32> {
        let x = normalize(x);
        if is_representable(x) { Some(F32 {value: x }) } else{ None }
    }

    fn ser(&self) {
        assert!(is_representable(self.value)); // Guaranteed by all constructors
        LittleEndian::write_f32(self.value)
    }

    fn de(bytes) -> Option<F32> {
        let x = LittleEndian::read_f32(bytes);
        if !is_representable(x) {
            warn!("Someone tries to send us hand-crafted bad messages!");
            return None;
        }
        Some(x)
    }
}
```

vitvakatu:
@matklad your proposed solution forbids zeros (both positive and negative), because is_normal will return false for them.

matklad:
👍 ,yep, should be `x.is_normal() || (x == 0 && x.signum() == 1.0)`. The nice thing is, even though my implementation contains a bug, `normalize/validate` split still allows one to avoid *binary* serialization problems (JSON has to be investigated separately anyway I think).

vitvakatu:
I think we need to merge it now and return to this later, when we'll have strong strategy around floating point numbers.

matklad:
The docstrings are slightly inconsistent with the implementation now, because they don't mention negative zero, and because the docstring on the struct does not mention subnormals.

vitvakatu:
@matklad indeed, fixed now

PR #416, opened 2018-01-08
Remove unused `services_configs` field

matklad:
This just removes some dead code

### DIFF
--- a/exonum/src/helpers/fabric/details.rs
+++ b/exonum/src/helpers/fabric/details.rs
@@ -445,7 +445,6 @@ impl Command for Finalize {
                     ..Default::default()
                 },
                 mempool: Default::default(),
-                services_configs: Default::default(),
             }
         };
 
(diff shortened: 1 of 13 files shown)


matklad:
This is not dead code at all, it turns out. The service can get a node config in service Factory via

```
let node_config: NodeConfig = context.get("node_config").unwrap();
```

Note how this first uses a known string key, and then `Deserialize` to make this API completely undiscoverable :) 

Would it be a reasonable idea to add

```Rust
impl Context {
    fn node_config(&self) -> Result<NodeConfig, Box<Error>> { 
        self.get("node_config") 
    }
    fn service_config<S: Service, C: Deserialize>(&self) -> Result<C, Box<Error>> {
        self.node_config().service_configs[S::NAME].try_into()
    }
}
```
?

matklad:
>Would we close this pullrequest?

Sure! I'll send my thoughs in the for of PR of how we can improve the API later today!

PR #417, opened 2018-01-10
Introduce strongly-typed Context keys

matklad:
So this was born out of https://github.com/exonum/exonum/pull/416#issuecomment-356197978.

The plan to add `Context::service_config` is not perfect, because context can very well lack `service_config` at all. I think what we really want here is something like this:

```Rust
pub trait ServiceFactory: 'static {
    #[allow(unused_variables)]
    fn command(&mut self, command: CommandName) -> Option<Box<CommandExtension>>;
    fn make_service(&mut self, run_context: &Context) -> Box<Service>;
}

pub trait ServiceFactoryWithConfig: ServiceFactory {
    type Config;
    fn make_service(&mut self, run_context: &Context, config: Self::Config) -> Box<Service>;
}
```

This is probably doable, and even in a nice way, despite the problem of object-safe associated types (which I believe is surmountable in a nice way by [this trick](https://github.com/exonum/exonum/pull/385#issuecomment-356329350)).

However, this PR in particular proposes a smaller and more localized fix: we actually can statically type all available `Context` keys. That is, we, instead of stringly-typed API, can provide users with a set of named constants, which allow to extract config values *of particular type*. 

This PR implements it only for `NodeConfig` and `AbstractConfig`, if this is something what we want, I'll convert all other string keys to this pattern.

### DIFF
--- a/exonum/src/helpers/fabric/mod.rs
+++ b/exonum/src/helpers/fabric/mod.rs
@@ -28,12 +28,15 @@ use self::internal::NotFoundInMap;
 pub use self::builder::NodeBuilder;
 pub use self::details::{Run, Finalize, GenerateNodeConfig, GenerateCommonConfig, GenerateTestnet};
 pub use self::shared::{AbstractConfig, NodePublicConfig, CommonConfigTemplate, NodePrivateConfig};
+pub use self::context_key::ContextKey;
 
 mod shared;
 mod builder;
 mod details;
 mod internal;
 mod clap_backend;
+#[macro_use]
+mod context_key;
 
 /// Default port value.
 pub const DEFAULT_EXONUM_LISTEN_PORT: u16 = 6333;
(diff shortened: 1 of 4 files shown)


exonum/src/helpers/fabric/mod.rs
```diff
@@ -191,6 +224,11 @@ impl Context {
         }
     }
 
+    /// Gets the variable from the context.
+    pub fn get2<'de, T: Deserialize<'de>>(&self, key: Key<T>) -> Result<T, Box<Error>> {
```
matklad: Obviously, this will be named `get` in real implementation. Note the crucial property: when the user supplies some key, the don't need to specify `T` explicitly, and they can't specify the wrong T even if they want to.

matklad:
Relevant [funny image](https://i.pinimg.com/474x/20/26/bb/2026bb36cfabf0c1b7812b8aa48a24c2--how-to-create-a-stables.jpg)

matklad:
So this is "production ready" I think, modulo missing doc comments for various configs that we actually used: I haven't myself wrapped my head around what is the purpose of each key :)

exonum/src/helpers/fabric/key.rs
```diff
@@ -0,0 +1,56 @@
...
+    // These fields are public so that `config_key`
+    // macro works outside of this crate. It should be
+    // replaced with `const fn`, once it is stable.
+    #[doc(hidden)]
+    pub __name: &'static str,
+    #[doc(hidden)]
+    pub __phantom: PhantomData<T>,
+}
+
+// We need explicit `impl Copy` because derive won't work if `T: !Copy`.
```
vitvakatu: I wonder if we can use [`derivative`](https://crates.io/crates/derivative) for similar cases.

matklad: Hm, interesting... I think this is a very rare case, so I would prefer to avoid pulling in a proc-macro just for this.

exonum/src/helpers/fabric/key.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+impl<T> Key<T> {
+    /// Name of this key.
+    pub fn name(&self) -> &str {
+        self.__name
+    }
+}
+
+#[macro_export]
```
slowli: Add documentation before merging, please. For example, the following doctest could reveal that the current macro implementation is not precisely correct:

```rust
#[macro_use] extern crate exonum;
use exonum::helpers::fabric::Key as ConfigKey;

fn main() {
    const KEY: ConfigKey<String> = config_key!("foo");
    // Should probably use the key somehow
}
```

matklad: Yep, to bad `no_missing_docs` does not work for macros 🙃 

I am also not sure about naming here: the ideal names would be `Key` and `key!`, but macros are not namespaced yet, so `key!` potentially can collide with something else, and that's why I choose to name it `config_key` instead.

Perhaps this fear is unfounded, and `key!` would be a fine name? What do you all think?

matklad:
Added minimal docs for the particular keys: they are a bit silly, because I don't yet know which configs mean what, but at least they mention relevant command names :-) 

Open question about naming:

* `Key` and `key!` -- optimal names, imo, but `key!` risks collision
* `ContextKey` and `context_key!` -- longer, but no collision and nodes in the direction of `Context`.
* `ConfigKey` and `config_key!` -- a slight variation of the previous one.

matklad:
And here's a proof-of-concept implementation of ServiceFactory with associated config, which also is related to the discussion about associated types in transactions: https://github.com/matklad/exonum/commit/316c73053da7816a6b4d7946bd88f73fa203716a

DarkEld3r:
> Open question about naming:

In C/C++ world where macros also doesn't belong to namespaces it is highly common to use project name as prefix. So it can be `exonum_key!` or `exonum_context_key!` (I suppose `key!` is too generic even if we talk only about our project).

matklad:
@DarkEld3r that's an interesting idea, but I don't think that this pattern is widespread in Rust, and, if we go with `exonum_key!`, we probably should use `exonum_message!` and `exonum_encoding_struct!` for consistency. I'll switch the names for `context_key` now, as it seems sufficiently specific, and I suggest we use this name unless someone has strong objections :)

PR #419, opened 2018-01-10
Use a consistent order of bytes and bits in DBKey [ECR-217]

alekseysidorov:
This PR fixes the bits order in the `DBKey`. 

- Bug fix (non-breaking change which fixes an issue)

- Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:


alekseysidorov:
And question for discussion: do we really need possibility to dump the `ProofMapIndex` structure?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -170,17 +171,20 @@ impl DBKey {
...
             let to = min((self.to + 7) / 8, (other.to + 7) / 8);
-            let max_len = min(self.len(), other.len());
 
             for i in from..to {
                 let x = self.data[i as usize] ^ other.data[i as usize];
                 if x != 0 {
-                    let tail = x.leading_zeros() as u16;
-                    return min(i * 8 + tail - self.from, max_len);
+                    let tail = x.trailing_zeros() as u16;
+                    max_len = min(i * 8 + tail - self.from, max_len);
```
matklad: I think `return min()` would read slightly better in this case, and will allow to reduce the difference in this function to just `s/leading/trailing/`, because you won't need `mut` anymore.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -259,52 +263,99 @@ impl PartialEq for DBKey {
...
+                let i = (byte * 8 + bit) as u16;
+                match i {
+                    _ if i < self.from => write!(&mut bits, "_")?,
+                    _ if i >= self.to => write!(&mut bits, "_")?,
+                    _ => {
+                        write!(
+                            &mut bits,
+                            "{}",
+                            if (1 << bit) & chunk == 0 { '0' } else { '1' }
+                        )?
```
matklad: Nit: it is possible to reduce strength from `write!`  to [push](https://doc.rust-lang.org/std/string/struct.String.html?search=#method.push)

matklad:
>And question for discussion: do we really need possibility to dump the ProofMapIndex structure?

I have no opinion about this particular case, because I have not worked with exonum proofs yet, but I Imagine that having a string representation might be useful for debugging? Could it lead to accidental leakage of secrets though? 

Another nice use of `Debug` is to write tests: instead of writing a ton of `assert_eq!(foo.get(i), 1)` assertions, one can dump `foo` to string and compare strings, which makes tests slightly more readable sometimes.

slowli:
- Does this PR require changes to the light client? Seems yes, since it changes serialization of `DBKey`, and correspondingly the calculation of node hashes in Merkle Patricia proofs, right?
- Can we afford to break compatibility with the old implementation? Do you have any migration procedures in mind?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -259,52 +263,99 @@ impl PartialEq for DBKey {
...
-                "{}",
-                match self.get(i) {
-                    ChildKind::Left => '0',
-                    ChildKind::Right => '1',
+        let mut bits = String::with_capacity(KEY_SIZE * 8);
+        for byte in 0..self.data.len() {
+            let chunk = self.data[byte];
+            for bit in (0..8).rev() {
+                let i = (byte * 8 + bit) as u16;
+                match i {
```
vitvakatu: I believe it'd be more readable to use if statement here.

dmitry-timofeev:
@alekseysidorov, just FYI in case you haven't seen this already: https://wiki.bf.local/display/EXN/Flat+proofs+for+ProofMapIndex

dmitry-timofeev:
> And question for discussion: do we really need possibility to dump the ProofMapIndex structure?

Might be valuable in tests to see how this structure looks like. We have a [simple printer](https://github.com/exonum/exonum-java-binding/blob/master/exonum-java-proofs/src/main/java/com/exonum/binding/storage/proofs/map/MapProofTreePrinter.java) that we used in our tests.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -259,52 +263,99 @@ impl PartialEq for DBKey {
...
+
+    assert_eq!(buf[0], LEAF_KEY_PREFIX);
+    assert_eq!(buf[33], 0);
+    assert_eq!(&buf[1..33], &[250; 32]);
+    assert_eq!(key2, key);
+}
+
+#[test]
+fn test_dbkey_storage_key_branch() {
+    let mut key = DBKey::leaf(&[255; 32]);
```
dmitry-timofeev: Do you use `leaf` here because you don't have a static factory method for a _branch_?

alekseysidorov: Yes.

dmitry-timofeev: Won't it make the code clearer (I don't know how frequently it is used) to _add it_?

alekseysidorov: I want to refactor this code in upcoming PR later.

alekseysidorov:
@slowli 

- Yes, it need to update the light client.
- Yes, we do not have long term production solutions at this moment. The only possible migration procedure is to re-create the blockchain from scratch from existing transactions.
So I consider that this PR is need to merge as soon as possible.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - The `Transaction` trait now inherit `ExonumJson`. (#402)
 - The list of peer connections is now restored to the last state after the process is restarted. (#378)
 - `message!` and `encoding_struct!` no longer require manual `SIZE` and offset specification.
+- The order of bytes and bits in the `DBKey` keys of `ProofMapIndex` became consistent. Note: this change will break old storages. (#419)
```
DarkEld3r: Maybe clarify the note a bit? For example, that breaking change is related to the proofs.

dmitry-timofeev: … break old storages using map proofs?

dmitry-timofeev: Or any storage uses a proof map for _internal_ purposes?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -259,52 +259,93 @@ impl PartialEq for DBKey {
...
     fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-        write!(f, "DBKey(")?;
-        for i in 0..self.len() {
-            write!(
-                f,
-                "{}",
-                match self.get(i) {
-                    ChildKind::Left => '0',
-                    ChildKind::Right => '1',
+        let mut bits = String::with_capacity(KEY_SIZE * 8);
```
slowli: Are you sure that the capacity is sufficient? Seems like `|` symbols will require additional `KEY_SIZE` capacity.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -259,52 +259,93 @@ impl PartialEq for DBKey {
 
 impl ::std::fmt::Debug for DBKey {
```
slowli: IMO, this implementation could be tested; if not for anything else, then for simply showing how output is intended to look in a [TDD](https://en.wikipedia.org/wiki/Test-driven_development) fashion. I'm not sure whether this aligns with best practices for Rust coding.

alekseysidorov: Ready

PR #420, opened 2018-01-15
DBKey refactoring

alekseysidorov:
### Description

Make DBKey great again.

-  Other (CI-improvements, docs fixes, version bumps, clippy, rustfmt)

### Checklist:


exonum/src/storage/proof_map_index/key.rs
```diff
@@ -76,53 +67,109 @@ impl ::std::ops::Not for ChildKind {
...
+    fn set_end(&mut self, end: u16) {
+        let max_len = (KEY_SIZE * 8) as u16;
+        assert!(end <= max_len);
+        // Update DBKey kind and right bound.
+        if end == max_len {
+            self.bytes[DB_KEY_SIZE - 1] = 0;
+            self.bytes[0] = LEAF_KEY_PREFIX;
+        } else {
+            self.bytes[0] = BRANCH_KEY_PREFIX;
+            self.bytes[DB_KEY_SIZE - 1] = end as u8;
```
matklad: Nit: assignments are in different order in the if's branches :)

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -76,53 +67,109 @@ impl ::std::ops::Not for ChildKind {
...
+    /// Create a path from the given key.
+    pub fn new<K: ProofMapKey>(key: &K) -> DBKey {
         debug_assert_eq!(key.size(), KEY_SIZE);
 
-        let mut data = [0; KEY_SIZE];
-        key.write(&mut data);
+        let mut data = [0; DB_KEY_SIZE];
+        data[0] = LEAF_KEY_PREFIX;
+        key.write(&mut data[1..KEY_SIZE + 1]);
+        data[KEY_SIZE + 1] = 0;
```
matklad: `KEY_SIZE + 1` is written as `DB_KEY_SIZE - 1` down the line. They are equivalent, right? Let's use one of them then?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -76,53 +67,109 @@ impl ::std::ops::Not for ChildKind {
...
+        };
     }
+}
 
-    // TODO: terrible hack, try to remove this (ECR-22)
-    #[doc(hidden)]
-    pub fn set_from(&mut self, from: u16) {
-        self.from = from
+/// The bits representation of the `DBKey`.
+pub trait KeyBitRange {
```
matklad: Is this trait intended to be `pub` (available to service implementers)? Perhaps it makes sense to make it `pub(crate)`?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -76,53 +67,109 @@ impl ::std::ops::Not for ChildKind {
...
+        self.end() == self.start()
     }
+    /// Get bit at position `idx`.
+    fn bit(&self, idx: u16) -> ChildKind;
+    /// Returns the new `DBKey` with the given left border.
+    fn start_from(&self, idx: u16) -> Self;
+    /// Shortens this DBKey to the specified length.
+    fn prefix(&self, pos: u16) -> Self;
+    /// Return object which represents a view on to this slice (further) offset by `i` bits.
+    fn suffix(&self, pos: u16) -> Self;
```
matklad: `i` in docstring, `pos` in implementation

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -76,53 +67,109 @@ impl ::std::ops::Not for ChildKind {
...
+    /// Returns true if we starts with the same prefix at the whole of `Other`
```
matklad: s/Other/other/

exonum/src/storage/proof_map_index/key.rs
matklad: So looks like there's an invariant binding `bytes[0]` and `bytes[DB_KEY_SIZE - 1]`: the first is `LEAF_KEY_PREFIX` iff the second is zero. If there is such invariant, could we use a single byte instead?

Should we check this invariant in `from_raw` method?

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -28,8 +28,8 @@ impl Serialize for DBKey {
     {
         let mut repr = String::with_capacity(KEY_SIZE * 8);
         let bslice = self;
-        for ind in 0..self.to() - self.from() {
-            match bslice.get(ind) {
+        for ind in 0..self.end() - self.start() {
```
matklad: `self.len()`?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -131,82 +178,84 @@ impl DBKey {
...
+        key.start = self.start;
+        key.set_end(self.start + pos);
+        key
     }
 
-    /// Shortens this DBKey to the specified length.
-    pub fn truncate(&self, size: u16) -> DBKey {
-        debug_assert!(self.from + size <= self.to);
+    fn suffix(&self, pos: u16) -> Self {
+        debug_assert!(self.start() + pos <= self.end());
```
matklad: The difference between `start_from` and `suffix` is pretty subtle :( Do we really need both of this methods? Should we implement `suffix` in terms of `start_from` as `fn suffix(&self, pos: u16) -> Self { self.start_from(self.start() + pos) }`

exonum/src/storage/proof_map_index/key.rs
alekseysidorov: Yes, the byte looks redundant, but I do not want to break the database structure even more.

I think yes.

matklad: >Yes, the byte looks redundant, but I do not want to break the database structure even more.

I think if we already break the storage in any way, any additional breakage is "free" migration-wise, because you have to recreate the database anyway. So it makes sense to batch all storage-breaking changes, and, if it makes sense to get rid of redundancy, this is the best opportunity to do this. 

I don't know whether getting rid of redundancy is worth it in this case though :-)

Hm, just a tangential thought, but we can leave a couple of unused bytes in the key for future use :-)

DarkEld3r:
I like `ProofPath` name. :+1:

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -75,52 +69,117 @@ impl ::std::ops::Not for ChildKind {
...
+            start: 0,
+        }
     }
 
-    /// Returns length of the `DBKey`.
-    pub fn len(&self) -> u16 {
-        self.to - self.from
+    /// Sets the right border of the bit range.
+    fn set_end(&mut self, end: u16) {
+        let max_len = (KEY_SIZE * 8) as u16;
```
slowli: It seems this can be changed to a `const`.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -75,52 +69,117 @@ impl ::std::ops::Not for ChildKind {
...
-        let chunk = self.data[(pos / 8) as usize];
+    fn end(&self) -> u16 {
+        if self.is_leaf() {
+            KEY_SIZE as u16 * 8
+        } else {
+            u16::from(self.bytes[PROOF_PATH_LEN_POS])
+        }
+    }
+
+    fn bit(&self, idx: u16) -> ChildKind {
```
slowli: It seems like `bit` and `common_prefix` methods can be moved to `BitRange` as default methods; they don't use `ProofPath` intrinsics. Is there some reason why they are not in `BitRange`?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -130,142 +189,67 @@ impl DBKey {
...
-            to: self.from + size,
-        }
+    fn suffix(&self, pos: u16) -> Self {
+        self.start_from(self.start() + pos)
     }
 
-    /// Returns how many bits at the beginning matches with `other`
-    pub fn common_prefix(&self, other: &Self) -> u16 {
+    fn common_prefix(&self, other: &Self) -> u16 {
         // We assume that all slices created from byte arrays with the same length
```
slowli: Nitpick: this comment doesn't make much sense to me, and if we specify that `common_prefix` should work only with ranges starting from the same position, it could be safely removed.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -130,142 +189,67 @@ impl DBKey {
...
 }
 
-impl ::std::fmt::Debug for DBKey {
+impl ::std::fmt::Debug for ProofPath {
     fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
-        // 8 bits + '|' symbol per byte.
-        let mut bits = String::with_capacity(KEY_SIZE * 9);
-        for byte in 0..self.data.len() {
-            let chunk = self.data[byte];
+        let mut bits = String::with_capacity(KEY_SIZE * 8);
```
slowli: Wrong capacity as a result of incorrect merge?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -75,52 +69,117 @@ impl ::std::ops::Not for ChildKind {
...
+        self.end() == self.start()
     }
-
     /// Get bit at position `idx`.
-    pub fn get(&self, idx: u16) -> ChildKind {
-        debug_assert!(self.from + idx < self.to);
+    fn bit(&self, idx: u16) -> ChildKind;
+    /// Returns the new `ProofPath` with the given left border.
+    fn start_from(&self, idx: u16) -> Self;
+    /// Shortens this ProofPath to the specified length.
```
slowli: Nitpick: "shortens" sounds like the transformation is performed in-place; propose changing to something like "Returns a copy of this bit range shortened to the specified length".

exonum/src/storage/proof_map_index/mod.rs
```diff
@@ -583,11 +586,12 @@ where
     fn insert_branch(
         &mut self,
         parent: &BranchNode,
-        key_slice: &DBKey,
+        key_slice: &ProofPath,
```
ozkriff: Shouldn't this be renamed to something more "path"-related? It's not "key" any more.

exonum/src/storage/proof_map_index/key.rs
matklad: I am not sure about this whole message: looks like the valid range for `end` is `0, 1, 2..., 255, 256`, which is much lower than `u16`. Could we use `Option<u8>` for `end`?

slowli: @alekseysidorov Would it make sense to change `u16` to `Option<u8>` (or possibly, a special type alias, like `ProofPathLength` *everywhere* in the `BitsRange` interface, where `0...256` values should be accepted? Or do you think that would be an overkill?

alekseysidorov: u16 is only used to simplify the code. For example for `start` `u8` is enough, but is not enough for `end` or `len`. 
So I do not want to write code like following:
```rust
    fn len(&self) -> u16 {
        self.end() - self.start() as u16
    }
```
In this way i consider that `Option<u8>` everywhere is overkill.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -274,20 +265,47 @@ impl ::std::fmt::Debug for DBKey {
...
             .finish()
     }
 }
 
+impl StorageKey for ProofPath {
+    fn size(&self) -> usize {
+        PROOF_PATH_SIZE
+    }
+
+    fn write(&self, buffer: &mut [u8]) {
```
slowli: Is it possible to write a `ProofPath` that doesn't start with a 0 position? If no, maybe add `debug_assert_eq!` to check?

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -323,57 +321,55 @@ impl<V: fmt::Debug + StorageValue> MapProof<V> {
...
         let res: Option<&V> = match *self {
             LeftBranch {
                 left_node: ref proof,
-                left_key: ref left_slice_key,
+                left_key: ref left_path,
                 ..
             } => {
-                let left_slice = left_slice_key;
-                if !searched_slice.starts_with(left_slice) {
+                let left_path = left_path;
```
slowli: This doesn't seem right...

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -323,57 +321,55 @@ impl<V: fmt::Debug + StorageValue> MapProof<V> {
...
-    fn validate(&self, searched_slice: &DBKey) -> Result<Option<&V>, Error> {
+    fn validate(&self, searched_path: &ProofPath) -> Result<Option<&V>, Error> {
         use self::BranchProofNode::*;
 
         // if we inspect the topmost level of a proof
         let res: Option<&V> = match *self {
             LeftBranch {
                 left_node: ref proof,
-                left_key: ref left_slice_key,
+                left_key: ref left_path,
```
slowli: I wonder whether these renamings really make sense, given that fields in `BranchProofNode` variants are still named `*key`.

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -323,57 +321,55 @@ impl<V: fmt::Debug + StorageValue> MapProof<V> {
...
             }
             RightBranch {
                 right_node: ref proof,
-                right_key: ref right_slice_key,
+                right_key: ref right_path,
                 ..
             } => {
-                let right_slice = right_slice_key;
-                if !searched_slice.starts_with(right_slice) {
+                let right_path = right_path;
```
slowli: This doesn't seem right as well.

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -385,93 +381,85 @@ impl<V: fmt::Debug> BranchProofNode<V> {
...
+                left_key: ref left_path,
+                right_key: ref right_path,
                 ..
             } => {
-                let mut left_slice = left_slice_key.clone();
-                left_slice.set_from(0);
-                let mut right_slice = right_slice_key.clone();
-                right_slice.set_from(0);
-                if !left_slice.starts_with(parent_slice) || !right_slice.starts_with(parent_slice) {
+                let mut left_path = left_path.start_from(0);
```
slowli: Nitpick: Would it make sense to implement `Clone` for `ProofPath` instead of using `.start_from(0)`? This seems a bit hacky to me.

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -323,57 +321,55 @@ impl<V: fmt::Debug + StorageValue> MapProof<V> {
...
+                left_key: ref left_path,
+                right_key: ref right_path,
                 ..
             } => {
-                let left_slice = left_slice_key;
-                let right_slice = right_slice_key;
-                if searched_slice.starts_with(left_slice) ||
-                    searched_slice.starts_with(right_slice)
-                {
+                let left_path = left_path;
```
slowli: This doesn't look right either.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -129,143 +164,99 @@ impl DBKey {
...
-        DBKey {
-            data: self.data,
-            from: self.from,
-            to: self.from + size,
-        }
+    /// Returns the copy of this bit range with the given left border.
+    fn start_from(&self, pos: u16) -> Self;
+    /// Returns a copy of this bit range shortened to the specified length.
+    fn prefix(&self, len: u16) -> Self;
+    /// Returns the copy of this bit range where the start is shifted by the `len` bit to the right.
```
slowli: Nitpick: should probably be "by the `len` bits to the right".

exonum/src/storage/proof_map_index/key.rs
alekseysidorov: No, for branches we serialize `ProofPath` with non zero start position.

exonum/src/storage/proof_map_index/proof.rs
alekseysidorov: In my opinion clone that produce not an exact copy looks confusing.

slowli: Got it; for some reason I thought that the path here necessarily starts from 0. Although this begs another question - what is the expected behavior when `start_from` shifts the starting position to the left - it could be dealt with later, as discussed privately.

PR #421, opened 2018-01-15
WIP: Significantly improve MemoryDB performance

matklad:
This makes `MemoryDB::snapshot` O(1)

### DIFF
--- a/exonum/src/storage/memorydb.rs
+++ b/exonum/src/storage/memorydb.rs
@@ -14,7 +14,7 @@
 
 //! An implementation of `MemoryDB` database.
 
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 use std::clone::Clone;
 use std::collections::btree_map::BTreeMap;
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


exonum/src/storage/memorydb.rs
```diff
@@ -51,18 +51,16 @@ impl Database for MemoryDB {
...
 
     fn snapshot(&self) -> Box<Snapshot> {
-        Box::new(MemoryDB {
-            map: Arc::new(RwLock::new(self.map.read().unwrap().clone())),
-        })
+        Box::new(MemoryDB { map: Arc::clone(&self.map) })
     }
 
     fn merge(&mut self, patch: Patch) -> Result<()> {
+        let map = Arc::make_mut(&mut self.map);
```
DarkEld3r: Is it OK here to (possibly) make a clone?

matklad: I think yep: we are trading clones here for clones in `.snapshot` in the previous implementation. That is, if we do a `.clone` here, that means that refernce count is greater than one, and that means that we've done `.snapshot` before. That is, the total number of clones in this implementation will not be greater than in the original one.

matklad:
This indeed changes the semantics of `Clone`, and this is important, because it is used in `Database::clone`, so the pr, as is, **is incorrect**.

matklad:
Closing the PR because it is wrong. Instead, filed https://github.com/exonum/exonum/pull/422 which should the bug in this PR impossible. 

I think it is still possible to make memory snapshots O(1) in a straight-forward way, and I'll look into that after #422. (We shouldn't spend *much* time on `MemoryDB`, but I hope the fix would be straightforward, and would help if someone else decides to `wrk` the cyrptoruccency demo :) ).

PR #422, opened 2018-01-16
Database trait simplification

matklad:
This is just a refactoring to make `Databases` trait easier to implement correctly, and to make a bug, which was almost introduced in #421, impossible. 

* Remove `clone` method from the `Database` trait.
* Store `Arc<Database>` instead of `Box<Database>` in `Blockchain` (this also should make cloning the blockchain faster, because it removes an allocation).
* Change `merge` signature from `&mut self` to `&self` which is semantically more clear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `message!` and `encoding_struct!` no longer require manual `SIZE` and offset specification.
 - `from_raw(raw: RawMessage)` method is not part of the `Message` trait, and not an inherent method. (#427)
 - Log dependency was updated to 0.4, which can cause issues with previous versions. (#433)
+- The `Database` trait is simplified: it is not longer required to implement state-sharing `clone` method.
+  Instead, the `merge` method now takes a shared reference to `self`.
 
 ### Removed
 - Removed default `state_hash` implementation in the `Service` trait. (#399)
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -66,7 +66,7 @@ pub mod config;
 /// Only blockchains with the identical set of services and genesis block can be combined
 /// into the single network.
 pub struct Blockchain {
-    db: Box<Database>,
+    db: Arc<Database>,
     service_map: Arc<VecMap<Box<Service>>>,
     service_keypair: (PublicKey, SecretKey),
     api_sender: ApiSender,
@@ -94,7 +94,7 @@ impl Blockchain {
         }
 
         Blockchain {
-            db: storage,
+            db: storage.into(),
             service_map: Arc::new(service_map),
             service_keypair: (service_public_key, service_secret_key),
             api_sender,
@@ -463,7 +463,7 @@ impl fmt::Debug for Blockchain {
 impl Clone for Blockchain {
     fn clone(&self) -> Blockchain {
         Blockchain {
-            db: self.db.clone(),
+            db: Arc::clone(&self.db),
             service_map: Arc::clone(&self.service_map),
             api_sender: self.api_sender.clone(),
             service_keypair: self.service_keypair.clone(),
--- a/exonum/src/storage/db.rs
+++ b/exonum/src/storage/db.rs
@@ -248,6 +248,9 @@ pub trait Database: Send + Sync + 'static {
 
     /// Atomically applies a sequence of patch changes to the database.
     ///
+    /// Note that this method may be called concurrently from different threads, the
+    /// onus to guarantee atomicity is on the implementor of the trait.
+    ///
     /// # Errors
     /// If this method encounters any form of I/O or other error during merging, an error variant
     /// will be returned. In case of an error the method guarantees no changes were applied to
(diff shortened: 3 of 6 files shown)


DarkEld3r:
I like almost everything except: 

> Change `merge` signature from `&mut self` to `&self` which is semantically more clear.

Why is that? As for me, adding new changes (that is `merge` for) is a mutable operation, because it changes the database state.

matklad:
>Why is that? As for me, adding new changes (that is merge for) is a mutable operation, because it changes the database state.

This is a peculiarity of Rust. `mut` in `&mut` does not mean "mutable", it means "unique". This is best explained in this blog post: http://smallcultfollowing.com/babysteps/blog/2014/05/13/focusing-on-ownership/

My version would be that `&mut T` guarantees that you have the *single* reference to a particular `T`, and `&T` gives you one of potentially may references. By itself, this does not tell anything about mutability, only about aliasing. However usually (but not always!) a unique access  implies mutable access. However types like Cell, Mutex or Database allow mutation via `&T` references as well. 


The original signature of `merge(&mut self)` tells: "there can be only one concurrent merge, because the self is `&unique`, so you don't need to worry about mutexes and such" (and this was actually a "lie" because of the `clone` function, which allowed one to create two different `mut` handles to the same database and call `merge` concurrenty).

The signature `merge(&self)` tells: "this method needs to be reentrant and it is possible that it is simultaneously called from different threads even, so care must be taken to actually serialize the mutations".

DarkEld3r:
Unrelated note: can `let mut guard = self.map.write().unwrap();` be moved from the loop?

matklad:
>Unrelated note: can let mut guard = self.map.write().unwrap(); be moved from the loop?

Heh, it *must* be moved outside the loop, at the moment we are violating atomicity grantees or `merge`! And we also don't need `RwLock` for snapshot, we can hand out just `HashMap` (b/c it is cloned). I'll tackle that in a separate PR.

slowli:
My 2 cents on the subject (may be fairly uneducated, so sorry in advance):

- Do I understand correctly that the proposed changes treat `Database` as a kind of "smart pointer" trait, akin to `RwLock` / `RefCell` / etc.? (As far as I get, "smart pointers" are the types that can mutate via `&self`.) Are there precedents of "smart pointer" *traits* (not types) in the standard library / elsewhere? (I can't remember one off the top of my head, but as I said, I'm fairly uneducated in Rust 😥 ) If this is the case, I propose to make the docs crystal clear about that.
- Regarding `merge` method: similarly, IMO it could be specified pretty clearly that implementations **MUST** guarantee that concurrent merges are processed correctly, and that the core makes no steps to guarantee that (if this is the case, of course).
- Overall, how likely / sound do you think third-party implementations of `Database` would be? Maybe, we should just warn against them altogether, or maybe they are somehow impossible right now?

matklad:
Meta comment: I think these particular changes are not that important, because they are about the small implementation detail of exonum. However, I do think that "unique" vs "mutable" is an important question for the general design of Exonum: at least for me realizing that "mut === unique" helped tremendously in building mental models for Rust code. 

>Do I understand correctly that the proposed changes treat Database as a kind of "smart pointer" trait, akin to RwLock / RefCell / etc.? (As far as I get, "smart pointers" are the types that can mutate via &self.)

I wouldn't phrase it this way :-) The idea is to make the contract "merge must be reentrant, and it may be called from different threads simultaneously" more explicit. That means that to implement a Database, one would most likely need interior mutability, but it doesn't make Database itself a smart pointer. Not every type which can be mutated via `&T` is a smart pointer: a nice example would be `AtomicUsize`. 

I don't think there's a trait in a stdlib with a similar design, because stdlib does not usually use "global, aliasable" things. However, a similarlish design can be found in the global allocator RFC: 

https://github.com/rust-lang/rfcs/blob/master/text/1974-global-allocators.md#defining-an-allocator

The `Allocator` trait itself uses `&mut self` methods (because allocator can be local), but for the global allocator, you have to `impl<'a> Allocator for &'a T`, so that `&mut self` is `&mut &'a T` which is pretty much equivalent to `&self`. And this makes sense, because global allocator is definitely aliasable!


>Regarding merge method: similarly, IMO it could be specified pretty clearly that implementations MUST guarantee that concurrent merges are processed correctly, and that the core makes no steps to guarantee that (if this is the case, of course).

This is about the documentation, right? Agree that we should make it crystal clear, especially given that `MemoryDB` violates this contract 😆 

>Overall, how likely / sound do you think third-party implementations of Database would be? Maybe, we should just warn against them altogether, or maybe they are somehow impossible right now?

I think this is unlikely. This change is a pure internal refactoring, which does not add much externally visible value. 



As another point for `&T` skeptics, note how https://github.com/exonum/exonum/pull/422/commits/b4a1ba37e8ee5bec2ba69e80b544bd8a1af83964 did not changed implementations of `merge` at all, which means that we don't *need* unique access here.

matklad:
Discussed at today's meeting: seems like we agree that `&self` is correct here. 

It was also noticed that there are several `&mut self` methods on the `Blockchain`, which probably should also be `&self`: given a `b: &Blockchain`, one can always `b.clone().commit()`. But that is more questionable and should be a separate issue.

DarkEld3r:
One more thing: the public interface is changed, so it is technically a breaking change, right? I propose to update the changelog.

PR #423, opened 2018-01-16
Run dev mode

mvlabat:
### Description

Implements `run-dev` mode for running node without configuring.

- `run-dev` command is added

- `commands` vector of `NodeBuilder` is refactored to `HashMap` (maybe `BTreeMap` would be better?)

- command `execute` function now accepts the list of the commands (the aforementioned `HashMap`) in order to execute other subcommands for generating configs. I don't like this solution, it was the first straightforward way that came to my mind

- `Context` args setters were added. That's rather a crutch than a cool solution. (The problem is that the commands called inside `RunDev` depend on these args.) 

### Checklist:
- [x] Implement a working prototype of `run-dev` command
- [x] Cleanup artifacts folder
- [x] Add tests

Using `run-dev` command

### DIFF
--- a/exonum/src/helpers/fabric/clap_backend.rs
+++ b/exonum/src/helpers/fabric/clap_backend.rs
@@ -13,24 +13,29 @@
 // limitations under the License.
 
 use std::ffi::OsString;
+use std::collections::HashMap;
 
 use clap;
 
 use super::{Context, ArgumentType};
 use super::internal::{Feedback, CollectedCommand};
+use super::CommandName;
 
 pub struct ClapBackend;
 
 impl ClapBackend {
     // TODO: remove code duplication (ECR-164)
     #[doc(hidden)]
-    pub fn execute_cmd_string<I, T>(commands: &[CollectedCommand], line: I) -> Feedback
+    pub fn execute_cmd_string<I, T>(
+        commands: &HashMap<CommandName, CollectedCommand>,
+        line: I,
+    ) -> Feedback
     where
         I: IntoIterator<Item = T>,
         T: Into<OsString> + Clone,
     {
         let subcommands: Vec<_> = commands
-            .iter()
+            .values()
             .map(|command| ClapBackend::command_into_subcommand(command))
             .collect();
         let matches = clap::App::new("Exonum application based on fabric configuration.")
@@ -45,21 +50,16 @@ impl ClapBackend {
             .unwrap();
 
         let subcommand = matches.subcommand();
-        for command in commands {
-            if command.name() == subcommand.0 {
-                return command.execute(Context::new_from_args(
-                    command.args(),
-                    subcommand.1.expect("Arguments not found."),
-                ));
-            }
-        }
-
-        panic!("Subcommand not found");
+        let command = commands.get(subcommand.0).expect("Subcommand not found.");
+        command.execute(
+            commands,
+            Context::new_from_args(command.args(), subcommand.1.expect("Arguments not found.")),
+        )
     }
 
-    pub fn execute(commands: &[CollectedCommand]) -> Feedback {
+    pub fn execute(commands: &HashMap<CommandName, CollectedCommand>) -> Feedback {
         let subcommands: Vec<_> = commands
-            .iter()
+            .values()
             .map(|command| ClapBackend::command_into_subcommand(command))
             .collect();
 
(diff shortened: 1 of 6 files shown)


exonum/src/helpers/fabric/clap_backend.rs
```diff
@@ -72,31 +72,22 @@ impl ClapBackend {
...
             .args()
             .iter()
-            .map(|arg| {
+            .zip(1..)
+            .map(|(arg, index)| {
                 let clap_arg = clap::Arg::with_name(arg.name);
                 let clap_arg = match arg.argument_type {
-                    ArgumentType::Positional => {
-                        let arg = clap_arg.index(index);
-                        index += 1;
```
mvlabat: Revert such way of indexing, as we have to count only positional arguments. Thanks, @vldm, for noticing

exonum/src/helpers/fabric/details.rs
```diff
@@ -148,6 +153,106 @@ impl Command for Run {
     }
 }
 
+/// Command for running service in dev mode.
+pub struct RunDev;
+
+impl RunDev {
+    /// Returns the name of the `Run` command.
```
matklad: s/Run/RunDev/

matklad:
Question: is it significant that we actually run commands by name for `run-dev`, or will it be sufficient to have their effects applied? Will it be possible to extract the logic of `GenerateCommonConfig` / `GenerateNodeConfig` into a standalone function and call it from `RunDev` directly, instead of carefully preparing arguments/context and calling commands by their name?

mvlabat:
@matklad, yes, that's exactly what I think about how we can improve the code. I believe that implementing separate functions or even services (why not?) that can be reused is a good solution to the problem. I'll try to do that

mvlabat:
Services - here I mean not services in exonum sense, but just some other kind of abstraction

mvlabat:
@matklad, oh, I forgot. I see a problem here. Our commands can be extended with services. If a service extends a command (for generating config files, for instance) we cannot ignore it.
If calling functions that generate configs directly, we skip the part with extensions, so services may misbehave.

I think we'll have to stick to calling commands here. Though, IMO, this approach could be improved too.

exonum/src/helpers/fabric/details.rs
```diff
@@ -148,6 +153,121 @@ impl Command for Run {
...
+    ) -> Feedback {
+        let db_path = Self::artifacts_path("db", &context);
+        context.set_arg(DATABASE_PATH, db_path);
+        Self::cleanup(&context);
+
+        let node_config_path = Self::generate_config(commands, &context);
+        context.set_arg("NODE_CONFIG_PATH", node_config_path);
+
+        let new_context = exts(context);
+        // TODO: clean up artifacts directory.
```
vitvakatu: Can you remove it now?

exonum/tests/config.rs
```diff
@@ -191,3 +209,24 @@ fn test_generate_full_config_run() {
...
+fn test_run_dev() {
+    let artifacts_dir = ".exonum";
+    let db_dir = format!("{}/{}", artifacts_dir, "db");
+    let full_db_dir = full_tmp_folder(&db_dir);
+
+    fs::create_dir_all(Path::new(&full_db_dir)).expect("Expected db temp folder to be created.");
+    let old_db_file = full_tmp_name("1", &db_dir);
+    touch(&old_db_file);
+
+    let result = panic::catch_unwind(|| { run_dev(artifacts_dir); });
```
vitvakatu: Why do you need `catch_unwind` here?

exonum/src/helpers/fabric/details.rs
```diff
@@ -148,6 +153,121 @@ impl Command for Run {
...
+        let directory = ctx.arg::<String>("ARTIFACTS_DIR").unwrap_or_else(|_| {
+            String::from(".exonum")
+        });
+        PathBuf::from(&directory)
+    }
+
+    fn artifacts_path(inner_path: &str, ctx: &Context) -> String {
+        let mut path = Self::artifacts_directory(ctx);
+        path.push(inner_path);
+        String::from(path.to_str().expect("Expected correct path"))
```
vitvakatu: supernit: The matter of personal taste, but why don't you use `into()` everywhere?

exonum/src/helpers/fabric/details.rs
```diff
@@ -148,6 +153,121 @@ impl Command for Run {
...
+        context.set_arg(DATABASE_PATH, db_path);
+        Self::cleanup(&context);
+
+        let node_config_path = Self::generate_config(commands, &context);
+        context.set_arg("NODE_CONFIG_PATH", node_config_path);
+
+        let new_context = exts(context);
+        // TODO: clean up artifacts directory.
+        commands
+            .get(Run::name())
```
vitvakatu: Oops, shouldn't it be `RunDev`?

vitvakatu: Oh nevermind, I see it now.

exonum/src/helpers/fabric/details.rs
mvlabat: Yeah, I see it now, that it's more favorable to use `into()` here. Ok, thanks, I'll change that )

alekseysidorov:
Remember that `btc-anchoring` has complex bootstrap procedure and please test your changes on it.

exonum/tests/config.rs
mvlabat: Frankly, I've just copy-pasted that without proper thinking. ) Now I understand it's really needed for cleaning up the temp directory, that contains generated configs for tests. And all the assertions should be wrapped into this `catch_unwind`, so we don't skip teardown process.
So there's something to fix, thanks for noticing

PR #424, opened 2018-01-18
Experimental: automatically generate `/transactions` endpoint

matklad:
This is **an experimental** PR to fix ECR-294 (getting rid of [boilerplate](https://github.com/exonum/cryptocurrency/blob/7ddba70423be5f7272d5ca9c13963301bedf7acd/src/lib.rs#L250-L265) required to define a service accepting transactions). I'd like to gauge consensus about the general approach before embarking onto the production-quality implementation. 

Problem: when defining a service, there are two pieces of boilerplate

*  deserializing transaction from an http request.
* deserializing transaction from a raw message (and we have a real bug in this boilerplate in one of our services).

The proposed solution is to allow a service to declare a set of transactions it accepts, and then automatically implement `tx_from_raw` method of service and a single `/transactions` HTTP endpoint for POSTing all trasactions. The endpoint thing is open for a debate: it can be argued that `/transactions` is not RESTfull and is more akin to RPC (on the other way, transactions are nouns in REST sense, and `GET /transactions/{tx_hash}` might be useful). 

However, to associated several types of transactions with a type of a service, we need to solve "Trait Objects with associated types" problem, so this PR proposes an approach to this as well.

Here are the major moving parts:

* `TransactionSet` trait and `transaction_set!` macro allows to create a type-level set of trasactions. They abstract away [this pattern](https://github.com/exonum/exonum-testkit/blob/73cb36995f0dc2117291dd5e441d0688383daedc/tests/currency/cryptocurrency.rs#L200-L203)

* The definition of `Service` is not changed, but it's purpose is a bit different now: it is a **low-level** object (in a dynamic-dispatch sense) that is used by the exonum core. While service implementors *could* implement `Service` directly for maximum control and flexibility, it is advisable to implement a higher level trait.

* `TransactionService` is such trait: it uses fancy associated types and constants to simplify the definition of the service, and it can be converted to `Box<Service>`.

* `ObserverService` is another flavor of such trait: it's a simple service *without* transactions. I included only to make sure that it is possible to provide different *base traits* for service authors. Note that `TransactionService` and `ObsercerService` don't have any supertraits, and are only loosely coupled to `Services`, and can co-evolve independently.

### DIFF
--- a/exonum/src/blockchain/service/srv.rs
+++ b/exonum/src/blockchain/service/srv.rs
@@ -1,11 +1,9 @@
 extern crate core;
 
 use serde_json::Value;
-use serde::de::{DeserializeOwned};
 
 use iron::prelude::*;
 use iron::Handler;
-use bodyparser;
 use router::Router;
 
 use serde_json;
@@ -22,18 +20,17 @@ use super::{Transaction, ServiceContext, ApiContext, Service};
 use std::marker::PhantomData;
 
 /// TODO
-pub trait TransactionSet: DeserializeOwned + Into<Box<Transaction>> + Clone {
+pub trait TransactionSet {
     /// TODO
     fn tx_from_raw(raw: RawTransaction) -> Result<Box<Transaction>, MessageError>;
+    /// TODO
+    fn tx_from_request(request: &mut Request) -> Result<Box<Transaction>, ApiError>;
 }
 
 #[macro_export]
 macro_rules! transaction_set {
     ( $name:ident { $($tx:ident),* } ) => {
-        #[derive(Deserialize, Clone)]
-        #[serde(untagged)] // :(
         pub enum $name {
-            $($tx($tx),)*
         }
 
         impl $crate::blockchain::TransactionSet for $name {
@@ -47,13 +44,27 @@ macro_rules! transaction_set {
 
                 return Err($crate::encoding::Error::IncorrectMessageType { message_type })
             }
-        }
 
-        impl Into<Box<Transaction>> for $name {
-            fn into(self) -> Box<Transaction> {
-                match self {$(
-                   $name::$tx(tx) => Box::new(tx),
-                )*}
+            //TODO: reexport & $crate::
+            fn tx_from_request(request: &mut $crate::iron::Request) -> Result<Box<$crate::blockchain::Transaction>, $crate::api::ApiError> {
+                use $crate::iron::prelude::*;
+
+                #[derive(Deserialize, Clone)]
+                #[serde(untagged)] // :(
+                enum Any {
+                    $($tx($tx),)*
+                }
+
+                match request.get::<$crate::bodyparser::Struct<Any>>() {
+                    Ok(None) => Err($crate::api::ApiError::IncorrectRequest("Empty request body".into()))?,
+                    Err(e) => Err($crate::api::ApiError::IncorrectRequest(Box::new(e)))?,
+                    Ok(Some(any)) => {
+                        match any {$(
+                            Any::$tx(tx) => Ok(Box::new(tx)),
+                        )*}
+                    }
+                }
+
             }
         }
     }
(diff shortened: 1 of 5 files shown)


matklad:
Here's an example how new service would looks like in practice: https://github.com/matklad/exonum-testkit/commit/f11f35cc1b6cf5950c37005ab8c9bc0f2eaa186b (~~usual grumble about **all** services not in the core repository~~)

exonum/src/blockchain/service/srv.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /// TODO
+    fn wire_private_api(&self, _router: &mut Router, _ctx: &ApiContext) {
+    }
+
+    /// TODO
+    fn into_service(self) -> Box<Service> where Self: Sized {
+        Box::new(TransactionServiceImpl(self))
+    }
```
matklad: This is an interesting method. Ideally, we'd like to be able to define  `impl<S: TransactionService> Into<Box<Service>> for S`, but it's possible to have only one such impl, which makes defining several base traits impossible. To put it in other way, nothing prevents a user from implementing *both* `TransactionService` and `ObserverService` for a single type, so we can't generically convert them to a `Box<Service>`.

exonum/src/blockchain/service/srv.rs
```diff
@@ -0,0 +1,247 @@
...
+                return Err($crate::encoding::Error::IncorrectMessageType { message_type })
+            }
+
+            //TODO: reexport & $crate::
+            fn tx_from_request(request: &mut $crate::iron::Request) -> Result<Box<$crate::blockchain::Transaction>, $crate::api::ApiError> {
+                use $crate::iron::prelude::*;
+
+                #[derive(Deserialize, Clone)]
+                #[serde(untagged)] // :(
+                enum Any {
```
slowli: I wonder whether it would make sense to implement this method akin to `tx_from_raw`, i.e., by initially parsing a `serde_json::Value` object and matching its `message_id` field. This would allow to avoid linear dependency of the parsing time on the number of transaction types. Would it be too expensive?

exonum/src/blockchain/service/srv.rs
```diff
@@ -0,0 +1,247 @@
...
+macro_rules! transaction_set {
+    ( $name:ident { $($tx:ident),* } ) => {
+        pub enum $name {
+        }
+
+        impl $crate::blockchain::TransactionSet for $name {
+            fn tx_from_raw(raw: RawTransaction) -> Result<Box<$crate::blockchain::Transaction>, $crate::encoding::Error> {
+                let message_type = raw.message_type();
+                $(
+                    if $tx::message_id() == message_type {
```
slowli: Could it make sense to use `match` instead of `if`s here? IMO, it would read more naturally (not sure about performance, but intuitively, `match` leaves more room for optimizations).

matklad: `match` does not work here because `message_id` is a function. Match would work though with associated constants, and I'd love to change message_id to such constant.

exonum/src/blockchain/service/srv.rs
matklad: Yeah, I plan to look into serialization closely, the `:(` here is for a reason :) We really should teach serde to use the value of message_id for choosing the variant. Now I believe this is even **incorrect** if two different transactions share the same set of fields. 

I don't think that literally producing a `json::Value` is the right approach (json object is much heavier than the corresponding message), but I am almost sure that there is *some* way to convince `serde` to do what we want it to do.

slowli: Maybe extracting *only* the `message_id` field value and skipping all other data a-la [this serde example](https://serde.rs/stream-array.html) could work? 🤔

slowli: Re: transactions with the same fields. We've discussed this before, and it seems that it would work correctly because a transaction deserialization logic expects a particular `message_id` value and fails otherwise. Still seems inefficient.

matklad:
@slowli regarding the extension part, I think an interesting idea is to have at least one such "extension" trait in the core and make it the primary way to define services. That is, I am thinking about something along the following lines:

```Rust
// This is today's service, which serves as a low-level abstraction
// and is used as `Box<LowLevelService>` inside exonum
trait LowLevelService {
  fn service_id(&self) -> u32;
  fn tx_from_raw(&self, raw) -> Box<Transaction>;
  ...
}

// A default high-level trait which service authors would generally use to implement services 
trait Service { 
    const ID: u32;
    type Transactions;
    ....
}
```

That is, I think it's important that service authors don't interact with `LowLevelService` directly by default: this should allow us to evolve the `LowLevel` independently of the surface interface we present to the service implementors. This sounds like an over-engineering though :)

matklad:
superseded by https://github.com/exonum/exonum/pull/457

PR #426, opened 2018-01-18
Make MAX_MESSAGE_LEN configurable

mvlabat:
### Description

This PR makes `MAX_MESSAGE_LEN` configurable, so we get rid of this hardcoded const.

~~**Important note.**
I had to duplicate this parameter in both configs: `ConsensusConfig` and `NetworkConfiguration`. The latter doesn't (de)serialize it, so we actually store it only in `genesis.consensus` section (as all nodes must share this value). This tricky workflow is needed for convenient passing it to `MessagesCodec`.~~

### DIFF
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -68,6 +68,11 @@ pub struct ConsensusConfig {
     pub timeout_adjuster: TimeoutAdjusterConfig,
 }
 
+impl ConsensusConfig {
+    /// Default value for max_message_len.
+    pub const DEFAULT_MESSAGE_MAX_LEN: u32 = 1024 * 1024; // 1 MB
+}
+
 impl Default for ConsensusConfig {
     fn default() -> Self {
         ConsensusConfig {
@@ -75,7 +80,7 @@ impl Default for ConsensusConfig {
             status_timeout: 5000,
             peers_timeout: 10_000,
             txs_block_limit: 1000,
-            max_message_len: 1024 * 1024, // 1 MB
+            max_message_len: Self::DEFAULT_MESSAGE_MAX_LEN,
             timeout_adjuster: TimeoutAdjusterConfig::Constant { timeout: 500 },
         }
     }
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -62,6 +62,8 @@ pub struct ConsensusConfig {
     pub peers_timeout: Milliseconds,
     /// Maximum number of transactions per block.
     pub txs_block_limit: u32,
+    /// Maximum message length (in bytes).
+    pub max_message_len: usize,
     /// `TimeoutAdjuster` configuration.
     pub timeout_adjuster: TimeoutAdjusterConfig,
 }
@@ -73,6 +75,7 @@ impl Default for ConsensusConfig {
             status_timeout: 5000,
             peers_timeout: 10_000,
             txs_block_limit: 1000,
+            max_message_len: 1024 * 1024, // 1 MB
             timeout_adjuster: TimeoutAdjusterConfig::Constant { timeout: 500 },
         }
     }
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -63,7 +63,7 @@ pub struct ConsensusConfig {
     /// Maximum number of transactions per block.
     pub txs_block_limit: u32,
     /// Maximum message length (in bytes).
-    pub max_message_len: usize,
+    pub max_message_len: u32,
     /// `TimeoutAdjuster` configuration.
     pub timeout_adjuster: TimeoutAdjusterConfig,
 }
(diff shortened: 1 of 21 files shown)


exonum/src/events/codec.rs
```diff
@@ -21,10 +21,11 @@ use tokio_io::codec::{Decoder, Encoder};
 use messages::{HEADER_LENGTH, MessageBuffer, RawMessage};
 use super::error::other_error;
 
-pub const MAX_MESSAGE_LEN: usize = 1024 * 1024; // 1 MB
-
 #[derive(Debug)]
-pub struct MessagesCodec;
+pub struct MessagesCodec {
+    /// Maximum message length (in bytes), gets populated from `NetworkConfiguration`.
+    pub max_message_len: usize,
```
DarkEld3r: Should this field be public?
I prefer to make a separate constructor and getter.

exonum/src/blockchain/config.rs
```diff
@@ -62,6 +62,8 @@ pub struct ConsensusConfig {
     pub peers_timeout: Milliseconds,
     /// Maximum number of transactions per block.
     pub txs_block_limit: u32,
+    /// Maximum message length (in bytes).
+    pub max_message_len: usize,
```
matklad: Let's make it `u32`? `ConsensusConfig` is serialize, so it would be nice if it didn't depend on bitness.

exonum/src/events/codec.rs
mvlabat: As discussed, made it just private and initializable with constructor

exonum/src/events/codec.rs
```diff
@@ -37,11 +44,11 @@ impl Decoder for MessagesCodec {
         }
         // Check payload len
         let total_len = LittleEndian::read_u32(&buf[6..10]) as usize;
-        if total_len > MAX_MESSAGE_LEN {
+        if total_len > self.max_message_len {
             return Err(other_error(format!(
                 "Received message is too long: {}, maximum allowed length is {}",
```
matklad: I would add `bytes` to the error message: `maximum allowed length is {} bytes`.

exonum/src/blockchain/config.rs
mvlabat: Sounds reasonable, np

exonum/src/events/codec.rs
```diff
@@ -37,11 +44,11 @@ impl Decoder for MessagesCodec {
         }
         // Check payload len
         let total_len = LittleEndian::read_u32(&buf[6..10]) as usize;
-        if total_len > MAX_MESSAGE_LEN {
+        if total_len as u32 > self.max_message_len {
```
vldm: remove `as usize`  and `as u32`

slowli:
Is it possible to test the change within this PR? I can think of creating several messages of particular sizes for different values of the max message size setting (undersized, exactly sized and oversized) and look how they are processed. Maybe, I'm missing something, but I don't see such tests right now.

exonum/src/events/codec.rs
mvlabat: @vldm, there are two places where we have to cast it to `usize` again. See the code below `// Read message` comment. I think it doesn't matter here.

mvlabat:
Sorry guys :D Tests have arrived! Reviews - dismissed :)

exonum/src/events/tests.rs
```diff
@@ -259,6 +268,33 @@ fn test_network_big_message() {
...
+    e2.wait_for_connect();
+
+    e2.connect_with(first);
+    e1.wait_for_connect();
+
+    e1.send_to(second, m1.clone());
+    assert_eq!(e2.wait_for_message(), m1);
+
+    e2.send_to(first, m2.clone());
+    assert!(e1.receives_error());
```
vitvakatu: nit: can be replaced with `e1.wait_for_event().is_err()`

mvlabat: Oh, jeez, indeed...

exonum/src/events/tests.rs
```diff
@@ -259,6 +261,33 @@ fn test_network_big_message() {
     assert_eq!(e2.wait_for_disconnect(), first);
 }
 
+#[test]
+fn test_network_max_message_len() {
+    let first = "127.0.0.1:17202".parse().unwrap();
+    let second = "127.0.0.1:17303".parse().unwrap();
+
+    let m1 = raw_message(15, ConsensusConfig::DEFAULT_MESSAGE_MAX_LEN as usize - 1000);
+    let m2 = raw_message(16, ConsensusConfig::DEFAULT_MESSAGE_MAX_LEN as usize + 1000);
```
DarkEld3r: I prefer to declare variables closer to the usage if possible. :smile: Additionally `m1` and `m2` can have more meaningful names. For example, `too_big_message`.

Should we test the message with exactly `DEFAULT_MESSAGE_MAX_LEN` size?

PR #429, opened 2018-01-20
[Fix] Config: Do not swallow error message while toml loading

### LINKED ISSUE #428: exonum::helpers::config::ConfigFile error reporting swallows the useful error message
In https://github.com/exonum/exonum/blob/master/exonum/src/helpers/config.rs

the method, `ConfigFile::load()` swallows the useful error returned from the toml crate that may describe a formatting error in the toml file. It maps it to a very generic error. It feels better to just return the error returned from the toml crate.

vitvakatu:
Fixes #428

Long story short:

- Not a breaking change
- No need in updating docs
- No need in testing

### DIFF
--- a/exonum/src/helpers/config.rs
+++ b/exonum/src/helpers/config.rs
@@ -14,13 +14,42 @@
 
 //! Loading and saving TOML-encoded configurations.
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::io::{Read, Write};
 use std::fs::{self, File};
 use std::error::Error;
+use std::fmt;
 
 use serde::{Serialize, Deserialize};
 use toml;
+
+#[derive(Debug)]
+struct DeserializeError {
+    path: PathBuf,
+    inner: toml::de::Error,
+}
+
+impl DeserializeError {
+    pub fn new(path: PathBuf, inner: toml::de::Error) -> Self {
+        Self { path, inner }
+    }
+}
+
+impl fmt::Display for DeserializeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Could not read {}: {}", self.path.display(), self.inner)
+    }
+}
+
+impl Error for DeserializeError {
+    fn description(&self) -> &str {
+        "Could not read toml config."
+    }
+
+    fn cause(&self) -> Option<&Error> {
+        Some(&self.inner)
+    }
+}
 
 /// Implements loading and saving TOML-encoded configurations.
 #[derive(Debug)]
@@ -36,7 +65,9 @@ impl ConfigFile {
         let mut file = File::open(path.as_ref())?;
         let mut toml = String::new();
         file.read_to_string(&mut toml)?;
-        toml::de::from_str(&toml).map_err(|e| Box::new(e) as Box<Error>)
+        toml::de::from_str(&toml).map_err(|e| {
+            Box::new(DeserializeError::new(path.as_ref().to_owned(), e)) as Box<Error>
+        })
     }
 
     /// Saves TOML-encoded file.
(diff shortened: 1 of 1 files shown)


vldm:
Can you attach error output before and after this fix?

vitvakatu:
@vldm sure!
With this change:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { inner: ErrorInner { kind: Wanted { expected: "a value", found: "a newline" }, line: Some(0), col: 14, message: "", key: [] } }', libcore/result.rs:916:5
```
Without:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: InvalidData, error: StringError("Unable to decode toml file") }', libcore/result.rs:916:5
```

vldm:
Error before this fix looks more user friendly, but after, it looks debug/developer friendlier. Can we create a custom error type with containt both this advantages?

matklad:
>Can we create a custom error type with containt both this advantages?

If we go for it, than I suggest to also include the path to the TOML file into the error message. 

And, if we are going to create our own wrapper, would it be possible to give the `failure` crate a try for this single use case? Or should we wait for a more organized error-improvement effort to add failure as a dependency?

vitvakatu:
> If we go for it, than I suggest to also include the path to the TOML file into the error message.

I had that idea, will see tomorrow.
> And, if we are going to create our own wrapper, would it be possible to give the failure crate a try for this single use case? Or should we wait for a more organized error-improvement effort to add failure as a dependency?

I don't think we should use `failure` for a single case, it will not give us anything except problems. At the same time, I'll be also excited to see error handling improvement in future (with `failure` or not).

vitvakatu:
Now output looks like this:
```
Display: Could not read toml config.
Path: test.toml
Error: expected a value, found a newline at line 1
Debug: DeserializeError { path: "test.toml", inner: Error { inner: ErrorInner { kind: Wanted { expected: "a value", found: "a newline" }, line: Some(0), col: 14, message: "", key: [] } } }
Description: Could not read toml config.
Cause: Some(Error { inner: ErrorInner { kind: Wanted { expected: "a value", found: "a newline" }, line: Some(0), col: 14, message: "", key: [] } })
```

Test code:
```rust
fn main() {
    let result: Result<TestConfig, _> = ConfigFile::load("test.toml");
    if let Err(e) = result {
        println!("Display: {}", e);
        println!("Debug: {:?}", e);
        println!("Description: {}", e.description());
        println!("Cause: {:?}", e.cause());
    }
}
```

exonum/src/helpers/config.rs
```diff
@@ -14,14 +14,48 @@
...
+    pub fn new(path: PathBuf, inner: toml::de::Error) -> Self {
+        Self { path, inner }
+    }
+}
+
+impl fmt::Display for DeserializeError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "Could not read toml config.\nPath: {}\nError: {}",
```
alekseysidorov: Is the `\n` suitable for Windows?

vitvakatu: I don't have an opportunity to check, so I'll just remove them ;D

matklad: Note that `writeln!` and `println!` also use `\n` on windows, and established tools like Cargo also just use `\n`: https://github.com/rust-lang/cargo/blob/d4e0b71c05df7768aabf95198cf606672788935d/src/cargo/lib.rs#L149

So, yes, `\n` works for windows just fine (unless you literally use `notepad.exe` in which case you have a more serious problem anyway), and we, imo, should stick to `\n` as well. 


(and a related piece of knowledge, forward slashes in paths also work on windows)

vitvakatu: Actually I don't think we need explicit newlines there. The output feels more dense and solid without them.

matklad: Yeah, I don't think we need `\n` in this particular case! By the way, I think we can compress the message a bit more: `"Could not read {}: {}"`.

PR #430, opened 2018-01-21
Fix a bug with codec that cause node to panic.

vldm:
### Description
Regression bug after tokio land.

### Checklist:

Execute this bash command:
`echo -e '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00' | nc 127.0.0.1 6333`

### DIFF
--- a/exonum/src/events/codec.rs
+++ b/exonum/src/events/codec.rs
@@ -37,6 +37,7 @@ impl Decoder for MessagesCodec {
         }
         // Check payload len
         let total_len = LittleEndian::read_u32(&buf[6..10]) as usize;
+
         if total_len > MAX_MESSAGE_LEN {
             return Err(other_error(format!(
                 "Received message is too long: {}, maximum allowed length is {}",
@@ -44,7 +45,14 @@ impl Decoder for MessagesCodec {
                 MAX_MESSAGE_LEN
             )));
         }
-
+        if total_len < HEADER_LENGTH {
+            return Err(other_error(format!(
+                "Received malicious message with insufficient \
+                size in header: {}, expected header size {}",
+                total_len,
+                HEADER_LENGTH
+            )));
+        }
         // Read message
         if buf.len() >= total_len {
             let data = buf.split_to(total_len).to_vec();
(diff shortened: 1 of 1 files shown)


exonum/src/events/codec.rs
```diff
@@ -64,3 +71,33 @@ impl Encoder for MessagesCodec {
...
+            _ => panic!("Wrong input"),
+        };
+    }
+
+    #[test]
+    fn decode_message_small_size_in_header() {
+        let data = vec![0u8, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+        let mut bytes: BytesMut = data.as_slice().into();
+        let mut codec = MessagesCodec;
+        match codec.decode(&mut bytes) {
```
vitvakatu: nit: can be replaced with `assert!(code.decode(&mut bytes).is_err()`

exonum/src/events/codec.rs
```diff
@@ -37,14 +37,21 @@ impl Decoder for MessagesCodec {
...
             return Err(other_error(format!(
                 "Received message is too long: {}, maximum allowed length is {}",
                 total_len,
                 MAX_MESSAGE_LEN
             )));
         }
-
+        if total_len < HEADER_LENGTH {
+            return Err(other_error(format!(
+                "Received malicious message with small size in header: {}, expected header size {}",
```
vitvakatu: nit: I'm not sure about the word `small`. 
`insufficient` maybe?

exonum/src/events/codec.rs
```diff
@@ -64,3 +72,30 @@ impl Encoder for MessagesCodec {
...
+        let mut bytes: BytesMut = data.as_slice().into();
+        let mut codec = MessagesCodec;
+        match codec.decode(&mut bytes) {
+            Ok(Some(ref r)) if r == &RawMessage::new(MessageBuffer::from_vec(data)) => {}
+            _ => panic!("Wrong input"),
+        };
+    }
+
+    #[test]
+    fn decode_message_small_size_in_header() {
```
matklad: Either the test for `total_len > MAX_MESSAGE_LEN` is missing (then let's add it?) or it is somewhere else (then let's move this test over there as well?).


Also, this might be a nice use case for [proptest](https://github.com/AltSysrq/proptest) library.

mvlabat: @matklad, the test for `total_len > MAX_MESSAGE_LEN` resides in `src/events/tests.rs` (`test_network_max_message_len`). Yea, I think, we should combine them in one file. This one seems to be better, because here we can check what error exactly we receive. While in `src/events/tests.rs` we can only check that there is some error, not knowing which one specifically.

PR #433, opened 2018-01-23
Update logger version

alekseysidorov:
Just update logger version.

- Not a breaking change
- No need in updating docs
- No need in testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,11 +23,11 @@ matrix:
     - env: FEATURE=non-fatal-checks
   fast_finish: true
   include:
-    - rust: nightly-2017-12-11
+    - rust: nightly-2018-01-22
       env: FEATURE=benchmarks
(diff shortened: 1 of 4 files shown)


alekseysidorov:
There are two question:

- Lets format time in the ISO standard format.
- Maybe the `<unknown file>` and `<unknown module> are more human readable.

DarkEld3r:
> Let's format time in the ISO standard format.

I think this is a nice idea for a separate pull request. :smile:

matklad:
Note that log 0.3 and 0.4 can coexist in the single binary, but don't work exactly great with each other, so I think at least minimal testing to check that logs to work is required.

DarkEld3r:
I propose to mention potential 0.3 and 0.4 issues in the changelog. Something like "Log dependency was updated to 0.4, which can cause issues with previous versions".

PR #436, opened 2018-01-24
Use union merge strategy for changelog

DarkEld3r:


.gitattributes
```diff
@@ -0,0 +1 @@
+CHANGELOG merge=union
```
ozkriff: `CHANGELOG.md`?

PR #439, opened 2018-01-26
Polish docs

slowli:
This PR edits several doc comments for clarity. A specific problem that it solves is the misleading naming of constants in the example for the `message!` macro, but it goes much deeper.

I will continue adding commits in this PR, but I would like for some version of it to be included in the upcoming 0.5 release.

Other (docs fixes).

### Checklist:

- [ ] ~Tests for the changes have been added~ (not applicable)

Fetch and checkout the PR locally and build the docs using `cargo doc --no-deps`. Then see the docs corresponding to the introduced changes.


exonum/src/encoding/mod.rs
```diff
@@ -12,23 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! `encoding` is a lazy serialization library,
-//! it allows to keep struct serialized in place, and deserialize fields on demand.
+//! `encoding` is a lazy serialization library supporting zero-copy (de)serialization
```
matklad: What meaning does the word "lazy" has here? I think this is **eager** serialization library, in that all messages are validated immediately, so deserialization is O(N) (in contrast to cap'n'protos O(1) deserialization).

Perhaps replace lazy with zero-copy, or just drop it?

exonum/src/encoding/mod.rs
```diff
@@ -12,23 +12,26 @@
...
-//! - Body - dynamic sized, known only after parsing header, part.
+//! Structures are in the root of any serializable Exonum object.
+//! Binary representation of structures is split into two main parts:
 //!
-//! For easy creating this structures,
-//! in exonum you can use macros `message!` and `encoding_struct!`
-//! #Examples
-//! Imagine structure with just two field's
+//! - **Header:** a fixed-sized part
+//! - **Body:** dynamic-sized part, known only after parsing the header
```
matklad: s/dynamic-sized/dynamically sized/? As in https://doc.rust-lang.org/beta/reference/dynamically-sized-types.html

slowli:
Ready for review. I've concentrated on most "public-facing" interfaces as of now, such as `Service`, `Transaction`, `Database`, `Snapshot`,  `message!` and `encoding_struct!` macros, etc. Besides the `message!` example, I've changed other 2:

- The `Service` example is consolidated and made more complete
- The `StorageKey` example is fixed; i would assess the previous version as wrong (the docs talk about how BE encoding should be used for integers for natural sorting - which doesn't work with signed integers, btw - but then  LE was used in the example).

exonum/src/blockchain/service.rs
```diff
@@ -32,128 +32,177 @@ use node::{ApiSender, Node, State, TransactionSend};
...
+    /// checking the message signature (via [`verify_signature`]) and, possibly,
+    /// other internal constraints. `verify` has no access to the blockchain state;
+    /// checks involving the blockchains state must be preformed in [`execute`](#tymethod.execute).
+    ///
+    /// If a transaction fails `verify`, it is considered incorrect and cannot be included into
     /// any correct block proposal. Incorrect transactions are never included into the blockchain.
     ///
     /// *This method should not use external data, that is, it must be a pure function.*
+    ///
+    /// [`verify_signature`]: ../messages/trait.Message.html#method.verify_signature
```
matklad: FYI, the nightly Rust recently gained ability to refer to other items via paths like `exonum::messages::Message::verify_signature`, so we'll be able to get rid of this hand-linking some time in the future.

slowli:
@DarkEld3r Do I need to add anything to the changelog? On the one hand, this PR is confined to docs; on the other, it fixes a couple of bugs (in the docs).

As a tangentially related note: I've noticed that the 1.23 toolchain outputs warnings when generating docs, which is related to [the transition to a new Markdown parser](https://github.com/rust-lang/rust/issues/44229) in `rustdoc`. You can see the warnings in Travis builds, e.g., [here](https://travis-ci.org/exonum/exonum/jobs/334355382). I did not deal with this issue in this PR.

DarkEld3r:
@slowli I have added changelog entry into #437, please take a look if everything is OK. Plus I'm ready to merge this pull request as soon as you update it.

PR #442, opened 2018-01-30
Introduce CryptoHash trait

matklad:
A **refactoring** PR.

Problem:

`Message::hash` and `StorageValue::hash` are two different functions which must behave the same nonetheless, which introduces a non-trivial invariant into the code. 

Solution:

Extract `hash` method to a separate small trait and use it everywhere. 

Questions:

- What is the best name for this trait? The natural choice is `Hash`, but it clashes both with the `Hash` struct and with `std::hash::Hash`. Another alternative is `Digest`, but it doesn't match the `fn hash` name. And changing `fn hash` to `fn digest` requires a much larger diff, and `fn hash(&self) -> Hash` reads better than `fn digest(&self) -> Hash`. My preferred solution is to use the current `CryptoHash`.

- Quite a few types implement an inherent version of the `hash` method (for example, all of `encoding_struct!` generated types). Should they only implement a trait instead? I would say yes, but I want to gather consensus before doing this larger refactoring.

This is better be merged after the release :)

### DIFF
--- a/exonum/src/crypto.rs
+++ b/exonum/src/crypto.rs
@@ -26,10 +26,12 @@ use sodiumoxide::crypto::sign::ed25519::{gen_keypair as gen_keypair_sodium, keyp
                                          PublicKey as PublicKeySodium,
                                          SecretKey as SecretKeySodium, Seed as SeedSodium,
                                          Signature as SignatureSodium, State as SignState};
-use sodiumoxide::crypto::hash::sha256::{hash as hash_sodium, Digest, State as HashState};
+use sodiumoxide::crypto::hash::sha256::{hash as hash_sodium, Digest as DigestSodium,
+                                        State as HashState};
 use sodiumoxide;
 use serde::{Serialize, Serializer};
 use serde::de::{self, Deserialize, Deserializer, Visitor};
+use byteorder::{ByteOrder, LittleEndian};
 
 use encoding::serialize::FromHex;
 
@@ -127,6 +129,17 @@ pub fn verify(sig: &Signature, data: &[u8], pubkey: &PublicKey) -> bool {
 pub fn hash(data: &[u8]) -> Hash {
     let dig = hash_sodium(data);
     Hash(dig)
+}
+
+/// A common trait for the ability to compute a
+/// cryptographic hash.
+pub trait CryptoHash {
+    /// Returns a hash of the value.
+    ///
+    /// The hashing strategy must satisfy the basic requirements of cryptographic hashing:
+    /// equal values must have the same hash and not equal values must have different hashes
+    /// (except for negligible probability).
+    fn hash(&self) -> Hash;
 }
 
 /// Initializes the sodium library and chooses faster versions of the primitives if possible.
@@ -393,7 +406,7 @@ implement_public_sodium_wrapper! {
 /// # drop(hash_from_data);
 /// # drop(default_hash);
 /// ```
-    struct Hash, Digest, HASH_SIZE
+    struct Hash, DigestSodium, HASH_SIZE
 }
 
 implement_public_sodium_wrapper! {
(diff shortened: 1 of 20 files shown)


matklad:
Also, an interesting observation:

The `Message` does not inherit from the `StorageValue` trait, however all messages are in fact `StorageValue`s. Moreover, `StorageValue: Sized`. Looks like the original idea was to make `Message: StorageValue`, and that failed because `Message` must be object safe, and `StorageValue` is not. However, `StorageValue` could be made object safe, if we move `: Sized` bound to the `from_bytes` method.

If we do `Message: StorageValue` refactoring, than the problem of two hashes would be solved without introduction of a new trait. However, I still feel that extracting hashable types into a separate trait makes sense: there are no strong arguments, but small one-method traits are generally preferred in Rust.

exonum/src/crypto.rs
```diff
@@ -129,6 +131,17 @@ pub fn hash(data: &[u8]) -> Hash {
     Hash(dig)
 }
 
+/// A common trait for the ability to compute a
+/// cryptographic hash.
+pub trait CryptoHash {
```
slowli: Tbh, I would like something like `Hashable` more than `*Hash`. `*Hash` looks like a result of hashing, which is strengthened by the fact that we use `Hash` for a result of hashing. OTOH, `Hash` from the standard library denotes something that can be hashed, so I'm not really sure.

exonum/src/crypto.rs
```diff
@@ -529,6 +542,97 @@ impl Default for Hash {
     }
 }
 
+impl CryptoHash for u8 {
+    fn hash(&self) -> Hash {
+        hash(&[*self])
+    }
+}
+
+impl CryptoHash for u16 {
```
slowli: Do you think it could make sense to use a macro to implement the trait for integers? IMO, it could decrease the amount of copy-pasted code, but could be an overkill.

DarkEld3r:
I like both `CryptoHash` name (because it adds additional semantical context comparing to plain `Hash`) and the idea of making `StorageValue` `Sized`.

I'm going to make a more thorough review (after the release). :sweat_smile:

alekseysidorov:
I think if for generate implementations for scalars by macros, then the code will become more clear.

exonum/src/crypto.rs
```diff
@@ -529,6 +542,97 @@ impl Default for Hash {
...
+
+impl CryptoHash for i64 {
+    fn hash(&self) -> Hash {
+        let mut v = [0; 8];
+        LittleEndian::write_i64(&mut v, *self);
+        hash(&v)
+    }
+}
+
+impl CryptoHash for () {
```
slowli: I cannot help but notice that `()` and `Hash` implementations of the trait are somewhat inconsistent: a SHA-256 hash of the empty byte array does not equal `[0; 32]`, and it's not obvious why `Hash` values shouldn't be hashed.

exonum/src/crypto.rs
matklad: Trait names in Rust, especially for single-method traits, are verbs: it is `Clone`, `Debug` and not `Clonable` and `Debuggable`. So I am pretty strongly against `able`izms :)

slowli: Half-serious note: It could be most logical to rename `Hash` to `Digest` / `CryptoDigest` then, as [it's already named](https://dnaq.github.io/sodiumoxide/sodiumoxide/crypto/hash/sha256/struct.Digest.html) in `sodiumoxide`.

exonum/src/crypto.rs
matklad: Hm, fascinating! 🤔 

I would think that hashes should be hashed: one of the requirements of the cryptographic hash functions is that you can't reconstruct the original value significantly more efficient than by just enumerating all possible values, and identity function certainly does not have this property. 

For `()`, hashing an empty array for consistency with `StorageValue` also makes sense. Basically, hashing should work as if there were a blanket `impl<T: StorageValue> CryptoHash for T { ... }`.

Should we make both changes? I would say yes, but his would be a storage breaking change :( 

Should we make these changes in this PR? I'd say no :)

exonum/src/crypto.rs
matklad: So that this looks like

```
trait CryptoDigest {
    fn digest(&self) -> Digest;
}
```

? 

Yeah, I think I like these names slightly more than `Hash` ones. The diff will be massive though, but that's OK :-)

matklad:
Should we rename `Hash` struct to `Digest`, `CryptoHash` to `CryptoDigest` and `fn hash` to `fn digest()`?

```rust
trait CryptoDigest {
    fn digest(&self) -> Digest;
}
```

Poll 👇

exonum/src/crypto.rs
slowli: I meant rather something like

```rust
trait CryptoHash {
    fn hash(&self) -> CryptoDigest;
}
```

(or the variant without `Crypto` prefixes, as per [avoiding stuttering guidelines](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-[rfc-356])). This would be most close to `sodiumoxide`; it uses [`hash(&[u8]) -> Digest`](https://dnaq.github.io/sodiumoxide/sodiumoxide/crypto/hash/sha256/fn.hash.html).

exonum/src/crypto.rs
slowli: Funnily enough, I made the opposite decision in #443. Quite probably, I'm overly fascinated with meta-programming 😄

exonum/src/crypto.rs
matklad: I wonder what was the original reasoning behind renaming `Hash` to `Digest` though?..

Overall, I don't have particularly strong feelings here: any combination of names would do, and the status quo has the obvious advantage of being a status quo.

However, if we do change the names, I would like to change the method's name from `hash` to `digest`, to avoid a clash with `std::hash::Hash::hash`

slowli: If I understand correctly, the most proper term for the result of hashing is *hash digest*; calling it *hash* is a colloquialism (see [Wikipedia](https://en.wikipedia.org/wiki/Hash_function)). I'm not particularly against staying with `Hash` because of the status quo, even as `Hash` vs `CryptoHash` distinction looks a bit confusing imo.

exonum/src/crypto.rs
DarkEld3r: I'm also fond of metaprogramming, but rust macros disgust me (yeap, even comparing with C++ template metaprogramming). :worried:

exonum/src/crypto.rs
DarkEld3r: Storage breaking is OK for me.

Separate PR is a must. Would you like to do that? There are other people if not. :smile:

matklad: Hm, @vldm pointed out that `Message::hash` and `StorageValue::hash` have slightly different usages.

`StorageValue::hash` is primarily used in indices like `ValueSetIndex`, and for this use-case you probably don't need strong cryptography guarantees, besides "effectively unique" hash. That is, it is OK to use identify function, and it is even probably ok to *not* hash integers and gain some performance out of it.

`Message::hash` on the other hand is used to identify transactions, and must actually be a hash of raw bytes.

However, using the message-style hash for both use-cases seems valid (and this is done for integers already), so I suggest that we do use approach in this PR, and do change `CryptoHash` for `()` and `Hash` in a follow up PR.

PR #443, opened 2018-01-30
Storage key for signed ints

slowli:
This PR changes the `StorageKey` implementation for signed integer types in order to order them in the natural order. Previously, these type were encoded in big-endian, which means that say `-1i8` (encoded as `0xff`) was considered greater than `127i8` (encoded as `0x7f`).

After the change, the signed integer value is shifted to the range of the corresponding signed integer. It's easy to see that this is the *only* way to order signed ints in the natural order without increasing the size of allocated bytes for the int.

Technically, this is a breaking change. However, I'm having doubts anyone has used `StorageKey` for signed ints because of the unnatural sorting order. Obviously, this needs to be discussed further.

### Checklist:

There are unit tests included into the PR. Maybe, it could make sense to create an integration test with a `MapIndex` with signed integer keys.

### DIFF
--- a/exonum/src/storage/keys.rs
+++ b/exonum/src/storage/keys.rs
@@ -115,110 +115,64 @@ impl StorageKey for u8 {
     }
 }
 
-/// Uses big-endian encoding.
-impl StorageKey for u16 {
-    fn size(&self) -> usize {
-        2
-    }
-
-    fn write(&self, buffer: &mut [u8]) {
-        BigEndian::write_u16(buffer, *self)
-    }
-
-    fn read(buffer: &[u8]) -> Self {
-        BigEndian::read_u16(buffer)
-    }
-}
-
-/// Uses big-endian encoding.
-impl StorageKey for u32 {
-    fn size(&self) -> usize {
-        4
-    }
-
-    fn write(&self, buffer: &mut [u8]) {
-        BigEndian::write_u32(buffer, *self)
-    }
-
-    fn read(buffer: &[u8]) -> Self {
-        BigEndian::read_u32(buffer)
-    }
-}
-
-/// Uses big-endian encoding.
-impl StorageKey for u64 {
-    fn size(&self) -> usize {
-        8
-    }
-
-    fn write(&self, buffer: &mut [u8]) {
-        BigEndian::write_u64(buffer, *self)
-    }
-
-    fn read(buffer: &[u8]) -> Self {
-        BigEndian::read_u64(buffer)
-    }
-}
-
-/// **Not sorted in the natural order.**
+/// Uses encoding with the values mapped to `u8`
+/// by adding the corresponding constant (`128`) to the value.
 impl StorageKey for i8 {
     fn size(&self) -> usize {
         1
     }
 
     fn write(&self, buffer: &mut [u8]) {
-        buffer[0] = *self as u8
+        buffer[0] = self.wrapping_add(i8::min_value()) as u8;
     }
 
     fn read(buffer: &[u8]) -> Self {
-        buffer[0] as i8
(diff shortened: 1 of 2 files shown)


exonum/src/storage/keys.rs
```diff
@@ -115,110 +111,64 @@ impl StorageKey for u8 {
...
     }
 
     fn write(&self, buffer: &mut [u8]) {
-        buffer[0] = *self as u8
+        buffer[0] = self.wrapping_add(i8::min_value()) as u8;
     }
 
     fn read(buffer: &[u8]) -> Self {
-        buffer[0] as i8
+        buffer[0].wrapping_add(i8::min_value() as u8) as i8
```
matklad: Nit: I would expect `write` and `read` to be complimentary and to use `add` / `sub`, even though the effect should be exactly the same.

exonum/src/storage/keys.rs
```diff
@@ -276,3 +226,69 @@ impl StorageKey for String {
...
+    // Number of samples for fuzz testing
+    const FUZZ_SAMPLES: usize = 10_000;
+
+    macro_rules! test_storage_key_for_int_type {
+        (full $type:ident, $size:expr => $test_name:ident) => {
+            #[test]
+            fn $test_name() {
+                // Roundtrip
+                let mut buffer = [0u8; $size];
+                for x in $type::min_value()..$type::max_value() {
```
matklad: This misses the `max_value` value itself. I don't know a beautiful way to write it correctly, only a kludge of `.chain(once($type::max_value()))`

exonum/src/storage/keys.rs
```diff
@@ -276,3 +226,69 @@ impl StorageKey for String {
...
+                }
+            }
+        };
+        (fuzz $type:ident, $size:expr => $test_name:ident) => {
+            #[test]
+            fn $test_name() {
+                use rand::{Rng, thread_rng};
+                let mut rng = thread_rng();
+
+                // Fuzzed roundtrip
```
matklad: Let's explicitelly check min and max value as well?

exonum/src/storage/keys.rs
```diff
@@ -276,3 +226,69 @@ impl StorageKey for String {
...
+                    let x: $type = rng.gen();
+                    x.write(&mut buffer);
+                    assert_eq!($type::read(&buffer), x);
+                }
+
+                // Fuzzed ordering
+                let (mut x_buffer, mut y_buffer) = ([0u8; $size], [0u8; $size]);
+                let mut vals: Vec<$type> = rng.gen_iter().take(FUZZ_SAMPLES).collect();
+                vals.sort();
+                for w in vals.windows(2) {
```
matklad: Are we using the `itertools` create already? If we do, that we can

```
for (x, y) in vals.tuples() {
}
```

exonum/src/storage/keys.rs
```diff
@@ -276,3 +226,69 @@ impl StorageKey for String {
...
+                    let (x, y) = (w[0], w[1]);
+                    x.write(&mut x_buffer);
+                    y.write(&mut y_buffer);
+                    assert!(x_buffer < y_buffer);
```
matklad: Hm, do the tests pass? It should read `<=` I would thing?

slowli: Seems like it works because of a small sample size 🙃 I've increased it 10x. Regarding `proptest`, I think it requires overhaul of testing in general, so I would prefer not to deal with it rn.

exonum/src/storage/keys.rs
DarkEld3r: We aren't using `itertools` for now. Adding dependency for just one line seems an overkill to me. If you see more use-cases, then why not.

PR #446, opened 2018-01-31
Add API to switch consensus interaction on/off [ECR-293]

ozkriff:
~~what's left to be done:~~
- [x] ~~Fix code's formatting~~
- [x] ~~Squash the commits and rebase on top of the master branch.~~
- [x] ~~Make `testkit` compile again (after a rebase on top of master)~~

Questions:
- Are things named ok?
- Should I add more tests?
- Should I add more logs?
- Is it ok to expose `Sandbox::set_time`? Should it be `&mut self` or just `&self`?
- Is it ok that `Sandbox::process_events` is now public?

You can test it locally using something like this:
- `curl -H "Content-Type: application/json" -X POST -d '{"enabled":false}' http://127.0.0.1:8010/api/system/v1/consensus_enabled`


exonum/src/node/mod.rs
```diff
@@ -110,6 +112,8 @@ pub struct NodeHandler {
     /// Known peer addresses.
     // TODO: move this into peer exchange service
     pub peer_discovery: Vec<SocketAddr>,
+    /// Does this node participates in the consensus?
```
vitvakatu: participate

exonum/src/api/private/system.rs
```diff
@@ -148,6 +147,16 @@ impl SystemApi {
         self.info.clone()
     }
 
+    fn get_consensus_enabled_info(&self) -> bool {
+        self.shared_api_state.is_enabled()
+    }
```
matklad: I think the naming conventions for getter/setters is `foo/set_foo`? Also, the `info` word does not any information, so I suggest renaming to `consensus_enabled/set_consensus_enabled`.

exonum/src/blockchain/service.rs
```diff
@@ -340,6 +341,18 @@ impl SharedNodeState {
...
+    pub fn is_enabled(&self) -> bool {
+        let state = self.state.read().expect("Expected read lock.");
+        state.is_enabled
+    }
+
+    /// Informs internal state about node's halting.
+    pub fn update_is_enabled(&self, is_enabled: bool) {
+        let mut state = self.state.write().expect("Expected read lock.");
+        state.is_enabled = is_enabled;
+    }
```
matklad: Similar here, `enabled/set_enabled`.

I am not sure if the styleguide required to have the `is_` prefix... At least in Cargo, we don't use `is_`: https://github.com/rust-lang/cargo/blob/df11070cb574a64b167d36b012df50cf62e061b9/src/cargo/util/config.rs#L502-L516. However, we don't use `rustfmt` either 😅

exonum/src/node/consensus.rs
```diff
@@ -708,6 +717,9 @@ impl NodeHandler {
 
     /// Handles request timeout by sending the corresponding request message to a peer.
     pub fn handle_request_timeout(&mut self, data: &RequestData, peer: Option<PublicKey>) {
+        if !self.is_enabled {
+            return;
+        }
```
matklad: This is most likely over engineered, but a fun way to avoid this repetitive `is_enabled` checks would be to have, instead of `is_enabled`, the field like

```
consensus_handler: Option<ConsensusHander>
```

and make these `handle_xxx` methods the methods of `ConsensusHandler`, so that you simply can't call them if consensus is not enabled.

exonum/src/node/mod.rs
```diff
@@ -662,9 +672,7 @@ impl TransactionSend for ApiSender {
             return Err(io::Error::new(io::ErrorKind::Other, msg));
         }
         let msg = ExternalMessage::Transaction(tx);
-        self.0.clone().send(msg).wait().map(drop).map_err(
-            into_other,
-        )
+        self.send_external_message(msg)
```
matklad: 👍

exonum/src/api/private/system.rs
ozkriff: I've used `get_` and `_info` here to match the surrounding code. This naming convention is used in the current file and in [`public/system.rs`](https://github.com/exonum/exonum/blob/0ff5c1a81/exonum/src/api/public/system.rs#L71-L99)

matklad: :( 

We should really rename those as well...

exonum/src/node/consensus.rs
vitvakatu: Personally I think it'll be over-engineering.

exonum/src/node/events.rs
```diff
@@ -49,16 +52,40 @@ impl NodeHandler {
     fn handle_api_event(&mut self, event: ExternalMessage) {
         match event {
             ExternalMessage::Transaction(tx) => {
+                if !self.is_enabled {
+                    info!("Ignoring the transaction {:?}", tx);
```
vitvakatu: Hmm, I'm not sure, maybe we should also print *why* the transaction is ignored. It can be confusing otherwise.
But consider this as a nitpick

exonum/src/node/events.rs
```diff
@@ -49,16 +52,40 @@ impl NodeHandler {
...
+                    info!("Ignoring the connect message to {}", address);
+                    return;
+                }
                 info!("Send Connect message to {}", address);
                 self.connect(&address);
             }
+            ExternalMessage::Enable(value) => {
+                let s = if value { "enabled" } else { "disabled" };
+                if self.is_enabled == value {
+                    info!("Node is already is already {}", s);
```
vitvakatu: Wow, too many "is already" for my taste 🤔

exonum/src/api/private/system.rs
DarkEld3r: I also strongly dislike such naming. This function should be named `is_consensus_enabled`.

I understand your concern about naming consistency, although prefer to make at least some improvements, even if that leads to inconsistency. All these things can be renamed in a separate pull request if you insist. 😄

exonum/src/blockchain/service.rs
DarkEld3r: I like `is_` prefix. 😄 Plus we have things like `is_empty`, `is_char_boundary`, etc. in the standard library.

sandbox/src/sandbox.rs
```diff
@@ -297,8 +302,24 @@ impl Sandbox {
         self.broadcast_to_addrs(msg, self.addresses.iter().skip(1));
     }
 
+    pub fn broadcast_nopanic<T: Message>(&self, msg: &T) -> Result<(), String> {
```
DarkEld3r: It seems that spellcheck doesn't like "nopanic", neither do I. 😄 Maybe follow the `from`/`try_from` pattern (`try_broadcast`)?

exonum/src/api/private/system.rs
ozkriff: I'll create another issue+PR for this.

exonum/src/blockchain/service.rs
ozkriff: I'll create another issue+PR for this.

exonum/src/api/private/system.rs
```diff
@@ -186,8 +195,40 @@ impl Api for SystemApi {
...
+                        "Required boolean parameter 'enabled' is missing".into(),
+                    ))?
+                }
+            }
+        };
+
         router.get("/v1/peers", peers_info, "peers_info");
         router.post("/v1/peers", peer_add, "peer_add");
         router.get("/v1/network", network, "network_info");
+        router.get(
```
slowli: This is subjective, but I don't quite like single-purpose minimalistic endpoints if there is no way to aggregate them (e.g., GraphQL). In this particular case, it seems that the consensus switch could be included into the `/v1/network` endpoint.

exonum/src/api/private/system.rs
```diff
@@ -186,8 +195,40 @@ impl Api for SystemApi {
...
 
+        let self_ = self.clone();
+        let consensus_enabled_info = move |_: &mut Request| -> IronResult<Response> {
+            let info = self_.get_consensus_enabled_info();
+            self_.ok_response(&::serde_json::to_value(info).unwrap())
+        };
+
+        let self_ = self.clone();
+        let consensus_enabled_set = move |req: &mut Request| -> IronResult<Response> {
+            let map = req.get_ref::<Params>().unwrap();
```
slowli: Do I understand correctly that the parameter is parsed from the URL rather than the body of the request? Is this intended? I'd say that using the body could be more flexible and is more commonplace for POST requests.

ozkriff: `curl -H "Content-Type: application/json" -X POST -d '{"enabled":false}' http://127.0.0.1:8010/api/system/v1/consensus_enabled`

I've used this for tests, so no, I'm pretty sure that it uses the body for parameters.

slowli: Judging from [docs](https://docs.rs/params/0.8.0/params/struct.Params.html), it seems that `Params` uses multiple sources available. Are you sure this is appropriate? It seems that the `bodyparser` crate could be used for more targeted parsing.

DarkEld3r: ^ @aleksuss

aleksuss: Agree with @slowli. `bodyparser` crate is more appropriate. Take a look at this example:
https://github.com/exonum/cryptocurrency/blob/master/src/lib.rs#L276

ozkriff: Ok. I'll create another issue+PR for this.

exonum/src/api/private/system.rs
ozkriff: Yeah, but this is supposed to be a just temporary solution for node's control.

PR #448, opened 2018-02-01
Move testkit inside exonum repository preserving history

DarkEld3r:
I have the following repository structure in mind:
```
└── exonum (monorepo)
    ├── exonum
    │   ├── ...
    │   └── ...
    ├── testkit
    │   ├── ...
    │   └── ...
    ├── services
    │   ├── configuration
    │   │   ├── ...
    │   │   └── ...
    │   ├── time
    │   │   ├── ...
    │   │   └── ...
    │   └── ...
    ├── demos
    │   ├── cryptocurrency
    │   │   ├── ...
    │   │   └── ...
    │   └── ...
    ├── .travis.yml
    ├── ...
    └── README.md
```

PR #453, opened 2018-02-01
Integrate spell-checker and fix issues [ECR-357]

vitvakatu:
Probably this should wait til #448 landing


exonum/src/api/mod.rs
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 //! `RESTful` API and corresponding utilities.
+// spell-checker:ignore Tful
```
ozkriff: Hmmm, is this really a good idea to add all these annotations into the Rust sources?

vitvakatu: If we look at this specific example, `Tful` can be a typo in another file, so it's better to ignore it only locally.

vitvakatu: In case we decide to remove all the lints from source files, you'll be forced to add every test string (like "faedjaef") to the dictionary, which seems weird.

ozkriff: Maybe Cspell has something like an external config for such per-file settings? :(

vitvakatu: No, there's no. Either global config, or inline lints

ozkriff: That's sad. Ok, in that case, inline directives for rare special cases seem fine.

matklad: Though, almost none of the `ignore`s in this PR seem like special cases to me. Would adding `RESTful` to the dictionary work in this case? 

We have quite a few Alexeys around here, so perhaps it shouldn't be a special case either: https://github.com/exonum/exonum/pull/453/files#diff-586e073566ba414fe35c6b99e404b3a5R15? :) 

And like almost all other cases seem similar :) 


>Tful can be a typo in another file

I would say that probability of this is negligible, and I personally would prefer a `Tful` typo in other file to this `ignore: ` thing. Like, supressing the typo creates more noise than the typo itself.

vitvakatu: Probably my mistake, I've tried `RESTful` before and it didn't work. 
But it works now though, so okay.

vitvakatu: > We have quite a few Alexeys around here, so perhaps it shouldn't be a special case either: 

What do you mean? Should I add exception to the dictionary?
Sandbox code is one special case because of its comments. I don't have enough info to fix them, but there're a lot of exceptional words in there.

DarkEld3r: I think that the comments about "Alexey" can be easily removed.

vitvakatu: @DarkEld3r done

vitvakatu:
It looks like Travis has old npm version, needs investigation

slowli:
The problem is seemlingly not with npm (package manager), but with Node / JS interpreter (it does not understand `async` keyword). You may use `nvm use 8` to switch to Node 8, which understands `async`.

vitvakatu:
@slowli thank you!

PR #456, opened 2018-02-01
Add `SystemTime` to storage [ECR-420]

SerhiiOryshych:


exonum/src/storage/keys.rs
```diff
@@ -388,4 +410,86 @@ mod tests {
...
+
+        let (mut buffer1, mut buffer2) = ([0u8; 12], [0u8; 12]);
+        for _ in 0..FUZZ_SAMPLES {
+            let time1 = SystemTime::now();
+            let time2 = time1 + Duration::new(0, 1);
+            time1.write(&mut buffer1);
+            time2.write(&mut buffer2);
+            assert!(buffer1 < buffer2);
+        }
+    }
```
matklad: I am not sure, but I think that this test will *not* fail, if we reverse the order in which we write seconds and nanos. If this is true, that it's better to add a test that can detect such behavior. 

For example, generating `FUZZ_SAMPLE`s *random* times and than checking that the ordering is correct should work I guess?

exonum/src/storage/keys.rs
```diff
@@ -227,6 +229,26 @@ impl StorageKey for String {
...
+        let duration = self.duration_since(UNIX_EPOCH).unwrap();
+        let secs = duration.as_secs();
+        let nanos = duration.subsec_nanos();
+        secs.write(&mut buffer[0..8]);
+        nanos.write(&mut buffer[8..12]);
+    }
+
+    fn read(buffer: &[u8]) -> Self {
+        let secs = u64::read(&buffer[0..8]);
+        let nanos = u32::read(&buffer[8..12]);
```
matklad: I think `BigEndian::read_u64` would be slightly more clear here.

matklad: I am not sure, but I think we must check that naonos <= 10**9? (at least a debug_assert is required I think).

exonum/src/storage/values.rs
```diff
@@ -240,3 +241,59 @@ impl StorageValue for String {
...
 }
+
+/// Uses little-endian encoding.
+impl CryptoHash for SystemTime {
+    fn hash(&self) -> Hash {
+        let duration = self.duration_since(UNIX_EPOCH).unwrap();
+        let secs = duration.as_secs();
+        let nanos = duration.subsec_nanos();
+
+        let mut buffer = vec![0; 12];
```
matklad: I am not sure that allocating a twelve element vector is a great idea :) Could we use `[0u8; 12]` here?

exonum/src/storage/keys.rs
```diff
@@ -227,6 +229,26 @@ impl StorageKey for String {
     }
 }
 
+impl StorageKey for SystemTime {
+    fn size(&self) -> usize {
+        12
+    }
+
+    fn write(&self, buffer: &mut [u8]) {
+        let duration = self.duration_since(UNIX_EPOCH).unwrap();
```
vitvakatu: let's replace `unwrap` with `expect`

exonum/src/storage/values.rs
```diff
@@ -240,3 +241,59 @@ impl StorageValue for String {
...
+        UNIX_EPOCH + Duration::new(secs, nanos)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_storage_key_for_system_time_round_trip() {
```
slowli: The test name is wrong; should be `test_storage_value...`.

exonum/src/storage/keys.rs
slowli: :+1: I'd say "fuzz" implies randomness involved.

exonum/src/storage/keys.rs
```diff
@@ -388,4 +410,86 @@ mod tests {
         // Notice the different order of values compared to the previous test
         assert_eq!(index.values().collect::<Vec<_>>(), vec![100, 200]);
     }
+
+    #[test]
+    fn test_storage_key_for_system_time_round_trip() {
+        use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+        let times = [
```
slowli: I think we should test for the other extreme compared to `UNIX_EPOCH`, and extreme values of nanoseconds.

exonum/src/storage/keys.rs
```diff
@@ -227,6 +229,26 @@ impl StorageKey for String {
     }
 }
 
+impl StorageKey for SystemTime {
```
slowli: I'd specify how serialization is implemented.

exonum/src/storage/keys.rs
slowli: Yeah, if `SystemTime` can be received from an external source, we need to do *something* with the fact that `nanos` can be greater than `1e9`. If left unchecked, it could break canonicity property of serialization:

```rust
let t = UNIX_EPOCH + Duration::new(0, 1_000_000_001);
let other_t = UNIX_EPOCH + Duration::new(1, 1);
assert_eq!(t, other_t);
```

DarkEld3r: > I think BigEndian::read_u64 would be slightly more clear here.

Not sure. In a hypothetical situation when we decide to store everything using `LittleEndian` it will be one more place that should be fixed. I prefer to reuse the implementation.

DarkEld3r: @slowli As far as I can see, `Duration` [performs normalization internally](https://doc.rust-lang.org/std/time/struct.Duration.html#method.new): 
> If the number of nanoseconds is greater than 1 billion (the number of nanoseconds in a second), then it will carry over into the seconds provided.

slowli: @DarkEld3r Yeah, I know, this is what my example is about 🙂 My problem is that `(0u64, 1_000_000_001u32)` and `(1u64, 1u32)` encode exactly the same `SystemTime` while having different binary presentations (which, afaik, is our definition of non-canonicity). That is, the roundtrip with `SystemTime` does not necessarily work, at the very least.

cc @vldm @DenisKolodin

exonum/src/storage/keys.rs
matklad: > I'd say "fuzz" implies randomness involved.

Generating radom samples and than adding one nanosecond to them would not help to check that seconds and nanons are compared in the correct order. It's important to compare two random value.

exonum/src/storage/keys.rs
matklad: @slowli I believe that the current code is correct, because it is assumed that bytes inside the storage are trusted (I am not sure about the validity of this assumption, but this is how the code works today). We don't validate messages we read from storage either. 

However, I am all for adding an assertion here, for the purposes of documenting the assumption (and for a 100% impossible case of a bug somewhere, which violates the assumption :) )

slowli: I've mistaken `StorageKey` / `StorageValue` with message-related stuff; so the concerns regarding canonicity should be addressed to it. However, there is one more question regarding ordering of `SystemTime` in `StorageKey`s: without checking `nanos` bounds, it would lead to incorrect ordering in some cases (e.g., `(0u64, 2_000_000_001u32)` is greater than `(1u64, 0u32)`). So it looks like the assertion is a must.

exonum/src/storage/keys.rs
```diff
@@ -431,15 +442,32 @@ mod tests {
...
+                    rng.gen::<u32>() % 1_000_000_000,
+                );
+            let time2 = UNIX_EPOCH +
+                Duration::new(
+                    rng.gen::<u64>() % (i32::max_value() as u64),
+                    rng.gen::<u32>() % 1_000_000_000,
+                );
             time1.write(&mut buffer1);
             time2.write(&mut buffer2);
-            assert!(buffer1 < buffer2);
```
matklad: assert_eq!(
    time1.cmp(&time2),
    buffer1.cmp(&buffer2),
)

exonum/CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ All notable changes to this project will be documented in this file. The project
   around a type (e.g., `struct QuirkyI32Key(i32)`) and implement `StorageKey` for it
   using big endian encoding. Then, use the wrapper instead of the int type in indices.
   See the unit tests for `StorageKey` for an example. (#443)
+  
+### New features
+- `StorageKey` and `StorageValue` are implemented for `SystemTime`. (#456)
```
DarkEld3r: I propose to add the word "traits": `StorageKey` and `StorageValue` _traits_ are...

exonum/src/storage/keys.rs
```diff
@@ -24,6 +26,13 @@ use crypto::{Hash, PublicKey, HASH_SIZE, PUBLIC_KEY_LENGTH};
 /// will not sort signed integer types in the natural order; therefore, they are
 /// mapped to the corresponding unsigned type by adding a constant to the source value.
 ///
+/// The `SystemTime` (de)serialization is only possible for the values greater than
+/// 1970-01-01 00:00:00 UTC. Since the `SystemTime` type saves time in the form
+/// `(seconds: u64, nanoseconds: u32)`, the total occupied size of `SystemTime` in the storage
+/// is 12 bytes, the time is stored as a buffer of bytes `buffer[0..8]: &[u8]`.
```
matklad: `buffer[0..8]: &[u8]`. -> `buffer[0..12]: &[u8]`. ?

exonum/src/storage/keys.rs
```diff
@@ -24,6 +26,13 @@ use crypto::{Hash, PublicKey, HASH_SIZE, PUBLIC_KEY_LENGTH};
 /// will not sort signed integer types in the natural order; therefore, they are
 /// mapped to the corresponding unsigned type by adding a constant to the source value.
 ///
+/// The `SystemTime` (de)serialization is only possible for the values greater than
```
slowli: IMO, this paragraph better belongs to the `impl StorageKey for SystemTime` block.

exonum/src/storage/keys.rs
slowli: I don't quite get the meaning of "the time is stored..." part; imo, it could be safely removed.

exonum/src/storage/keys.rs
```diff
@@ -24,6 +26,13 @@ use crypto::{Hash, PublicKey, HASH_SIZE, PUBLIC_KEY_LENGTH};
...
+/// 1970-01-01 00:00:00 UTC. Since the `SystemTime` type saves time in the form
+/// `(seconds: u64, nanoseconds: u32)`, the total occupied size of `SystemTime` in the storage
+/// is 12 bytes, the time is stored as a buffer of bytes `buffer[0..8]: &[u8]`.
+/// Accordingly, the seconds are stored as a buffer of bytes `buffer[0..8]: &[u8]`,
```
slowli: I'd rephrase

> The seconds are stored in the first 8 bytes as per the `StorageKey` implementation for `u64`, and nanoseconds in the remaining 4 bytes as per the `StorageKey` implementation for `u32`.

exonum/src/storage/keys.rs
```diff
@@ -390,4 +424,106 @@ mod tests {
...
+            index.iter_from(&UNIX_EPOCH).collect::<Vec<_>>(),
+            vec![(x2, y2), (x1, y1)]
+        );
+        assert_eq!(
+            index
+                .iter_from(&(UNIX_EPOCH + Duration::new(20, 0)))
+                .collect::<Vec<_>>(),
+            vec![(x1, y1)]
+        );
+        assert_eq!(
```
slowli: Nitpick: I feel this assert is redundant (it tests the same case as the previous one). Maybe, change it to match `x1` exactly?

exonum/src/storage/values.rs
```diff
@@ -240,3 +241,72 @@ impl StorageValue for String {
         String::from_utf8(value.into_owned()).unwrap()
     }
 }
+
+/// Uses little-endian encoding.
+impl CryptoHash for SystemTime {
```
slowli: It seems other `CryptoHash` implementations are in `crypto` module. Is there a particular reason this one is here?

PR #459, opened 2018-02-02
Testkit server

slowli:
Reincarnation of a pull request in the `exonum-testkit` repository. Allows to run the testkit with an interface similar to that of a real Exonum node, but with an additional testkit-specific APIs controlling creation of blocks and the like.

The server mode could be especially useful together with [the blockchain explorer](https://github.com/exonum/blockchain-explorer), which could implement testkit-specific UI. Also, it could allow to perform reproducible client-side testing.

New feature (non-breaking change which adds functionality).

### Checklist:

There are some tests present in the PR, which cover the API from the Rust side and from JS side. Probably needs more of both.


slowli:
This PR could substantially benefit from moving the cryptocurrency tutorial into this repo; then, the server example in this PR could use it directly instead of duplicating the code. See, I'm all for a mono-repository now 😄

slowli:
Sort of ready for review. Couple of questions for discussion:

- Is the REST API for the testkit adequate? I use the `DELETE` method to rollback the blockchain even after checkpoint refactoring, as this is idempotent and seems more intuitive.
- The `server` example in the testkit uses the cryptocurrency tutorial crate. If I understand correctly, this will require to publish it as a crate, which is consciously avoided as of now. Alternatives not resulting in extensive copy-paste are appreciated. I can use another service from integration tests and add it to the example with the help of `#[path = "..."]` trickery, but I feel that an example using a "real" service would be more practical (maybe, use `exonum-time`?).
- Would it make sense to integrate the testkit server as an even more developer-oriented analogue of the `run-dev` CLI command? (This would require to move CLI to a separate crate to avoid circular dependencies.)

testkit/examples/server/.gitignore
```diff
@@ -0,0 +1,2 @@
+/node_modules
+/npm-debug.log
```
DarkEld3r: Personally I prefer to have a single `.gitignore` at the repository root. What do you think?

exonum/src/explorer/mod.rs
```diff
@@ -294,32 +299,32 @@ impl BlockWithTransactions {
...
+    pub fn iter(&self) -> EagerTransactions<T> {
         self.transactions.iter()
     }
 }
 
 /// Iterator over transactions in [`BlockWithTransactions`].
 ///
 /// [`BlockWithTransactions`]: struct.BlockWithTransactions.html
-pub type EagerTransactions<'a> = ::std::slice::Iter<'a, CommittedTransaction>;
+pub type EagerTransactions<'a, T> = ::std::slice::Iter<'a, CommittedTransaction<T>>;
```
DarkEld3r: I don't like paths starting with `::`. :upside_down_face: Would you mind adding `use std::slice`?

testkit/examples/server/index.js
```diff
@@ -0,0 +1,255 @@
...
+async function getBlockchainHeight () {
+  let response = await fetch(TESTKIT_URL + '/v1/status')
+  response = await response.json()
+  return response.height
+}
+
+describe('CurrencyService', function () {
+  this.slow(500)
+
+  // // // // Service constants // // // //
```
DarkEld3r: I have little understanding of JavaScript, but such comments still confuse me. Isn't it possible (and reasonable) to split a single file into several modules?

testkit/examples/server/.gitignore
slowli: I don't have a particularly strong opinion on the topic. IMO, "local" `.gitignore`s can be more understandable and easier to maintain.

DarkEld3r: I don't have strong arguments, but I have the exactly opposing opinion. :upside_down_face: As for me, it is harder to maintain several files.

@aleksuss @ozkriff @vitvakatu @anyone
Opinions?

vitvakatu: Same thing, one gitignore to rule the world

testkit/src/handler.rs
```diff
@@ -0,0 +1,433 @@
...
+            ))?,
+        };
+        if height == 0 {
+            Err(ApiError::BadRequest(
+                "Cannot rollback past genesis block".into(),
+            ))?;
+        }
+
+        if self.height().0 >= height {
+            let rollback_blocks = (self.height().0 - height + 1) as usize;
```
vitvakatu: It might be the matter of taste, but I think it's strange. We're using `Height` starting with `0`, but you have to pass `1` when you need to rollback to genesis block.
```rust
// Test roll-back to the genesis block
        request::delete(
            "http://localhost:3000/v1/blocks/1",
            Headers::new(),
            &handler,
        ).unwrap();
        {
            let testkit = testkit.read().unwrap();
            assert_eq!(testkit.height(), Height(0)); /// <- What? I want "1" here!
        }
```

slowli: I'm not entirely sure that `DELETE` method in general is the best choice here (maybe, use something like `POST .../testkit/v1/rollback`?), but if we stick with it, it seems natural to me that `DELETE .../blocks/1` deletes the first block (i.e., the block immediately after the genesis block - the same URL scheme the explorer is using).

vitvakatu: Yep, it's okay when you read docs, but first time I was like wtf, it's mistake!

testkit/src/handler.rs
```diff
@@ -0,0 +1,433 @@
...
+        let status = TestKitStatus {
+            height: self.height(),
+            configuration: self.configuration_change_proposal(),
+            next_configuration: self.next_configuration().cloned(),
+        };
+        ok_response(&status)
+    }
+
+    fn handle_create_block(&mut self, req: &mut Request) -> IronResult<Response> {
+        #[derive(Clone, Debug, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
slowli: `Clone` is needed for `bodyparser::Struct` to work.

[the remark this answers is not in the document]
vitvakatu: My mistake, I was 100% sure it's gonna work (as it compiles, but fails when `test`ing)

PR #461, opened 2018-02-05
[Fix] Implement ExonumJsonDeserialize for floats

vitvakatu:
This fix will allow to use `Vec<F32>` and `Vec<F64>` in `message!` macros.

### DIFF
--- a/exonum/src/encoding/float.rs
+++ b/exonum/src/encoding/float.rs
@@ -22,7 +22,7 @@ use super::Result as EncodingResult;
 use super::Error as EncodingError;
 use encoding::{CheckedOffset, Field, Offset};
 use encoding::serialize::WriteBufferWrapper;
-use encoding::serialize::json::ExonumJson;
+use encoding::serialize::json::{ExonumJson, ExonumJsonDeserialize};
 
 /// Wrapper for the `f32` type that restricts non-finite
 /// (NaN, Infinity, negative zero and subnormal) values.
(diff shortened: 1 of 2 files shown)


exonum/src/encoding/float.rs
```diff
@@ -243,6 +243,13 @@ impl ExonumJson for F32 {
     }
 }
 
+impl ExonumJsonDeserialize for F32 {
+    fn deserialize(value: &Value) -> Result<Self, Box<Error>> {
+        let number = value.as_f64().ok_or("Can't cast json as float")?;
+        Ok(Self::new(number as f32))
```
matklad: This certainly should do some validation so as not to panic!?

vitvakatu: Should we? afaik `as` will never panic, can you provide the example of the opposite?

matklad: `new` panics for NaNs?

vitvakatu: Nevermind, I see it now.

matklad: Hm, even if the author of code forgets that new can panic, perhaps we should not have the new method at all? Or make it return an option?

vitvakatu: :D The whole story about floats is a complete nightmare. But we need them anyway, and new method seems to be well documented. (how am I to read docs?).

PR #463, opened 2018-02-06
ARCHITECTURE.md

matklad:
This PR is ** documentation to improve contributor experience**. It adds an ARCHITECTURE.md file which gives a one-page description of the large moving parts of Exonum. The primary purpose of this code is to allow someone to dive into code.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,83 @@
+# Exonum Architecture
+
+This document describes the high-level architecture of Exonum. Its target
+audience is developers of Exonum core. To learn how to *use* Exonum, see
+[documentation].
+
+[documentation]: https://exonum.com/doc/
+
+This document is provided on the best effort basis — it is not guaranteed to be
+up-to-date and complete. The purpose of this document is to highlight the most
+important types, modules and functions in the code, to allow new contributors to
+dive into Exonum quickly. It is assumed that the reader is familiar with core
+concepts of Exonum, such as [services] and [transactions].
+
+[services]: https://exonum.com/doc/architecture/services/
+[transactions]: https://exonum.com/doc/architecture/transactions/
+
+## Code organization
+
+The repository is a single Cargo workspace of several related packages:
+
+- `exonum`, most of the code lives here,
+
+- `sandbox`, where integration tests live,
+
+- `testkit`, a framework for testing Exonum services.
+
+This document currently describes only the `exonum` crate.
+
+## Storage
+
+Storage module can be found at `src/storage`. It provides an abstraction over an
+embedded key-value store: the `Database` trait. The keys and values are just
+slices of bytes: `&[u8]`.
+
+To transform raw bytes into native Rust data structures, the `StorageKey` and
+`StorageValue` traits are used.
+
+On top of raw storage, collections like sets, lists and maps are provided: see
+various `*Index` structs. Of particular interest are `storage/proof_map_index`
+and `storage/proof_list_index` modules, which provide Merkelized collections,
+capable of providing compact proofs for search queries.
+
+Note that indexes are used both by the Exonum core and by the services. The core
+uses indexes to store the internal blockchain state, which is described in the
+next section. The services define and use indexes to store arbitrary
+service-specific data.
+
+## Blockchain
+
+The format of the block of the Exonum blockchain is described in
+`src/blockchain/block.rs`. Blocks are stored in the `Database`. The schema at
+`src/blockchain/schema.rs` describes the indexes used for blocks, transactions
+and some other data used by Exonum core.
+
+The heart of the `Blockchain` is `Blockchain::create_patch` method, which, given
+a list of transactions, executes them and produces a new `Block`.
+
+## Node
+
+The entities responsible for bringing all the pieces together and connecting
+blockchain and services to the external world are `Node` and `NodeHandler`. Note
+that there are several `impl` blocks for `NodeHandler`.
+
+The entry point is `Node::run`, and the event dispatch loop is started in
+`Node::run_handler`.
+
+The event dispatch itself happens in `src/node/base.rs` module.
+
+Events specific to the consensus protocol messages are handled in the
+`src/node/consensus.rs` module. The messages themselves are defined in
+the `src/messages/protocol.rs` module.
+
+`Node` is also responsible for starting an HTTP API server. The API is assembled
+from the built-in part, specified in `src/api` and parts, provided by each
+service. The entry point of API construction is `create_private_api_handler` and
+`create_public_api_handler` functions.
+
+## Configuration
+
+The code responsible for assembling a node with a particular set of services and
+configuration lives in `src/helpers/fabric`. `NodeBuilder::run` is the "`fn
+main`" of Exonum.


ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
+# Exonum Architecture
+
+This document describes the high-level architecture of Exonum. Its target
+audience is the developers of Exonums core. To learn how to *use* Exonum, see
```
dmitry-timofeev: Exonum**s** ?

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+- `testkit`, a framework for testing Exonum services.
+
+
+This document currently describes only the `exonum` crate.
+
+
+## Storage
+
+Storage module can be found at `src/storage`. It provides an abstracton over an
+embeded key value store: the `Database` trait. The keys and values of are just
```
dmitry-timofeev: embedded?

dmitry-timofeev: The keys and values ~~of~~ are …

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+slices of bytes: `&[u8]`.
+
+To transform raw bytes into native Rust datastructures, the `StorageKey` and
```
dmitry-timofeev: Won't it be better to use active voice here to explicitly say _who_ uses these traits? The core doesn't need them, but a storage user -- does.

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+`StorageValue` traits are used.
+
+On top of raw storage, collections like sets, lists and maps are provided: see
```
dmitry-timofeev: The fact that everyone names these things as they want (collections, tables, views, indices), tells us that an "index" is a terrible name.

I welcome everyone aboard: https://jira.bf.local/browse/ECR-437

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+various `*Index` structs. Of particular interest are `storage/proof_map_index`
+and `storage/_proof_list_index` modules, which provide merkelized collections,
```
dmitry-timofeev: _proof -- leading underscore?

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+capable of providing compact proofs for search queries.
```
dmitry-timofeev: search or read?

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+
+Indexes are used by the core of Exonum to store the blockchain state. Service
+imlementors may use indexes to store service-specific data.
```
dmitry-timofeev: implementors (missing p)

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+## Blockchain
+
+`Blockchain` struct wraps a `Database` and provides access to blockchain related
+tables. The heart of the `Blockchain` is `Blockchain::create_patch` method,
+which, given a list of transaction, executes them and produces a new `Block`.
```
dmitry-timofeev: a list of trnasaction**s**

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+## Node
+
+The entities responsible for brining all the pieces together and connecting
```
dmitry-timofeev: brin**g**ing

ARCHITECTURE.md
```diff
@@ -0,0 +1,73 @@
...
+blockchain and services to external world are `Node` and `NodeHandler`. Note
+that there are several `impl` blocks for `NodeHandler`.
+
+The entry point is `Node::run`, and the event dispatch loop is started in
```
dmitry-timofeev: Shan't we expand on the kinds of events this thing handles?

ARCHITECTURE.md
matklad: @dmitry-timofeev: core itself uses storage to store the blockchain itself, so it does make use of StorageKey/Value. 

I'll try to rephrase this a bit and give more concrete pointers to usage sites inside the Core.

ARCHITECTURE.md
matklad: ➕  💯

ARCHITECTURE.md
```diff
@@ -0,0 +1,74 @@
...
+To transform raw bytes into native Rust data structures, the `StorageKey` and
+`StorageValue` traits are used.
+
+On top of raw storage, collections like sets, lists and maps are provided: see
+various `*Index` structs. Of particular interest are `storage/proof_map_index`
+and `storage/proof_list_index` modules, which provide merkleized collections,
+capable of providing compact proofs for search queries.
+
+Indexes are used by the core of Exonum to store the blockchain state. Service
+implementors may use indexes to store service-specific data.
```
slowli: I'd refer the reader to the built-in indexes, i.e., `blockchain/schema.rs`, right here rather than later. The notion that the core has built-in indexes is not trivial.

ARCHITECTURE.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+## Blockchain
+
+Blockchain state is stored in the `Database`. The data format is described in
```
slowli: Nitpick: Do you mean the format for the data stored by the core specifically? If so, maybe mention it explicitly?

ARCHITECTURE.md
```diff
@@ -0,0 +1,74 @@
+# Exonum Architecture
+
+This document describes the high-level architecture of Exonum. Its target
+audience is the developers of Exonum core. To learn how to *use* Exonum, see
+https://exonum.com/doc/.
+
+This document is provided on the best effort bases -- it is not guaranteed to be
```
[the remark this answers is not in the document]
matklad: It should be — rather than – though :-)

ARCHITECTURE.md
```diff
@@ -0,0 +1,74 @@
...
+up-to-date and complete. The purpose of this document is to highlight the most
+important types, modules and functions in the code, to allow new contributors to
+dive into Exonum quickly. It is assumed that the reader is familiar with core
+concepts of Exonum, such as services and transactions.
+
+
```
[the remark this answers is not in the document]
matklad: I usually separate sections (`##`) in markdown docs with two blank lines: there's no difference for rendering, but this looks better in raw text form, because it stands out more than just a single blank line between paragraphs.

ARCHITECTURE.md
```diff
@@ -0,0 +1,74 @@
...
+## Code organization
+
+The repository is a single Cargo workspace of several related packages:
```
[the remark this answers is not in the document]
matklad: There's no right answer here really because the terminology used is contradictory :( However, I personally would prefer a `package` here. For the reference, the current mess looks like this:

- `crate` is, formally, a compilation unit. It is a tree of modules, which starts at `lib.rs` / `main.rs` / `example/foo.rs` and includes other files via `mod foo` declarations. `crate` is a term, specific to the Rust language. 

- `package` is "cohesive unit of distribution" of rust code. It is a term, specific to Cargo. A package typically consists of several crates: the library in `src/lib.rs`, the binary in `src/main.rs`, the integration tests and examples. The `package` terminology is used in `Cargo.toml` (it begins with a `[package]` section) and in certain flags (`cargo test --package exonum`).

- `target`: confusingly, this is how Cargo calls crates which constitute a package. I.e., there's a library target, a binary target, and `--all-targets` flag.

- `target-tripple`: this specifies the identity of architecture for cross-compiling. It is often referred to as simply "target".

- `crate`: confusingly, this is how the `crates.io` website calls the `package`. 

The Rust book also seems to use package: https://doc.rust-lang.org/book/second-edition/ch14-03-cargo-workspaces.html

ARCHITECTURE.md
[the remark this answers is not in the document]
slowli: Not necessarily; both en dash and em dash can be used in English typography, and there are no definitive rules which one should be preferred. See [Wikipedia](https://en.wikipedia.org/wiki/Dash#En_dash_versus_em_dash). A funny thing is: I feel that in this particular case, a semicolon would be better anyway.

ARCHITECTURE.md
```diff
@@ -0,0 +1,87 @@
...
+and some other data used by Exonum core.
+
+The heart of the `Blockchain` is `Blockchain::create_patch` method, which, given
+a list of transactions, executes them and produces a new `Block`.
+
+
+## Node
+
+The entities responsible for bringing all the pieces together and connecting
+blockchain and services to external world are `Node` and `NodeHandler`. Note
```
Maria-Nosyk: "to external world" --> "to the external world"

ARCHITECTURE.md
```diff
@@ -0,0 +1,84 @@
...
+that there are several `impl` blocks for `NodeHandler`.
+
+The entry point is `Node::run`, and the event dispatch loop is started in
+`Node::run_handler`.
+
+The event dispatch itself happens in `src/node/base.rs` module. 
+
+Events specific to the consensus protocol messages are handled in the
+`src/node/consensus.rs` module. The messages themselves are defined in
+`src/messages/protocol.rs` module.
```
DarkEld3r: I'm not sure how it should be, but you are using "the" article with the first module name and no article with the second one.

ARCHITECTURE.md
```diff
@@ -0,0 +1,84 @@
...
+
+The repository is a single Cargo workspace of several related packages:
+
+- `exonum`, most of the code lives here,
+
+- `sandbox`, where integration tests live,
+
+- `testkit`, a framework for testing Exonum services.
+
+
```
dmitry-timofeev: Why two blanks?

ARCHITECTURE.md
```diff
@@ -0,0 +1,84 @@
...
+This document currently describes only the `exonum` crate.
+
+## Storage
+
+Storage module can be found at `src/storage`. It provides an abstraction over an
+embedded key value store: the `Database` trait. The keys and values are just
```
dmitry-timofeev: key-value is separated with a hyphen

ARCHITECTURE.md
```diff
@@ -0,0 +1,84 @@
...
+
+To transform raw bytes into native Rust data structures, the `StorageKey` and
+`StorageValue` traits are used.
+
+On top of raw storage, collections like sets, lists and maps are provided: see
+various `*Index` structs. Of particular interest are `storage/proof_map_index`
+and `storage/proof_list_index` modules, which provide Merkelized collections,
+capable of providing compact proofs for search queries.
+
+Note that indexes are used both by the Exonum core and by the services. The core
```
dmitry-timofeev: Which services (the seems redundant, @Maria-Nosyk please correct me if I'm wrong)?

dmitry-timofeev:
@DarkEld3r you might consider adding "Check if ARCHITECTURE.md needs to be updated" to your pre-release check-list :wink:

ARCHITECTURE.md
matklad: I think "the services which were discussed in `the reader is familiar with core concepts of Exonum`", so `the` seems slightly more correct to me.

PR #464, opened 2018-02-06
Added a method `handle_execute` in `Service` trait. [ECR-651]

aleksuss:
### Description

This feature add a possibility to modify `Fork` in the service before `handle_commit` invoke.

- [ ] Bug fix (non-breaking change which fixes an issue)

- [x] New feature (non-breaking change which adds functionality)

### Checklist:

### DIFF
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -275,6 +275,13 @@ impl Blockchain {
                 schema.tx_location_by_tx_hash_mut().put(hash, location);
             }
 
+            // Invoke handle_execute for each service
+            fork.checkpoint();
+            for service in self.service_map.values() {
+                service.handle_execute(&mut fork);
+            }
+            fork.commit();
+
             // Get tx & state hash
             let (tx_hash, state_hash) = {
                 let state_hashes = {
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -253,10 +253,7 @@ impl Blockchain {
                 );
 
                 fork.checkpoint();
-
-                let r = panic::catch_unwind(panic::AssertUnwindSafe(|| { tx.execute(&mut fork); }));
-
-                match r {
+                match panic::catch_unwind(panic::AssertUnwindSafe(|| tx.execute(&mut fork))) {
                     Ok(..) => fork.commit(),
                     Err(err) => {
                         if err.is::<Error>() {
@@ -276,11 +273,26 @@ impl Blockchain {
             }
 
             // Invoke handle_execute for each service
-            fork.checkpoint();
             for service in self.service_map.values() {
-                service.handle_execute(&mut fork);
+                fork.checkpoint();
+                match panic::catch_unwind(panic::AssertUnwindSafe(
+                    || service.handle_execute(&mut fork),
+                )) {
+                    Ok(..) => fork.commit(),
+                    Err(err) => {
+                        if err.is::<Error>() {
+                            // Continue panic unwind if the reason is StorageError
+                            panic::resume_unwind(err);
+                        }
+                        fork.rollback();
+                        error!(
+                            "handle_execute failed for a service: {} with error: {:?}",
+                            service.service_name(),
+                            err
+                        );
+                    }
+                }
             }
-            fork.commit();
 
             // Get tx & state hash
             let (tx_hash, state_hash) = {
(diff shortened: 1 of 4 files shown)


exonum/src/blockchain/mod.rs
```diff
@@ -275,6 +275,13 @@ impl Blockchain {
                 schema.tx_location_by_tx_hash_mut().put(hash, location);
             }
 
+            // Invoke handle_execute for each service
+            fork.checkpoint();
+            for service in self.service_map.values() {
+                service.handle_execute(&mut fork);
```
matklad: This calls into user code, which can panic!

matklad: Hm, interesting, the same problem happens with `handle_commit` as well...

What actually is the expected behavior here? Seems unfortunate if the bug in a single service (perhaps even specific to a particular transaction) brings down the whole exonum...

slowli: Why doesn't this handle panics in the same way as transactions do?

DarkEld3r: We need (very) carefully revise guaranties that we can and want to provide.

There are many places where panic can cause problems: `Transaction::verify` and almost all `Service` methods.

DarkEld3r:
I also prefer a more refined solution, but the proposed workaround is better than no solution at all. As for me it is OK to add this method now and improve it step by step (or propose an entirely new solution).

slowli:
@DarkEld3r If the proposed handler is a temporary solution, I think it could make sense to hide it from the docs, and add a comment that the handler in the present state is essentially a hack and should not be used unless completely necessary (and if used, should not modify the consensus part of the storage).

exonum/src/blockchain/tests.rs
```diff
@@ -273,15 +272,109 @@ fn handling_tx_panic_storage_error(blockchain: &Blockchain) {
...
+        unimplemented!()
+    }
+
+    fn handle_execute(&self, fork: &mut Fork) {
+        let mut index = ListIndex::new(IDX_NAME, fork);
+        index.push(1);
+    }
+}
+
+struct ServicePanic;
```
slowli: Nitpick: to avoid excessive copy-paste, I'd suggest using macros, but I know that @DarkEld3r is not very fond of them 🙃

exonum/src/blockchain/tests.rs
```diff
@@ -273,15 +272,109 @@ fn handling_tx_panic_storage_error(blockchain: &Blockchain) {
...
+    fn tx_from_raw(&self, _raw: RawTransaction) -> Result<Box<Transaction>, MessageError> {
+        unimplemented!()
+    }
+
+    fn handle_execute(&self, _fork: &mut Fork) {
+        panic!(Error::new("42"));
+    }
+}
+
+fn handling_handle_execute(blockchain: &Blockchain, db: &mut Box<Database>) {
```
slowli: Why are these functions called `handling*`? IMO, `assert*` would be more appropriate.

exonum/src/blockchain/tests.rs
```diff
@@ -298,6 +391,18 @@ mod memorydb_tests {
         )
     }
 
+    fn create_blockchain_with_service(_: &Path, service: Box<Service>) -> Blockchain {
+        let service_keypair = gen_keypair();
+        let api_channel = mpsc::channel(1);
+        Blockchain::new(
+            Box::new(MemoryDB::new()),
```
slowli: Why doesn't this use `create_database`? +The comment about macros is applicable here as well (albeit a bit off-topic). You could look at #380 for a reference how I envision tests with macros.

exonum/src/blockchain/mod.rs
```diff
@@ -275,6 +272,28 @@ impl Blockchain {
                 schema.tx_location_by_tx_hash_mut().put(hash, location);
             }
 
+            // Invoke handle_execute for each service
+            for service in self.service_map.values() {
+                fork.checkpoint();
+                match panic::catch_unwind(panic::AssertUnwindSafe(
```
slowli: Nitpick: This formatting looks a bit uncomfortable to me; I'd propose to use a variable for the `catch_unwind` result here.

DarkEld3r:
@slowli We've just had a little discussion with @aleksuss and @vldm. As a result we decided that use-cases, for which these chnages are needed, can be handled differently. Thus, I propose not to merge it: we have time for a more refined solution.

exonum/src/blockchain/tests.rs
aleksuss: Because `create_database` required an additional variable `Path` 😄

DarkEld3r:
@aleksuss Could you please resolve conflicts (update to master branch) and rename `handle_execute` method to `execute`.

DarkEld3r:
Superseded by #667.

PR #465, opened 2018-02-06
Move exonum-time inside exonum

SerhiiOryshych:


services/exonum-time/README.md
```diff
@@ -0,0 +1,85 @@
+# exonum-time
+
+[![Build Status](https://travis-ci.org/exonum/exonum-time.svg?branch=master)](https://travis-ci.org/exonum/exonum-time)
```
vitvakatu: Hmm... Need to fix links, right?

services/exonum-time/README.md
```diff
@@ -0,0 +1,85 @@
...
+// Gets the time of validator with a public key equal to `public_key`. 
+let validator_time = time_schema.validators_time().get(&public_key);
+```
+
+## License
+
+`exonum-time` is licensed under the Apache License (Version 2.0). See [LICENSE](LICENSE) for details.
+
+[service]: examples/simple_service.rs
+[exonum-testkit]: https://github.com/exonum/exonum-testkit
```
vitvakatu: Need to fix link

.travis.yml
```diff
@@ -107,9 +110,11 @@ script: |
       cd exonum && cargo fmt -- --write-mode=diff && cd .. &&
       cd sandbox && cargo fmt -- --write-mode=diff && cd .. &&
       cd testkit && cargo fmt -- --write-mode=diff && cd .. &&
+      cd services/exonum-time && cargo fmt -- --write-mode=diff && cd ../.. &&
       cspell sandbox/{src,examples,tests}/**/*.rs &&
       cspell exonum/{src,benches,tests}/**/*.rs &&
-      cspell testkit/{src,examples,tests}/**/*.rs
+      cspell testkit/{src,examples,tests}/**/*.rs &&
+      cspell services/exonum-time{src, examples, tests}/**/*.rs
```
slowli: Missing `/` after `time`?

Cargo.toml
```diff
@@ -3,5 +3,6 @@ members = [
     "exonum",
     "sandbox",
     "testkit",
+    "services/exonum-time"
```
slowli: I wonder if it makes sense to call the directory `exonum-time`; I'd prefer naming it simply `time`.

README.md
```diff
@@ -14,7 +14,7 @@ This is the main Exonum repository that includes
 * [Exonum testing framework](https://github.com/exonum/exonum/blob/master/testkit/README.md).
 * Services:
   * Configuration service (currently in [exonum-configuration](https://github.com/exonum/exonum-configuration) repository).
-  * Time service (currently in [exonum-time](https://github.com/exonum/exonum-time) repository).
+  * [Time service](https://github.com/exonum/exonum/blob/master/services/exonum-time/README.md).
```
slowli: Nitpick: I think it could make sense to use a relative link here: `[Time service](services/exonum-time/README.md)`.

services/exonum-time/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "exonum-time"
+version = "0.5.0"
+authors = ["The Exonum Team <exonum@bitfury.com>"]
+homepage = "https://github.com/exonum/exonum-time"
```
slowli: Wrong link.

.travis.yml
```diff
@@ -32,6 +32,8 @@ matrix:
       env: FEATURE=clippy-sandbox
     - rust: nightly-2018-01-22
       env: FEATURE=clippy-testkit
+    - rust: nightly-2018-01-22
+      env: FEATURE=clippy-exonum-time
```
slowli: Something that bugged me for a while:

- I think it could be better to define jobs in the Travis config (see #459, where I've done this). IMO, it's more readable than having `switch`es in `install` and `script` hooks.
- Does it really make sense to have multiple `clippy` and test jobs? I agree that clippy jobs should be separated from test ones, because they seemingly overwrite each other's artifacts; but I see little sense in splitting clippy / test jobs among services.

@DarkEld3r, could you comment on the topic?

services/exonum-time/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[dependencies]
+exonum = { version = "0.5.1", path = "../../exonum" }
+iron = "0.6.0"
+bodyparser = "0.8.0"
+router = "0.6.0"
+serde = "1.0"
+serde_json = "1.0"
+serde_derive = "1.0"
+pretty_assertions = "0.4.0"
```
slowli: Just noticed: this is an ordinary dependency? Where is it used? 🤔

Cargo.toml
DarkEld3r: Funny thing: we discussed that with @SerhiiOryshych, but decided to use full crate name. On the other hand `teskit`'s directory isn't named `exonum-teskit`.

README.md
DarkEld3r: There is (or was?..) a problem with relative links in the crates.io, but that isn't applied to the root readme.

.travis.yml
DarkEld3r: I had in mind the "optimization" of Travis build script, but thought of it as a separate task/pull request. 

I also think that jobs are easier to read and maintain, so we can switch to them already if @SerhiiOryshych doesn't mind.

Not sure about merging `clippy` jobs into one: it can take a long time, but it worth to try.

services/exonum-time/Cargo.toml
DarkEld3r: `pretty_assertions` is a [special library](https://github.com/colin-kiegel/rust-pretty-assertions): it should not be used directly, but it makes the assertions output nicer. Honestly, I'm not sure if we need that, so other opinions are welcome.

Perhaps, it makes sense to add it as `[dev-dependencies]`. In that case, it should be done for all our crates.

services/time/README.md
```diff
@@ -0,0 +1,85 @@
...
+// Gets the time of validator with a public key equal to `public_key`. 
+let validator_time = time_schema.validators_time().get(&public_key);
+```
+
+## License
+
+`exonum-time` is licensed under the Apache License (Version 2.0). See [LICENSE](LICENSE) for details.
+
+[service]: examples/simple_service.rs
+[exonum-testkit]: ../../exonum/testkit
```
slowli: Nitpick: This link definition is seemingly unused.

DarkEld3r:
Looks like additional work is needed because of markdown linter integration. :upside_down_face:

PR #470, opened 2018-02-08
Update travis script. [ECR-675]

SerhiiOryshych:


.travis.yml
```diff
@@ -80,86 +65,87 @@ install:
...
+  - env: FEATURE=test-sandbox
+    script:
+    - cargo test -p sandbox --verbose
+  - env: FEATURE=test-testkit
+    script:
+    - cargo test -p exonum-testkit --verbose
+    - cargo run -p exonum-testkit --example timestamping
+    - cargo run -p exonum-testkit --example configuration_change
+  - env: FEATURE=test-time
+    script:
```
matklad: Are we sure that we need separate `FEATURE=test-core`, `FEATURE=test-testkit`, etc.?

With multiple features, we'll be building exonum itself several times for each job. With `cargo test --all`, the total amount of work will be smaller.

.travis.yml
```diff
@@ -80,86 +65,87 @@ install:
...
+    - cd sandbox && cargo clippy --verbose -- -D warnings
+  - <<: *clippy-core
+    env: FEATURE=clippy-testkit
+    script:
+    - cd testkit && cargo clippy --verbose -- -D warnings
+  - <<: *clippy-core
+    env: FEATURE=clippy-time
+    script:
+    - cd services/time && cargo clippy --verbose -- -D warnings
```
matklad: Would `cargo clippy --all` work? It seems to work locally.

.travis.yml
slowli: 👍 . @SerhiiOryshych Could you consolidate the present jobs a single test job and a single clippy job to measure the impact on the build time? You can use `test-core` / `clippy-core` for `FEATURE` values in order to reuse existing caches.

.travis.yml
```diff
@@ -80,86 +65,59 @@ install:
...
+    rust: nightly-2018-01-22
+    install:
+    - cargo clippy --version | grep $CLIPPY_VERS || cargo install clippy --force --vers $CLIPPY_VERS
+    script:
+    - cargo clippy --all
+
+  # Tests
+  - env: FEATURE=test-core
+    script:
+    - cargo test -- all
```
slowli: Should be `--all`. Also, you still need to run examples (I suggest to run examples in separate commands to get the full picture of possible failures).

matklad: Are we sure that we want to *run* examples in CI? This will prevent us from creating an example which runs a server.

`cargo tests` builds the examples, and I think that is as good as you can get without restricting what examples should do?

matklad: No, it's not possible to run several binaries (example or usual ones) with a single command: unlike tests, they can not terminate, and they can use different CLI interface.

PR #472, opened 2018-02-09
Specify Rust version for CI explicitelly

matklad:
Let's pin a specific version of Rust, so that we *know* which is the minimal supported version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,9 @@ addons:
     - librocksdb
 
 rust:
-  - stable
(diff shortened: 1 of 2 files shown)


DarkEld3r:
How about adding a badge (http://shields.io/) to the root README.md with this version? Plus a comment that these versions should be updated simultaneously.

ozkriff:
Like this one: https://img.shields.io/badge/rustc-1.23+-blue.svg

rust 1.23+

PR #474, opened 2018-02-09
Use failure for error handling

matklad:
This switches error handling from `std::Error` to the [`failure`](https://github.com/withoutboats/failure) crate.

How to use failure (abridged edition):
- use `failure::Error` instead of `Box<Error>`,
- to define an error, add `derive(Fail)` to a struct/enum. Consider adding `#[cause]` and `#[fail(display="format")` attributes,
- consider using `.context` method to improve error messages for enduser-visible expected errors.

Here's the high-level list of changes in the PR:

1) `storage::Error` implements `Fail` (I've also changed the naming inside storage module to `module::Item` style)

2) The `fabric` module uses `failure::Error` instead of `Box<Error>`. This is a breaking change. We also should get slightly better error messages when reading config fails due to IO errors (message now includes the file name).

3) `ApiError` implements fail. This is the bulk of the changes from this PR. The sole purpose of `ApiError` is to be converted to `IronResponse`, so I've removed couple of variants like `FromHex`, and replaced them with `BadRequest`. I've also tweaked the way we generate status codes for ApiError: looks like previously we used a single HTTP code for all errors. I've also extracted some code duplicates from the `api` module so as to have a single place to tweak error handling. 

Note that serialization infra still uses `Box<Error>`. @vldm is furiously refactoring serialization, so I tried to avoid touching it =)

I'll write a changelog now :-)

### DIFF
--- a/exonum-dictionary.txt
+++ b/exonum-dictionary.txt
@@ -15,6 +15,7 @@ byteorder
 checkpointed
 clippy
 clonned
+compat
 concat
 counterintuitive
 cryptocurrency
--- a/exonum/CHANGELOG.md
+++ b/exonum/CHANGELOG.md
@@ -44,6 +44,21 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   - Remove `ID` constants.
   - Replace `TYPE` constants with a single `SERVICE_ID` constant.
 
+- Several variants were removed from `ApiError` enum. (#474)
+
+  Migration path:
+
+  - Use generic `ApiError::BadRequest` variant or create `IronError` directly.
+
+- `CommandExtension` uses `failure::Error` instead of `Box<std::error::Error>`
+  for errors. (#474)
+
+  Migration path:
+
+  - `std::error::Error` can be converted to `failure::Error` via `.into()` method.
+
+- `storage::Error` implements `failure::Fail` instead of `std::error::Error`. (#474)
+
 ### New features
 
 - `StorageKey` and `StorageValue` traits are implemented for `SystemTime`. (#456)
--- a/exonum/CHANGELOG.md
+++ b/exonum/CHANGELOG.md
@@ -63,6 +63,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `StorageKey` and `StorageValue` traits are implemented for `SystemTime`. (#456)
 - `StorageValue` and `CryptoHash` traits are implemented for `bool`. (#385)
+- `Height` implements `std::str::FromStr`. (#474)
 
 ### Bug fixes
 
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -277,7 +277,7 @@ where
     let fragment = params.find(name).ok_or_else(|| {
         ApiError::BadRequest(format!("Required parameter '{}' is missing", name))
     })?;
-    let value: T = T::from_str(fragment).map_err(|e| {
+    let value = T::from_str(fragment).map_err(|e| {
         ApiError::BadRequest(format!("Invalid '{}' parameter: {}", name, e))
     })?;
     Ok(value)
@@ -291,7 +291,7 @@ where
     let map = request.get_ref::<params::Params>().unwrap();
     let value = match map.find(&[name]) {
         Some(&params::Value::String(ref param)) => {
-            let value: T = T::from_str(param).map_err(|e| {
+            let value = T::from_str(param).map_err(|e| {
                 ApiError::BadRequest(format!("Invalid '{}' parameter: {}", name, e))
             })?;
             Some(value)
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -19,6 +19,7 @@ use std::marker::PhantomData;
 use std::io;
 use std::collections::BTreeMap;
 use std::fmt;
+use std::str::FromStr;
 
 use iron::IronError;
 use iron::prelude::*;
@@ -275,4 +276,19 @@ pub trait Api {
 
     /// Used to extend Api.
     fn wire<'b>(&self, router: &'b mut Router);
+}
+
+fn url_fragment<T>(request: &Request, name: &str) -> Result<T, ApiError>
+where
+    T: FromStr,
+    T::Err: ::std::error::Error + Send + Sync + 'static,
+{
+    let params = request.extensions.get::<Router>().unwrap();
+    let fragment = params.find(name).ok_or_else(|| {
+        ApiError::BadRequest(format!("Required parameter '{}' is missing", name).into())
+    })?;
+    let value: T = T::from_str(fragment).map_err(|e| {
+        ApiError::BadRequest(format!("Invalid '{}' parameter: {}", name, e).into())
+    })?;
+    Ok(value)
 }
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -28,6 +28,7 @@ use iron::headers::Cookie;
 use hyper::header::{ContentType, SetCookie};
 use cookie::Cookie as CookiePair;
 use router::Router;
+use params;
 use serde_json;
 use serde::{Serialize, Serializer};
 use serde::de::{self, Visitor, Deserialize, Deserializer};
@@ -291,4 +292,32 @@ where
         ApiError::BadRequest(format!("Invalid '{}' parameter: {}", name, e).into())
     })?;
     Ok(value)
+}
+
+fn optional_param<T>(request: &mut Request, name: &str) -> Result<Option<T>, ApiError>
+where
+    T: FromStr,
+    T::Err: ::std::error::Error + Send + Sync + 'static,
+{
+    let map = request.get_ref::<params::Params>().unwrap();
+    let value = match map.find(&[name]) {
+        Some(&params::Value::String(ref param)) => {
+            let value: T = T::from_str(param).map_err(|e| {
+                ApiError::BadRequest(format!("Invalid '{}' parameter: {}", name, e).into())
+            })?;
+            Some(value)
+        }
+        _ => None,
+    };
+    Ok(value)
+}
+
+fn required_param<T>(request: &mut Request, name: &str) -> Result<T, ApiError>
+where
+    T: FromStr,
+    T::Err: ::std::error::Error + Send + Sync + 'static,
+{
+    optional_param(request, name)?.ok_or_else(|| {
+        ApiError::BadRequest(format!("Required parameter '{}' is missing", name).into())
+    })
 }
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -106,7 +106,7 @@ impl From<ApiError> for IronError {
             let mut map = BTreeMap::new();
             map.insert("debug", format!("{:?}", e));
             map.insert("description", e.to_string());
-            ::serde_json::to_string_pretty(&map).unwrap()
+            serde_json::to_string_pretty(&map).unwrap()
         };
         IronError::new(e.compat(), (code, body))
     }
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -33,7 +33,7 @@ use serde::de::{self, Visitor, Deserialize, Deserializer};
 
 use crypto::{PublicKey, SecretKey, Hash};
 use encoding::serialize::{FromHex, FromHexError, ToHex, encode_hex};
-use storage::{Result as StorageResult, Error as StorageError};
+use storage;
 
 pub mod public;
 pub mod private;
@@ -46,7 +46,7 @@ pub enum ApiError {
     /// Service error.
     Service(Box<::std::error::Error + Send + Sync>),
     /// Storage error.
-    Storage(StorageError),
+    Storage(storage::Error),
     /// Converting from hex error.
     FromHex(FromHexError),
     /// Input/output error.
@@ -80,7 +80,7 @@ impl ::std::error::Error for ApiError {
         match *self {
             ApiError::Service(ref error) |
             ApiError::IncorrectRequest(ref error) => error.description(),
-            ApiError::Storage(ref error) => error.description(),
+            ApiError::Storage(ref error) => "storage error", //FIXME
             ApiError::FromHex(ref error) => error.description(),
             ApiError::Io(ref error) => error.description(),
             ApiError::FileNotFound(_) => "File not found",
@@ -106,8 +106,8 @@ impl From<io::Error> for ApiError {
     }
 }
 
-impl From<StorageError> for ApiError {
-    fn from(e: StorageError) -> ApiError {
+impl From<storage::Error> for ApiError {
+    fn from(e: storage::Error) -> ApiError {
         ApiError::Storage(e)
     }
 }
@@ -215,7 +215,7 @@ pub trait Api {
         &self,
         request: &'a Request,
         key: &str,
-    ) -> StorageResult<Vec<u8>> {
+    ) -> storage::Result<Vec<u8>> {
         if let Some(&Cookie(ref cookies)) = request.headers.get() {
             for cookie in cookies.iter() {
                 if let Ok(c) = CookiePair::parse(cookie.as_str()) {
@@ -227,7 +227,7 @@ pub trait Api {
                 }
             }
         }
-        Err(StorageError::new(
+        Err(storage::Error::new(
             format!("Unable to find value with given key {}", key),
         ))
     }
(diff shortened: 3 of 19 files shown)


exonum/src/api/mod.rs
```diff
@@ -27,77 +28,62 @@ use iron::headers::Cookie;
...
-    FromHex(FromHexError),
+    #[fail(display = "Storage error: {}", _0)]
+    Storage(
+        #[cause]
+        storage::Error
+    ),
+
     /// Input/output error.
+    #[fail(display = "IO error: {}", _0)]
     Io(::std::io::Error),
```
vitvakatu: Shouldn't there be `#[cause]` lint?

exonum/src/api/mod.rs
```diff
@@ -295,3 +275,46 @@ pub trait Api {
...
+                ApiError::BadRequest(format!("Invalid '{}' parameter: {}", name, e))
+            })?;
+            Some(value)
+        }
+        _ => None,
+    };
+    Ok(value)
+}
+
+fn required_param<T>(request: &mut Request, name: &str) -> Result<T, ApiError>
```
vitvakatu: Hmm, I wonder if we can make these functions `pub`. They're quite useful

vitvakatu:
I feel uncomfortable about removing `NotFound` error type. :confused:

exonum/src/api/mod.rs
```diff
@@ -27,77 +28,65 @@ use iron::headers::Cookie;
...
-    Io(::std::io::Error),
+    #[fail(display = "IO error: {}", _0)]
+    Io(
+        #[cause]
+        ::std::io::Error
+    ),
+
     /// File not found.
+    #[fail(display = "File not found")]
     FileNotFound(Hash),
```
slowli: May be off-topic, but `File*` errors have bugged me for a long time. It seems like they were introduced as a sort of a hack for some demo or project; IMO, they violate abstraction.

exonum/src/api/mod.rs
```diff
@@ -106,37 +95,31 @@ impl From<io::Error> for ApiError {
...
+        body.insert("description", e.to_string());
         let code = match e {
             ApiError::FileExists(hash) |
             ApiError::FileNotFound(hash) => {
                 body.insert("hash", encode_hex(&hash));
                 status::Conflict
             }
+            ApiError::BadRequest(..) => status::BadRequest,
+            ApiError::InternalError(..) => status::InternalServerError,
             _ => status::Conflict,
```
slowli: `Unauthorized` and `NotFound` have obvious HTTP status codes as well. Also, I don't think that `Conflict` returned here matches its description in [RFC 7231](https://tools.ietf.org/html/rfc7231#page-60). I'd rather return `BadRequest` as a catch-all code.

exonum/src/api/mod.rs
slowli: Agree. They could be used within this very PR in testkit integration tests. Also, just an idea (maybe, a stupid one): create a trait encompassing this functions and implement it for `Request`.

exonum/src/api/mod.rs
```diff
@@ -295,3 +278,46 @@ pub trait Api {
...
+where
+    T: FromStr,
+    T::Err: ::std::error::Error + Send + Sync + 'static,
+{
+    let params = request.extensions.get::<Router>().unwrap();
+    let fragment = params.find(name).ok_or_else(|| {
+        ApiError::BadRequest(format!("Required parameter '{}' is missing", name))
+    })?;
+    let value: T = T::from_str(fragment).map_err(|e| {
+        ApiError::BadRequest(format!("Invalid '{}' parameter: {}", name, e))
```
slowli: Could it make sense to create separate `ApiError` variants related to parameter processing? IMO, it could be clearer.

exonum/src/api/mod.rs
matklad: Yeah, the previous logic (which I more or less preserved) seems completely wrong. 

In fact, it is so wrong that I am suspicious that maybe this is deliberate? Could it be the case that we want to return the same status code for the purposes of information hiding?

exonum/src/api/mod.rs
matklad: Moderately strongly disagree here :-)

The sole purpose of `ApiError` is to be converted to `IronError`, whose sole purpose is to be converted to non-200/300 HTTP response. So the domain object that `ApiError` is modeling is actually an unsuccessful HTTP response, and there's no  specific status code for "invalid parameter".

I am not sure that `ApiError` enum is actually needed at all: it is just a disguised DSL to make working with hyper more pleasant. This enum has a ton of variants, but we never match it. Each ApiError instance created is almost instantly turned into IronError via a From impl and `?`.

For the same reason, I am not sure that exposing `url_fragment` and `requred/optional_parameter` is a good idea: this helpers have nothing to do with Exonum and are strictly iron specific. 

But perhaps practicality beats purity and we should make them pub with a comment like "yeah, this is a bad API which has nothing to do with Exonum, and it should not be considered stable, but you may use it, if you find it useful"?

exonum/src/api/mod.rs
matklad: Imo, this, the status code situation, and the copy-paste implementation of API (which I somewhat fixed here, but which is likely present in real-life servers as well) means that someone needs to take a holistic look at the HTTP API we expose and clean it up :-) I don't want to do it in this PR though, because it is orthogonal to "let's use failure and bury std::error::Error" goal :)

exonum/src/api/mod.rs
```diff
@@ -295,3 +278,46 @@ pub trait Api {
...
+
+fn optional_param<T>(request: &mut Request, name: &str) -> Result<Option<T>, ApiError>
+where
+    T: FromStr,
+    T::Err: ::std::error::Error + Send + Sync + 'static,
+{
+    let map = request.get_ref::<params::Params>().unwrap();
+    let value = match map.find(&[name]) {
+        Some(&params::Value::String(ref param)) => {
+            let value: T = T::from_str(param).map_err(|e| {
```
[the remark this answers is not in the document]
matklad: No, it's not necessary, though I find that it nicely complements the necessary vague `value` name :)

exonum/src/api/mod.rs
matklad: Replied in https://github.com/exonum/exonum/pull/474#discussion_r167296411

TL;DR: this belongs to an HTTP-API crate, and not to the Exonum per se, so I'd prefer not to expose it, or expose with an "warning, unstable API" in the docs.

exonum/src/api/mod.rs
matklad: More reasonable status codes in https://github.com/exonum/exonum/pull/474/commits/bdfa835aad8da9205ccc8072fc71786af8655037

exonum/src/api/mod.rs
```diff
@@ -106,37 +95,47 @@ impl From<io::Error> for ApiError {
...
+            // Note that `status::Unauthorized` does not fit here, because
+            //
+            // > A server generating a 401 (Unauthorized) response MUST send a
+            // > WWW-Authenticate header field containing at least one challenge.
+            //
+            // https://tools.ietf.org/html/rfc7235#section-4.1
+            ApiError::Unauthorized => status::Forbidden,
+
+            ApiError::BadRequest(..) |
+            ApiError::FileTooBig => status::BadRequest,
```
vitvakatu: Hmmmm... I don't understand such interpretation.

testkit/tests/currency/cryptocurrency.rs
```diff
@@ -219,7 +219,7 @@ impl CryptocurrencyApi {
         if let Some(wallet) = self.wallet(&public_key) {
             self.ok_response(&serde_json::to_value(wallet).unwrap())
         } else {
-            Err(IronError::new(ApiError::NotFound, (
+            Ok(Response::with((
```
vitvakatu: Yeah, now I understand why I don't like this code. There is more appropriate [way](https://docs.rs/exonum/0.5.1/exonum/api/trait.Api.html#method.not_found_response) to do that

exonum/src/api/mod.rs
DarkEld3r: Why not `FileTooBig` error?

exonum/src/api/mod.rs
```diff
@@ -106,37 +95,47 @@ impl From<io::Error> for ApiError {
...
-        body.insert("debug", format!("{:?}", e));
-        body.insert("description", e.description().to_string());
         let code = match e {
-            ApiError::FileExists(hash) |
-            ApiError::FileNotFound(hash) => {
-                body.insert("hash", encode_hex(&hash));
-                status::Conflict
+            ApiError::FileExists(..) => status::Conflict,
+            ApiError::FileNotFound(..) |
+            ApiError::NotFound => status::NotFound,
```
DarkEld3r: `FileNotFound` isn't used here because it contains `Hash`, right?

exonum/src/api/mod.rs
```diff
@@ -106,37 +95,47 @@ impl From<io::Error> for ApiError {
...
+            map.insert("description", e.to_string());
+            match e {
+                ApiError::FileExists(hash) |
+                ApiError::FileNotFound(hash) => {
+                    map.insert("hash", encode_hex(&hash));
+                }
+                _ => (),
             }
-            _ => status::Conflict,
+            ::serde_json::to_string_pretty(&map).unwrap()
```
DarkEld3r: Is it only me, who doesn't like `::` prefixes? :smile:

exonum-dictionary.txt
```diff
@@ -15,6 +15,7 @@ byteorder
 checkpointed
 clippy
 clonned
+compat
```
DarkEld3r: This is the thing that I would never like in rust: short names. :disappointed:

exonum/src/api/mod.rs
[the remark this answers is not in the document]
DarkEld3r: Personally I prefer to remove that parameter, it is duplicated really near from each other: `... :T = T::...`. Not insisting, though.

exonum/src/api/private/system.rs
```diff
@@ -17,12 +17,11 @@ use std::collections::HashMap;
 
 use router::Router;
 use iron::prelude::*;
-use params::{Params, Value as ParamsValue};
 
 use crypto::PublicKey;
 use node::{ExternalMessage, ApiSender};
 use blockchain::{Service, Blockchain, SharedNodeState};
-use api::{Api, ApiError};
+use api::{Api, ApiError, required_param};
```
DarkEld3r:
```rust
use api::{..., self};
api::required_param(...);
```
vs
```rust
use api::{..., required_param};
required_param(...);
```
Do we have some conventions here?..

exonum/src/api/public/blockchain_explorer.rs
```diff
@@ -12,16 +12,12 @@
...
-use std::str::ParseBoolError;
-
-use params::{Params, Value};
 use router::Router;
 use iron::prelude::*;
 
 use blockchain::{Blockchain, Block};
 use explorer::{BlockInfo, BlockchainExplorer};
-use api::{Api, ApiError};
+use api::{Api, ApiError, url_fragment, required_param, optional_param};
```
DarkEld3r: Same here. :smile:

exonum/src/storage/error.rs
```diff
@@ -14,11 +14,9 @@
 
 //! An implementation of `Error` type.
 
-use std::fmt;
-use std::error;
-
 /// The error type for I/O operations with storage.
-#[derive(Debug, Clone)]
```
DarkEld3r: Is there a motivation behind removing `Clone`?

exonum/src/api/mod.rs
alekseysidorov: This error is never used in code, why not to rename or remove it in this or following PR?

exonum/src/api/mod.rs
```diff
@@ -27,77 +28,65 @@ use iron::headers::Cookie;
...
     /// Not found.
+    #[fail(display = "Not found")]
     NotFound,
+
     /// File too big.
+    #[fail(display = "File too big")]
     FileTooBig,
+
     /// File already exists.
+    #[fail(display = "File exists")]
```
alekseysidorov: This error is also looks like rudiment.

exonum/src/api/mod.rs
matklad: >This error is never used in code, why not to rename or remove it in this or following PR?

Indeed 😆 ! I've just noticed that it is a dead code effectively! Will remove all unused options: if the consumer wants to have some custom IronError, they can just create a custion IronError directly :)

exonum/src/api/private/system.rs
```diff
@@ -156,31 +155,16 @@ impl SystemApi {
...
-                }
-                _ => {
-                    Err(ApiError::IncorrectRequest(
-                        "Required parameter of peer 'ip' is missing".into(),
-                    ))?
-                }
-            }
+            let addr: SocketAddr = required_param(req, "ip")?;
+            self_.node_channel.peer_add(addr).map_err(ApiError::from)?;
+            self_.ok_response(&::serde_json::to_value("Ok").unwrap())
```
aleksuss: What about to move a `::serde_json::to_value(&str).unwrap()` into `Api::ok_response` method ?

vitvakatu: Let's do it in separate PR

slowli: Maybe, it makes sense to use [`json!` macro](https://docs.serde.rs/serde_json/macro.json.html), like `json!("Ok")`?

exonum/src/api/private/system.rs
aleksuss: I guess we DON'T 😄

DarkEld3r: Should we? :smile:

exonum-dictionary.txt
matklad: That's from the `failure` crate. But yeah, `TyCtxt` level of compression of `rustc` deferentially makes code harder to read....

exonum/src/api/mod.rs
matklad: Yep, it definitely  is not required here! Although, I personally like absolute paths: you can understand them without context (that is, you don't need to scan use declarations)

exonum/src/storage/error.rs
matklad: No, Clone is back!

exonum/src/api/public/system.rs
```diff
@@ -19,8 +19,7 @@ use node::state::TxPool;
 use blockchain::{Blockchain, SharedNodeState};
 use crypto::Hash;
 use explorer::{BlockchainExplorer, TxInfo};
-use api::{Api, ApiError};
-use encoding::serialize::FromHex;
+use api::{Api, ApiError, url_fragment};
```
DarkEld3r: `self` instead of `url_fragment`? :smile:

exonum/CHANGELOG.md
```diff
@@ -44,6 +44,21 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   - Remove `ID` constants.
   - Replace `TYPE` constants with a single `SERVICE_ID` constant.
 
+- Several variants were removed from `ApiError` enum.
```
DarkEld3r: Please add reference to the pull request.

exonum/CHANGELOG.md
```diff
@@ -44,6 +44,21 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+
+  Migration path:
+
+  - Use generic `ApiError::BadRequest` variant or create `IronError` directly.
+
+- `CommandExtension` uses `failure::Error` instead of `Box<std::error::Error>`
```
DarkEld3r: Same here.

exonum/CHANGELOG.md
```diff
@@ -44,6 +44,21 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+  for errors.
+
+  Migration path:
+
+  - `std::error::Error` can be converted to `failure::Error` via `.into()` method.
+
+- `storage::Error` implements `failure::Fail` instead of `std::error::Error`.
```
DarkEld3r: And here?..

Should all the above be somehow combined?..

matklad: I don't know :-)

On the one hand, they are all part of a single "migrate to failure" PR. On the other hand, each one is a separate user-facing piece of the API.

matklad:
Should be good to go?

Here's the last commit, which moves utils to the `API` crate: https://github.com/exonum/exonum/pull/474/commits/8925ed632757fa59c76e592a4ece1efffb9b1050 (heh, I really want [a code review system with diff of diffs and brains](https://www.janestreet.com/tech-talks/code-review/))

PR #480, opened 2018-02-12
Minor: correct job names in .travis.yml

matklad:
No functional changes, just rename travis jobs to better reflect reality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ env:
   - ROCKSDB_LIB_DIR=/usr/lib/x86_64-linux-gnu
   - SNAPPY_LIB_DIR=/usr/lib/x86_64-linux-gnu
   matrix:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -90,16 +90,16 @@ jobs:
     - find . -not -path "./3rdparty/*" -and -not -path "./node_modules/*" -name "*.md" | xargs ./node_modules/.bin/markdownlint --config .markdownlintrc
 
   # Clippy linting
-  - &clippy-core
-    env: FEATURE=clippy-core
+  - &clippy
```
slowli: It seems that this node identifier isn't used anywhere. Remove?

matklad: I actually have no idea how borrowing works in YAML, so let's try removing it and see if anything breaks :-)

PR #481, opened 2018-02-12
Extend committed response

DarkEld3r:

### DIFF
--- a/services/configuration/src/api.rs
+++ b/services/configuration/src/api.rs
@@ -50,13 +50,13 @@ pub struct ConfigInfo {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct ProposePost {
+pub struct ProposeResponse {
     pub tx_hash: Hash,
     pub cfg_hash: Hash,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct VotePost {
+pub struct VoteResponse {
     pub tx_hash: Hash,
 }
 
@@ -114,7 +114,10 @@ impl PublicConfigApi {
     fn get_votes_for_propose(&self, config_hash: &Hash) -> VotesInfo {
         let snapshot = self.blockchain.snapshot();
         let configuration_schema = ConfigurationSchema::new(&snapshot);
-        if configuration_schema.propose_data_by_config_hash().contains(config_hash) {
+        if configuration_schema
+            .propose_data_by_config_hash()
+            .contains(config_hash)
+        {
             Some(configuration_schema.get_votes(config_hash))
         } else {
             None
@@ -229,10 +232,7 @@ impl<T> PrivateConfigApi<T>
 where
     T: TransactionSend + Clone,
 {
-    fn put_config_propose(
-        &self,
-        cfg: StoredConfiguration,
-    ) -> Result<ProposePost, ApiError> {
+    fn put_config_propose(&self, cfg: StoredConfiguration) -> Result<ProposeResponse, ApiError> {
         let cfg_hash = cfg.hash();
         let config_propose = TxConfigPropose::new(
             &self.config.0,
@@ -242,16 +242,16 @@ where
         let tx_hash = config_propose.hash();
         let ch = self.channel.clone();
         ch.send(Box::new(config_propose))?;
-        let res = ProposePost { tx_hash, cfg_hash };
+        let res = ProposeResponse { tx_hash, cfg_hash };
         Ok(res)
     }
 
-    fn put_config_vote(&self, cfg_hash: &Hash) -> Result<VotePost, ApiError> {
+    fn put_config_vote(&self, cfg_hash: &Hash) -> Result<VoteResponse, ApiError> {
         let config_vote = TxConfigVote::new(&self.config.0, cfg_hash, &self.config.1);
         let tx_hash = config_vote.hash();
         let ch = self.channel.clone();
         ch.send(Box::new(config_vote))?;
-        let res = VotePost { tx_hash };
+        let res = VoteResponse { tx_hash };
         Ok(res)
     }
 }
--- a/services/configuration/src/api.rs
+++ b/services/configuration/src/api.rs
@@ -25,38 +25,38 @@ use exonum::node::TransactionSend;
 use exonum::encoding::serialize::json::reexport as serde_json;
 use exonum::helpers::Height;
 
-use super::{StorageValueConfigProposeData, TxConfigPropose, TxConfigVote, ConfigurationSchema};
+use super::{ProposeData, TxConfigPropose, TxConfigVote, ConfigurationSchema};
 
-pub type ApiResponseVotesInfo = Option<Vec<Option<TxConfigVote>>>;
+pub type VotesInfo = Option<Vec<Option<TxConfigVote>>>;
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct ApiResponseConfigHashInfo {
+pub struct ConfigHashInfo {
     pub hash: Hash,
     pub config: StoredConfiguration,
     pub propose: Option<Hash>,
-    pub votes: ApiResponseVotesInfo,
+    pub votes: VotesInfo,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct ApiResponseProposeHashInfo {
+pub struct ProposeHashInfo {
     pub hash: Hash,
-    pub propose_data: StorageValueConfigProposeData,
+    pub propose_data: ProposeData,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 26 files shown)


services/configuration/src/config_api.rs
```diff
@@ -261,8 +263,8 @@ where
         let ch = self.channel.clone();
         ch.send(Box::new(config_propose))?;
         let res = ApiResponseProposePost {
-            tx_hash: tx_hash,
-            cfg_hash: cfg_hash,
+            tx_hash,
+            cfg_hash,
```
matklad: FWIW, IDEA can apply `foo: foo -> foo` fix automatically to the whole project :)

services/configuration/src/config_api.rs
```diff
@@ -73,26 +75,34 @@ pub struct PublicConfigApi {
...
-        let res = ApiResponseConfigHashInfo {
-            hash: actual_cfg.hash(),
-            config: actual_cfg,
-        };
-        Ok(res)
+    fn get_actual_config(&self) -> Result<ApiResponseConfigHashInfo, ApiError> {
+        let snapshot = self.blockchain.snapshot();
+        let configuration_schema = Schema::new(&snapshot);
+        let actual_cfg = configuration_schema.actual_configuration();
+        Ok(self.get_config_with_proofs(actual_cfg))
```
matklad: Why this function returns a `Result`? There are no `?`?

services/configuration/src/config_api.rs
```diff
@@ -73,26 +75,34 @@ pub struct PublicConfigApi {
...
-                config: cfg,
+        let following_cfg = match Schema::new(&self.blockchain.snapshot()).following_configuration() {
+            Some(val) => val,
+            None => {
+                return Ok(None);
             }
-        });
-        Ok(res)
+        };
+        Ok(Some(self.get_config_with_proofs(following_cfg)))
```
matklad: Same here: no question marks, could probably return just Option?

services/configuration/src/config_api.rs
```diff
@@ -70,54 +72,52 @@ pub struct PublicConfigApi {
...
-        ) {
-            None => None,
-            Some(_) => Some(configuration_schema.get_votes(cfg_hash)),
-        };
-
-        Ok(res)
+        configuration_schema
+            .propose_data_by_config_hash()
+            .get(config_hash)
+            .map(|_| configuration_schema.get_votes(config_hash))
```
matklad: Seems highly suspicious that it just ignores `propose_data`. I think it is required for some reason, but perhaps a code comment, explaining why we need this lookup, would make this slightly less mysterious? :)

DarkEld3r: I would love to see such comment too... :smile:

@alekseysidorov @slowli @vldm 
Can anyone remember why it is needed? Is it just a refactoring artifact that can be safely removed?

slowli: I think I understand the general idea: the code checks if there exists a proposal with the given `config_hash` and returns votes for it only if it exists. The reasoning for a preliminary existence check is that `get_votes()` method uses a group of indexes distinguished by a prefix to store / retrieve data, and I think (not 100% sure) that there are no methods in Exonum API to check whether a particular index within the family exists.

Still, if I'm correct, the code could be refactored to use `contains(config_hash)` instead of `get`.

services/configuration/src/config_api.rs
```diff
@@ -70,54 +72,52 @@ pub struct PublicConfigApi {
...
-    fn get_actual_config(&self) -> Result<ApiResponseConfigHashInfo, ApiError> {
-
-        let actual_cfg = Schema::new(&self.blockchain.snapshot()).actual_configuration();
-        let res = ApiResponseConfigHashInfo {
-            hash: actual_cfg.hash(),
-            config: actual_cfg,
-        };
-        Ok(res)
+    fn get_config_with_proofs(&self, config: StoredConfiguration) -> ApiResponseConfigHashInfo {
+        let propose = ConfigurationSchema::new(&self.blockchain.snapshot())
```
slowli: Nitpick: ref'ing a snapshot seems superfluous here.

services/configuration/src/config_api.rs
```diff
@@ -173,50 +173,43 @@ impl PublicConfigApi {
...
     fn get_all_committed(
         &self,
         previous_cfg_hash_filter: Option<Hash>,
         actual_from_filter: Option<Height>,
-    ) -> Result<Vec<ApiResponseConfigHashInfo>, ApiError> {
+    ) -> Vec<ApiResponseConfigHashInfo> {
         let snapshot = self.blockchain.snapshot();
         let general_schema = Schema::new(&snapshot);
         let index = general_schema.configs_actual_from();
         let committed_configs = {
```
slowli: It seems like this block has a single statement (a long one, but still). Is this intentional / required by the formatter, or an artifact? IMO, unnecessary indentation should be avoided.

services/configuration/src/config_api.rs
```diff
@@ -173,50 +173,43 @@ impl PublicConfigApi {
...
 
     fn get_all_committed(
         &self,
         previous_cfg_hash_filter: Option<Hash>,
         actual_from_filter: Option<Height>,
-    ) -> Result<Vec<ApiResponseConfigHashInfo>, ApiError> {
+    ) -> Vec<ApiResponseConfigHashInfo> {
         let snapshot = self.blockchain.snapshot();
         let general_schema = Schema::new(&snapshot);
         let index = general_schema.configs_actual_from();
```
slowli: Nitpick: `index` seems such a non-descriptive name...

services/configuration/src/config_api.rs
```diff
@@ -173,50 +173,43 @@ impl PublicConfigApi {
...
         let committed_configs = {
             index
                 .into_iter()
                 .map(|reference| {
-
-                    let cfg_hash = reference.cfg_hash();
-                    let cfg = general_schema.configs().get(cfg_hash).expect(&format!(
+                    let config_hash = reference.cfg_hash();
+                    general_schema.configs().get(config_hash).expect(&format!(
```
slowli: Does calling `configs()` method repeatedly introduce any sort of overhead? Even if not, initializing the configs table once would look more readable to me.

services/configuration/src/config_api.rs
DarkEld3r: I want as many such "nitpicks" as possible: they makes the code better. :+1:

services/configuration/src/config_api.rs
DarkEld3r: This is unfortunately needed to fight the borrow checker.

slowli: Huh, it seems to work locally for me. I'm talking about the following simple change:

```rust
let committed_configs = index.into_iter()
    .map(|reference| { /* ... */ })
    // ...
    .collect::<Vec<_>>();
committed_configs
```

Btw, there is a similar obsolete block in `get_all_proposes`.

services/configuration/src/api.rs
```diff
@@ -25,36 +25,38 @@ use exonum::node::TransactionSend;
...
     pub committed_config: Option<StoredConfiguration>,
-    pub propose: Option<StorageValueConfigProposeData>,
+    pub propose: Option<ProposeData>,
 }
 
-pub type ApiResponseVotesInfo = Option<Vec<Option<TxConfigVote>>>;
-
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct ApiResponseProposePost {
+pub struct ProposePost {
```
slowli: I'd use `ProposeResponse`.

services/configuration/src/api.rs
```diff
@@ -25,36 +25,38 @@ use exonum::node::TransactionSend;
...
     pub tx_hash: Hash,
     pub cfg_hash: Hash,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct ApiResponseVotePost {
+pub struct VotePost {
```
slowli: `VoteResponse`.

services/configuration/src/config_api.rs
DarkEld3r: Have you tried to run Clippy after that? :wink: 
We have `#[cfg_attr(feature = "cargo-clippy", allow(let_and_return))]` annotations in many places because of this issue. I dislike both variants equally.

slowli: But even with a block, this still *is* a let-and-return, only not recognized by clippy (yet?). 

In general, do you think iterator lifetimes could be an issue with Exonum? I've not looked into the issue in depth, but it seems that the problem is that the iterators over indexes are constrained by the lifetime of the index (default behavior), whereas they could be constrained by the lifetime of a snapshot / fork.

DarkEld3r: Yes, I suppose, but it should be checked more thoroughly. I has created an [issue](https://jira.bf.local/browse/ECR-737) for that, but I don't know when we will have time to look at that.

PR #482, opened 2018-02-12
Update CONTRIBUTING.md

### LINKED ISSUE #109: Update CONTRIBUTING.md with code conventions

DarkEld3r:
Any suggestions are welcome.

Closes #109.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,269 +1,79 @@
-# Contributing to Exonum Core
+# Contributing to Exonum
 
-This contribution guide is partially derived from
-[Bitcoin Contribution Guide](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md)
+Exonum welcomes contribution from everyone in the form of [suggestions], [bug
+reports] or pull requests. This document gives some guidance if you are
+thinking of helping us.
 
-The Exonum Core project and its separate modules operate an open contributor
-model where anyone is welcome to contribute towards development
-in the form of peer review, testing and patches.
-This document explains the practical process and guidelines for contributing.
-The guidelines are the same for other Exonum repositories.
+[Project overview] and [documentation] can help you to better understand current
+state of the project.
 
-Firstly in terms of structure, there is no particular concept of
-"Core developers" in the sense of privileged people.
-Open source often naturally revolves around meritocracy where
-longer term contributors gain more trust from the developer community.
-However, some hierarchy is necessary for practical purposes.
-As such there are repository "maintainers" who are responsible for merging
-pull requests as well as a "lead maintainer" who is responsible
-for the release cycle, overall merging, moderation and appointment of maintainers.
+You can always ask for help or guidance in our [gitter] channel. There is also
+a separate [channel][gitter-ru] for our Russian-speaking users.
 
-## Contributor Workflow
+## Quick Start
 
-The codebase is maintained using the "contributor workflow" where everyone
-without exception contributes patch proposals using "pull requests".
-This facilitates social contribution, easy testing and peer review.
+Install Rust and mandatory dependencies according to our [installation guide],
+then you can build the project and run the tests:
 
-To contribute a patch, the workflow is as follows:
-
-- Fork repository
-- Create topic branch
-- Commit patches
-
-The project coding conventions in the **TODO: insert here a link to developer notes**
-[developer notes](doc/developer-notes.md) must be adhered to.
-
-In general [commits should be atomic](https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention)
-and diffs should be easy to read. For this reason do not mix any
-formatting fixes or code moves with actual code changes.
-
-Commit messages should be verbose by default consisting of a short subject line
-(50 chars max), a blank line and detailed explanatory text as separate paragraph(s);
-unless the title alone is self-explanatory (like "Corrected typo in init.rs")
-then a single title line is sufficient. Commit messages should be helpful
-to people reading your code in the future, so explain the reasoning for your decisions.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,269 +1,54 @@
...
+- [cargo audit]
+- [cargo-deadlinks]
+- [cargo-outdated]
+- [rustfmt 0.9.0]
+- [cspell]
+- [markdownlint-cli]
 
-The project coding conventions in the **TODO: insert here a link to developer notes**
-[developer notes](doc/developer-notes.md) must be adhered to.
+You can set up and run these tools locally (see [Travis script] for the details.
```
vitvakatu: Missing closing bracket

CONTRIBUTING.md
```diff
@@ -1,269 +1,54 @@
...
+reports] or pull requests. This document gives some guidance if you are 
+thinking of helping us.
 
-The Exonum Core project and its separate modules operate an open contributor
-model where anyone is welcome to contribute towards development
-in the form of peer review, testing and patches.
-This document explains the practical process and guidelines for contributing.
-The guidelines are the same for other Exonum repositories.
+[Project overview] and [documentation] can help you to better understand what's
+going on here.
```
Maria-Nosyk: "what's going on here." may be it's better to make it more bookish like "current state of the project"

matklad:
I suggest adding a quick start section, which shows how to get code up and running:

## Quick Start

Make sure you have Rust installed: https://rustup.rs/

Make sure that the following mandatory native dependencies are installed: 
* (I am not sure what native deps are mandatory, I don't think it's really important to mention optional dependensies like librocksdb here)


    $ git clone https://github.com/exonum/exonum
    $ cd exonum
    $ cargo test --all

CONTRIBUTING.md
```diff
@@ -1,269 +1,70 @@
...
-- have a clear use case, fix a demonstrable bug or serve the greater good
-    of the project (for example refactoring for modularisation);
-- be well peer reviewed;
-- have unit tests and sandbox tests where appropriate;
-- follow code style guidelines;
-- not break the existing test suite;
-- where bugs are fixed, where possible, there should be unit tests demonstrating
-    the bug and also proving the fix. This helps prevent regression.
+The [good first issue] label can be used to find the easy issues. If you want
+to work on an issue, please leave a comment so that we can assign it to you.
```
matklad: I am mildly against 

>If you want to work on an issue, please leave a comment so that we can assign it to you.

clause, for two reasons:

1) It is only occasionally helpful, because clashes rarely happen in practice. 
2) It discourages contributors, because it asks for commitment to fix the issue. That is, it "forbids" the following workflow: "I'll want to fix an issue, but I am not sure about my abilities to do it, so I'll just *try* to fix it locally and, if I am successful, I'll send a PR. If, however, I fail to fix the bug, nobody knows that about that!"

It might be a good idea for *ourselves* though to mark formerly "help wanted" issues if we decide to fix them ourselves.

CONTRIBUTING.md
```diff
@@ -1,269 +1,70 @@
...
 
-## Release Policy
+- Don't use `debug!` log level.
 
-The project leader is the release manager for each Exonum Core release.
+  It is convenient to use `debug!` when you develop some feature and are only
+  interested in your logging output.
 
-## Copyright
+- ?
```
matklad: The conventions section does not seem very helpful, because the most important conventions are enforced by CI anyway. I suggest just to drop this section 😆

DarkEld3r: I cannot agree here. :smile: It happens relatively often for us when a new team member becomes aware of some "rules" only on the review stage. We need to document such conventions somewhere, so why not here?

CONTRIBUTING.md
DarkEld3r: Removed. :+1:

CONTRIBUTING.md
matklad: Sgtm!

CONTRIBUTING.md
```diff
@@ -1,269 +1,79 @@
...
-
-## "Decision Making" Process
-
-The following applies to code changes to the Exonum Core project and related
-projects such as Exonum BTC Anchoring Service.
+## Finding something to fix or improve
 
-Whether a pull request is merged into Exonum Core rests with the project merge
-maintainers and ultimately the project lead.
+The [good first issue] label can be used to find the easy issues.
```
slowli: I'm not quite sure we should advertise this label before there are any issues tagged with it.

DarkEld3r: I don't think that there will be many potential contributors right now. :wink:

DarkEld3r:
> Should the contributing guide removed from the exonum-doc repository to avoid confusion?

I'm not sure. Perhaps a link to this document can be used?

CONTRIBUTING.md
```diff
@@ -1,269 +1,79 @@
...
 
-Maintainers will take into consideration if a patch is in line with the general
-principles of the project; meets the minimum standards for inclusion;
-and will judge the general consensus of contributors.
+## Linters
 
-In general, all pull requests must:
+Notice that the repository uses a set of linters and static analysis:
```
Maria-Nosyk: "linters and static analysis" --> "linters for static analysis"

PR #483, opened 2018-02-13
Fix hash implementation for `()`. [ECR-442]

SerhiiOryshych:

### DIFF
--- a/exonum/CHANGELOG.md
+++ b/exonum/CHANGELOG.md
@@ -59,6 +59,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `storage::Error` implements `failure::Fail` instead of `std::error::Error`. (#474)
 
+- `CryptoHash` for `()` now correctly calculates a hash of an empty byte array
+  instead of returning `Hash::zero()`. (#483)
+
 ### New features
 
 - `StorageKey` and `StorageValue` traits are implemented for `SystemTime`. (#456)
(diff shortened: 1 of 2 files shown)


DarkEld3r:
Please add a changelog entry.

exonum/CHANGELOG.md
```diff
@@ -59,6 +59,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `storage::Error` implements `failure::Fail` instead of `std::error::Error`. (#474)
 
+- `CryptoHash` for `()` now correctly calculated hash instead
```
vitvakatu: calculates?

exonum/src/crypto.rs
```diff
@@ -612,7 +612,7 @@ impl CryptoHash for i64 {
 
 impl CryptoHash for () {
     fn hash(&self) -> Hash {
-        Hash::zero()
+        hash(&[])
```
slowli: I wonder if it could make sense to return a constant here. Is this doable without `const fn` support?

DarkEld3r: I suppose it's not possible without things like `lazy_static`.

matklad: >I suppose it's not possible without things like lazy_static.

I'd say that transmuting an `[u8; 32]` would be a nice solution

matklad: Provided that we add a tests that checks that the the result of transmute is indeed equal to hash(&[]).

PR #484, opened 2018-02-13
Add basic fuzz testing suit [ECR-203]

vitvakatu:

### DIFF
--- a/exonum/fuzz/.gitignore
+++ b/exonum/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
--- a/exonum/fuzz/.gitignore
+++ b/exonum/fuzz/.gitignore
@@ -1,4 +1,0 @@
-
-target
-corpus
-artifacts
--- a/exonum/fuzz/Cargo.toml
+++ b/exonum/fuzz/Cargo.toml
@@ -8,14 +8,9 @@ publish = false
 [package.metadata]
 cargo-fuzz = true
 
-[dependencies.exonum]
-path = ".."
-[dependencies.libfuzzer-sys]
-git = "https://github.com/rust-fuzz/libfuzzer-sys.git"
-
-# Prevent this from interfering with workspaces
-[workspace]
-members = ["."]
+[dependencies]
+exonum = { path = ".." }
+libfuzzer-sys = { git = "https://github.com/rust-fuzz/libfuzzer-sys.git" }
 
 [[bin]]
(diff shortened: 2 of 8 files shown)


exonum/fuzz/.gitignore
```diff
@@ -0,0 +1,4 @@
+
+target
+corpus
```
ozkriff: why not add these ignores to repo's `.gitignore`? So they all can be found in one place.

```text
./fuzz/target
./fuzz/corpus
./fuzz/artifacts
```

vitvakatu: The only reason is that it's file was generated automatically by `cargo-fuzz`. Will fix

exonum/fuzz/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+cargo-fuzz = true
+
+[dependencies.exonum]
+path = ".."
+[dependencies.libfuzzer-sys]
+git = "https://github.com/rust-fuzz/libfuzzer-sys.git"
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
```
matklad: Perhaps an `exclude` in the parent Cargo.toml would be more direct?

exonum/fuzz/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+authors = ["Automatically generated"]
+publish = false
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies.exonum]
+path = ".."
+[dependencies.libfuzzer-sys]
+git = "https://github.com/rust-fuzz/libfuzzer-sys.git"
```
matklad: Why not 

```
[dependencies]
exonum = { path = ".." }
libfuzzer-sys = { git = "..." }
```

vitvakatu:
> Hm, should we somehow integrate it with CI?

No, it's probably not possible and it isn't the goal.

services/configuration/src/lib.rs
```diff
@@ -255,7 +256,7 @@ where
     /// 2. Config, previous to a `StoredConfiguration` is referenced by
     /// `previous_cfg_hash` in `StoredConfiguration`.
     ///
-    /// - Table **index** is validator_id - position of a validator's `PublicKey`
+    /// - Table **index** is validator_id - position of a validators `PublicKey`
```
mvlabat: Singular form then?

PR #487, opened 2018-02-14
circleci: basic macos build [ECR-443]

ozkriff:
This PR adds some basic macOS CI testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,9 @@ jobs:
   build_and_test:
     macos:
       xcode: "9.0"
+    environment:
+      - ROCKSDB_LIB_DIR: /usr/local/lib
+      - SNAPPY_LIB_DIR: /usr/local/lib
     steps:
       - checkout
       - run: curl https://sh.rustup.rs -sSf | sh -s -- -y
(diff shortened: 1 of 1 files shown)


aleksuss:
I guess you should add env variables for precompiled libraries `libsodium` and `librocksdb`.

.circleci/config.yml
```diff
@@ -0,0 +1,20 @@
+version: 2
+
+jobs:
+  build_and_test:
+    macos:
+      xcode: "9.0"
```
aleksuss:
```
environment:
  - ROCKSDB_LIB_DIR: /usr/local/lib
  - SNAPPY_LIB_DIR: /usr/local/lib
```

vitvakatu:
Shouldn't we do the same thing as in Travis: https://github.com/exonum/exonum/blob/f81d4535a4173651e8cbb28d7b00cfdbd2db0b45/.travis.yml#L104

matklad:
>Shouldn't we do the same thing as in Travis: 

I think we can drop `--all` from `cargo test` on travis: `--all` is implied for virtual manifests. OTOH, `cargo fmt` and `cargo clippy` seem to need an explicit `--all` :(

vitvakatu:
@matklad actually I meant running examples as well
cc @ozkriff

DarkEld3r:
@vitvakatu Running examples would be nice, but we have pretty harsh limits on CircleCI, so I suppose it can be dropped (for now).

PR #489, opened 2018-02-14
WIP: Move encoding into a separate crate [ECR-711]

DarkEld3r:


exonum/src/lib.rs
```diff
@@ -27,6 +27,8 @@ extern crate exonum_rocksdb as rocksdb;
 #[macro_use]
 extern crate exonum_profiler;
 #[macro_use]
+extern crate exonum_encoding as encoding;
```
matklad: I think this might want to be `pub extern crate` for `$crate::encoding` in macros to work?

exonum/encoding/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+serde_json = "1.0.2"
+toml = "0.4.2"
+clap = "2.25.0"
+term = "0.4.6"
+colored = "1.5.2"
+env_logger = "0.5.3"
+hyper = "0.10.13"
+cookie = "0.10.1"
+router = "0.6.0"
+iron = "0.6.0"
```
matklad: We probably don't need most of these here :)

DarkEld3r: Yeap, it is "in progress" for a reason. 😉

exonum/components/encoding/src/serialize/json.rs
```diff
@@ -27,15 +27,14 @@ use std::error::Error;
 
 use serde_json::value::{Value, from_value};
 use bit_vec::BitVec;
-use hex::FromHex;
+use hex::{FromHex};
```
alexander-irbis: Скобки остались лишние

DarkEld3r:
Postponed (waiting for changes in the serialization).

PR #491, opened 2018-02-16
Update tx_from_raw method description [ECR-704]

DarkEld3r:


slowli:
I'd suggest to reference a "standard" implementation of the method using `TransactionSet`:

```rust
fn tx_from_raw(&self, raw: RawTransaction) -> Result<Box<Transaction>, EncError> {
    let tx = MyTransactions::tx_from_raw(raw)?;
    Ok(tx.into())
}
```

It is already present in the service example, so you could just provide a link, I suppose.

exonum/src/blockchain/service.rs
```diff
@@ -148,6 +148,18 @@ pub trait Service: Send + Sync + 'static {
...
+    ///
+    /// Service should return an error in the following cases (see `MessageError` for more details):
+    /// - Incorrect transaction identifier.
+    /// - Incorrect data layout.
+    ///
+    /// Service _shouldn't_ perform signature check or logical validation of the transaction: these
+    /// operations should be performed in the `Transaction::verify` and `Transaction::execute`
+    /// methods respectively.
+    ///
+    /// `transactions!` macro generates code that allows simple implementation, see [Service][1]
```
slowli: This doc will be on the `Service` page, so it may make sense to provide a link to the page fragment:

```markdown
...see [the `Service` example above](#examples).
```

exonum/src/blockchain/service.rs
```diff
@@ -148,6 +148,18 @@ pub trait Service: Send + Sync + 'static {
...
 
     /// Tries to create a `Transaction` from the given raw message.
+    ///
+    /// Service should return an error in the following cases (see `MessageError` for more details):
+    /// - Incorrect transaction identifier.
+    /// - Incorrect data layout.
+    ///
+    /// Service _shouldn't_ perform signature check or logical validation of the transaction: these
+    /// operations should be performed in the `Transaction::verify` and `Transaction::execute`
+    /// methods respectively.
```
slowli: Nitpick: `verify` may perform logical validation of the transaction (provided it does not depend on external factors), so I'd remove "respectively".

exonum/src/blockchain/service.rs
```diff
@@ -148,6 +148,17 @@ pub trait Service: Send + Sync + 'static {
     fn state_hash(&self, snapshot: &Snapshot) -> Vec<Hash>;
 
     /// Tries to create a `Transaction` from the given raw message.
+    ///
```
dmitry-timofeev: Shan't we explicitly say that the _only_ thing that the core guarantees is that `service_id` is equal to the id of this service (and, maybe, the message has proper size), and therefore, the user implementing the interface **must** handle invalid transactions that may come from byzantine nodes?

PR #492, opened 2018-02-16
Remove failed transaction body from logging [ECR 765]

DarkEld3r:


exonum/src/blockchain/mod.rs
```diff
@@ -256,42 +256,7 @@ impl Blockchain {
                     "BUG: Cannot find transaction in pool.",
                 );
 
-                fork.checkpoint();
```
vitvakatu: Hmmm... Are you sure?

DarkEld3r: :sweat_smile:

exonum/src/blockchain/mod.rs
```diff
@@ -492,3 +457,41 @@ impl Clone for Blockchain {
...
+
+    let tx_hash = tx.hash();
+    let tx_result = match catch_result {
+        Ok(execution_result) => {
+            match execution_result {
+                Ok(()) => fork.commit(),
+                Err(ref e) => {
+                    // Unlike panic, transaction failure isn't that rare, so logging the
+                    // whole transaction body is an overkill: it can be relatively big.
+                    info!("{:?} transaction execution failed: {:?}", tx_hash, e);
```
alekseysidorov: In my opinion warnin level is more suitable.

DarkEld3r: I'm not sure (but not entirely against it). :smile: 

I think that from the framework's perspective this is definitely OK. User decides to return error and we have nothing to do with that: we cannot fix it in _our_ code.
Furthermore, it can even be OK for the service author. For example, users can try to transfer more currency than they have and it isn't a problem (except for the front-end that can probably prevent such things).

@slowli @matklad @alekseysidorov @aleksuss Opinions?

slowli: I certainly think that a transaction returning an unsuccessful execution result should not be treated as an extraordinary event. If a service wants to log that something went *really* wrong, it can use the macros provided by the log crate (which will probably be replaced by the `ExecutionContext` logging methods in the future if/when we implement it).

PR #494, opened 2018-02-16
Make CI stricter

slowli:
This PR returns missing `-D warnings` argument for clippy, and denies warnings from the Rust compiler as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,12 +97,12 @@ jobs:
     install:
     - cargo clippy --version | grep $CLIPPY_VERS || cargo install clippy --force --vers $CLIPPY_VERS
     script:
-    - cargo clippy --all
+    - cargo clippy --all -- -D warnings
 
   # Tests
   - env: FEATURE=test
     script:
-    - cargo test --all
+    - RUSTFLAGS="$RUSTFLAGS -D warnings" cargo test --all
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -97,10 +97,12 @@ jobs:
     install:
     - cargo clippy --version | grep $CLIPPY_VERS || cargo install clippy --force --vers $CLIPPY_VERS
     script:
-    - cargo clippy --all
+    - cargo clippy --all -- -D warnings
 
   # Tests
   - env: FEATURE=test
+    install:
+    - export RUSTFLAGS="$RUSTFLAGS -D warnings"
```
slowli: Does it make more sense to apply `-D warnings` in this way (on the per-job basis), or would it make more sense to redefine the global setting of the `RUSTFLAGS` variable? The first way is more flexible, but can be easier to forget: e.g., `-D warnings` is not active for benchmarking job (not sure it should).

matklad: I'd say let's define `RUSTFLAGS` once globally, and, if/when we find a need to disable it for a particular job, switch to per-job setting.

DarkEld3r: > if/when we find a need to disable it for a particular job, switch to per-job setting.

I would even prefer to disable such warnings locally with attributes.

PR #495, opened 2018-02-17
Move cryptocurrency example [ECR-714]

DarkEld3r:


examples/cryptocurrency/.gitignore
```diff
@@ -0,0 +1,5 @@
+/target/
+**/*.rs.bk
+.idea/
+1.log
+cmake-build-debug/
```
matklad: Some of these should be removed or moved to the root .gitignore.

examples/cryptocurrency/.travis.yml
```diff
@@ -0,0 +1,42 @@
...
+- rustup toolchain install $NIGHTLY_RUST
+- cargo +$NIGHTLY_RUST clippy --version | grep $CLIPPY_VERS || cargo +$NIGHTLY_RUST
+  install clippy --force --vers $CLIPPY_VERS
+- if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
+- if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install libsodium rocksdb jq; fi
+script:
+- cargo fmt -- --write-mode=diff
+- cargo +$NIGHTLY_RUST clippy -- -D warnings
+- cargo test
+- cd examples && bash test.sh
```
matklad: We probably don't need a separate travis.yml and Cargo.lock?

examples/cryptocurrency/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "exonum-cryptocurrency"
+version = "0.5.0"
+authors = ["The Exonum Team <exonum@bitfury.com>"]
+repository = "https://github.com/exonum/exonum"
+readme = "README.md"
+license = "Apache-2.0"
+
+[dependencies]
+exonum = "0.5.0"
```
matklad: `exonum = { path = }`?

DarkEld3r:
@matklad I prefer to reorganize tests as a separate task/pull request. 😉

examples/cryptocurrency/README.md
```diff
@@ -0,0 +1,58 @@
+# Cryptocurrency Tutorial
+
+[![Build status][travis-image]][travis-url]
+[![Gitter][gitter-image]][gitter-url]
+
+[travis-image]: https://img.shields.io/travis/exonum/cryptocurrency.svg?style=flat-square
+[travis-url]: https://travis-ci.org/exonum/cryptocurrency
```
matklad: Urls are wrong. Perhaps just remove the badges from the readme?

DarkEld3r: I'm not sure. On the one hand, we still have badges in other readme files. On the other hand, it is probably enough to have them in the root readme.

@slowli What do you think?

slowli: I don't have a particularly strong opinion on this topic. IMO, any of the two options would do; the badges allow to quickly check CI status, but they are slightly misleading because CI tasks are not split along with packages.

DarkEld3r:
I have created a task for cryptocurrency tests reogranization: https://jira.bf.local/browse/ECR-782.

PR #496, opened 2018-02-18
Make config service better

slowli:
This PR refactors the configuration service, making it (hopefully) more understandable and analyzable.


services/configuration/src/errors.rs
```diff
@@ -0,0 +1,207 @@
...
+                    $($from::$variant => $to::$variant,)+
+                }
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+#[repr(u8)]
+enum CommonErrorCode {
```
slowli: I'm not sure how to deal with common error codes. Any suggestions how to minimize copypaste would be appreciated.

services/configuration/src/errors.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// Next configuration is already scheduled.
+    AlreadyScheduled = 0,
+    /// The sender of the transaction is not among the active validators.
+    UnknownSender,
+    /// The configuration in the proposal does not reference the currently active configuration.
+    InvalidConfigRef,
+    /// Current blockchain height exceeds the height of the proposal activation.
+    ActivationInPast,
+    /// The same configuration is already proposed.
+    AlreadyProposed = 32,
```
slowli: Not sure how codes should be assigned.

services/configuration/src/transactions.rs
```diff
@@ -0,0 +1,343 @@
...
+    fn verify(&self) -> bool {
+        self.verify_signature(self.from())
+    }
+
+    /// See [`ProposeErrorCode`] for the description of emitted error codes.
+    ///
+    /// [`ProposeErrorCode`]: enum.ProposeErrorCode.html
+    fn execute(&self, fork: &mut Fork) -> ExecutionResult {
+        let (cfg, cfg_hash) = self.precheck(fork.as_ref()).map_err(|err| {
+            error!("Discarding propose {:?}: {}", self, err);
```
slowli: Is this adequate error handling?

services/configuration/src/transactions.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        self.save(fork, &cfg, cfg_hash);
+        trace!("Put propose {:?} to config_proposes table", self);
+        Ok(())
+    }
+}
+
+impl Vote {
+    /// Checks if this vote encodes a special "absence of vote" variant.
+    pub fn is_none(&self) -> bool {
```
slowli: Should there be `is_some`?

services/configuration/src/api.rs
```diff
@@ -61,22 +61,47 @@ pub struct VoteResponse {
...
-            .get_propose(&config.hash())
+/// Filter for stored configurations.
+struct Filter {
+    previous_cfg_hash: Option<Hash>,
+    actual_from: Option<Height>,
+}
+
+impl Filter {
+    /// Checks if a supplied configuration satisfies this filter.
+    fn test(&self, cfg: &StoredConfiguration) -> bool {
```
matklad: Alternative name suggestion: `fn matches`.

services/configuration/src/api.rs
```diff
@@ -61,22 +61,47 @@ pub struct VoteResponse {
...
+        if let Some(ref prev) = self.previous_cfg_hash {
+            if cfg.previous_cfg_hash != *prev {
+                return false;
+            }
+        }
+
+        if let Some(ref from_height) = self.actual_from {
```
matklad: No `ref` is necessary, `Height` is Copy.

services/configuration/src/api.rs
```diff
@@ -199,35 +190,29 @@ impl PublicConfigApi {
...
-                    config,
-                    previous_cfg_hash_filter,
-                    actual_from_filter,
-                )
-            })
-            .map(|config| self.get_config_with_proofs(config))
+            .filter(|config| filter.test(config))
+            .map(|config| self.config_with_proofs(config))
             .collect();
         committed_configs
```
matklad: Looks it's possible to remove `let committed_configs` variable, because it is immediately returned?

services/configuration/src/lib.rs
```diff
@@ -12,673 +12,127 @@
...
-use exonum::helpers::fabric::{ServiceFactory, Context};
-use exonum::crypto::{CryptoHash, Signature, PublicKey, Hash};
-use exonum::messages::{Message, RawTransaction};
-use exonum::storage::{Fork, ProofListIndex, ProofMapIndex, Snapshot, StorageValue};
-use exonum::encoding::{Field, Error as StreamStructError};
-use exonum::encoding::serialize::json::reexport as serde_json;
-
-/// Configuration service http api.
-pub mod api;
+use exonum::blockchain::{Service as ServiceTrait, Transaction, ApiContext};
```
matklad: An alternative would be 

```
use exonom::blockchain::{self, Transaction}

impl blockchain::Service for Service 
```

services/configuration/src/schema.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// by its configuration hash.
+    pub fn votes_by_config_hash(&self, config_hash: &Hash) -> ProofListIndex<&Snapshot, MaybeVote> {
+        ProofListIndex::with_prefix(VOTES, gen_prefix(config_hash), self.view.as_ref())
+    }
+
+    /// Returns a `Propose` transaction with a particular configuration hash.
+    pub fn propose(&self, cfg_hash: &Hash) -> Option<Propose> {
+        self.propose_data_by_config_hash().get(cfg_hash).map(
+            |propose_data| propose_data.tx_propose(),
+        )
```
matklad:
```
self.propose_data_by_config_hash().get(cfg_hash)?.tx_propose()
```

services/configuration/src/schema.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    /// Returns a list of votes for the proposal corresponding to the given configuration hash.
+    #[cfg_attr(feature = "cargo-clippy", allow(let_and_return))]
+    pub fn votes(&self, cfg_hash: &Hash) -> Vec<Option<Vote>> {
+        let votes_by_config_hash = self.votes_by_config_hash(cfg_hash);
+        let votes = votes_by_config_hash.iter().map(MaybeVote::into).collect();
+        votes
```
matklad: `votes` variable does not seem to be necessary. Given that the clippy warning is silence it seems that this is intentional, but it doesn't seem to make code more readable?

services/configuration/src/tests/api.rs
```diff
@@ -19,30 +19,30 @@ use exonum::helpers::{Height, ValidatorId};
...
 
-    fn get_all_committed(
+    fn all_committed(
         &self,
         previous_cfg_hash_filter: Option<Hash>,
         actual_from_filter: Option<Height>,
     ) -> Vec<ConfigHashInfo>;
 
-    fn get_votes_for_propose(&self, cfg_hash: &Hash) -> VotesInfo;
+    fn votes_for_propose(&self, cfg_hash: &Hash) -> VotesInfo;
```
matklad: :+1: :+1: :+1:  for renames!

services/configuration/src/transactions.rs
```diff
@@ -0,0 +1,320 @@
...
+
+/// Checks if there is enough votes for a particular configuration hash.
+fn enough_votes_to_commit(snapshot: &Snapshot, cfg_hash: &Hash) -> bool {
+    let actual_config = CoreSchema::new(snapshot).actual_configuration();
+
+    let schema = Schema::new(snapshot);
+    let votes = schema.votes_by_config_hash(cfg_hash);
+    let votes_count: usize = votes
+        .iter()
+        .map(|vote| if !vote.is_none() { 1 } else { 0 })
```
matklad: `!vote.is_none` -> `vote.is_some()`. Maybe even `.iter().fiter(|vote| vote.is_some()).count()`.

services/configuration/src/transactions.rs
```diff
@@ -0,0 +1,320 @@
...
+        if let Some(following) = following_config {
+            Err(AlreadyScheduled(following))?;
+        }
+
+        if let Some(validator_id) = validator_index(snapshot, self.from()) {
+            let vote = Schema::new(snapshot)
+                .votes_by_config_hash(self.cfg_hash())
+                .get(validator_id as u64)
+                .unwrap();
+            if !vote.is_none() {
```
matklad: `!vote.is_none()` -> `vote.is_some()`

services/configuration/src/transactions.rs
```diff
@@ -0,0 +1,320 @@
...
+                Err(AlreadyVoted)?;
+            }
+        } else {
+            Err(UnknownSender)?;
+        }
+
+        let propose = Schema::new(snapshot).propose(self.cfg_hash()).ok_or_else(
+            || {
+                UnknownConfigRef(*self.cfg_hash())
+            },
```
matklad: Braces are not necessary?

services/configuration/src/api.rs
slowli: Regarding this and other places that have silenced `let_and_return` lint: The code *won't* compile without an additional variable, because the borrow checker thinks that the index is still borrowed after `collect()` (but not in the case the returned value is bound to a variable). I can't explain why this happens, and I would like to get rid of it.

services/configuration/src/transactions.rs
slowli: They are added by the formatter, unfortunately.

services/configuration/src/api.rs
matklad: Fascinating! I also can't explain what's going on here, but `#![feature(nll)]` fixes this, so I think this might be just a bug/limitation of the current borrow checker. 

However, it looks like the iterators here don't actually need lifetimes to start with? Both `MemoryDB` and `RocksDB` don't actually make use of the lifetime.

slowli:
Ready for review, assuming that updating tests (to check error codes) will be done in a separate PR. Several questions for discussion (some I've outlined previously):

- What should be the strategy for assigning error codes? They may be entirely different for `Propose` and `Vote` transactions; or have a common part, but be otherwise different; or have intersecting "non-common" part (this last approach is currently implemented).
- Should there be any descriptions attached to error codes?
- Is error handling (in particular, logging) adequate?

Further improvements:
- Testing for error codes and error logging.
- Personally, I think the `api` module should be made private and all structures defined within it could in some cases be removed entirely (i.e., replaced with `json!` macro; seems possible for `*Response`).

matklad:
Yeah, error definitions definitely look a bit awkward. A lot of this, of course, is mandated by our `u8` error code :unamused: 

It seems to me that we use these errors for two purposes:

* logging
* converting them to `u8`

In particular, we don't match on the errors anywhere. 

So, as an alternative, I would suggest a simpler approach, which seems to be enough for both tasks: 

Define a *single* error enum for the whole service, implement `ExecutionError` for it directly (btw, should we use `ExecutionError::with_description` somewhere?)

Hm, now that I think about it, perhaps even this would work? Use `ExecutionError` directly instead of a separate type, and provide *functions* for constructing appropriate `ExecutionError`s, which do logging at the instance of error construction.

slowli:
@matklad 

> Define a single error enum for the whole service

This is functionally equivalent to the "have a common part, but be otherwise different" approach I've described previously :slightly_smiling_face: 

> we don't match on the errors anywhere

Yet, but we will in testing, and the users of the service might want to match them as well, or at least have them in a consistent form.

services/configuration/src/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Common error types for `Propose` and `Vote`.
+#[derive(Debug, Fail)]
+pub(crate) enum Error {
+    #[fail(display = "Next configuration is already scheduled: {:?}", _0)]
+    AlreadyScheduled(StoredConfiguration),
+
+    #[fail(display = "Not authored by a validator")]
+    UnknownSender,
```
DarkEld3r: Should public key of the "unknown sender" will displayed?

services/configuration/src/errors.rs
```diff
@@ -0,0 +1,114 @@
...
+    UnparseableConfig(
+        #[cause]
+        JsonError
+    ),
+
+    #[fail(display = "Does not reference known config with hash {:?}", _0)]
+    UnknownConfigRef(Hash),
+
+    #[fail(display = "Validator already voted for a referenced proposal")]
+    AlreadyVoted,
```
DarkEld3r: Same here. Probably it can be useful to know who exactly sends several votes.

services/configuration/src/schema.rs
```diff
@@ -0,0 +1,273 @@
...
+// limitations under the License.
+
+//! Storage schema for the configuration service.
+
+use std::borrow::Cow;
+
+use exonum::blockchain::gen_prefix;
+use exonum::encoding::Field;
+use exonum::crypto::{self, CryptoHash, Hash, PublicKey, Signature};
+use exonum::storage::{Fork, ProofListIndex, ProofMapIndex, Snapshot, StorageValue};
```
DarkEld3r: Technically, exonum is an external crate here and we have an agreement that such using should be before standard ones.

exonum-dictionary.txt
```diff
@@ -59,11 +59,13 @@ nanos
 nodelay
 oneshot
 openssl
+parseable
```
ozkriff: Maybe it's better to use "parsable"?

vitvakatu: ... or not use it at all. `Unparseable` can be replaced with `invalid` imo

services/configuration/src/api.rs
```diff
@@ -249,48 +234,46 @@ where
...
 }
 
-impl Api for PublicConfigApi {
+impl Api for PublicApi {
     fn wire(&self, router: &mut Router) {
-
         let self_ = self.clone();
         let config_actual = move |_: &mut Request| -> IronResult<Response> {
-            let info = self_.get_actual_config();
+            let info = self_.actual_config();
```
DarkEld3r: I don't like all these `let self_ = self.clone()`. What do you think about creating separate methods that takes `self` by value (similar to [this](https://github.com/exonum/exonum/blob/master/exonum/src/api/public/blockchain_explorer.rs#L126))?

services/configuration/src/errors.rs
slowli: I assume that the error is output close to a faulty transaction (which is the case for error logging, in particular), so this information will be redundant - the sender will be displayed as a part of the transaction.

services/configuration/src/tests/mod.rs
```diff
@@ -20,46 +20,43 @@ use exonum::storage::StorageValue;
...
-    fn find_propose(&self, config_hash: Hash) -> Option<TxConfigPropose>;
+    fn find_propose(&self, config_hash: Hash) -> Option<Propose>;
 }
 
 impl ConfigurationTestKit for TestKit {
     fn configuration_default() -> Self {
         TestKitBuilder::validator()
             .with_validators(4)
-            .with_service(ConfigurationService::new())
+            .with_service(ConfigurationService {})
```
mvlabat: Shall we make it a unit-like struct?

PR #500, opened 2018-02-19
Rework badges in readmes [ECR-593]

vitvakatu:
What do you think?

[Main readme](https://github.com/vitvakatu/exonum/tree/ecr_593/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 [![dependency status](https://deps.rs/repo/github/exonum/exonum/status.svg)](https://deps.rs/repo/github/exonum/exonum)
 
 **Project info:**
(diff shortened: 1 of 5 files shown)


ozkriff:
What about adding explicit group names, something like
image
?

DarkEld3r:
I like the idea of the group names. As for me, it is quite the opposite of noise: it helps to navigate through badges.

vitvakatu:
Another question, do we need crates versions in readmes?

ozkriff:
docs.rs badge should already show the latest version:
https://docs.rs/exonum/badge.svg

PR #502, opened 2018-02-19
API: Update Blocks-in-range endpoint Attempt 2 [ECR-791]

quicky84:
This PR is directly related to #351: appends suggested changes, optimizes the code, and adds requested test.

### DIFF
--- a/exonum/src/explorer/mod.rs
+++ b/exonum/src/explorer/mod.rs
@@ -77,7 +77,7 @@ pub enum TxStatus {
 }
 
 /// Range information. Range borders are included.
-#[derive(Debug, Serialize)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct Range {
     /// Left border of the range, >=0.
     pub from: u64,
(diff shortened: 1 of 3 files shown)


exonum/src/explorer/mod.rs
```diff
@@ -76,6 +76,24 @@ pub enum TxStatus {
     },
 }
 
+/// Range information. Range borders are included.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Range {
```
vitvakatu: Can't we use `std::Range`?

quicky84: https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html this one?

vitvakatu: https://doc.rust-lang.org/std/ops/struct.Range.html this? Or you need inclusive one?

quicky84: By the logic of the block exploration, it makes sense to work with closed intervals.

vitvakatu: Sadly, I don't think it's possible as it's nightly-only API atm.

vitvakatu:
Replaced by #509

PR #505, opened 2018-02-20
Fix `TestKitApi::send` signature

alekseysidorov:
This PR fixes problem with signature of `TestKitApi::send` method, which previosly did not accept `Box<Transaction>`.

### DIFF
--- a/testkit/CHANGELOG.md
+++ b/testkit/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
  - Modified signature of the `TestKitApi::send` method, which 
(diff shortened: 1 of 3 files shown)


slowli:
Could you please add a test where `Box<Transaction>` is passed to the method in order to prevent regression in future?

testkit/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
```
DarkEld3r: What do you think about the new core changelog format? As for me, it would be better to use it everywhere, so instead of "changed" section" there probably should be "bug fixes". Maybe such format is too much for our other projects, so we can list the improvements without sections?..

PR #507, opened 2018-02-20
Fix `TransactionInfo` JSON serialization

slowli:
This PR fixes the format of transaction information returned by the blockchain explorer, to use `"type": "unknown" | "in-pool" | "committed"` tag instead of returning data like `{ "in-pool": { "content": ... } }`.

### DIFF
--- a/testkit/tests/counter/main.rs
+++ b/testkit/tests/counter/main.rs
@@ -567,6 +567,6 @@ fn test_explorer_transaction() {
                 .is_ok()
         );
     } else {
-        // Already proven false by `assert_contains_all`
+        panic!("Invalid transaction info format, object expected");
     }
 }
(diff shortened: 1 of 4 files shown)


testkit/tests/counter/main.rs
```diff
@@ -513,48 +519,54 @@ fn test_explorer_transaction() {
...
         let block = block.block;
         assert!(
-            info.location_proof
+            location_proof
                 .validate(*block.tx_hash(), u64::from(block.tx_count()))
                 .is_ok()
         );
     } else {
-        panic!("Transaction should be committed");
+        // Already proven false by `assert_contains_all`
```
DarkEld3r: I'm insisting on panic here because `assert_contains_all` implementation can be changed during some refactoring. Other solution: return `serde_json::map::Map` from the `assert_contains_all` (which should probably be renamed in that case), so `if let` will not be needed.

PR #509, opened 2018-02-21
Update blocks in range endpoint [ECR-794, ECR-791]

vitvakatu:


quicky84:
Thanks for taking this over. I am not completely happy with the default range [start, end) because it goes against the logic. Say there is only the genesis block. then we have this in the test
```
let blocks: Vec<Block> = api.get(ApiKind::Explorer, "v1/blocks?count=10");
let response: BlocksRange = api.get(ApiKind::Explorer, "v1/blocks?count=10");
let (blocks, range) = (response.blocks, response.range);
assert_eq!(blocks.len(), 1);
...
assert_eq!(range.start, 0);
assert_eq!(range.end, 1);
```
which, if no extra info is provided, reads "blocks 0 and 1 were explored to find the required blocks" and "the id of the last block is 1", both statements are false.
Why not have `Range` with `start` and `end` as a custom struct and then get rid of it in favor of `RangeInclusive<Idx>` as soon as it is stable?

vitvakatu:
@quicky84 the reason behind this change is keeping API surface small and use standard ways instead of hand-made ones. The logic now is a bit strange indeed, but ranges (and their behavior) are well familiar to every Rust programmer, so I don't think it's the problem.
I'll create a task to make a transition to `InclusiveRange`.

PR #512, opened 2018-02-21
Remove time tutorial

slowli:
This PR removes the tutorial from the time service and replaces it with a link to [the Exonum website](https://exonum.com/doc/advanced/time).

Also, there are some possibly controversial changes: changing the badges for the time service README - I feel that the old ones are either unnecessary or duplicate ones in the main README. Instead, I left service-specific badges - the latest crate version and the crate docs.

### Discussion

I'm not sure whether the HTTP endpoint description belongs in the crate docs, the Exonum website docs, or elsewhere. From the maintenance perspective, it would be easier to keep it in the crate docs (e.g., as a separate section in the service description).


services/time/README.md
```diff
@@ -1,10 +1,7 @@
 # exonum-time
 
-[![Travis Build Status](https://img.shields.io/travis/exonum/exonum/master.svg?label=Linux%20Build)](https://travis-ci.org/exonum/exonum)
```
vitvakatu: The idea behind all those badges is simple: they're visible on crates.io page and provide some basic info about crate (that it's tested, have a copy-left license and requires specific Rust version). At the same time, crates.io badge is pretty useless (the last version is already visible on docs.rs badge and it's the only thing you can find on crates.io and not in readme).

vitvakatu: So I intend to disagree with this change..

slowli:
I got an idea about HTTP API (which goes against what I just did in the config service, but I don't mind reverting it): Have API be described as `PublicApi` and `PrivateApi` structures within the crate, with the following public interface:

- Constructor `new(_: &ApiContext) -> Self`
- Public methods corresponding 1-to-1 to endpoints, with signatures like `get_smth(req: RequestType) -> ResponseType`. The endpoints would be documented right there

If you don't mind, I'll try to elaborate on this concept and apply it to the time service and/or config service to see how it may work.

slowli:
@DarkEld3r The task has escalated into a complete rewrite of the `api` module (currently elaborating the interface; hope to complete a first draft during the weekend) :sweat_smile: So, you can merge this PR if you think it would be appropriate.

PR #513, opened 2018-02-21
Clean up after #496

slowli:
This PR cleans up some mistakes in #496, such as incorrect documentation of transactions. It also provides the changelog.

The documentation errors was noticed by [the `deadlinks` lint](https://travis-ci.org/exonum/exonum/jobs/344503829#L1732). Unfortunately, it seems that the documentation for `Option::as_ref` method (mentioned on the `MaybeVote` page because it implements `Deref<Target = Option<Vote>>`) is copied verbatim from the `std` docs, which leads to 3 more incorrect links discovered by `deadlinks`. Honestly, I don't know how to best fix this.

I can't help but notice that the changelog does not describe #481 (and possibly other changes as well). Should I include them as well, or will they be dealt with separately?

### DIFF
--- a/services/configuration/CHANGELOG.md
+++ b/services/configuration/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+<!-- cspell:ignore ZEROVOTE -->
+
 All notable changes to this project will be documented in this file.
 The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
@@ -7,7 +9,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Breaking changes
 
-- Most types renamed to avoid [stuttering] (#496):
+- Most types renamed to avoid stuttering (see [here][stuttering] for
+  an explanation of the term) (#496):
 
   - `ConfigurationService` to `Service`
   - `ConfigurationServiceFactory` to `ServiceFactory`
@@ -19,6 +22,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
   Check the crate documentation for more details.
 
+  **Migration path:** Rename imported types from the crate, using aliases
+  or qualified names if necessary: `use exonum_configuration::Service as ConfigService`.
+
 [stuttering]: https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-[rfc-356]
 
 - Multiple APIs are no longer public (#496):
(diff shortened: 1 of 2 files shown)


DarkEld3r:
> Should I include them as well, or will they be dealt with separately?

As you wish, but I would be grateful if you dealt with them. 🙃

services/configuration/CHANGELOG.md
```diff
@@ -7,10 +7,35 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+
+  Check the crate documentation for more details.
+
+- `ZEROVOTE` structure is replaced with the `MaybeVote` type, which is now used
+  instead of `Vote` in the schema method signatures. The storage format itself
+  is unchanged (#496).
+
+### New features
+
+- Implemented error handling based on error codes (#496).
```
DarkEld3r: I'm not sure which formatting is better, but in the core changelog dot is placed before the pull request number. For example:
> Implemented error handling based on error codes. (#496)

services/configuration/CHANGELOG.md
```diff
@@ -7,10 +7,35 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+  - `ConfigurationServiceFactory` to `ServiceFactory`
+  - `TxConfigPropose` to `Propose`
+  - `TxConfigVote` to `Vote`
+  - `ConfigurationSchema` to `Schema`
+  - `StorageValueConfigProposeData` to `ProposeData`
+  - `CONFIG_SERVICE` constant to `SERVICE_ID`
+
+  Check the crate documentation for more details.
+
+[stuttering]: https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-[rfc-356]
```
matklad: FWIW, it's not an *authoritative* style guide, and the recent libraries guidelines intentionally do not include this rule: https://rust-lang-nursery.github.io/api-guidelines/naming.html.

So, looks like the state of the art is "it is unclear, which is better, io::Error or IoError, use your own judgment": https://github.com/rust-lang-nursery/futures-rfcs/pull/2#issuecomment-364271680

slowli: Yeah, I intended to use a link primarily to explain the term, not to appeal to authority. As a side note, `Error`, `Result` and some other names are special, because they shadow types included into *every* scope.

services/configuration/CHANGELOG.md
```diff
@@ -1,16 +1,53 @@
 # Changelog
 
+<!-- cspell:ignore ZEROVOTE -->
```
DarkEld3r: What do you think about ignoring it locally? `ZEROVOTE` is removed, so we shouldn't encounter it in other places.

services/configuration/CHANGELOG.md
```diff
@@ -1,16 +1,53 @@
...
+  is unchanged (#496).
+
+### New features
 
-- `CONFIG_PROPOSE_MESSAGE_ID` and `CONFIG_VOTE_MESSAGE_ID` constants are no
-  longer public.
+- Information about configurations by `/v1/configs/actual`, `/v1/configs/following`
+  and `/v1/configs/committed` endpoints is extended with the hash of the corresponding
+  proposal and votes for the proposal (#481).
+- Implemented error handling based on error codes (#496).
```
DarkEld3r: I propose to separate list items by a new line.

PR #514, opened 2018-02-22
Make log output more human readable

alekseysidorov:
Use `rfc2822` for time formatting and also reduce visual noise.
Also this PR will break the benchmark scripts.

Here example of a new output:
Example of a new output

### DIFF
--- a/exonum/CHANGELOG.md
+++ b/exonum/CHANGELOG.md
@@ -70,9 +70,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `Connect` message has been extended with a user agent string, which breaks
   binary compatibility with previous versions. (#362)
 
-- Log output becomed more human readable. Now it uses `rfc2822` for time formatting.
-  It will break scripts that analyze the log output.  
-  (#514)
+- Log output become more human-readable. Now it uses `rfc2822` for time formatting.
+  This change can break scripts that analyze the log output. (#514)
 
 ### New features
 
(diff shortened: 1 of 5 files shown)


exonum/src/helpers/mod.rs
```diff
@@ -16,11 +16,12 @@
 
 use std::env;
 use std::io::{self, Write};
-use std::time::{SystemTime, UNIX_EPOCH};
+use std::time::SystemTime;
 
 use log::{Level, Record, SetLoggerError};
 use env_logger::{Builder, Formatter};
 use colored::*;
+use chrono::prelude::*;
```
DarkEld3r: As far as I can see, you are using only two types (`DateTime` and `Local`). I propose to replace `use chrono::prelude::*` with `use chrono::{DateTime, Local}`.

exonum/CHANGELOG.md
```diff
@@ -70,6 +70,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `Connect` message has been extended with a user agent string, which breaks
   binary compatibility with previous versions. (#362)
 
+- Log output becomed more human readable. Now it uses `rfc2822` for time formatting.
+  It will break scripts that analyze the log output.  
+  (#514)
```
DarkEld3r: I propose to reword a little: 
> Log output become more human-readable. Now it uses `rfc2822` for time formatting. This change can break scripts that analyze the log output.

DarkEld3r:
@vitvakatu Currently benchmark scripts aren't located in this repository and they should be updated anyway.

PR #515, opened 2018-02-22
Add meaningful transaction errors for exonum-time[ECR-685]

SerhiiOryshych:


services/time/src/lib.rs
```diff
@@ -111,62 +113,114 @@ transactions! {
...
+        self.verify_signature(self.pub_key())
+    }
+
+    fn execute(&self, view: &mut Fork) -> ExecutionResult {
+        // The transaction must be signed by the validator.
+        self.signed_by_validator(view.as_ref(), self.pub_key())
+            .map_err(|err| err)?;
+
+        self.update_validator_time(view, self.pub_key(), self.time())
+            .map_err(|err| err)?;
```
matklad: Hm, why do we need these no-op `map_err`s?

services/time/src/lib.rs
```diff
@@ -111,62 +113,114 @@ transactions! {
...
+}
+
+#[derive(Debug, Fail)]
+enum Error {
+    #[fail(display = "Not authored by a validator")]
+    UnknownSender,
+
+    #[fail(display = "The validator time is longer than the proposed one")]
+    ValidatorTimeIsLonger,
+}
```
matklad: Currently, `ErrorCode` and `Error` enums are exact duplicates. I understand that, in the future, we might want to extend `Error` with variants with data, but I would suggest using a single `enum` until that time.

services/time/src/lib.rs
```diff
@@ -111,62 +113,99 @@ transactions! {
...
+    }
+}
+
+impl TxTime {
+    fn signed_by_validator(&self, snapshot: &Snapshot, key: &PublicKey) -> ExecutionResult {
+        let keys = Schema::new(&snapshot).actual_configuration().validator_keys;
+        let signed = keys.iter().any(|k| k.service_key == *key);
         if !signed {
-            return Ok(());
+            Err(Error::UnknownSender)?;
```
DarkEld3r: I propose to rewrite the following way:
```rust
if keys.iter().any(|k| k.service_key == *key) {
    Err(Error::UnknownSender)
} else {
    Ok(())
}
```

services/time/src/lib.rs
```diff
@@ -111,62 +113,99 @@ transactions! {
...
             // Write the time for the validator.
-            _ => {
-                schema.validators_times_mut().put(
-                    self.pub_key(),
-                    self.time(),
-                )
-            }
+            _ => schema.validators_times_mut().put(key, proposed_time),
         }
+        Ok(())
```
DarkEld3r: Result can be returned directly from the match:
```rust
match schema.validators_times().get(key) {
    Some(time) if time >= proposed_time => {
        Err(Error::ValidatorTimeIsLonger)
    }
    _ => {
        schema.validators_times_mut().put(key, proposed_time);
        Ok(())
    }
}
```

services/time/src/lib.rs
```diff
@@ -111,62 +113,100 @@ transactions! {
...
-    }
+#[derive(Debug, Fail)]
+#[repr(u8)]
+enum Error {
+    #[fail(display = "Not authored by a validator")]
+    UnknownSender = 0,
 
-    fn execute(&self, view: &mut Fork) -> ExecutionResult {
-        let validator_keys = Schema::new(&view).actual_configuration().validator_keys;
+    #[fail(display = "The validator time is longer than the proposed one")]
```
slowli: I don't think "longer" can be used in this context. Perhaps, "greater" would be better? @Maria-Nosyk

services/time/src/lib.rs
```diff
@@ -111,62 +113,100 @@ transactions! {
...
-            validator.service_key == *self.pub_key()
-        });
+impl From<Error> for ExecutionError {
+    fn from(value: Error) -> ExecutionError {
+        ExecutionError::new(value as u8)
+    }
+}
+
+impl TxTime {
+    fn signed_by_validator(&self, snapshot: &Snapshot, key: &PublicKey) -> ExecutionResult {
```
slowli: - Cannot `key` be taken from `self`?
- I don't quite like the method name; maybe, `check_signed_by_validator` could be better?

services/time/src/lib.rs
```diff
@@ -111,62 +113,100 @@ transactions! {
...
+impl Transaction for TxTime {
+    fn verify(&self) -> bool {
+        self.verify_signature(self.pub_key())
+    }
+
+    fn execute(&self, view: &mut Fork) -> ExecutionResult {
+        // The transaction must be signed by the validator.
+        self.signed_by_validator(view.as_ref(), self.pub_key())?;
+
+        self.update_validator_time(
```
slowli: Again, if `update_validator_time` is a method, then `pub_key` and `time` can be taken from `self`.

services/time/src/lib.rs
```diff
@@ -111,68 +113,98 @@ transactions! {
...
             }
         }
+    }
+
+    fn update_consolidated_time(fork: &mut Fork) {
+        let keys = Schema::new(&fork).actual_configuration().validator_keys;
+        let mut schema = TimeSchema::new(fork);
 
         // Find all known times for the validators.
         let mut validator_times: Vec<SystemTime>;
```
slowli: I suggest using the following code here:

```rust
let validator_times = {
    let idx = schema.validators_times();
    let mut times: Vec<SystemTime> = idx.iter()
        .filter_map(|(public_key, time)| {
            keys.iter()
                .find(|validator| validator.service_key == public_key)
                .map(|_| time)
        })
        .collect();
    times.sort();
    times.reverse();
    times
};
```

(You may use `sort_by` - it's probably faster, although I find it less understandable).

PR #516, opened 2018-02-23
Rework configuration tutorial with docker [ECR-697]

vitvakatu:


services/configuration/docker/example-start.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+# Usage:
+# example-start.sh <number of nodes>
+
+node_count=$1
+port_start=8000
+port_end=$((port_start + (2 * node_count) - 1))
+
+docker run -p ${port_start}-${port_end}:${port_start}-${port_end} vitvakatu/exonum-configuration-example:latest $node_count
```
aleksuss: Is this a right path `vitvakatu/` ?

services/configuration/docker/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM rust:1.23.0
+
+WORKDIR /usr/src
```
aleksuss: Why would you prefer to use a path `/usr/src` ?

services/configuration/doc/testnet-api-tutorial.md
```diff
@@ -12,106 +38,48 @@ cargo install --example configuration
 `exonum` crate system dependencies and rust toolchain configuration -
 [exonum install instructions](https://github.com/exonum/exonum/blob/master/INSTALL.md).
 
-## Running testnet
+#### Generate testnet directory and config
+
+`4` is a required indexed parameter and stands for the number of nodes in
```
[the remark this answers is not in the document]
vitvakatu: Seems reasonable, but I can't imagine how to explain the distribution of ports for every node. Will try though

services/configuration/docker/Dockerfile
vitvakatu: Well, actually, I don't know. I've used the example from official rust docker image. Is there any drawbacks?

aleksuss: Actually this path is used for storing some source files. Linux kernel sources e.g. I'd prefer some thing in `home` directory. E.g. `/home/exonum` Basically it is not important. Just an advice.

PR #520, opened 2018-02-25
WIP: Refactor service APIs

slowli:
The changes to the `api` module I've mentioned in #512.

For now, the changes are non-breaking - you may either use the new approach to API or the old one. This is for the sake of flexibility: if the interfaces change (which most probably *will* happen), there will be no need to overhaul the entire repo.


examples/cryptocurrency/src/lib.rs
```diff
@@ -229,93 +215,97 @@ impl Transaction for TxTransfer {
...
+    ///     }, {
+    ///         "pub_key": bob_pubkey,
+    ///         "name": "Bob",
+    ///         "balance": "100",
+    ///     }])
+    /// );
+    /// # }
+    /// ```
+    @(ID = "wallets")
+    pub GetWallets(()) -> Vec<Wallet>;
```
matklad: Can we infer `ID` from the name of endpoint itself?

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,743 @@
...
+
+    // `pub(..)` visibility specifier
+    // XXX: `pub(in ..)` processing is essentially a hack
+    (
+        $(#[$attr:meta])*
+        @(ID = $id:expr) pub($(in)* $vis:path) $name:ident($req:ty) -> $resp:ty;
+    ) => {
+        $(#[$attr])*
+        #[derive(Debug, Clone)]
+        pub(in $vis) struct $name($crate::blockchain::Blockchain);
```
matklad: So this creates a bunch of code behind the scene to basically wrap a function. I wonder if we can avoid this code generations and use functions directly instead? I imagine the API to look like

```
        let api = ApiBuilder::new(ctx)
            .add_transactions::<TimeTransactions>()
            .add_get("time", |blockchain: &Blockchain, req: ()| -> ApiResult<SystemTime> { ... });
```

That is, we can use `Fn(&Blochain, U) -> ApiResult<V> where U: Deserialize, V: Serialize` as a `Handler` (or, as an alternative, with `RequestCtx` instead of `Blockchain`, for future extensibility). The main benefit is that there's no macros, so the user doesn't have to decipher what does a CamelCase named structofunction actually do :-) It also allows to close over some over data, and, for example, to gain access to a service-local state from a handler, which might come useful for some services.

matklad:
Might be interesting to try to port `api/public` and `api/private` to this as well!

exonum/src/api/ext.rs
slowli: I'll post a more-or-less detailed memo with design considerations internally approx. tomorrow, so extended discussion will be there :upside_down_face: The use of one (singleton) struct per request in the current design is intentional; it makes it easy to locate all params of the request - it suffices to look at the docs, namely implementation of `EndpointSpec`. And it's on the separate page, too. Additionally, this page may (and imo should) contain at least one example allowing to fix the JSON format for requests / responses of the endpoint for client devs (look at the cryptocurrency demo for an example).

exonum/src/api/iron.rs
```diff
@@ -0,0 +1,148 @@
...
+/// TODO
+pub fn into_handler(api: ServiceApi) -> Box<Handler> {
+    // Can an endpoint be used in `GET` HTTP requests?
+    fn can_get(e: &BoxedEndpoint) -> bool {
+        e.method() == Method::Get
+    }
+
+    // Can an endpoint be used in `POST` HTTP requests?
+    fn can_post(_: &BoxedEndpoint) -> bool {
+        true
```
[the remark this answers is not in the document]
slowli: I see as intentional (albeit discussable) feature - some user agents (and maybe proxy servers too) limit the maximum length of URLs, so using `POST` for "heavyweight" requests that still have `GET` semantics (i.e., are safe) could be a nice feature. Also, if I understand correctly, the body can be compressed (signaled via [`Content-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding)), while the URL can't.

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,743 @@
...
+        impl $crate::api::ext::EndpointSpec for $name {
+            type Request = $req;
+            type Response = $resp;
+
+            const METHOD: $crate::api::ext::Method = $crate::api::ext::Method::Get;
+            const ID: &'static str = $id;
+        }
+
+        impl AsRef<$crate::blockchain::Blockchain> for $name {
+            fn as_ref(&self) -> &$crate::blockchain::Blockchain {
```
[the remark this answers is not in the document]
slowli: I don't like it either, but it seems that there can't be multiple inherent implementations (`impl Type { ... }`), so by defining one we would rob the developers of the possibility to define their own inherent methods...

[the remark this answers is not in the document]
matklad: Hm, I believe multiple impl blacks are allowed?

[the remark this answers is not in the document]
slowli: Yeah, seem to work; don't know why I thought they didn't. Will remove `AsRef` then.

examples/cryptocurrency/src/lib.rs
alekseysidorov: In my opinion syntax like `@ (ID =" wallets ")` is inconsistent and a quite unclear. I propose to use rust attribute syntax:
```rust
#[exonum_api(id = "wallets")]
pub get_wallets() -> Vec<Wallet>
```

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,542 @@
...
+/// The specification for the "standard" transaction sink.
+///
+/// Transaction sinks can be added to service APIs using the [`set_transactions`]
+/// method in `ServiceApi`; see it for more details.
+///
+/// [`add_transactions`]: struct.ServiceApi.html#method.set_transactions
+pub const TRANSACTIONS: Spec = Spec { id: "transactions" };
+
+/// API-related errors.
+#[derive(Debug, Fail)]
```
alekseysidorov: In according to failure release plan is more suitable to use `display_derive` instead of use `[fail(display = "")]`
https://boats.gitlab.io/blog/post/2018-02-22-failure-1.0/

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,542 @@
...
+    /// A transaction processed by API has failed to be sent.
+    #[fail(display = "Failed to send transaction: {}", _0)]
+    TransactionNotSent(
+        #[cause]
+        io::Error
+    ),
+
+    /// Generic server-side error.
+    #[fail(display = "Internal server error: {}", _0)]
+    InternalError(Box<::std::error::Error + Send + Sync>),
```
alekseysidorov: Why not `failure::Error`?

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,542 @@
...
+
+    /// Gets the public key used for signing operations by services.
+    pub fn service_public_key(&self) -> &PublicKey {
+        self.blockchain.service_public_key()
+    }
+}
+
+/// Context supplied to mutating endpoints.
+#[derive(Debug)]
+pub struct MutContext<'a> {
```
alekseysidorov: It sounds a little bit inconsistent with the `ReadContext`.

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,542 @@
...
+    inner: ReadContext<'a>,
+}
+
+impl<'a> ::std::ops::Deref for MutContext<'a> {
```
alekseysidorov: What do you think about the using https://crates.io/crates/derive_deref?

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,542 @@
...
+            .api_sender()
+            .send(transaction.into())
+            .map_err(ApiError::from)
+    }
+
+    /// Queues a transaction for signing, sending over the network and including
+    /// into the blockchain.
+    ///
+    /// The transaction is signed with the service secret key and returned.
+    pub fn sign_and_send(&self, message: &RawMessage) -> Result<Hash, ApiError> {
```
alekseysidorov: Is this code using only for transactions from the service?

exonum/src/api/ext.rs
slowli: Yeah, I don't quite like the naming of context types. Maybe, `EndpointContext` / `MutEndpointContext`? Or use shorter names, assuming that devs will qualify if necessary (e.g., `api::Context` / `api::MutContext`)?

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,542 @@
...
+    pub fn new() -> Self {
+        ServiceApi::default()
+    }
+
+    /// Adds an instantiated endpoint.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the API already contains an endpoint with the same identifier.
+    pub fn insert(&mut self, spec: Spec, endpoint: Endpoint) {
```
alekseysidorov: Why not associate `Spec` with the `Endpoint`?

Somethink like this:
```rust
use serde::de::DeserializeOwned;
use serde::Serialize;
use exonum::api::ApiError;

pub trait HttpConstEndpoint {
    fn path(&self) -> String;
    fn handle(&self, args: String) -> Result<String, ApiError>;
}

pub trait ConcreteConstEndpoint: Send + Sync + 'static {
    const PATH: &'static str;
    type Arguments: DeserializeOwned;
    type Value: Serialize;

    fn handle(&self, args: Self::Arguments) -> Result<Self::Value, ApiError>;
}

struct Wrapper<T: ConcreteConstEndpoint>(T);

impl<T: ConcreteConstEndpoint> HttpConstEndpoint for Wrapper<T> {
    fn path(&self) -> String {
        String::from(T::PATH)
    }
    fn handle(&self, args: String) -> Result<String, ApiError>
    {
        let args: T::Arguments = ::serde_json::from_str(&args).unwrap();
        let value = json!(self.0.handle(args).unwrap());
        Ok(::serde_json::to_string(&value).unwrap())
    }
}

fn to_http_endpoint<T: ConcreteConstEndpoint>(endpoint: T) -> Box<HttpConstEndpoint> {
    let wrapper = Wrapper(endpoint);
    Box::new(wrapper) as Box<HttpConstEndpoint>
}
```

slowli:
To comment a little bit on changes in bb5a5a6a92df09ad78515b463604a557d86e41e8 .. b5adf4a55d1d83edf8a881a4859da94c37e40895:

- The interface of the core APIs (`api::ext`) has got much saner. There is still room for improvement, obviously, but I don't expect *substantial* overhauls now.
- I've reworked `ApiSender`. It is now logically tied with the `Blockchain` type (it's even moved correspondingly). It also now has ability to queue transactions for signing with the service private key (quite hack'y because we don't have a concept of unsigned transactions). In turn, this means that this secret key is stashed in a dedicated signer on blockchain creation, and is no longer part of the blockchain instance or any contexts it provides to the services.

Features that seem easy enough to implement in this PR:

- Public / private visibility in the endpoint `Spec`.
- Less hacks in the signer (if possible).

Further developments:

- Specifying service API as `Service::api(&self) -> ServiceApi`
- Making `BlockchainApi` as a map `Service` -> `ServiceApi` and supplying it to the adapters (e.g., via a blockchain method).
- Sane Iron adapter interface / lifecycle.
- Sane configuration for the Iron adapter.

slowli:
After some hesitation, I've brought strongly-typed endpoints back (named `TypedEndpoint` as of 59698b006f25b9408df6a4e052042b6dea532d11). I don't quite like the related changes in `Context`, namely that restrictions to mutable methods are enforced in the runtime rather than through the type system:

```rust
struct Misconfigured;

impl TypedEndpoint for Misconfigured {
    type Arg = u64;
    type Output = Hash;

    const ID: &'static str = "send";
    const VIS: Visibility = Visibility::Private;
    // Need to manually specify:
    // const CONSTANT: bool = false;

    fn call(&self, ctx: &Context, arg: u64) -> ApiResult<Hash> {
        let tx = MyTx::new(arg, &crypto::gen_keypair().1);
        let tx_hash = tx.hash();
        ctx.as_mut().send(tx)?; // will panic in `as_mut()`
        Ok(tx_hash)
    }
}
```

OTOH, this seems to be inevitable future of some context operations; and it's better than the upstream design.

exonum/src/api/ext.rs
```diff
@@ -0,0 +1,751 @@
...
+    pub fn service_public_key(&self) -> &PublicKey {
+        self.blockchain.service_public_key()
+    }
+
+    /// Accesses the mutable context.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if invoked in a constant endpoint.
+    pub fn as_mut(&self) -> &MutContext<'a> {
```
alekseysidorov: So can we avoid to write code that can panic?
Maybe we can split `TypedEndpoint` to the `TypedConstEndpoint` and `TypedReadEndpoint` with the `&Context` and `&mut Context` arguments?

slowli:
@alekseysidorov I've already had both your suggestions at 68b39d4d28eeb0be60b5dc8c216e41b3b08e61b8 :upside_down_face: (There is some point in preserving commit history, after all.)

I'm moderately against using `&Context` / `&mut Context` distinction, because it is misleading and un-Rust'ful: there is no technical or semantic reasons for `send` and other "mutable" methods to require `&mut self`, so I'd rather go with 2 different types (as it's the case with `Endpoint::new` / `Endpoint::create_mut`). Regarding 2 typed endpoint traits, I'm more positive, but they have their own share of problems:

- You either have to extract the types and constants to an aux trait or duplicate code
- `wire` method will need to be duplicated anyway
- Generic `From<T: _> for Endpoint` will need to be removed, as there is no way to guarantee that some type will not implement *both* mutable and constant endpoint traits (so such code would simply not compile). Thus, you'll need to write typed endpoint -> `Endpoint` conversions in a roundabout manner.

slowli:
OK, it seems that I've finally got the typed endpoint hierarchy; it is similar to inheritance in `Fn` / `FnMut` closures. Looks quite good, imo. Also, I've changed my decision on `&mut Context`; "mutable" methods of the context may potentially require `&mut self` in the future, and using a new type may be an overkill on its own right.

slowli:
Closing as deprecated. I'm currently working on PoC, which uses ZeroMQ to create minimalistic *SDK*, which could be used to implement web APIs, oracles, clients, etc. (maybe even transaction processors eventually) - all decoupled from the core. I expect the first draft to be ready at the next week. The draft may have serious limitations, such as absence of the `tokio` reactor for ZeroMQ message processing, but the general functionality should be ready. I hope to achieve some level of compatibility with #727; most probably, by moving web API-related methods from `Service` into a separate trait.

Link for the initiated: https://wiki.bf.local/pages/viewpage.action?pageId=10093217.

PR #523, opened 2018-02-26
Compare `round_timeout` and `timeout_adjuster.timeout` [ECR-650]

SerhiiOryshych:


matklad:
Should this include a test?

PR #526, opened 2018-02-27
Add shutdown endpoint [ECR-692]

DarkEld3r:


DarkEld3r:
It "works" because the node shuts down, but I get the following error:
> thread 'main' panicked at 'Node return error: Error { repr: Custom(Custom { kind: Other, error: StringError("An error in the `Network` thread occurred: can\'t cancel routine") }) }'

Additionally, I'm not happy with events processing.

exonum/src/api/private/system.rs
```diff
@@ -12,12 +12,13 @@
...
 
-use std::net::SocketAddr;
-use std::collections::HashMap;
-
+use serde_json;
 use router::Router;
 use iron::prelude::*;
 
+use std::net::SocketAddr;
+use std::collections::HashMap;
```
matklad: Not sure there's a real rule for  it, but it seems that std imports are usually first?

exonum/src/events/internal.rs
```diff
@@ -68,6 +65,19 @@ impl InternalPart {
                             .map_err(|_| panic!("Can't execute jump to round"));
                         to_box(fut)
                     }
+                    InternalRequest::Shutdown => {
+                        let internal_tx = internal_tx.clone();
+                        let fut = Ok(())
+                            .into_future()
+                            .and_then(move |_| {
```
matklad: Why do we need `Ok(()).and_then`? This should be equivalent to `futures::lazy`?

exonum/src/api/private/system.rs
DarkEld3r: Nope: https://github.com/exonum/exonum/blob/master/CONTRIBUTING.md#conventions
:smile:

DarkEld3r:
@alekseysidorov Nice idea! I will update tests as a separate pull request: https://jira.bf.local/browse/ECR-847 :wink:

PR #527, opened 2018-02-27
Fix bug with the extra creation of the genesis configuration. [ECR-826]

alekseysidorov:
Before this, PR, `create_genesis_block` was called after each` Node` creation.

### DIFF
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -45,10 +45,10 @@ use vec_map::VecMap;
 use byteorder::{ByteOrder, LittleEndian};
 use mount::Mount;
 
-use crypto::{self, Hash, CryptoHash, PublicKey, SecretKey};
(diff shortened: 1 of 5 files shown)


exonum/src/blockchain/mod.rs
```diff
@@ -174,8 +174,18 @@ impl Blockchain {
         Schema::new(&self.snapshot()).last_block()
     }
 
+    /// This method checks whether the genesis block has already been created.
```
DarkEld3r: I propose to remove "this method" part:
> Checks whether the genesis block has already been created.

exonum/src/blockchain/mod.rs
```diff
@@ -174,8 +174,18 @@ impl Blockchain {
...
+    pub fn is_initialized(&self) -> bool {
```
DarkEld3r: Should this method be public? I suppose there is no harm in that, but will it be really useful? I think it is better to have minimalistic public API, because it is easy to extend it, but otherwise it will be a breaking change.

PR #529, opened 2018-03-01
Employ 'bodyparser' crate [ECR-688]

ozkriff:
This PR employs `bodyparser` crate for parsing data sections of POST-requests.

Motivation: https://github.com/exonum/exonum/pull/446/files#r166044729

Should I add a changelog item?

### DIFF
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -33,6 +33,7 @@ use serde_json;
 use serde::{Serialize, Serializer};
 use serde::de::{self, Visitor, Deserialize, Deserializer};
 use failure::Fail;
+use bodyparser;
 
 use crypto::{PublicKey, SecretKey};
 use encoding::serialize::{FromHex, FromHexError, ToHex, encode_hex};
@@ -226,6 +227,18 @@ pub trait Api {
         self.optional_param(request, name)?.ok_or_else(|| {
             ApiError::BadRequest(format!("Required parameter '{}' is missing", name))
         })
+    }
+
+    /// Deserializes request's body as a struct of type `T`.
+    fn parse_body<T: 'static>(&self, req: &mut Request) -> Result<T, ApiError>
+    where
+        T: Clone + for<'de> Deserialize<'de>,
+    {
+        match req.get::<bodyparser::Struct<T>>() {
+            Ok(Some(param)) => Ok(param),
+            Ok(None) => Err(ApiError::BadRequest("Body is empty".into())),
+            Err(e) => Err(ApiError::BadRequest(format!("Invalid struct: {}", e))),
+        }
     }
 
     /// Loads hex value from the cookies.
--- a/exonum/src/api/private/system.rs
+++ b/exonum/src/api/private/system.rs
@@ -154,8 +154,13 @@ impl SystemApi {
 
     fn handle_peer_add(self, router: &mut Router) {
         let peer_add = move |request: &mut Request| -> IronResult<Response> {
-            let address: SocketAddr = self.required_param(request, "ip")?;
-            self.node_channel.peer_add(address).map_err(ApiError::from)?;
+            #[derive(Serialize, Deserialize, Clone, Debug)]
+            struct PeerAddInfo {
+                ip: SocketAddr,
+            }
+
+            let PeerAddInfo { ip } = self.parse_body(request)?;
+            self.node_channel.peer_add(ip).map_err(ApiError::from)?;
             self.ok_response(&serde_json::to_value("Ok").unwrap())
         };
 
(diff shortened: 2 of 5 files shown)


ozkriff:
Hmmm, still can't figure out how to fix these tests (see CI logs). :-\

exonum/src/api/mod.rs
```diff
@@ -198,34 +198,16 @@ pub trait Api {
         Ok(value)
     }
 
-    /// Deserializes an optional parameter from request body or get-parameters
-    fn optional_param<T>(&self, request: &mut Request, name: &str) -> Result<Option<T>, ApiError>
+    /// Deserializes request's body as a struct of type `T`.
+    fn parse<T: 'static>(&self, req: &mut Request) -> Result<T, ApiError>
```
slowli: Nitpick: `parse` is a rather non-descriptive name; I'd suggest something like `parse_body`.

exonum/src/api/mod.rs
```diff
@@ -198,34 +198,16 @@ pub trait Api {
...
-    fn required_param<T>(&self, request: &mut Request, name: &str) -> Result<T, ApiError>
-    where
-        T: FromStr,
-        T::Err: fmt::Display,
-    {
-        self.optional_param(request, name)?.ok_or_else(|| {
-            ApiError::BadRequest(format!("Required parameter '{}' is missing", name))
-        })
+        match req.get::<bodyparser::Struct<T>>() {
+            Ok(Some(param)) => Ok(param.clone()),
```
slowli: Are you sure that `clone()` is required here? I haven't tested it, but it seems that `req.get` returns a value.

exonum/src/api/private/system.rs
```diff
@@ -90,6 +90,16 @@ struct PeersInfo {
     outgoing_connections: HashMap<SocketAddr, IncomingConnection>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+struct EnabledInfo {
+    pub enabled: bool,
```
slowli: These fields don't need to be public, do they? Additionally, I'd move definitions into functions where they are used, but I don't insist on that.

exonum/src/api/private/system.rs
```diff
@@ -162,8 +172,8 @@ impl Api for SystemApi {
 
         let self_ = self.clone();
         let peer_add = move |req: &mut Request| -> IronResult<Response> {
-            let addr: SocketAddr = self_.required_param(req, "ip")?;
-            self_.node_channel.peer_add(addr).map_err(ApiError::from)?;
+            let info: PeerAddInfo = self_.parse(req)?;
```
slowli: It seems you can destructure here like `let PeerAddInfo { ip } = ...` Would look much more awesome, imo.

exonum/src/api/public/blockchain_explorer.rs
```diff
@@ -33,6 +33,13 @@ enum TransactionInfo {
     Committed(TxInfo),
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+struct ResponseInfo {
+    pub count: u64,
+    pub latest: Option<u64>,
+    pub skip_empty_blocks: Option<bool>,
```
slowli: I'd rather try to use [`#[serde(default = ..)]`](https://serde.rs/attr-default.html) here; using a `null` value for `skip_empty_blocks` in JSON would be rather strange.

services/configuration/src/api.rs
```diff
@@ -230,12 +231,7 @@ impl PublicApi {
     }
 
     fn retrieve_filter(&self, request: &mut Request) -> Result<Filter, ApiError> {
-        let previous_cfg_hash: Option<Hash> = self.optional_param(request, "previous_cfg_hash")?;
-        let actual_from: Option<Height> = self.optional_param(request, "actual_from")?;
-        Ok(Filter {
-            previous_cfg_hash,
-            actual_from,
-        })
+        self.parse(request).or(Ok(Filter::default()))
```
slowli: Are you sure this is right? As written now, his method *never* fails, including the case when the body is malformed or non-existent.

exonum/src/api/mod.rs
```diff
@@ -198,34 +198,16 @@ pub trait Api {
         Ok(value)
     }
 
-    /// Deserializes an optional parameter from request body or get-parameters
-    fn optional_param<T>(&self, request: &mut Request, name: &str) -> Result<Option<T>, ApiError>
```
slowli: Are you sure `optional_param` and `required_param` should be removed? They were public and (imo) quite useful.

vitvakatu: I thought that the intention behind this PR was to replace these methods with more convinient tool?

ozkriff:
Updated. Changelog entry is added.

PR #530, opened 2018-03-01
Add shutdown tests [ECR-850], [ECR-847]

alekseysidorov:
- Fix "can't cancel routine" error.
 - Add test for shutdown request.
 - Add test for a possible regression in "create_genesis_block" behaviour.

### DIFF
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -82,8 +82,8 @@ pub struct Blockchain {
 
 impl Blockchain {
     /// Constructs a blockchain for the given `storage` and list of `services`.
-    pub fn new(
-        storage: Box<Database>,
+    pub fn new<D: Into<Arc<Database>>>(
+        storage: D,
         services: Vec<Box<Service>>,
         service_public_key: PublicKey,
         service_secret_key: SecretKey,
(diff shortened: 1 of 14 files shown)


exonum/src/events/network.rs
```diff
@@ -311,11 +312,6 @@ impl RequestHandler {
                         let fut = cancel_sender
                             .take()
                             .ok_or_else(|| other_error("shutdown twice"))
```
vitvakatu: Is it real scenario? I mean, with this change, cancel_handler will return `Ok` in any situation, won't it?

alekseysidorov: It seems to be no way to do it twice.

exonum/src/blockchain/mod.rs
```diff
@@ -82,8 +82,8 @@ pub struct Blockchain {
 
 impl Blockchain {
     /// Constructs a blockchain for the given `storage` and list of `services`.
-    pub fn new(
-        storage: Box<Database>,
+    pub fn new<D: Into<Arc<Database>>>(
+        storage: D,
```
DarkEld3r: Technically, this is a breaking change, right?

alekseysidorov: I do not think so. In old code you can pass `Box::new(MemoryDb::new))` but now you need to convert `MemoryDb` into `Arc<Database>` manually or just pass `MemoryDb::new` I you do not need customization.

matklad:
Do we have a tests where a shutdown message is send twice? Should we have one?

PR #532, opened 2018-03-02
Make $transaction_set public by default [ECR-845]

alekseysidorov:
This PR fixes usage scenarions of transactions` set in separate modules.

### DIFF
--- a/exonum/src/blockchain/transaction.rs
+++ b/exonum/src/blockchain/transaction.rs
@@ -428,6 +428,7 @@ pub trait TransactionSet
 /// ```
 #[macro_export]
 macro_rules! transactions {
+    // Variant with the private enum.
     {
         $(#[$tx_set_attr:meta])*
         $transaction_set:ident {
@@ -444,11 +445,7 @@ macro_rules! transactions {
                 }
             )*
         }
-    }
-
-    =>
-
-    {
+    } => {
         messages! {
             const SERVICE_ID = $service_id;
             $(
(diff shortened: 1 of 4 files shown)


slowli:
I'd rather enable the user to make a concious choice by supplying default / `pub` and `pub(???)` versions of the macro, so it could be used like:

```rust
transactions! {
    // also `pub(super)`, `pub(in foo::bar)`, etc.
    pub(crate) Any {
      // ...
    }
}
```

See [here](https://github.com/exonum/exonum/blob/f37e944357d9c9b945e9bce3b77bb7fb8047b3a4/exonum/src/api/ext.rs#L472) for how it could be implemented (maybe, it's a hacky way, but it seems to work).

Also, it could make sense to use the visibility specifiers for transaction types themselves in order for te code to look less magical.

Also, this feature is easy to test, so I'd say it should be.

exonum/src/blockchain/transaction.rs
```diff
@@ -443,11 +445,46 @@ macro_rules! transactions {
...
+    };
+    // Variant with the public enum without restrictions.
     {
+        $(#[$tx_set_attr:meta])*
+        pub $transaction_set:ident {
+            const SERVICE_ID = $service_id:expr;
+
+            $(
+                $(#[$tx_attr:meta])*
+                struct $name:ident {
```
slowli: You can replace this with `struct $name:ident { $($def:tt)* }` for brevity.

PR #533, opened 2018-03-03
Release 0.6.0

DarkEld3r:


services/configuration/CHANGELOG.md
```diff
@@ -52,6 +54,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Implemented error handling based on error codes (#496).
 
+- Update to the [Exonum 0.6.0] release (#533).
```
slowli: I don't really think that this is a new feature; I'd rather characterize it as "internal improvement".

services/configuration/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "exonum-configuration"
-version = "0.5.0"
+version = "0.6.0"
```
slowli: I'm against releasing configuration until the new functionality (e.g., error handling) is covered with tests. If a release is necessary for some reason, I suggest to make it a release candidate: `0.6.0-rc.1`.

services/time/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "exonum-time"
-version = "0.5.0"
+version = "0.6.0"
```
slowli: Same as with configuration, but for a different reason, which I shouldn't disclose - we *know* there is an issue with this service in the current form.

services/configuration/CHANGELOG.md
vitvakatu: I think we should not mention that fact at all

alekseysidorov:
Relese?

services/configuration/CHANGELOG.md
DarkEld3r: It is controversial, but it can often be the only change in a new release, so I prefer to leave it. Plus it is (currently) always a breaking change, so I'm not sure that "internal improvements" section is a good place for that.

PR #535, opened 2018-03-04
Make blockchain explorer better

slowli:
This PR modifies API of the blockchain explorer in a way that simplifies its use in Rust environment (e.g., in the testkit, where it could be used to check status of transaction execution).

Ultimately, I see this PR as preparation for moving the explorer into a separate logical entity (e.g., a library + Exonum service), but I'd like to approach this task incrementally.


.travis.yml
```diff
@@ -132,6 +132,11 @@ jobs:
     - cd $TRAVIS_BUILD_DIR/examples/cryptocurrency/examples && ./test.sh
     - cd $TRAVIS_BUILD_DIR/examples/cryptocurrency/examples && ./test.ps1
 
+  # Additional tests
+  - env: FEATURE=test-aux
+    script:
+    - cd $TRAVIS_BUILD_DIR/exonum && cargo test --features doctests --doc explorer
```
slowli: This is needed for heavyweight blockchain initialization, which is shared among all explorer doctests. I would love to get rid of it, but don't know how.

exonum/src/explorer/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+#[cfg(any(test, feature = "doctests"))]
+#[doc(hidden)]
+pub mod tests;
+
+/// Transaction parsing result.
+type ParseResult = Result<Box<Transaction>, encoding::Error>;
+
+/// Range of `Height`s.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct HeightRange(pub Bound<Height>, pub Bound<Height>);
```
slowli: Could be accomplished with [`RangeArgument`](https://doc.rust-lang.org/std/collections/range/trait.RangeArgument.html), but it's unstable as of now.

exonum/src/explorer/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+///         "txs": *block.transaction_hashes(),
+///     })
+/// );
+/// # } // main
+/// # #[cfg(not(feature = "doctests"))] fn main() {}
+/// ```
+#[derive(Debug)]
+pub struct BlockInfo<'a> {
+    explorer: &'a BlockchainExplorer<'a>,
+    block: Block,
```
slowli: Imo, it would make sense to rename this field header, but it would break backward compatibility with JSON serialization. A possible compromise is renaming the field in the serialization code.

exonum/src/explorer/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+        self.block.tx_count() as usize
+    }
+
+    /// Is this block empty (i.e., contains no transactions)?
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Returns a list of precommits for this block.
+    pub fn precommits(&self) -> Ref<[Precommit]> {
```
slowli: Is this adequate complexity for the task? The reason I use it is to make `BlockInfo` load as little data from the blockchain from the get-go as possible in order to make it viable as an `Item` for `Blocks` iterator.

exonum/src/explorer/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+
+        BlockWithTransactions {
+            block,
+            precommits,
+            transactions,
+        }
+    }
+}
+
+impl<'a> Deref for BlockInfo<'a> {
```
slowli: Do `Deref` implementations for `BlockInfo` and `BlockWithTransactions` make sense?

exonum/src/explorer/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+    /// Exclusive range of blocks.
+    pub range: Range<Height>,
+    /// Blocks in the range.
+    pub blocks: Vec<Block>,
+}
+
+/// Blockchain explorer.
+pub struct BlockchainExplorer<'a> {
+    snapshot: Box<Snapshot>,
+    transaction_parser: Box<'a + Fn(RawMessage) -> ParseResult>,
```
slowli: The external lifetime is relatively easy to get rid of (by cloning the `Blockchain` in the constructor), but imo it serves a purpose: it makes it impossible to update the blockchain while there is an instance of explorer for it; i.e., it makes use of Rust borrow checker.

exonum/src/explorer/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+    fn fmt(&self, formatter: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        formatter
+            .debug_struct("Blocks")
+            .field("ptr", &self.ptr)
+            .field("back", &self.back)
+            .finish()
+    }
+}
+
+impl<'a> Iterator for Blocks<'a> {
```
slowli: Under-tested for now.

testkit/tests/counter/main.rs
```diff
@@ -627,11 +643,22 @@ fn test_explorer_transaction_statuses() {
...
     ]);
+    assert!(block[0].status().is_ok());
+    assert!({
+        let err = block[1].status().unwrap_err();
+        err.error_type() == ErrorType::Code(0) &&
+            err.description() == Some("Adding zero does nothing!")
+    });
+    assert!({
+        let err = block[2].status().unwrap_err();
+        err.error_type() == ErrorType::Panic &&
```
slowli: It could make sense to write some helpers, e.g., `err.is_panic()` and `err.has_code(..)`.

exonum/src/explorer/mod.rs
andreyluchin: I think it is a bad idea to use `Deref` trait to emulate polymorphism. To quote [the documentation](https://doc.rust-lang.org/std/ops/trait.Deref.html):

> Because of this, Deref should only be implemented for smart pointers to avoid confusion.

 It might be confusing for rust beginners and it generally feels like a hack. We are supposed to embrace Rust language design choices (like absence of inheritance) not to hack around it with `Deref`.

I think that a simple `block(&self)` getter is enough, maybe `block_mut(&mut self)` and `into_block(self)` too if necessary. This is how it is done all over the standard library and is more familiar to rust developers.

slowli: Got it; I had similar concerns initially. Will rework to use `block(&self)` and `into_block(self)` (or maybe `*header`; imo, it would be a better naming choice).

DarkEld3r:
@slowli 
Looks pretty good, but I'm heavily concerned with additional complexity (feature, additional build job, auxiliary code, etc.) added _only_ for the examples. Clear documentation is very important, but isn't it an overkill here?

testkit/tests/counter/main.rs
DarkEld3r: Perhaps, `assert_matches` can help?

exonum/src/explorer/tests.rs
```diff
@@ -0,0 +1,543 @@
...
+
+use blockchain::{ExecutionError, ExecutionResult, Service, TransactionSet};
+use crypto::{self, CryptoHash, PublicKey, SecretKey};
+use encoding::Error as EncodingError;
+use messages::{Message, RawTransaction, ServiceMessage};
+use node::ApiSender;
+use storage::{Fork, MemoryDB, Snapshot};
+use super::*;
+
+// // // Transactions // // //
```
DarkEld3r: Please don't use such comments. :upside_down_face: 
(@ozkriff ping)

CHANGELOG.md
```diff
@@ -60,6 +60,23 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+  In particular:
+
+  - The explorer now allows to iterate over blocks in the blockchain
+    in the given height range, replacing old `blocks_range` method
+  - `block_info` and `tx_info` methods of the explorer are renamed to
+    `block` and `transaction`, respectively
+  - `TransactionInfo` moved from the `api::public` module to the `explorer` module
+  - `BlocksRange` moved from the `explorer` module to the `api::public` module
+  - `TxInfo` is renamed to `CommittedTransaction`
+  - `BlockInfo` has its fields made private, instead providing getter methods
```
mvlabat: The last sentence sounds complicated, imo. What if we just say?
- `BlockInfo` fields are private now, yet accessible with getter methods

mvlabat: Btw, as our changelog list items are complete sentences we should use periods at the end (as we already did :) )

DarkEld3r: I agree with @mvlabat (especially with the comment about dots), but I want to merge this pull request already, so we will deal with this issue in a separate pull request. :pray:

PR #536, opened 2018-03-05
Add TransactionResult tests [ECR-833]

SerhiiOryshych:


services/time/tests/test_exonum_time.rs
```diff
@@ -353,7 +378,13 @@ fn test_creating_transaction_is_not_validator() {
 
     let (pub_key, sec_key) = gen_keypair();
     let tx = TxTime::new(SystemTime::now(), &pub_key, &sec_key);
-    testkit.create_block_with_transactions(txvec![tx]);
+    testkit.create_block_with_transactions(txvec![tx.clone()]);
+
+    assert_transaction_result(
+        testkit.core_schema().transaction_results().get(&tx.hash()),
+        0,
+        None,
```
ozkriff: I don't like these `None`s, but not sure if this is important for tests

services/time/tests/test_exonum_time.rs
```diff
@@ -24,19 +25,35 @@ fn assert_storage_times_eq<T: AsRef<Snapshot>>(
...
+    code: u8,
+    description: Option<&str>,
+) {
+    match result {
+        Some(Err(e)) => {
+            assert_eq!(e.error_type(), TransactionErrorType::Code(code));
+            assert_eq!(e.description(), description);
+        }
+        _ => {
+            assert!(false);
```
ozkriff: Maybe just use `expect` instead of this match?

testkit/src/lib.rs
```diff
@@ -584,6 +584,11 @@ impl TestKit {
         self.blockchain.snapshot()
     }
 
+    /// Returns the information schema for core.
+    pub fn core_schema(&self) -> CoreSchema<Box<Snapshot>> {
```
slowli: Nitpick: I don't think the addition of this method is strictly necessary; you could've used

```rust
fn assert_transaction_result<S: AsRef<Snapshot>, T: Transaction>(
    snapshot: S,
    transaction: &T,
    expected_code: u8,
)
```

services/time/tests/test_exonum_time.rs
```diff
@@ -24,19 +25,35 @@ fn assert_storage_times_eq<T: AsRef<Snapshot>>(
...
+            validators_times.get(public_key),
             expected_validators_times[i]
         );
     }
 }
 
+fn assert_transaction_result(
+    result: Option<TransactionResult>,
+    code: u8,
+    description: Option<&str>,
```
slowli: As far as I see, this is never used in tests. Imo, it could make more sense to *return* `Option<&str>` on success in order to be able to check it with various `str` methods.

services/time/tests/test_exonum_time.rs
```diff
@@ -385,7 +420,12 @@ fn test_transaction_time_less_than_validator_time_in_storage() {
     let time1 = time0 - Duration::new(10, 0);
     let tx1 = TxTime::new(time1, pub_key, sec_key);
 
-    testkit.create_block_with_transactions(txvec![tx1]);
+    testkit.create_block_with_transactions(txvec![tx1.clone()]);
+    assert_transaction_result(
+        testkit.core_schema().transaction_results().get(&tx1.hash()),
+        1,
```
slowli: Please use `Error::ValidatorTimeIsGreater as u8` for clarity.

services/time/tests/test_exonum_time.rs
```diff
@@ -24,19 +25,35 @@ fn assert_storage_times_eq<T: AsRef<Snapshot>>(
...
+    code: u8,
+    description: Option<&str>,
+) {
+    match result {
+        Some(Err(e)) => {
+            assert_eq!(e.error_type(), TransactionErrorType::Code(code));
+            assert_eq!(e.description(), description);
+        }
+        _ => {
+            panic!("Expected Err(), found Ok()");
```
slowli: Nitpick: This is also executed if `result` is `None`, so the panic description is not strictly correct.

services/time/tests/test_exonum_time.rs
```diff
@@ -24,19 +25,39 @@ fn assert_storage_times_eq<T: AsRef<Snapshot>>(
...
     }
 }
 
+fn assert_transaction_result<S: AsRef<Snapshot>, T: Transaction>(
+    snapshot: S,
+    transaction: &T,
+    expected_code: u8,
+) -> Option<String> {
+    let result = Schema::new(snapshot).transaction_results().get(
+        &transaction
```
vitvakatu: What a weird formatting? Though it passed CI

PR #537, opened 2018-03-05
Make TransactionInfo public

alekseysidorov:
This enum is useful for `testkit` and other purposes.

### DIFF
--- a/exonum/src/api/public/blockchain_explorer.rs
+++ b/exonum/src/api/public/blockchain_explorer.rs
@@ -31,7 +31,7 @@ const MAX_BLOCKS_PER_REQUEST: u64 = 1000;
 pub enum TransactionInfo {
     /// Transaction with the given id is absent in the blockchain.
     Unknown,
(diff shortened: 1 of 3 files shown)


exonum/src/api/public/blockchain_explorer.rs
```diff
@@ -25,11 +25,18 @@ use helpers::Height;
...
-#[derive(Serialize)]
+/// Information about the transaction.
+#[derive(Debug, Serialize)]
 #[serde(tag = "type", rename_all = "kebab-case")]
-enum TransactionInfo {
+pub enum TransactionInfo {
+    /// Transaction with the given id is absent in the blockchain.
     Unknown,
-    InPool { content: JsonValue },
+    /// Transaction in the memory pool but not yet committed to the blockchain.
```
vitvakatu: nitpick: Transaction **is** in the memory pool

DarkEld3r: Plus "...pool, but not...".

PR #538, opened 2018-03-05
Added DbOptions [ECR-839]

vitvakatu:
Abstract way to adjust settings of RocksDB via node config.

TODO:
- [x] Test max_open_files feature
- [x] Add changelog entry (technically, it's breaking change)

### DIFF
--- a/exonum/benches/storage.rs
+++ b/exonum/benches/storage.rs
@@ -27,7 +27,7 @@ mod tests {
     use rand::{Rng, thread_rng, XorShiftRng, SeedableRng};
     use tempdir::TempDir;
     use exonum::storage::{Database, MemoryDB};
-    use exonum::storage::{RocksDB, RocksDBOptions};
+    use exonum::storage::{RocksDB, DbOptions};
     use exonum::storage::{ProofMapIndex, ProofListIndex};
     use exonum::storage::proof_map_index::PROOF_MAP_KEY_SIZE as KEY_SIZE;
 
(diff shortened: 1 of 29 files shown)


exonum/src/storage/rocksdb.rs
```diff
@@ -42,6 +42,37 @@ pub struct RocksDB {
     db: Arc<rocksdb::DB>,
 }
 
+/// Options for `RocksDB`.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct RocksDBOptions {
```
DarkEld3r: I propose to name it `DBOptions` (`DbOptions`?).

exonum/benches/storage.rs
```diff
@@ -27,7 +27,7 @@ mod tests {
     use rand::{Rng, thread_rng, XorShiftRng, SeedableRng};
     use tempdir::TempDir;
     use exonum::storage::{Database, MemoryDB};
-    use exonum::storage::{RocksDB, RocksDBOptions};
+    use exonum::storage::{RocksDB, DbOptions};
```
DarkEld3r: Lets merge several `exonum::storage` imports into one.

exonum/src/storage/options.rs
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+
+//! Abstract settings for databases.
+
+/// Options for database.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct DbOptions {
+    /// Number of open files that can be used by the DB.
+    ///
+    /// Defaults to `-1`, which means opened files are always kept open.
```
DarkEld3r: I understand that this is an implementation detail of RocksDB, but I don't like such "magic constants". What do you think about using `Options<u32>` here (and converting `None` to the `-1` when needed)?

vitvakatu: Sounds reasonable. Don't know if we would be able to implement abstract API for any storage though.

PR #545, opened 2018-03-07
travis: Remove non-fatal-checks, move cargo-deadlinks to 'lints' job [ECR-822]

ozkriff:
There's no point in having `cargo-deadlinks` as a non-fatal check - everyone just ignores it.

This PR fixes some of the warnings.
Though, there're some bugs in `rustdoc`, so a few more hacks were added to suppress them for now.

```
https://docs.rs/exonum-configuration/0.6.0/exonum_configuration/struct.MaybeVote.html

Linked file at path /home/ozkriff/bitfury/exonum/target/std/string/struct.String.html does not exist!
Linked file at path /home/ozkriff/bitfury/exonum/target/std/primitive.usize.html does not exist!
Linked file at path /home/ozkriff/bitfury/exonum/target/doc/exonum_configuration/enum.Option.html does not exist!
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,6 @@ rust:
   - 1.23.0
 
 matrix:
-  allow_failures:
-  - env: FEATURE=non-fatal-checks
   fast_finish: true
 
 cache:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -22,8 +22,6 @@ rust:
   - 1.23.0
 
 matrix:
-  allow_failures:
-  - env: FEATURE=non-fatal-checks
```
DarkEld3r: Should the whole `matrix` section be removed along with `fast_finish`?

PR #552, opened 2018-03-10
Remove unused environment variables from the Travis script

DarkEld3r:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ env:
   - DEADLINKS_VERS=0.3.0
   - SODIUM_VERS=1.0.13
   - CARGO_INCREMENTAL=1
+  - RUSTFLAGS="-D warnings"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,11 +41,8 @@ env:
   - CLIPPY_VERS=0.0.187
   - RUSTFMT_VERS=0.9.0
   - DEADLINKS_VERS=0.3.0
-  - OUTDATED_VERS=0.7.0
   - SODIUM_VERS=1.0.13
   - CARGO_INCREMENTAL=1
-  - RUSTFLAGS="-C link-dead-code -D warnings"
```
vitvakatu: I remember that this line were added by @slowli in order to deny all warnings. Why we're removing it now?

DarkEld3r: Nice catch: I should only remove the `link-dead-code` flag that was needed for coverage.

PR #553, opened 2018-03-12
Reorganize cryprocurrency tests [ECR-782]

DarkEld3r:


examples/cryptocurrency/tests/constants/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Shared constants used both in api and transactions logic tests.
+
+/// Alice's wallets name.
+pub const ALICE_NAME: &str = "Alice";
+/// Bob's wallet name.
+pub const BOB_NAME: &str = "Bob";
```
ozkriff: Why `constants/mod.rs` instead of just `constants.rs`?

DarkEld3r: https://doc.rust-lang.org/book/second-edition/ch11-03-test-organization.html#submodules-in-integration-tests 🙃

ozkriff: Interesting official hack, didn't know about that.

examples/cryptocurrency/src/lib.rs
```diff
@@ -210,20 +251,29 @@ impl Transaction for TxTransfer {
     /// [`TxCreateWallet`]: struct.TxCreateWallet.html
     fn execute(&self, view: &mut Fork) -> ExecutionResult {
```
vitvakatu: Shouldn't we also check for `sender == receiver`?

DarkEld3r: I'm not sure: this seems logically wrong, but technically gives no problem, right? We probably don't want to add unnecessary details to the tutorial.

@slowli What do you think?

slowli: If I understand correctly, transactions with `sender == receiver` don't pass the `verify()` test, so during transaction execution, it is guaranteed that these two aren't equal.

examples/cryptocurrency/src/lib.rs
```diff
@@ -172,6 +174,43 @@ transactions! {
...
+    ///
+    /// Can be emitted by `TxCreateWallet`.
+    #[fail(display = "Wallet already exists")]
+    WalletAlreadyExists = 0,
+
+    /// Sender doesn't exist.
+    ///
+    /// Can be emitted by `TxTransfer`.
+    #[fail(display = "Sender doesn't exist")]
+    SenderNotExist = 1,
```
slowli: Maybe, use something like `AbsentSender` / `NonexistentSender` / `SenderNotFound`? `SenderNotExist` looks ungrammatic to me.

examples/cryptocurrency/src/lib.rs
```diff
@@ -210,20 +251,29 @@ impl Transaction for TxTransfer {
...
+
+        let receiver = match schema.wallet(self.to()) {
+            Some(val) => val,
+            None => Err(Error::ReceiverNotExist)?,
+        };
+
+        let amount = self.amount();
+        if sender.balance() >= amount {
+            let sender = sender.decrease(amount);
+            let receiver = receiver.increase(amount);
```
slowli: If I understand correctly, this instruction may lead to a panic (`u64` overflow, to be exact). I'm not quite sure whether this should be an error, or panicking is OK in this case.

examples/cryptocurrency/tests/constants/mod.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright 2017 The Exonum Team
```
slowli: Nitpick: 2018?

examples/cryptocurrency/src/lib.rs
DarkEld3r: Originally it was named `SenderDoesntExist", but the spell-checker wasn't happy.
Fixed.

examples/cryptocurrency/src/lib.rs
DarkEld3r: As for me, panic is OK here.

PR #557, opened 2018-03-13
Rewrite to chrono::DateTime [ECR-821]

vitvakatu:
-> What happen?
Rewriting `exonum-time` from `SystemTime` to `chrono::DateTime`. `SystemTime` can't be used in storages and messages anymore, but it's still in use in ~~dark and scary~~ some internals (no, we can't rewrite those places to `chrono`, because ~~I tried and it's complete nightmare~~ `tokio` and a lot of different things are tangled with `SystemTime` and `chrono` is not that usable in some scenarios). 

-> Drawbacks?
As a drawback, there're a lot of `into` calls in the code related to time operations, but personally I don't think it's huge problem.

-> For what? I loved `SystemTime`!
We've found some weird behavior in one of the tests under Windows. After short investigation we've released that `SystemTime` has different precision and (it's more important) different ranges on different platforms. Most likely this situation won't change, so we decided to use platform-independent solution (as we need to guarantee predictability across all platforms).

### DIFF
--- a/exonum/src/blockchain/tests.rs
+++ b/exonum/src/blockchain/tests.rs
@@ -16,6 +16,7 @@
 
 use rand::{thread_rng, Rng};
 use serde_json;
+use chrono::{DateTime, Utc};
 
 use blockchain::{Blockchain, Service, Snapshot, Schema, Transaction, ExecutionResult};
 use crypto::{gen_keypair, Hash, CryptoHash};
@@ -107,16 +108,16 @@ fn test_u64() {
 
 #[test]
 fn test_system_time() {
-    use std::time::{SystemTime, UNIX_EPOCH};
+    use std::time::UNIX_EPOCH;
     encoding_struct! {
         struct Test {
-            some_test: SystemTime,
+            some_test: DateTime<Utc>,
         }
     }
     let test_data = r##"{"some_test":{"nanos":0,"secs":"0"}}"##;
 
 
-    let test = Test::new(UNIX_EPOCH);
+    let test = Test::new(UNIX_EPOCH.into());
     let data = ::serde_json::to_string(&test).unwrap();
     assert_eq!(data, test_data);
 }
--- a/exonum/src/blockchain/tests.rs
+++ b/exonum/src/blockchain/tests.rs
@@ -16,7 +16,7 @@
 
 use rand::{thread_rng, Rng};
 use serde_json;
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Utc, TimeZone};
 
 use blockchain::{Blockchain, Service, Snapshot, Schema, Transaction, ExecutionResult};
 use crypto::{gen_keypair, Hash, CryptoHash};
@@ -102,23 +102,22 @@ fn test_u64() {
     }
     let test_data = r##"{"some_test":"1234"}"##;
     let test = Test::new(1234);
-    let data = ::serde_json::to_string(&test).unwrap();
+    let data = serde_json::to_string(&test).unwrap();
     assert_eq!(data, test_data);
 }
 
 #[test]
-fn test_system_time() {
-    use std::time::UNIX_EPOCH;
+fn test_date_time() {
     encoding_struct! {
         struct Test {
             some_test: DateTime<Utc>,
         }
     }
+
     let test_data = r##"{"some_test":{"nanos":0,"secs":"0"}}"##;
 
-
-    let test = Test::new(UNIX_EPOCH.into());
-    let data = ::serde_json::to_string(&test).unwrap();
+    let test = Test::new(Utc.timestamp(0, 0));
+    let data = serde_json::to_string(&test).unwrap();
     assert_eq!(data, test_data);
 }
 
--- a/exonum/src/encoding/fields.rs
+++ b/exonum/src/encoding/fields.rs
@@ -16,7 +16,7 @@
 
 use std::mem;
 use std::net::{SocketAddr, SocketAddrV4, Ipv4Addr};
-use std::time::{SystemTime, Duration, UNIX_EPOCH};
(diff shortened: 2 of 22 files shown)


vitvakatu:
CI passed, rebased onto master

exonum/src/blockchain/tests.rs
```diff
@@ -61,16 +62,16 @@ fn test_u64() {
 
 #[test]
 fn test_system_time() {
-    use std::time::{SystemTime, UNIX_EPOCH};
+    use std::time::UNIX_EPOCH;
     encoding_struct! {
         struct Test {
-            some_test: SystemTime,
+            some_test: DateTime<Utc>,
```
aleksuss: What do you think about to create a type aliasing like: `type DateTimeUtc = DateTime<Utc>` and use it everywhere in the code?

ozkriff: Hmmm, it doesn't seem worth it to me

vitvakatu: I'm not sure. Are there other opinions?

DarkEld3r: As for me, type alias makes sense if it is more concise (for example, `DateTime` instead of `DateTime<Utc>`) or it adds more contextual meaning (for example `TransactionTime`).

I'm happy with DateTime<Utc>  by the way.

exonum/src/blockchain/tests.rs
```diff
@@ -61,16 +62,16 @@ fn test_u64() {
 
 #[test]
 fn test_system_time() {
-    use std::time::{SystemTime, UNIX_EPOCH};
+    use std::time::UNIX_EPOCH;
```
DarkEld3r: It seems that the test name is no longer correct. :upside_down_face:

exonum/src/encoding/fields.rs
```diff
@@ -264,23 +264,21 @@ implement_pod_as_ref_field! {Signature}
...
-impl<'a> Field<'a> for SystemTime {
+impl<'a> Field<'a> for DateTime<Utc> {
     fn field_size() -> Offset {
-        (mem::size_of::<u64>() + mem::size_of::<u32>()) as Offset
+        (mem::size_of::<i64>() + mem::size_of::<u32>()) as Offset
     }
 
-    unsafe fn read(buffer: &'a [u8], from: Offset, to: Offset) -> SystemTime {
-        let secs = LittleEndian::read_u64(&buffer[from as usize..from as usize + 8]);
+    unsafe fn read(buffer: &'a [u8], from: Offset, to: Offset) -> Self {
```
DarkEld3r: :+1:

exonum/src/encoding/serialize/json.rs
```diff
@@ -192,27 +193,25 @@ impl<'a> ExonumJson for &'a str {
...
-        let duration = self.duration_since(UNIX_EPOCH)?;
-        let duration = DurationHelper {
-            secs: duration.as_secs().to_string(),
-            nanos: duration.subsec_nanos(),
+        let timestamp = TimestampHelper {
+            secs: self.timestamp().to_string(),
+            nanos: self.timestamp_subsec_nanos(),
         };
-        Ok(::serde_json::to_value(&duration)?)
+        Ok(::serde_json::to_value(&timestamp)?)
```
DarkEld3r: `use serde_json`? :upside_down_face:

exonum/src/node/consensus.rs
```diff
@@ -924,7 +924,7 @@ impl NodeHandler {
             round,
             propose_hash,
             block_hash,
-            self.system_state.current_time(),
+            self.system_state.current_time().into(),
```
DarkEld3r: Can we change the `SystemStateProvider` trait definition?

exonum/src/blockchain/tests.rs
```diff
@@ -61,16 +62,16 @@ fn test_u64() {
...
         struct Test {
-            some_test: SystemTime,
+            some_test: DateTime<Utc>,
         }
     }
     let test_data = r##"{"some_test":{"nanos":0,"secs":"0"}}"##;
 
 
-    let test = Test::new(UNIX_EPOCH);
+    let test = Test::new(UNIX_EPOCH.into());
```
DarkEld3r: Are there any advantages of using `UNIX_EPOCH` here instead of `Utc.timestamp(0, 0)`?

vitvakatu: Possibly provide user some example how to use `SystemTime` in messages? I dunno actually, will change.

exonum/src/blockchain/tests.rs
vitvakatu: I don't like `DateTime`, because if user use `chrono::DateTime` in his code, he will face errors like "DateTime is already imported into the scope". What he gonna do? Import it like `use chrono::DateTime as RealDateTimeNotProvidedWithExonumOne` :smile: ?
I'm also against `TransactionTime`, as it's completely wrong in terms of semantics.

DarkEld3r: `TransactionTime` was just an example of a "meaningful" name in some context. :upside_down_face:

exonum/src/storage/values.rs
```diff
@@ -404,24 +410,20 @@ mod tests {
 
     #[test]
     fn test_storage_value_for_system_time_round_trip() {
```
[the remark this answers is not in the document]
vitvakatu: Because it's impossible to create `DateTime` with those values.

exonum/src/node/consensus.rs
vitvakatu: Checked, no. Internals a tangled with `SystemTime` heavily, rewriting the whole thing will waste a lot of time without much profit.

PR #558, opened 2018-03-14
[Discussion]: use rustfmt provided by rustup

alekseysidorov:


.travis.yml
```diff
@@ -64,10 +63,11 @@ jobs:
   include:
   # Formatting & other lints that do not require compilation
   - env: FEATURE=lints
+    rust: 1.24.1
     install:
     - cargo-audit -V || cargo install cargo-audit --force
     - cargo-deadlinks -V | grep $DEADLINKS_VERS || cargo install cargo-deadlinks --vers $DEADLINKS_VERS --force
-    - rustfmt -V | grep $RUSTFMT_VERS || cargo install rustfmt --vers $RUSTFMT_VERS --force
+    - rustup component add rustfmt-preview
```
ozkriff: I think that we should at least print the version: `rustfmt --version`

DarkEld3r:
Generally, I like the new formatting, but I think that we should wait for a more mature release.

Additionally, the versioning must be handled properly: we should use concrete `rustfmt` version (if it is possible with `rustup`). Otherwise the formatting (and therefore the build) can be broken with the new release.

alekseysidorov:
`rustfmt` version is fixed by the given rust version. 
So we can just install it by the following commands without manual compilation:
```
rustup default 1.24.0
rustup component add rustfmt-preview
```

DarkEld3r:
I propose to wait (for stabilization) a little bit more. Additionally, it would be easier to apply formatting one more time than updating this pull request. :upside_down_face:

PR #560, opened 2018-03-15
Fixed multiple timeout warnings bug in configuration service [ECR-921]

andreyluchin:
Exonum has to warn users when 2 * propose timeout is less than round timeout. The warning happened during config deserialization which lead to warning being displayed more than necessary. The warning was changed to occur during the "post propose" api handling and on genesis configuration block creation.

### DIFF
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -81,7 +81,7 @@ impl ConsensusConfig {
         let propose_timeout = match self.timeout_adjuster {
             TimeoutAdjusterConfig::Constant { timeout } => timeout,
             TimeoutAdjusterConfig::Dynamic { max, .. } |
-            TimeoutAdjusterConfig::MovingAverage { max, .. } => max
+            TimeoutAdjusterConfig::MovingAverage { max, .. } => max,
         };
 
         if self.round_timeout <= 2 * propose_timeout {
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -74,6 +74,28 @@ pub struct ConsensusConfig {
 impl ConsensusConfig {
     /// Default value for max_message_len.
     pub const DEFAULT_MAX_MESSAGE_LEN: u32 = 1024 * 1024; // 1 MB
+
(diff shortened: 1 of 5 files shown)


exonum/src/blockchain/config.rs
```diff
@@ -74,6 +74,28 @@ pub struct ConsensusConfig {
 impl ConsensusConfig {
     /// Default value for max_message_len.
     pub const DEFAULT_MAX_MESSAGE_LEN: u32 = 1024 * 1024; // 1 MB
+
+    /// Checks if propose timeout is less than round timeout. Warns if fails.
+    pub fn check_timeout_recommendations(&self) {
```
vitvakatu: Shouldn't it be private function? Are there any usage scenarios when it's useful for users?

exonum/src/blockchain/config.rs
```diff
@@ -74,6 +74,28 @@ pub struct ConsensusConfig {
...
+                max
+            }
+            TimeoutAdjusterConfig::MovingAverage { max, .. } => {
+                max
+            }
+        };
+
+        if self.round_timeout <= 2 * propose_timeout {
+            warn!(
+                "It is recommended that round_timeout({}) be at least twice as large \
```
vitvakatu: `be` -> `is` or `should be`

Also I think you should explain numbers in parens, like "round_timeout (current: {})" and "propose_timeout (current: {})"

exonum/src/blockchain/config.rs
DarkEld3r: It is used in the `exonum-configuration` service. 🙃 

I propose to mark this  function as `doc(hidden)`. Additionally, I'm thinking of making this function (name) a little bit more generic. Probably, `validate_configuration` name would be better.

exonum/src/blockchain/config.rs
andreyluchin: Thanks for pointing that out. The problem with adding "current: " is that timeouts in the message are not necessarily current. There are 2 possible scenarios when the warning is called: 
1. Invalid timeouts are in genesis block (in this case they are indeed current);
2. Invalid timeouts are specified in a new propose (in this case they are not yet current (until propose is accepted by supermajority)).

PR #562, opened 2018-03-15
WIP: Fix `CryptoHash` for `Hash` [ECR-698]

SerhiiOryshych:
An attempt to fix the calculation of `CryptoHash` for `Hash`:

```rust
impl CryptoHash for Hash {
	fn hash(&self) -> Hash {
		hash(self.as_ref())
	}
}
```

Now wrapper is implemented over `Hash`:

```rust
Struct EntryHash(Hash);

impl CryptoHash for EntryHash {
	fn hash(&self) -> Hash {
		self.0
	}
}
```

`Hash` is changed to `EntryHash` in `ProofListIndex` and `ProofMapIndex`. But some tests fall in the `sandbox` (`sandbox::tests::test_sandbox_service_handle_commit` is one of them).
I need your help. Maybe I'm doing something wrong.
What can you advise to solve the problem?


exonum/src/blockchain/schema.rs
```diff
@@ -87,7 +87,7 @@ where
     }
 
     /// Returns table that represents a map from transaction hash into execution result.
-    pub fn transaction_results(&self) -> ProofMapIndex<&T, Hash, TransactionResult> {
+    pub fn transaction_results(&self) -> ProofMapIndex<&T, EntryHash, TransactionResult> {
```
slowli: What does the `Hash` vs `EntryHash` key type influence?

exonum/src/blockchain/schema.rs
```diff
@@ -277,12 +277,12 @@ where
 
     /// Returns configuration for given configuration hash.
     pub fn configuration_by_hash(&self, hash: &Hash) -> Option<StoredConfiguration> {
-        self.configs().get(hash)
+        self.configs().get(&EntryHash(*hash))
```
slowli: This is quite awkward. Could it make sense to use a method signature involving `Borrow` as in [standard collections](https://doc.rust-lang.org/1.21.0/std/collections/struct.BTreeMap.html#method.get)?

exonum/src/blockchain/schema.rs
```diff
@@ -277,12 +277,12 @@ where
...
     }
 
     /// Returns the `state_hash` table for core tables.
     pub fn core_state_hash(&self) -> Vec<Hash> {
-        vec![self.configs().merkle_root(), self.transaction_results().merkle_root()]
+        vec![self.configs().merkle_root().hash(), self.transaction_results().merkle_root().hash()]
```
slowli: I don't understand why `merkle_root` returns `EntryHash`, sorry.

exonum/src/storage/proof_list_index/proof.rs
```diff
@@ -17,7 +17,7 @@ use serde::ser::SerializeStruct;
 use serde::de::Error;
 use serde_json::{Error as SerdeJsonError, Value, from_value};
 
-use crypto::{Hash, hash};
+use crypto::{hash, CryptoHash, EntryHash};
```
slowli: I seriously doubt `EntryHash` should be placed in the `crypto` module; it concerns only storage implementation details.

DarkEld3r:
I agree with @slowli: this change has a huge impact on our code base without obvious advantages.

PR #567, opened 2018-03-19
Refactor Blockchain::tx_from_raw method to return Result instead of Option. [ECR-949]

pavel-mukhanov:
Refactor Blockchain::tx_from_raw method to return Result instead of Option.

`
pub fn tx_from_raw(&self, raw: RawMessage) -> Result<Box<Transaction>, MessageError> {
...
}
`

Motivation: for more specific error messages.

Also I found a small typo in **testnet-api-tutorial.md** and fixed it.


exonum/src/api/public/blockchain_explorer.rs
```diff
@@ -141,9 +142,13 @@ impl ExplorerApi {
             "Expected tx in database",
         );
 
-        Ok(self.blockchain.tx_from_raw(raw_tx.clone()).ok_or_else(|| {
-            ApiError::InternalError(format!("Service not found for tx: {:?}", raw_tx).into())
-        })?)
+        Ok(self.blockchain.tx_from_raw(raw_tx.clone()).or_else(
+            |error| {
+                Err(ApiError::InternalError(
+                    format!("{}, tx: {:?}", error.description(), raw_tx).into(),
```
vitvakatu: Hmmm... I don't think we should print `raw_tx`. `Debug` is auto-derived and it will be something like `RawMessage(sync::Arc<MessageBuffer { raw: [1, 123, 123, 321, 324, ....] }>)`

exonum/src/blockchain/mod.rs
```diff
@@ -141,11 +142,14 @@ impl Blockchain {
     ///
     /// - Blockchain has service with the `service_id` of given raw message.
     /// - Service can deserialize given raw message.
-    pub fn tx_from_raw(&self, raw: RawMessage) -> Option<Box<Transaction>> {
+    pub fn tx_from_raw(&self, raw: RawMessage) -> Result<Box<Transaction>, MessageError> {
         let id = raw.service_id() as usize;
-        self.service_map.get(id).and_then(|service| {
-            service.tx_from_raw(raw).ok()
-        })
+        self.service_map
```
slowli: Nitpick / debatable: imo, the following structure would be easier to digest:

```rust
let service = self.service_map
    .get(id)
    .unwrap_or_else(|| { Err(..) })?;
service.tx_from_raw(raw)
```

exonum/src/blockchain/mod.rs
```diff
@@ -352,8 +356,8 @@ impl Blockchain {
                 failure::err_msg("BUG: Cannot find transaction in database.")
             })?;
 
-            self.tx_from_raw(tx).ok_or_else(|| {
-                failure::err_msg("BUG: couldn't create tx from raw message")
+            self.tx_from_raw(tx).or_else(|_| {
```
slowli: Do we want to swallow the message description here?

exonum/src/node/consensus.rs
```diff
@@ -159,26 +160,32 @@ impl NodeHandler {
...
-                        error!(
-                            "Received block with already known transaction, block={:?}",
-                            block
-                        );
-                        return None;
-                    }
-                    profiler_span!("tx.verify()", {
-                        if !tx.verify() {
-                            error!("Incorrect transaction in block detected, block={:?}", block);
+                match self.blockchain.tx_from_raw(raw) {
```
slowli: Nitpick / debatable: imo, the following structure would look nicer (less indentation):

```rust
let tx = match self.blockchain.tx_from_raw(raw) {
    Ok(tx) => tx,
    Err(e) => {
        error!("{}, block={:?}", e.description(), block);
        return None;
    }
};

let hash = tx.hash();
// ...
```

exonum/src/api/public/blockchain_explorer.rs
```diff
@@ -141,9 +142,13 @@ impl ExplorerApi {
             "Expected tx in database",
         );
 
-        Ok(self.blockchain.tx_from_raw(raw_tx.clone()).ok_or_else(|| {
-            ApiError::InternalError(format!("Service not found for tx: {:?}", raw_tx).into())
-        })?)
+        Ok(self.blockchain.tx_from_raw(raw_tx.clone()).or_else(
```
slowli: Side note: I find it mildly worrying that there are multiple places with `tx_from_raw(raw_tx.clone())` pattern. Maybe, it would make more sense to pass a raw message by reference?

Also, imo this would be more readable without `Ok(..)` wrapping:

```rust
self.blockchain
    .tx_from_raw(raw_tx.clone())
    .map_err(|e| ...)
```

exonum/src/blockchain/mod.rs
```diff
@@ -141,11 +142,14 @@ impl Blockchain {
...
+    pub fn tx_from_raw(&self, raw: RawMessage) -> Result<Box<Transaction>, MessageError> {
         let id = raw.service_id() as usize;
-        self.service_map.get(id).and_then(|service| {
-            service.tx_from_raw(raw).ok()
-        })
+        self.service_map
+            .get(id)
+            .map(|service| service.tx_from_raw(raw))
+            .unwrap_or_else(|| {
+                Err(MessageError::IncorrectServiceId { service_id: id as u16 })
```
slowli: Technically speaking, this is incorrect use of `IncorrectServiceId`; its field denotes *expected* `service_id` and used as such in parsers inferred by the `messages!` / `transactions!` macro.

PR #568, opened 2018-03-19
Move sandbox into exonum core [ECR-302]

DarkEld3r:


exonum/src/lib.rs
```diff
@@ -63,12 +63,16 @@ extern crate tokio_retry;
 extern crate os_info;
 #[macro_use]
 extern crate failure;
+extern crate bodyparser;
+
+// Test dependencies.
+#[cfg(all(test, feature = "long_benchmarks"))]
+extern crate test;
```
vldm: can you move this extern crate inside "benches" ?

PR #569, opened 2018-03-20
WIP Checkpoint remake [ECR-975]

andreyluchin:
#### Before:
We used to rollback our CheckpointDb by merges count.
```rust
handler.merge();
...
handler.rollback(1);
```
Same goes for Testkit, where we rollback by block count.
```rust
testkit.create_block(...);
teskit.rollback(1);
```
#### After this pull request:
CheckpointDb and Testkit now require you to make a checkpoint to rollback to:
```rust
handler.checkpoint();
// Changes
handler.rollback();
```
Testkit inherits this new interface as well:
```rust
testkit.checkpoint();
// Changes
testkit.rollback();
```

#### Important points:
* there can only be one checkpoint at a time (this is enforced by **panicking**);
(this is not a problem for the existing code base, however may be necessary in the future)
* fuzzing tests for old `rollback(count)` and `rollback(blocks)` were removed because fuzzing happened on the argument (count or blocks) which has been removed in this pull request;

### DIFF
--- a/testkit/src/checkpoint_db.rs
+++ b/testkit/src/checkpoint_db.rs
@@ -78,7 +78,7 @@ impl<T: Database> CheckpointDbHandler<T> {
     ///
     /// # Panics
     ///
-    /// - Panics if another checkpoint was set before and have not been rollbacked to.
+    /// - Panics if another checkpoint was set before and have not been rolled back to.
     pub fn checkpoint(&self) {
         self.inner
             .write()
@@ -127,13 +127,13 @@ impl<T: Database> CheckpointDbInner<T> {
 
     fn merge(&mut self, patch: Patch) -> StorageResult<()> {
         if self.checkpoint_set {
-            self.merge_with_journaling(patch)
+            self.merge_with_journal_logging(patch)
         } else {
             self.db.merge(patch)
         }
     }
 
-    fn merge_with_journaling(&mut self, patch: Patch) -> StorageResult<()> {
+    fn merge_with_journal_logging(&mut self, patch: Patch) -> StorageResult<()> {
         // NB: make sure that **both** the db and the journal
         // are updated atomically.
         let snapshot = self.db.snapshot();
--- a/testkit/src/checkpoint_db.rs
+++ b/testkit/src/checkpoint_db.rs
@@ -74,11 +74,10 @@ pub struct CheckpointDbHandler<T> {
 }
 
 impl<T: Database> CheckpointDbHandler<T> {
-    /// Set a checkpoint for future rollback
+    /// Set a checkpoint for future rollback.
     ///
-    /// # Panics
-    ///
-    /// - Panics if another checkpoint was set before and have not been rolled back to.
+    /// Checkpoints and rollbacks are stack-based:
+    /// `rollback()` rolls back to the last defined `checkpoint()`.
     pub fn checkpoint(&self) {
         self.inner
             .write()
@@ -86,7 +85,7 @@ impl<T: Database> CheckpointDbHandler<T> {
             .checkpoint();
     }
 
-    /// Rolls back this database by undoing the latest `count` `merge()` operations.
+    /// Rolls back this database to the latest checkpoint.
     ///
     /// # Panics
     ///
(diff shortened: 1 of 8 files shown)


slowli:
I don't quite like the proposed approach (although I can live with it). In my opinion, the primary attractiveness for rollbacks is for client-side testing (aka testkit server; see #459). See [this JS script](https://github.com/slowli/exonum/blob/044b677a7da4543d00063f6467487c0993152dac/testkit/examples/server/index.js) for an example how I envision its usage; before each test, the blockchain [is commanded to be rolled back](https://github.com/slowli/exonum/blob/044b677a7da4543d00063f6467487c0993152dac/testkit/examples/server/index.js#L154) (in order to ensure reproducibility). Correspondingly, it could be useful to have an ability to roll the blockchain to a specific height - it is stateless and imo easier to manage than the concept of a single checkpoint.

testkit/src/checkpoint_db.rs
```diff
@@ -73,31 +74,50 @@ pub struct CheckpointDbHandler<T> {
...
     }
 }
 
 #[derive(Debug)]
+// Journal is used here to track patches that need to be applied (in order
+// that they appear) to db for it to become what it was on checkpoint.
+//
+// Insertion should happen to the front of the journal,
+// application of patches in the journal should happen
+// sequentially in order that they are in the VecDeque.
```
vitvakatu: `Insertion should occur at the front of the journal, while patches should apply sequentially in the order in which they are in the VecDeque.`

I'm not native, but I think it'll be better. Contact me if our opinions mismatch.

testkit/src/checkpoint_db.rs
```diff
@@ -73,31 +74,50 @@ pub struct CheckpointDbHandler<T> {
...
             .write()
             .expect("Cannot lock CheckpointDb for rollback")
-            .rollback(count);
+            .rollback();
     }
 }
 
 #[derive(Debug)]
+// Journal is used here to track patches that need to be applied (in order
+// that they appear) to db for it to become what it was on checkpoint.
```
vitvakatu: `Journal is used to track patches that need to be applied (in the order of their appearance) to the database in order to restore its state to a checkpoint.`

testkit/src/checkpoint_db.rs
```diff
@@ -106,6 +126,14 @@ impl<T: Database> CheckpointDbInner<T> {
...
 
     fn merge(&mut self, patch: Patch) -> StorageResult<()> {
+        if self.checkpoint_set {
+            self.merge_with_journal_logging(patch)
+        } else {
+            self.db.merge(patch)
+        }
+    }
+
+    fn merge_with_journal_logging(&mut self, patch: Patch) -> StorageResult<()> {
```
vitvakatu: I think `merge_with_logging` is better. `Journal` and `log` is the same thing, in fact.

testkit/src/checkpoint_db.rs
```diff
@@ -125,22 +153,24 @@ impl<T: Database> CheckpointDbInner<T> {
...
 
-    fn rollback(&mut self, count: usize) {
+    fn checkpoint(&mut self) {
         assert!(
-            self.journal.len() >= count,
-            "Cannot rollback {} changes; only {} checkpoints in the journal",
-            count,
-            self.journal.len()
+            !self.checkpoint_set,
+            "Checkpoint is already set. There can only be one checkpoint at a time."
```
vitvakatu: `Checkpoint has already been set`

testkit/src/checkpoint_db.rs
```diff
@@ -125,22 +153,24 @@ impl<T: Database> CheckpointDbInner<T> {
...
         );
+        self.checkpoint_set = true;
+    }
 
-        for _ in 0..count {
-            let patch = self.journal.pop().unwrap();
+    fn rollback(&mut self) {
+        assert!(self.checkpoint_set, "Checkpoint was not set yet.");
```
vitvakatu: `Checkpoint has not been set yet`

testkit/src/lib.rs
```diff
@@ -388,7 +388,13 @@ impl TestKit {
...
     }
 
-    /// Rolls the blockchain back for a certain number of blocks.
+    /// Sets a checkpoint for future [`rollback`](#method.rollback).
+    /// There can be only one checkpoint at a time.
+    pub fn checkpoint(&mut self) {
+        self.db_handler.checkpoint();
+    }
+
+    /// Rolls the blockchain back to a [`checkpoint`](#method.checkpoint) set before.
```
vitvakatu: `to _the_ checkpoint`

testkit/src/checkpoint_db.rs
andreyluchin: I'd prefer `in order that they appear` or maybe `in order in which they appear`, but pretty much agree with everything else. Thanks for helping with my terrible English grammar!

vitvakatu: Okay, but please add `the` preposition before `order`. `In order` and `in the order` have different meanings.

testkit/src/checkpoint_db.rs
```diff
@@ -73,31 +74,51 @@ pub struct CheckpointDbHandler<T> {
 }
 
 impl<T: Database> CheckpointDbHandler<T> {
+    /// Set a checkpoint for future rollback
```
DarkEld3r: Set**s** a checkpoint for future rollback.

testkit/src/checkpoint_db.rs
```diff
@@ -73,31 +74,51 @@ pub struct CheckpointDbHandler<T> {
...
+    ///
+    /// # Panics
+    ///
+    /// - Panics if another checkpoint was set before and have not been rolled back to.
```
DarkEld3r: ...has not been...

testkit/src/checkpoint_db.rs
```diff
@@ -73,31 +74,51 @@ pub struct CheckpointDbHandler<T> {
...
             .expect("Cannot lock CheckpointDb for rollback")
-            .rollback(count);
+            .rollback();
     }
 }
 
 #[derive(Debug)]
+// Journal is used to track patches that need to be applied
+// (in the order that they appear) to the database in order
+// to restore its state to a checkpoint.
```
DarkEld3r: I prefer to put comments before attributes. :upside_down_face:

PR #570, opened 2018-03-20
Refactor testkit get_err method to return ApiError [ECR-990]

mvlabat:
Testkit users may want to test against response status codes. In order to make it implementation agnostic, I decided to use `ApiError` as return value for `get_err` method.

Also one more dev dependency for `testkit` has been introduced: https://github.com/SimonSapin/rust-std-candidates/tree/master/matches

It helps a lot with checking destructured `ApiError`s without pattern matching boilerplate.

TODO: write more unit tests for covering `Into<TestkitApiError> for Response` implementation.


testkit/src/api.rs
```diff
@@ -41,6 +43,37 @@ pub enum ApiKind {
     Service(&'static str),
 }
 
+#[derive(Debug)]
+struct TestkitApiError(ApiError);
```
slowli: Unless I'm misunderstanding something, this wrapper is superfluous; you could use a custom function (instead of an `Into` trait) in order to convert `Response` to `ApiError`.

testkit/src/api.rs
```diff
@@ -15,12 +15,14 @@
 use exonum::blockchain::{SharedNodeState, Transaction};
 use exonum::node::{ApiSender, TransactionSend, create_public_api_handler,
                    create_private_api_handler};
-use iron::{IronError, Handler, Chain};
+use exonum::api::ApiError;
+use iron::{Response, IronError, Handler, Chain};
 use iron::headers::{ContentType, Headers};
-use iron::status::StatusClass;
```
slowli: Nitpick: Why not `use iron::status::{self, StatusClass}`?

testkit/tests/counter/main.rs
```diff
@@ -545,11 +547,12 @@ fn test_explorer_transaction() {
...
         TxIncrement::new(&pubkey, 5, &key)
     };
-    let info: Value = api.get_err(
+    let info: ApiError = api.get_err(
         ApiKind::Explorer,
         &format!("v1/transactions/{}", &tx.hash().to_string()),
     );
-    assert_eq!(info, json!({ "type": "unknown" }));
+    let error_body = json!({ "type": "unknown" }).to_string();
+    assert!(matches!(info, ApiError::NotFound(ref body) if *body == error_body));
```
slowli: Nitpick: Seems that `assert_matches!` fits better.

testkit/src/api.rs
```diff
@@ -41,6 +43,37 @@ pub enum ApiKind {
     Service(&'static str),
 }
 
+#[derive(Debug)]
+struct TestkitApiError(ApiError);
+
+impl Into<TestkitApiError> for Response {
```
vitvakatu: You should not implement `Into` where `From` is available. I can't see the reason why you can't use `From` here, but maybe I'm mistaken?

vitvakatu: Oh, okay, I see the reason now, so nevermind

testkit/src/api.rs
```diff
@@ -192,16 +225,14 @@ impl TestKitApi {
...
-        TestKitApi::get_internal(
-            &self.public_handler,
-            &format!("{}/{}", kind.into_prefix(), endpoint),
-            true,
-            true,
-        )
+    pub fn get_err(&self, kind: ApiKind, endpoint: &str) -> ApiError {
+        let url = format!("http://localhost:3000/{}/{}", kind.into_prefix(), endpoint);
+        let response = match request::get(&url, Headers::new(), &self.public_handler) {
+            Ok(response) |
```
slowli: I *think* I've written a line comment here, but somehow it got lost :thinking: It seems that the new version of `get_err` will actually attempt to create errors from non-erroneous (i.e., 20x) responses (which would differ from the current behavior: panicking if the response code is not erroneous; I'd argue that this is reasonable behavior). Could you add a test to fix that this is not the case?

mvlabat: @slowli, so, have I got you right? We should panic, if there's not erroneous code - I should preserve this behavior and write tests for it.

slowli: Yeah, that describes my intention.

exonum/src/api/mod.rs
```diff
@@ -65,6 +65,10 @@ pub enum ApiError {
     #[fail(display = "Bad request: {}", _0)]
     BadRequest(String),
 
+    /// Not found.
```
alekseysidorov: What is not found?

mvlabat: ? :)
https://en.wikipedia.org/wiki/HTTP_404

mvlabat:
@slowli, oh, sorry, have noticed your comment regarding IronError just now. I think, it'll complicate things, as for now we may send error status codes via `Ok` result (for example 404). So treating `IronError` only won't cover all the cases.
I have no idea why we do this.. maybe we should fix it, but it goes out of the task scope

testkit/tests/counter/main.rs
```diff
@@ -545,11 +548,12 @@ fn test_explorer_transaction() {
...
         TxIncrement::new(&pubkey, 5, &key)
     };
-    let info: Value = api.get_err(
+    let info: ApiError = api.get_err(
         ApiKind::Explorer,
         &format!("v1/transactions/{}", &tx.hash().to_string()),
     );
-    assert_eq!(info, json!({ "type": "unknown" }));
+    let error_body = json!({ "type": "unknown" }).to_string();
+    assert_matches!(info, ApiError::NotFound(ref body) if *body == error_body);
```
slowli: This code here imo signifies that the chosen approach is not perfect. The comparison `*body == error_body` is logically incorrect; a correct one would be

```rust
ApiError::NotFound(ref body) if serde_json::to_value(body).unwrap() ==
    json!({ "type": "unknown" })
```

Maybe, make `NotFound` variant in `ApiError` wrap `serde_json::Value`?

testkit/src/api.rs
```diff
@@ -280,4 +279,59 @@ impl TestKitApi {
...
     }
+
+    /// Converts iron Response to ApiError.
+    ///
+    /// # Panics
+    ///
+    /// - Panics if the response has a non-error response status.
+    fn response_to_api_error(response: Response) -> ApiError {
+        let status = response.status.expect("Status header is not set");
+        let status_class = status.class();
```
slowli: Would something like this make sense?

```rust
let error_converter: fn(String) -> _ = match status {
    status::Forbidden => |_| ApiError::Unauthorized,
    status::BadRequest => ApiError::BadRequest,
    status::NotFound => ApiError::NotFound,
    s if s.is_server_error() => |err| ApiError::InternalError(err.into()),
    s => panic!("Received non-error response status: {}", s.to_u16()),
};

let error = ...;
error_converter(error)
```

testkit/src/api.rs
```diff
@@ -280,4 +279,58 @@ impl TestKitApi {
...
+    ///
+    /// # Panics
+    ///
+    /// - Panics if the response has a non-error response status.
+    fn response_to_api_error(response: Response) -> ApiError {
+        let status = response.status.expect("Status header is not set");
+
+        fn error(response: Response) -> String {
+            let body = response::extract_body_to_string(response);
+            serde_json::from_str::<JsonValue>(&body)
```
slowli: I know this is probably my extra exposure to JS, but could it make sense to use [IIFE](https://en.wikipedia.org/wiki/Immediately-invoked_function_expression) here? (Or, to be less extravagant, another function.)

```rust
let body = response::extract_body_to_string(response);
(|| match serde_json::from_str::<JsonValue>(&body).ok()? {
    JsonValue::Object(ref object) if object.contains_key("description") => {
        Some(object["description"].as_str()?.to_owned())
    }
    JsonValue::String(string) => Some(string),
    _ => None,
})().unwrap_or(body)
```

Also note that I got rid of 2-way conversion `body -> value -> _: String` in the final match hand and changed the `_or_default()` logic - I don't see a clear case for losing the response body in some cases.

testkit/src/api.rs
```diff
@@ -280,4 +279,58 @@ impl TestKitApi {
...
+                .map(|value| match value {
+                    JsonValue::Object(object) => {
+                        object
+                            .get("description")
+                            .map(|value| value.as_str().map(String::from))
+                            .unwrap_or_else(|| serde_json::to_string(&object).ok())
+                            .unwrap_or_default()
+                    }
+                    JsonValue::String(string) => string,
```
slowli: Are there tests that test various parsing possibilities:

- [ ] Object with `description` field
- [x] Object without `description` field
- [ ] JSON string
- [ ] Valid JSON, but not a string or an object
- [x] non-JSON input

I've marked those that seem to be implemented right now (maybe, I've missed some).

testkit/src/api.rs
mvlabat: Looks pretty laconically indeed! I like this approach, as well as IIFE )

mvlabat: But `clippy` doesn't :(
(speaking of IIFE)

testkit/src/api.rs
```diff
@@ -280,4 +279,58 @@ impl TestKitApi {
...
         )
     }
+
+    /// Converts iron Response to ApiError.
+    ///
+    /// # Panics
+    ///
+    /// - Panics if the response has a non-error response status.
+    fn response_to_api_error(response: Response) -> ApiError {
+        let status = response.status.expect("Status header is not set");
```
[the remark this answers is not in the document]
mvlabat: Just tried to do this way, with such level of nesting it looks ugly...

testkit/src/api.rs
```diff
@@ -280,4 +279,97 @@ impl TestKitApi {
...
+
+        fn error(response: Response) -> String {
+            let body = response::extract_body_to_string(response);
+
+            fn match_error(body: &str) -> Option<String> {
```
slowli: Nitpick: I think this function could be moved one level up (i.e., at the same level as `error`) and named something like `description` or `extract_description`.

PR #572, opened 2018-03-21
Property testing for `ProofMapIndex`

slowli:
This PR includes [property testing](https://crates.io/crates/proptest) for proofs generated by `ProofMapIndex`.

### DIFF
--- a/exonum/tests/proof_map_index.rs
+++ b/exonum/tests/proof_map_index.rs
@@ -1,0 +1,266 @@
+// Copyright 2018 The Exonum Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


exonum/tests/proof_map_index.rs
```diff
@@ -0,0 +1,290 @@
...
+            ) {
+                let table: ProofMapIndex<_, [u8; 32], u64> =
+                    ProofMapIndex::new(INDEX_NAME, db.snapshot());
+                let proof = table.get_proof(key);
+                check_map_proof(proof, Some(key), &table);
+            }
+
+
+            #[test]
+            #[cfg_attr(feature = "cargo-clippy", allow(useless_format))]
```
DarkEld3r: It would be nice to get to the root of the problem and submit a bug report. 🙃

exonum/tests/proof_map_index.rs
```diff
@@ -0,0 +1,290 @@
...
+
+const INDEX_NAME: &str = "index";
+
+fn check_map_proof<T, K, V>(proof: MapProof<K, V>, key: Option<K>, table: &ProofMapIndex<T, K, V>)
+where
+    T: AsRef<Snapshot>,
+    K: ProofMapKey + PartialEq + ::std::fmt::Debug,
+    V: StorageValue + PartialEq + ::std::fmt::Debug,
+{
+    let entries = match key {
```
vitvakatu: Can be replaced with 
```rust
let entries = key.map(|key| {
        let value = table.get(&key).unwrap();
        (key, value)
    });
```

PR #573, opened 2018-03-21
Human-readable description of tx failures

quicky84:
Cryptocurrency Service now returns a human-readable description in case of a tx failure, instead of an empty string. Relevant tests are updated.
(Test unfortunately fails when two transaction with the same hash are added to the blockchain. This is expected, but an error must be generated and now it just panics)

### DIFF
--- a/examples/cryptocurrency/src/lib.rs
+++ b/examples/cryptocurrency/src/lib.rs
@@ -207,7 +207,8 @@ pub enum Error {
 
 impl From<Error> for ExecutionError {
     fn from(value: Error) -> ExecutionError {
-        ExecutionError::new(value as u8)
+        let description = format!("{}", &value);
+        ExecutionError::with_description(value as u8, description)
     }
 }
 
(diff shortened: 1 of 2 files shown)


examples/cryptocurrency/src/lib.rs
```diff
@@ -207,7 +207,8 @@ pub enum Error {
 
 impl From<Error> for ExecutionError {
     fn from(value: Error) -> ExecutionError {
-        ExecutionError::new(value as u8)
+        let description = format!("{}", &value);
```
ozkriff: hmm, why not https://doc.rust-lang.org/std/error/trait.Error.html#tymethod.description ?

quicky84: As far as I understand this is the reason
```
#[derive(Debug, Fail)]
#[repr(u8)]
pub enum Error {
...
}
```
this is not an `std::error::Error`. Of course you can implement the trait, but I worked in the context of the example.

vitvakatu:
Replaced by https://github.com/exonum/exonum/pull/581

PR #574, opened 2018-03-21
Organize cryptocurrency example into modules [ECR-993]

andreyluchin:
Does what the title says: instead of comment delimiters, file is organized into rust modules. Imports in tests were changed accordingly. Names for modules were chosen in accordance (somewhat) to `configuration` service from `exonum-configuration` crate.
#### Important points:
* Imports were split up for each module
* function `new` was created for `CryptocurrencyApi` to fix visibility issues caused by struct construction


examples/cryptocurrency/src/lib.rs
```diff
@@ -39,410 +37,451 @@ extern crate router;
...
-use exonum::api::{Api, ApiError};
-use iron::prelude::*;
-use iron::Handler;
-use iron::status::Status;
-use iron::headers::ContentType;
-use iron::modifiers::Header;
-use router::Router;
-
-// // // // // // // // // // CONSTANTS // // // // // // // // // //
+/// Constants
```
DarkEld3r: Please add (here and below) a dot at the end of the comment.

examples/cryptocurrency/src/lib.rs
```diff
@@ -39,410 +37,451 @@ extern crate router;
...
+pub mod constants {
```
ozkriff: I think that it would be great to get rid of `constants` module.
`SERVICE_ID` constant can be moved to `service` module and `INIT_BALANCE` - to `contracts` mod.

examples/cryptocurrency/src/lib.rs
```diff
@@ -39,410 +37,451 @@ extern crate router;
...
-    pub fn wallets_mut(&mut self) -> MapIndex<&mut Fork, PublicKey, Wallet> {
-        MapIndex::new("cryptocurrency.wallets", &mut self.view)
-    }
-}
+/// Transactions
+pub mod transactions {
+    use exonum::crypto::PublicKey;
 
-// // // // // // // // // // TRANSACTIONS // // // // // // // // // //
+    use super::constants::SERVICE_ID;
```
ozkriff: What's the point of `super::` in this case?

examples/cryptocurrency/src/lib.rs
```diff
@@ -39,410 +37,451 @@ extern crate router;
...
 
-transactions! {
-    CurrencyTransactions {
-        const SERVICE_ID = SERVICE_ID;
+    transactions! {
+        pub(crate) CurrencyTransactions {
```
ozkriff: `pub(crate)` why not just `pub`?

andreyluchin: This due to an issue with `#[deny(missing_docs]` and `transaction!` macro I ran into. It is described [here](https://github.com/exonum/exonum/pull/575).

examples/cryptocurrency/src/lib.rs
andreyluchin: Makes a lot of sense. This is similar to what we see in `exonum-configuration` service. Will change.

PR #575, opened 2018-03-21
transaction! macro docs bug

andreyluchin:
While making [#574](https://github.com/exonum/exonum/pull/574) I needed to make transaction enum public (from inside the macro) like so:
```rust
transactions! {
        /// Some docs
        pub CurrencyTransactions {
            const SERVICE_ID = SERVICE_ID;

            /// Transaction type for creating a new wallet.
            ///
            /// See [the `Transaction` trait implementation](#impl-Transaction) for details how
            /// `TxCreateWallet` transactions are processed.
            struct TxCreateWallet {
                /// Public key of the wallet's owner.
                pub_key: &PublicKey,
                /// UTF-8 string with the owner's name.
                name: &str,
            }

            /// Transaction type for transferring tokens between two wallets.
            ///
            /// See [the `Transaction` trait implementation](#impl-Transaction) for details how
            /// `TxTransfer` transactions are processed.
            struct TxTransfer {
                /// Public key of the sender.
                from: &PublicKey,
                /// Public key of the receiver.
                to: &PublicKey,
                /// Number of tokens to transfer from sender's account to receiver's account.
                amount: u64,
                /// Auxiliary number to guarantee [non-idempotence][idempotence] of transactions.
                ///
                /// [idempotence]: https://en.wikipedia.org/wiki/Idempotence
                seed: u64,
            }
        }
    }
...
```
I added `Some docs` on top because crate was marked with `#[deny(missing_docs)]`.

However compilation failed with the following message:
```
error: macro expansion ignores token `doc = r" Some docs"` and any following
  --> <transactions macros>:19:30
   |
19 | derive ( Clone , Debug ) ] $ ( $ tx_set_attr ) * pub enum $ transaction_set{
   |                              ^^^^^^^^^^^^^^^^^
   |
note: caused by the macro expansion here; the usage of `transactions!` is likelyinvalid in item context
  --> examples/cryptocurrency/src/lib.rs:130:5
   |
130| /     transactions! {
131| |         /// Some docs
132| |         pub CurrencyTransactions {
133| |             const SERVICE_ID = SERVICE_ID;
...  |
162| |         }
163| |     }
   | |_____^
```

Looking into `transaction!` macro source code few mistakes were found when expanding attributes (**consider** that docs get expanded into `#[doc=""]` attribute), which are easily fixed by changing every incorrect `$($tx_set_attr)*` into correct `$(#[$tx_set_attr])*`. (Done in this pull request).

After that I get this compilation error:
```
error: missing documentation for a variant
  --> examples/cryptocurrency/src/lib.rs:139:20
   |
20 | | //! for use in production.
   | |_______________________________________^
...
139|               struct TxCreateWallet {
   |  ____________________^
   |
note: lint level defined here
  --> examples/cryptocurrency/src/lib.rs:26:9
   |
26 | #![deny(missing_docs)]
   |         ^^^^^^^^^^^^

error: missing documentation for a variant
  --> examples/cryptocurrency/src/lib.rs:150:20
   |
20 | | //! for use in production.
   | |_______________________________________^
...
150|               struct TxTransfer {
   |  ____________________^
error: aborting due to 2 previous errors

error: Could not compile `exonum-cryptocurrency`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```
A temporary workaround seems to be marking transaction enum with `pub(crate)` instead of `pub` like so:
```rust
transactions! {
        pub(crate) CurrencyTransactions {
```
and removing docs, which compiles just fine.

This pull request is dedicated to fixing this inconvenience.

#### Checklist:
- [x] Wait for [#574](https://github.com/exonum/exonum/pull/574) merge
- [x] ~~Write tests~~ (**update**: testing for compilation errors is a bad idea)
- [x] Find out what's causing the bug
- [x] Fix it if possible


aleksuss:
Write tests please

alekseysidorov:
I vote for the third option.

exonum/Cargo.toml
```diff
@@ -58,6 +58,7 @@ exonum_flamer = { path = "../3rdparty/flamer", version = "0.1.6", optional = tru
 pretty_assertions = "0.5.0"
 tempdir = "0.3.5"
 lazy_static = "1.0.0"
+compiletest_rs = "*"
```
aleksuss: Actually I guess it's a bad practice to use `*` instead numeric version

vitvakatu: Actually I'm not sure about additional dependency. Do we really need it? Let's discuss it privately with core team.

PR #576, opened 2018-03-21
Add field attribute `#[serde(default)]` to optional fields [ECR-663]

SerhiiOryshych:


exonum/src/node/mod.rs
```diff
@@ -342,6 +343,7 @@ pub struct NodeConfig {
     /// Memory pool configuration.
     pub mempool: MemoryPoolConfig,
     /// Additional config, usable for services.
+    #[serde(default)]
     pub services_configs: BTreeMap<String, Value>,
     /// Optional database configuration.
     pub database: Option<DbOptions>,
```
vitvakatu: What about adding attribute for this field as well? (And then we can remove `Option`? :thinking: )

PR #579, opened 2018-03-23
Remove network_id from everywhere. [ECR-1016]

pavel-mukhanov:
`network_id` parameter is no longer needed, therefore we decided to remove it.

This change breaks **cryptocurrency** example and I also updated test json's.

**update:**  we decided to keep `HEADER_LENGTH=10` and reserve zero byte in `RawMessage`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `events` module becomes private. (#568)
 
 - `CryptoHash` trait is no longer implemented for `Hash`. (#579)
+
 - `network_id` attribute has been removed from `NodeInfo` and `RawMessage`. It
   affects message structure, now `HEADER_LENGTH` is `9` and all values shifted left
   by one.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `events` module becomes private. (#568)
 
 - `CryptoHash` trait is no longer implemented for `Hash`. (#579)
+- `network_id` attribute has been removed from `NodeInfo` and `RawMessage`. It
+  affects message structure, now `HEADER_LENGTH` is `9` and all values shifted left
+  by one.
 
 #### exonum-testkit
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `CryptoHash` trait is no longer implemented for `Hash`. (#578)
 
+- `network_id` attribute has been removed from `NodeInfo` and `RawMessage`.
+  `HEADER_LENGTH` remains the same, zero byte of `RawMessage` is now reserved and
+  always set to `0`. (#579)
+
 #### exonum-testkit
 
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -56,6 +56,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `CryptoHash` trait is no longer implemented for `Hash`. (#579)
 
+- `network_id` attribute has been removed from `NodeInfo` and `RawMessage`. It
+  affects message structure, now `HEADER_LENGTH` is `9` and all values shifted left
+  by one.
```
DarkEld3r: Please add the pull request number. :upside_down_face:

exonum/src/events/codec.rs
```diff
@@ -92,7 +92,7 @@ mod test {
 
     #[test]
     fn decode_message_valid_header_size() {
-        let data = vec![0u8, 0, 0, 0, 0, 0, 10, 0, 0, 0];
+        let data = vec![0u8, 0, 0, 0, 0, 10, 0, 0, 0, 0];
```
vldm: This test was about minimum valid header size, can you fix it?

exonum/src/encoding/segments.rs
```diff
@@ -189,7 +189,7 @@ impl<'a> SegmentField<'a> for RawMessage {
             });
         }
         let actual_size = slice.len() as Offset;
-        let declared_size: Offset = LittleEndian::read_u32(&slice[6..10]);
+        let declared_size: Offset = LittleEndian::read_u32(&slice[5..9]);
```
vldm: I'm afraid of the new header size. Probably, the alignment will be worse

DarkEld3r: I doubt this will affect performance.

slowli: Can we benchmark it somehow? It could be possible to come up with an alternative that just permanently sets the `network_id` byte to 0 (with an added benefit that the change would be backward-compatible).

DarkEld3r: On second thought, alignment can be important. Still I don't like `network_id`, so I see several solutions:
- Remove the `network_id` method, but leave header size as 10 bytes by adding a placeholder (`MessageWriter` should add it).
- Adjust other fields, for example change `version` from `u8` to `u16`.

CHANGELOG.md
```diff
@@ -120,6 +116,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Sandbox tests have been moved inside of the exonum core. (#568)
 
+- `network_id` attribute has been removed from `NodeInfo` and `RawMessage`.
```
DarkEld3r: I suppose this is still technically a breaking change. :warning:

CHANGELOG.md
```diff
@@ -54,7 +54,11 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `events` module becomes private. (#568)
 
-- `CryptoHash` trait is no longer implemented for `Hash`. (#579)
+- `CryptoHash` trait is no longer implemented for `Hash`. (#578)
+
+- `network_id` attribute has been removed from `NodeInfo` and `RawMessage`.
+  `HEADER_LENGTH` remains the same, zero byte of `RawMessage` is now reserved and
```
vitvakatu: Not `zero byte`, but `first byte`?

PR #580, opened 2018-03-23
example: Don't evaluate wallet in block

ivanovaleksey:
Hello,

while I was studying the example I noticed that `wallet` is evaluated in block.

I am not sure but it seems like there is no any special reason for that. So, I tried to remove the block and everything still worked fine.

Please, correct me if I am wrong here.


DarkEld3r:
Thank you for the pull request. This is probably a refactoring artifact. 🙃

ivanovaleksey:
Thanks for merging!

PR #582, opened 2018-03-26
CheckpointDb and handmade checkpoints [ECR-975]

andreyluchin:
Note: this is remake of https://github.com/exonum/exonum/pull/569 which was closed due to local technical difficulties.

This PR changes the interface of CheckpointDb from merge based rollbacks to  rollbacks to handmade checkpoints. Multiple checkpoints can be set and are handled on a stack-based basis (Rollback always rolls back to **the last** checkpoint).

## Before:
```rust
handler.rollback(3);
```
## Now:
```rust
handler.checkpoint();
// Mutate the database
handler.rollback();
```
or 
```rust
handler.checkpoint();
handler.checkpoint();
// Mutate the database
handler.rollback();
handler.rollback();
```

## Vision
Merge-based rollbacks was a very inconvenient way to use CheckpointDb and Testkit. It only makes sense to implement actual checkpoints to CheckpointDb.
## Issues for discussion
* There is no place in the existing codebase where multiple checkpoints are needed (besides tests for CheckpointDb itself)
* Stack-based rollbacks may not be the very convenient

### DIFF
--- a/testkit/src/checkpoint_db.rs
+++ b/testkit/src/checkpoint_db.rs
@@ -129,7 +129,7 @@ impl<T: Database> CheckpointDbInner<T> {
         self.db.merge(patch.clone())?;
         let mut rev_fork = self.db.fork();
 
-        // Reverse a patch to get a backup patch
+        // Reverse a patch to get a backup patch.
         for (name, changes) in patch {
             for (key, _) in changes {
                 match snapshot.get(&name, &key) {
@@ -143,7 +143,7 @@ impl<T: Database> CheckpointDbInner<T> {
             }
         }
 
-        // Should insert backup patches to the front of the backup (VecDeque)
+        // Should insert backup patches to the front of the backup (VecDeque).
         self.backup_stack
             .last_mut()
             .expect("`merge_with_logging` called before checkpoint has been set")
@@ -283,7 +283,7 @@ mod tests {
             );
         }
 
-        // Check that the old snapshot still corresponds to the same DB state
+        // Check that the old snapshot still corresponds to the same DB state.
         assert_eq!(snapshot.get("foo", &[]), Some(vec![2]));
         let snapshot = db.snapshot();
         assert_eq!(snapshot.get("foo", &[]), Some(vec![3]));
@@ -297,7 +297,7 @@ mod tests {
         fork.put("foo", vec![], vec![2]);
         db.merge(fork.into_patch()).unwrap();
 
-        // Both checkpoints are on purpose
+        // Both checkpoints are on purpose.
         handler.checkpoint();
         handler.checkpoint();
         let mut fork = db.fork();
@@ -331,7 +331,7 @@ mod tests {
             assert_eq!(backup.len(), 0);
         }
 
-        // Check that DB continues working as usual after a rollback
+        // Check that DB continues working as usual after a rollback.
         handler.checkpoint();
         let mut fork = db.fork();
         fork.put("foo", vec![], vec![4]);
--- a/testkit/src/checkpoint_db.rs
+++ b/testkit/src/checkpoint_db.rs
@@ -217,42 +217,27 @@ mod tests {
         assert_eq!(patch_set, expected_set);
     }
 
+    fn stack_len<T>(db: &CheckpointDb<T>) -> usize {
+        let inner = db.inner.read().unwrap();
+        inner.backup_stack.len()
+    }
+
     #[test]
     fn test_backup_stack() {
         let db = CheckpointDb::new(MemoryDB::new());
         let handler = db.handler();
 
-        {
-            let inner = db.inner.read().unwrap();
-            let stack = &inner.backup_stack;
-            assert_eq!(stack.len(), 0);
-        }
+        assert_eq!(stack_len(&db), 0);
         handler.checkpoint();
-        {
-            let inner = db.inner.read().unwrap();
-            let stack = &inner.backup_stack;
-            assert_eq!(stack.len(), 1);
-        }
+        assert_eq!(stack_len(&db), 1);
         handler.rollback();
-        {
-            let inner = db.inner.read().unwrap();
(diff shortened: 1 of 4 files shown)


testkit/src/checkpoint_db.rs
```diff
@@ -195,33 +218,83 @@ mod tests {
     }
 
     #[test]
-    fn test_checkpoint_db_basics() {
+    fn test_backup_stack() {
         let db = CheckpointDb::new(MemoryDB::new());
+        let handler = db.handler();
+
+        {
```
vitvakatu: nitpick: you can move these repeating pieces into helper function.

andreyluchin: I originally decided that this helper function will hurt consistency with code in latter tests (since they have a similar pattern which cannot be replaced by the same helper function). But on second thought, introducing a helper is a good idea. Will do

testkit/tests/inflating_currency/main.rs
```diff
@@ -168,16 +175,6 @@ fn fuzz_transfers_and_maybe_rollbacks(use_rollbacks: bool) {
         let total_balance: u64 = pubkeys.iter().map(|key| get_balance(&api, key)).sum();
         assert_eq!(total_balance, (USERS as u64) * testkit.height().0);
 
-        if use_rollbacks {
-            let rollback_blocks = rng.choose(&[0usize, 0, 0, 1, 2, 3]);
```
slowli: I think it could make sense to randomly generate and rollback checkpoints here.

testkit/src/checkpoint_db.rs
```diff
@@ -73,31 +74,39 @@ pub struct CheckpointDbHandler<T> {
...
-            .rollback(count);
+            .rollback();
     }
 }
 
 #[derive(Debug)]
 struct CheckpointDbInner<T> {
     db: T,
-    journal: Vec<Patch>,
+    backup_stack: Vec<VecDeque<Patch>>,
```
slowli: Nitpick: I don't particularly understand the reasoning behind renaming, but I'm OK with the new name.

testkit/tests/change_configuration.rs
```diff
@@ -66,23 +66,13 @@ fn test_configuration_and_rollbacks() {
...
     let new_config = proposal.stored_configuration().clone();
 
-    testkit.commit_configuration_change(proposal);
+    testkit.checkpoint();
 
-    testkit.create_blocks_until(Height(10));
-    assert_eq!(testkit.actual_configuration(), new_config);
-    testkit.rollback(2);
-    assert_eq!(testkit.actual_configuration(), old_config);
+    testkit.commit_configuration_change(proposal);
```
slowli: It makes sense to create another checkpoint after `commit_configuration_change` and test that its rollback does not influence accepting the new configuration.

testkit/src/checkpoint_db.rs
```diff
@@ -125,20 +143,25 @@ impl<T: Database> CheckpointDbInner<T> {
             }
         }
 
-        self.journal.push(rev_fork.into_patch());
+        // Should insert backup patches to the front of the backup (VecDeque).
+        self.backup_stack
+            .last_mut()
+            .expect("`merge_with_logging` called before checkpoint has been set")
+            .push_front(rev_fork.into_patch());
```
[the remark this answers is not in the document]
andreyluchin: Good point. I'm on it

testkit/src/checkpoint_db.rs
```diff
@@ -125,20 +143,25 @@ impl<T: Database> CheckpointDbInner<T> {
...
         assert!(
-            self.journal.len() >= count,
-            "Cannot rollback {} changes; only {} checkpoints in the journal",
-            count,
-            self.journal.len()
+            !self.backup_stack.is_empty(),
+            "Checkpoint has not been set yet"
         );
-
-        for _ in 0..count {
```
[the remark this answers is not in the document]
andreyluchin: In my vision, directly iterating with `drain` or `into_iter` is a much cleaner approach, I think `for _ in 0..count` is a waste of a for loop header, we also avoid using `assert!`. I don't see why would we want to preserve the old semantics.

testkit/src/checkpoint_db.rs
```diff
@@ -194,82 +217,136 @@ mod tests {
...
         fork.put("bar", vec![1], vec![4]);
+        fork.put("bar2", vec![5], vec![6]);
         db.merge(fork.into_patch()).unwrap();
         {
             let inner = db.inner.read().unwrap();
-            let journal = &inner.journal;
-            assert_eq!(journal.len(), 2);
-            check_patch(&journal[0], vec![("foo", vec![], Change::Delete)]);
+            let stack = &inner.backup_stack;
+            let recent_backup = stack.last().expect("There are not backups in the stack");
```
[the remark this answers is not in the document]
andreyluchin: This makes a lot of sense. On it

testkit/src/checkpoint_db.rs
andreyluchin: I find `journal` name very ambiguous. `backup_stack` on the other hand, gives an idea on two things:
* This `Vec` is used like a stack;
* It contains "backup" patches as opposed to "change" patches.

slowli: I've taken the name from [DB terminology](https://en.wikipedia.org/wiki/Transaction_log), but I see your point.

andreyluchin: From the [article](https://en.wikipedia.org/wiki/Transaction_log):
> Physically, a log is a file listing changes to the database, stored in a stable storage format.

This contradicts with our case because we don't store actual changes to the database but rather reversed changes.

testkit/tests/change_configuration.rs
andreyluchin: Can you please elaborate on it? We don't really have arguments to properly fuzz test and I find it unnecessary for this situation.

PR #583, opened 2018-03-27
Bunch transactions response [ECR-1029]

SerhiiOryshych:

### DIFF
--- a/exonum/src/messages/mod.rs
+++ b/exonum/src/messages/mod.rs
@@ -22,7 +22,7 @@ use crypto::PublicKey;
 use encoding::Error;
 use helpers::{Height, Round, ValidatorId};
 
-pub use self::raw::{RawMessage, MessageWriter, MessageBuffer, ServiceMessage, Message,
+pub use self::raw::{Message, MessageBuffer, MessageWriter, RawMessage, ServiceMessage,
                     HEADER_LENGTH, PROTOCOL_MAJOR_VERSION};
 pub use self::protocol::*;
 
@@ -55,6 +55,8 @@ pub enum Any {
     Request(RequestMessage),
     /// Transaction.
     Transaction(RawTransaction),
+    /// A batch of the transactions.
+    TransactionsBatch(TransactionsResponse),
 }
 
 /// Consensus message.
@@ -208,6 +210,9 @@ impl Any {
                 CONNECT_MESSAGE_ID => Any::Connect(Connect::from_raw(raw)?),
                 STATUS_MESSAGE_ID => Any::Status(Status::from_raw(raw)?),
                 BLOCK_RESPONSE_MESSAGE_ID => Any::Block(BlockResponse::from_raw(raw)?),
+                TRANSACTIONS_RESPONSE_MESSAGE_ID => Any::TransactionsBatch(
+                    TransactionsResponse::from_raw(raw)?
+                ),
 
                 PROPOSE_MESSAGE_ID => {
                     Any::Consensus(ConsensusMessage::Propose(Propose::from_raw(raw)?))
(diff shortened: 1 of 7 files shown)


exonum/src/node/consensus.rs
```diff
@@ -589,6 +590,36 @@ impl NodeHandler {
...
+            );
+            return;
+        }
+
+        if !msg.verify_signature(msg.from()) {
+            error!("Received request with incorrect signature, msg={:?}", msg);
+            return;
+        }
+
+        for tx in msg.transactions() {
```
slowli: Do I understand correctly that this code clones all raw messages? Doesn't look very good, tbh.

exonum/src/messages/mod.rs
```diff
@@ -55,6 +55,8 @@ pub enum Any {
     Request(RequestMessage),
     /// Transaction.
     Transaction(RawTransaction),
+    /// A bunch of transactions.
+    TransactionsBunch(TransactionsResponse),
```
slowli: `Bunch` intuitively seems wrong to me; `TransactionsBundle` or `TransactionsBatch` both sound better. @Maria-Nosyk?

Maria-Nosyk: bunch isn't absolutely wrong but has a slightly different connotation, besides batch is more common for us as I could notice from our docs

alekseysidorov:
What about the problem with the messages size limit? Single transaction can fit it, but a several in sum can be bigger than size limit.

exonum/src/messages/mod.rs
DarkEld3r: Personally, I prefer to use "batch" term here.

DarkEld3r:
@alekseysidorov 
Nice catch, but we already have this problem with `BlockResponse` message. I suppose we can deal with it separately.

exonum/src/node/consensus.rs
SerhiiOryshych: Yes, you are right. How do you think, how should this be done?

SerhiiOryshych:
@alekseysidorov 
I'm trying to verify this here: https://github.com/exonum/exonum/pull/583/files#diff-dece0e59f8e786debd5d53b790fcccfcR87.

exonum/src/messages/protocol.rs
```diff
@@ -63,6 +65,10 @@ pub const PEERS_REQUEST_MESSAGE_ID: u16 = PeersRequest::MESSAGE_ID;
 /// `BlockRequest` message id.
 pub const BLOCK_REQUEST_MESSAGE_ID: u16 = BlockRequest::MESSAGE_ID;
 
+/// The size of the part of the message intended only for transactions.
+pub const TRANSACTIONS_RESPONSE_UNOCCUPIED_MESSAGE_SIZE: u32 =
```
vldm: Can you rewrite it as local variable, and use real `MAX_MESSAGE_LEN` config?

exonum/src/node/consensus.rs
```diff
@@ -589,6 +590,36 @@ impl NodeHandler {
         }
     }
 
+    /// Handles raw transactions.
+    #[cfg_attr(feature = "flame_profile", flame)]
+    pub fn handle_txs(&mut self, msg: &TransactionsResponse) {
+        if msg.to() != self.state.consensus_public_key() {
+            error!(
+                "Received request that intended for another peer, to={}, from={}",
```
vldm: As i understand, this is not request.

exonum/src/node/consensus.rs
```diff
@@ -589,6 +590,36 @@ impl NodeHandler {
...
+                msg.to().to_hex(),
+                msg.from().to_hex()
+            );
+            return;
+        }
+
+        if !self.state.whitelist().allow(msg.from()) {
```
vldm: Probably we should ignore "batches" that was not requested perviously.

exonum/src/node/requests.rs
```diff
@@ -12,9 +12,13 @@
...
 // limitations under the License.
 
+use std::mem::replace;
+
 use messages::{BlockRequest, BlockResponse, Message, PrevotesRequest, ProposeRequest,
-               RequestMessage, TransactionsRequest};
+               RequestMessage, TransactionsRequest, TransactionsResponse};
 use blockchain::Schema;
+use crypto::{PUBLIC_KEY_LENGTH, SIGNATURE_LENGTH};
+use messages::HEADER_LENGTH;
```
vitvakatu: nitpick: what about grouping this import with that one above?

PR #584, opened 2018-03-28
Testkit/checkpointdb refactor

andreyluchin:
This PR contains some minor changes and refactoring to changes introduced in https://github.com/exonum/exonum/pull/582.

Also, the changes in `CheckpointDb` and `Testkit` interfaces were described in the `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,12 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum-testkit
 
+- `CheckpointDb` and `Testkit` have their `rollback` methods updated (#582):
+  - `rollback` by blocks in `Testkit` was removed;
+  - `rollback` by merge count in `CheckpointDb` was removed;
+  - `checkpoint` method was introduced to set checkpoints;
+  - new `rollback` rolls back to the last set checkpoint.
+
 - Testkit api now contains two methods to work with the transaction pool (#549):
   - `is_tx_in_pool` - for checking transaction existence in the pool;
   - `add_tx` - for adding a new transaction into the pool.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   - `checkpoint` method was introduced to set checkpoints;
   - new `rollback` rolls back to the last set checkpoint.
 
+  Migration path:
+  - Replace every old `rollback(blocks)` by a pair of `checkpoint()` and `rollback()`
+
 - Testkit api now contains two methods to work with the transaction pool (#549):
   - `is_tx_in_pool` - for checking transaction existence in the pool;
   - `add_tx` - for adding a new transaction into the pool.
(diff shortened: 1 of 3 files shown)


testkit/tests/change_configuration.rs
```diff
@@ -72,6 +72,16 @@ fn test_configuration_and_rollbacks() {
     testkit.create_blocks_until(Height(10));
     assert_eq!(testkit.actual_configuration(), new_config);
 
+    testkit.checkpoint();
+    testkit.create_block();
+    testkit.rollback();
+    assert_eq!(testkit.actual_configuration(), new_config);
+
+    testkit.checkpoint();
```
slowli: I don't quite understand why 2 instances of this code are required. IMO, it could make sense to setup one checkpoint immediately after a new config is committed:

```rust
testkit.commit_configuration_change(proposal);
testkit.checkpoint(); // <<< new code
testkit.create_blocks_until(Height(10));
assert_eq!(testkit.actual_configuration(), new_config);

testkit.rollback();
assert_eq!(testkit.actual_configuration(), old_config);
testkit.create_blocks_until(Height(10));
assert_eq!(testkit.actual_configuration(), new_config);

// other code...
```

testkit/src/checkpoint_db.rs
```diff
@@ -308,11 +304,9 @@ mod tests {
             let inner = db.inner.read().unwrap();
             let stack = &inner.backup_stack;
             assert_eq!(stack.len(), 2);
-            let backup = stack.last().expect("There are not backups in the stack");
+            let backup = &stack[1];
             assert_eq!(backup.len(), 1);
```
slowli: Tiny nitpick: you can write `assert_eq!(stack[1].len(), 1)` directly here and in similar places; imo, it doesn't hurt code readability.

CHANGELOG.md
```diff
@@ -62,6 +62,12 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum-testkit
 
+- `CheckpointDb` and `Testkit` have their `rollback` methods updated (#582):
```
slowli: `CheckpointDb` is not public, so I don't think we should write about changes in it here.

testkit/tests/change_configuration.rs
andreyluchin: This is not going to go through because of how `commit_configuration_change` in testkit works:
1. `commit_configuration_change` schedules configuration for update in the next block.
2. block is created, configuration updated
If testkit rollbacks this block here - we get `old_config` (since `new_config` is no longer in blockchain). However, `testkit.create_blocks_until(Height(10));` won't change it to `new config`, because we don't have it scheduled by `commit_configuration_change`.

CHANGELOG.md
```diff
@@ -62,6 +62,11 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum-testkit
 
+- Rollback mechanism in `Testkit` is reworked to work with checkpoints (#582):
+  - old `rollback` by blocks in `Testkit` was removed;
+  - `checkpoint` method was introduced to set checkpoints;
+  - new `rollback` rolls back to the last set checkpoint.
```
DarkEld3r: It would be nice to add a migration path. :upside_down_face:

andreyluchin: Totally missed that. WIll do

PR #586, opened 2018-03-28
Introduce '--validators-count' parameter to config template. [ECR-905]

pavel-mukhanov:
Introduce `--validators-count` parameter to config template.

**Use case**: 
```
cryptocurrency generate-template common.toml --validators-count 4
...
cryptocurrency finalize ... --public-configs pub_1.toml pub_2.toml pub_3.toml pub_4.toml
```

If count of provided `pub_*.toml` files differ from `validators-count` panic will be raised.

```
error: Number of validators configs does not match number of validator keys.
```

**For discussion:**
* Method `override_validators_count()` is  probably is not optimal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   a new configuration proposal. By default the number of votes is calculated
   as 2/3 + 1 of total validators count. (#546)
 
+- `validators-count` command-line parameter is added. Now, when generating
+  config template using `generate-template` command you must specify number of validators.
+
 #### exonum-time
 
 - `SystemTime` has been replaced with `chrono::DateTime<Utc>`, as it provides
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   as 2/3 + 1 of total validators count. (#546)
 
 - `validators-count` command-line parameter is added. Now, when generating
-  config template using `generate-template` command you must specify number of validators.
+  config template using `generate-template` command you must specify number of
+  validators. (#586)
 
 #### exonum-time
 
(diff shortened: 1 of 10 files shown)


exonum/src/helpers/fabric/shared.rs
```diff
@@ -65,6 +65,8 @@ pub struct CommonConfigTemplate {
     /// Services configuration.
     #[serde(default)]
     pub services_config: AbstractConfig,
+    /// General configuration
```
DarkEld3r: Please add a dot. :upside_down_face:

exonum/src/helpers/fabric/details.rs
```diff
@@ -554,8 +574,20 @@ impl Command for Finalize {
                 ConfigFile::load(path).expect("Failed to load validator public config.")
             })
             .collect();
+
         let (common, list, our) = Self::reduce_configs(public_configs, &secret_config);
 
+        let validators_count = common
+            .general_config
+            .get("validators_count")
+            .expect("validators_count is not found in common config.")
```
DENMROOT: Why not make it optional with default value ? Just to avoid breaking changes.

DarkEld3r: @DENMROOT I'm not sure that we want to "avoid breaking changes", at least for now. :upside_down_face: This will require some compromises and introduce additional complexity.

Additionally, as far as I know, we need this parameter to be explicit in order to prevent mistakes. With default value we are going to lose that advantage.

CHANGELOG.md
```diff
@@ -78,6 +78,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   a new configuration proposal. By default the number of votes is calculated
   as 2/3 + 1 of total validators count. (#546)
 
+- `validators-count` command-line parameter is added. Now, when generating
```
vitvakatu: We are using present perfect tense (in passive voice) in changelog, so:
```
`validators-count` command-line parameter has been added. Now, when generating config template using `generate-template` command, you must specify the number of validators.
```

exonum/src/helpers/fabric/details.rs
vitvakatu: validators_count not found in common config.

exonum/src/helpers/fabric/details.rs
```diff
@@ -554,8 +574,20 @@ impl Command for Finalize {
...
+            .as_integer()
+            .unwrap() as usize;
+
+        if validators_count != list.len() {
+            panic!("Number of validators configs does not match number of validator keys.");
```
vitvakatu: The number of validators configs does not match the number of validators keys.

PR #592, opened 2018-04-03
Modify method with_description in ExecutionError structure

VladYermakov:
Changed signature of method with_description in ExecutionError structure to
`pub fn with_description<T : Into<String>>(code: u8, description: T) -> Self`
Now we can pass &str arguments to this function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### Exonum core
 
-- `ExecutionError::with_description` method now takes `Into<String>` instead of `String` which allows to pass `&str` directly. (#592)
+- `ExecutionError::with_description` method now takes `Into<String>`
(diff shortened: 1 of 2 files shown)


exonum/src/blockchain/transaction.rs
```diff
@@ -656,12 +657,13 @@ mod tests {
...
 
         for value in &values {
-            let error = ExecutionError::with_description(value.0, value.1.to_owned());
+            let error = ExecutionError::with_description(value.0, value.1);
             assert_eq!(value.0, error.code);
             assert_eq!(value.1, error.description.unwrap());
         }
     }
 
+
```
ozkriff: ?

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### Exonum core
 
+- ExecutionError::with_description method now takes Into<String> instead of
```
vitvakatu: Please use fancy markdown syntax with \`sometext\`

PR #593, opened 2018-04-03
Make `transactions!` macro support empty body. [ECR-1133]

andreyluchin:
The following `transactions!` usage is now possible:
```rust
transactions!{}
```


alexander-irbis:
> I'm not convinced about usefulness of this change (it's superfluous feature to me), but PR is okay

It was suggested that this creates an unexpected "why?" for developers at the initial stage of creating a service. Although this is not critical at all, it is only a question of convenience.

PR #598, opened 2018-04-05
Enhance cli fabric

BitfuryLightning:
It proposal added possibility to extend default exonum's fabric cli.

### DIFF
--- a/exonum/src/helpers/fabric/builder.rs
+++ b/exonum/src/helpers/fabric/builder.rs
@@ -28,6 +28,8 @@ use super::details::{Run, RunDev, Finalize, GenerateNodeConfig, GenerateCommonCo
 use super::keys;
 use super::CommandName;
 
+use clap;
+
 /// `NodeBuilder` is a high level object,
 /// usable for fast prototyping and creating app from services list.
 #[derive(Default)]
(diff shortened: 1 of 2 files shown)


vldm:
@BitfuryLightning Hello, good start, but there some problems with API.
Our public api in fabric module encapsulate, so there should be no clap in public methods.

`pub fn execute<F>(commands: &HashMap<CommandName, CollectedCommand>, app: Option<clap::App>, callback: Option<F>)`

One possible way is to rewrite this code using our internal types.

Additionally we can implement "pseudo command" - "default" for this purposes, and use existing "fabric" pattern.

BitfuryLightning:
@vldm I want to give an opportunity integrate any clap's cli with exonum's fabric cli while I didn't want to expose fabric::internal module. Can you hint me, what i can do that in another way?

If i correctly understood you, you propose add default command and redetermine it outside if is necessary, but this way restricts outside usage. I mean I can't integrate my complex cli with exonum's fabric cli.

vldm:
@BitfuryLightning 
>I can't integrate my complex cli with exonum's fabric cli.
Usually you should not.
The one possible solution is to extend our fabric API  (add default command, allow external users to extend commands list).
More dirty solution is to make undocumented api, that allows you to integrate your CLI.

@DarkEld3r  do you have any thoughts about undocumented api?

DarkEld3r:
I'm OK with "hidden" (undocumented) api if this functionality is crucial. Otherwise it would be better to create our own abstraction level.

DarkEld3r:
Would be implemented as part of https://jira.bf.local/browse/ECR-72 and https://jira.bf.local/browse/ECR-1165 tasks.

PR #601, opened 2018-04-06
Logging middleware for testkit

slowli:
As a part of bringing #459 to shape, this PR introduces logging middleware for the testkit API and adds corresponding tests.


testkit/src/lib.rs
```diff
@@ -280,7 +281,7 @@ impl TestKitBuilder {
     /// Creates the testkit.
     pub fn create(self) -> TestKit {
         if self.logger {
-            exonum::helpers::init_logger().unwrap();
+            exonum::helpers::init_logger().ok();
```
vitvakatu: What's the purpose of this change?

testkit/src/api.rs
```diff
@@ -311,6 +307,96 @@ impl TestKitApi {
...
+            .unwrap_or_default(),
+        resp_status = response
+            .status
+            .as_ref()
+            .map(ToString::to_string)
+            .unwrap_or_else(|| "(no status)".to_string()),
+        resp_body = response_body
+            .as_ref()
+            .map(String::as_ref)
+            .unwrap_or_else(|| "(no body)")
```
vitvakatu: `unwrap_or`?

testkit/src/api.rs
```diff
@@ -311,6 +307,96 @@ impl TestKitApi {
...
+        method = request.method,
+        access = access,
+        url = url,
+        req_body = req_body
+            .map(|body| format!("\nBody: {}", body))
+            .unwrap_or_default(),
+        resp_status = response
+            .status
+            .as_ref()
+            .map(ToString::to_string)
```
vitvakatu: Why not using `String::as_ref` here too?

testkit/src/api.rs
```diff
@@ -311,6 +307,96 @@ impl TestKitApi {
...
+        })
+    };
+
+    trace!(
+        "{method} ({access}) /{url}{req_body}\nResponse: {resp_status}\n{resp_body}\n",
+        method = request.method,
+        access = access,
+        url = url,
+        req_body = req_body
+            .map(|body| format!("\nBody: {}", body))
```
vitvakatu: nitpick: I'm not a fan of avoiding **all** allocations in code, but you can easily avoid any of them in this part of the code (by the price of less readable message in case of no request body).

testkit/src/lib.rs
slowli: There is a separate commit dedicated to it :slightly_smiling_face: Basically, if you try to enable the logger in an integration test (say, `counter`) via

```rust
fn init_testkit() -> (TestKit, TestKitApi) {
    let testkit = TestKitBuilder::validator()
        .with_service(CounterService)
        .with_logger()
        .create();
    // ...
}
```

then all tests except the first one will panic on this `unwrap`.

testkit/src/api.rs
slowli: `response.status` has type `Option<Status>`, so `ToString::to_string` converts it to the string form. If I understand correctly, in order to use `String::as_ref`, you need to map `Option<Status>` to `Option<String>` beforehand, which is sort of clunky.

PR #549, opened 2018-03-09
Change mempool to save transactions in storage. [ECR-895]

vldm:
Save transactions in persistent pool instead of in memory `Map`. This allows node to recover safe after restart.

PR #608, opened 2018-04-11
Fix BlockResponse with known transaction bug

vldm:
After https://github.com/exonum/exonum/pull/549 `handle_block` ignore blocks with transactions that already in pool.

### DIFF
--- a/exonum/src/sandbox/consensus.rs
+++ b/exonum/src/sandbox/consensus.rs
@@ -20,12 +20,14 @@ use bit_vec::BitVec;
 use std::time::Duration;
 use std::collections::BTreeMap;
 
-use messages::{Connect, Message, PeersRequest, Precommit, Prevote, PrevotesRequest, Propose,
-               ProposeRequest, RawMessage, TransactionsRequest, TransactionsResponse, CONSENSUS};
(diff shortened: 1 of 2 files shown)


exonum/src/sandbox/consensus.rs
```diff
@@ -3074,6 +3076,90 @@ fn handle_tx_has_full_propose() {
     sandbox.add_time(Duration::from_millis(0));
 }
 
+/// HANDLE block response
+
+/// - should process block even if tx in pool
```
DarkEld3r: I would extend the description a little bit. Additionally, do we need a doc-comments here?

exonum/src/sandbox/consensus.rs
```diff
@@ -3074,6 +3076,90 @@ fn handle_tx_has_full_propose() {
...
+#[test]
+fn handle_block_response_tx_in_pool() {
+    let sandbox = timestamping_sandbox();
+
+    let tx = gen_timestamping_tx();
+
+    let propose = ProposeBuilder::new(&sandbox)
+        .with_duration_since_sandbox_time(sandbox.propose_timeout())
+        .build();
+    // this block with transactions should be in real
```
DarkEld3r: This comment isn't clear for me. :upside_down_face:

PR #612, opened 2018-04-12
Update dependencies versions

DarkEld3r:
`tokio-retry` will be updated separately: https://jira.bf.local/browse/ECR-1226


DarkEld3r:
It seems that `tokio-timer` version cannot be (easily) updated either.

PR #613, opened 2018-04-12
Add 'allow-origin' parameter to the finalize command [ECR-1225]

DarkEld3r:


exonum/src/node/mod.rs
```diff
@@ -245,28 +247,24 @@ impl<'de> de::Deserialize<'de> for AllowOrigin {
...
-#[test]
-fn allow_origin_serde() {
-    fn check(text: &str, allow_origin: AllowOrigin) {
-        #[derive(Serialize, Deserialize)]
-        struct Config {
-            allow_origin: AllowOrigin,
+impl FromStr for AllowOrigin {
+    type Err = failure::Error;
+
+    fn from_str(s: &str) -> Result<Self, <Self as FromStr>::Err> {
```
vitvakatu: `Self::Err` is sufficient here

DarkEld3r: This signature was generated by the IDE. 🙃

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### New features
+
+#### exonum
+
+- The `--allow-origin` parameter has been added to the `finalize` command.
```
vitvakatu: @Maria-Nosyk can you please take a look at this line? Is `The` at the beginning necessary?

alexander-irbis: This is a fairly common use of the article "the". For example, in the Java documentation: https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html : 

> _The `args` parameter_ represents any of the command line arguments that would normally be passed to the javac program and are outlined in the above Synopsis section.

> _The `out` parameter_ indicates where the compiler's diagnostic output is directed.

> If _the `-Xprefer:newer` option_ is used, then it reads the newer of the source or class file for a type (default). If _the `-Xprefer:source` option_ is used, then it reads the source file. 

But:
> Use _`-Xprefer:source`_ when you want to be sure that any annotation processors can access annotations declared with a retention policy of SOURCE.

PR #615, opened 2018-04-13
IPv6 Support [ECR-158]

andreyluchin:
To support IPv6 in exonum few things needed to be changed:
1. `SocketAddr` serialization;
2. IP address parsing in fabric;
3. IP address parsing in TOML (which happened to be needless since it's done with `serde` that already supports IPv6 in `SokcetAddr`).

While making this PR I found that we parse same data (IP addresses) differently in TOML (uses serde) and in command-line arguments (uses std), which seems odd.

## Update 2 (after collective review)
Changes introduced after the review:
1. `const` moved to the top;
2. `SIZE_DIFF` is now a standalone constant;
3. `IpMode` enum instead of `bool` in tests.

### Discussion topic
I didn't write any tests for IPv6 parsing in TOML. It is parsed by `serde` crate, so it is its responsibility to cover that and the functionality is `std::net::SocketAddr`'s responsibility, but I still have doubts about that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
+### New features
+
+- IPv6 addressing is now supported. (#615)
+
 ### Bug fixes
 
 #### exonum-cryptocurrency-advanced
@@ -12,8 +16,6 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Frontend has been updated to reflect latest backend changes. (#602 #611)
 
 ### Internal improvements
-
-- IPv6 addressing is now supported. (#615)
 
 #### exonum-time
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Internal improvements
 
+- IPv6 addressing is now supported. (#615)
+
 #### exonum-time
 
 - Split service components to separate modules. (#604)
@@ -170,8 +172,6 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `Display` trait is implemented for types from the `crypto` module. (#590)
 
 - `transactions!` macro now allows empty body. (#593)
-
-- IPv6 addressing is now supported by exonum. (#615)
 
 #### exonum-testkit
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `transactions!` macro now allows empty body. (#593)
 
+- IPv6 addressing is now supported by exonum. (#615)
+
 #### exonum-testkit
 
 - `create_block*` methods of the `TestKit` now return the information about
--- a/exonum/src/encoding/fields.rs
+++ b/exonum/src/encoding/fields.rs
@@ -20,7 +20,7 @@ use uuid::{self, Uuid};
 
 use std::mem;
 use std::result::Result as StdResult;
-use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 
 use crypto::{Hash, PublicKey, Signature};
 use helpers::{Height, Round, ValidatorId};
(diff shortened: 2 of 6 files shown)


exonum/src/encoding/fields.rs
```diff
@@ -304,43 +304,82 @@ impl<'a> Field<'a> for DateTime<Utc> {
...
+                octets.copy_from_slice(&buffer[addr_start..addr_start + ADDR_SIZE]);
+                IpAddr::V4(Ipv4Addr::from(octets))
+            }
+            IPV6_HEADER => {
+                const ADDR_SIZE: usize = mem::size_of::<Ipv6Addr>();
+                let mut octets: [u8; ADDR_SIZE] = mem::uninitialized();
+                octets.copy_from_slice(&buffer[addr_start..addr_start + ADDR_SIZE]);
+                IpAddr::V6(Ipv6Addr::from(octets))
+            }
+            header => panic!("Unknown header `{:X}` for SocketAddr", header),
```
vitvakatu: I wonder if we should mark that as `unreachable`. In theory, we always need to `check` before `read`.

andreyluchin: The motivation behind it was that it is still theoretically possible to reach that `match` branch. By panicking there instead of `unreachable!()` we don't get any extra overhead (besides binary size I guess), so I'd rather leave it there just in case.

exonum/src/encoding/fields.rs
```diff
@@ -304,43 +304,82 @@ impl<'a> Field<'a> for DateTime<Utc> {
     }
 }
 
+const HEADER_SIZE: usize = 1;
+const BODY_SIZE: usize = mem::size_of::<Ipv6Addr>();
```
vldm: `Ipv6Addr` is Rust structure which layout is unstable, so it's size probably can vary. Can you rewrite `BODY_SIZE` const to be just a 128 bit (or 16 bytes).

exonum/src/encoding/fields.rs
```diff
@@ -304,43 +304,82 @@ impl<'a> Field<'a> for DateTime<Utc> {
...
     unsafe fn read(buffer: &'a [u8], from: Offset, to: Offset) -> Self {
-        let mut octets = [0u8; 4];
-        octets.copy_from_slice(&buffer[from as usize..from as usize + 4]);
-        let ip = Ipv4Addr::from(octets);
-        let port = LittleEndian::read_u16(&buffer[from as usize + 4..to as usize]);
-        SocketAddr::V4(SocketAddrV4::new(ip, port))
+        let addr_start = from as usize + HEADER_SIZE;
+        let ip = match buffer[from as usize] {
+            IPV4_HEADER => {
+                const ADDR_SIZE: usize = mem::size_of::<Ipv4Addr>();
```
vldm: The same as above using `size_of` for rust struct is probably unsafe.

exonum/src/helpers/fabric/details.rs
```diff
@@ -335,22 +335,33 @@ impl GenerateNodeConfig {
...
 
     fn addr(context: &Context) -> (SocketAddr, SocketAddr) {
-        let addr = context.arg::<String>(PEER_ADDRESS).unwrap_or_default();
+        let addr_str = &context.arg::<String>(PEER_ADDRESS).unwrap_or_default();
 
-        let mut addr_parts = addr.split(':');
-        let ip = addr_parts.next().expect("Expected ip address");
-        if ip.len() < 8 {
-            panic!("Expected ip address in {}.", PEER_ADDRESS);
+        let maybe_addr = addr_str.parse::<SocketAddr>();
```
vldm: Can you remove duplication belowe, by using chain, like this:
```
let maybe_addr = addr_str.parse::<SocketAddr>()
                                         .unwrap_or_else( || {
     let ip = addr_str.parse::<IpAddr>().expect(error_msg)
     SocketAddr::new(ip, DEFAULT_EXONUM_LISTEN_PORT)
});
```

exonum/src/encoding/fields.rs
```diff
@@ -344,9 +344,7 @@ impl<'a> Field<'a> for SocketAddr {
         match *self {
             SocketAddr::V4(ref addr) => {
                 buffer[from as usize] = IPV4_HEADER;
-                let diff = (mem::size_of::<Ipv4Addr>() as isize
-                    - mem::size_of::<Ipv6Addr>() as isize)
-                    .abs() as usize;
+                let diff = (IPV4_SIZE as isize - IPV6_SIZE as isize).abs() as usize;
```
pavel-mukhanov: Maybe it's better to introduce a constant here?

exonum/src/encoding/fields.rs
```diff
@@ -20,12 +20,22 @@ use uuid::{self, Uuid};
...
 use crypto::{Hash, PublicKey, Signature};
 use helpers::{Height, Round, ValidatorId};
 use super::{CheckedOffset, Error, Offset, Result};
 
+const SOCKET_ADDR_HEADER_SIZE: usize = 1;
+const PORT_SIZE: usize = 2;
+
+const IPV4_SIZE: usize = 4;
+const IPV6_SIZE: usize = 16;
+const SIZE_DIFF: usize = 12;
```
slowli: Would it be possible to define this constant as `IPV6_SIZE - IPV4_SIZE`?

exonum/src/encoding/fields.rs
```diff
@@ -309,38 +319,66 @@ impl<'a> Field<'a> for DateTime<Utc> {
...
+        let port = LittleEndian::read_u16(&buffer[to as usize - PORT_SIZE..to as usize]);
+        SocketAddr::new(ip, port)
     }
 
     fn write(&self, buffer: &mut Vec<u8>, from: Offset, to: Offset) {
         match *self {
-            SocketAddr::V4(addr) => {
-                buffer[from as usize..to as usize - 2].copy_from_slice(&addr.ip().octets());
+            SocketAddr::V4(ref addr) => {
+                buffer[from as usize] = IPV4_HEADER;
```
slowli: Does Rust or the `write` method provides any guarantee as to the uninitialized values of bytes in `buffer` (as in this case, with 12 uninitialized bytes after the IPv4 address)? Could it be a source of malleability / failed round trip? Maybe, it would make sense that these padding bytes are equal to 0 i the `check` method?

exonum/tests/config.rs
```diff
@@ -200,14 +210,17 @@ fn test_generate_template() {
     }
 }
 
-#[test]
 #[cfg_attr(feature = "cargo-clippy", allow(needless_range_loop))]
```
slowli: Is this attribute really needed here? The `needless_range_loop` lint doesn't seem to be triggered by the test code. Btw, the legitimate triggering that occurs in the `finalize_config` function could be avoided by rewriting the `for` loop in it as

```rust
for &conf in PUB_CONFIG.iter().take(count) {
    override_validators_count(conf, count, folder);
    variables.push(full_tmp_name(conf, folder));
}
```

CHANGELOG.md
```diff
@@ -171,6 +171,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `transactions!` macro now allows empty body. (#593)
 
+- IPv6 addressing is now supported by exonum. (#615)
```
vitvakatu: `by exonum` is superfluous here, I think :upside_down_face:

andreyluchin:
#### More changes
As @slowli pointed out, 12 padding bytes in case of IPv4 are much better off being set to 0 during `write` and checked with `check` to be 0's.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Internal improvements
 
+- IPv6 addressing is now supported. (#615)
```
DarkEld3r: I would rather say that this is a "new feature". :upside_down_face:

Issue #606, opened 2018-04-11
Replace `SystemTime` with `chrono::DateTime`


DarkEld3r:
We have mostly switched to the `chrono::DateTime<Utc>` (#557), but `SystemTime` is still used in `/exonum/exonum/src/helpers/metrics.rs` module. This usage should be replaced too. See #514 for the example.

PR #620, opened 2018-04-14
Replace SystemTime::now() with Utc::now()

jihoonson:
Fixes https://github.com/exonum/exonum/issues/606. `Utc::now`() is used instead.


jihoonson:
@aleksuss @vitvakatu @DarkEld3r thank you for the review!

PR #635, opened 2018-04-17
Setup Appveyor [ECR-1090] 

vitvakatu:
Time limit for build is 1 hour in free version, and at the moment it takes almost 50 minutes for `cargo test --all`

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,6 +4,7 @@ install:
   - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe
   - rustup-init -yv --default-toolchain stable --default-host x86_64-pc-windows-msvc
   - set PATH=%PATH%;%USERPROFILE%\.cargo\bin
+  - set RUST_BACKTRACE=1
   - rustc -vV
   - cargo -vV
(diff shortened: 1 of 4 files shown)


.appveyor.yml
```diff
@@ -0,0 +1,17 @@
...
+  - rustc -vV
+  - cargo -vV
+
+build: false
+
+test_script:
+  - cargo test --all
+
+cache:
+  - '%USERPROFILE%\.cargo'
```
ozkriff:
```yaml
cache:
    - target -> Cargo.lock
    - C:\Users\appveyor\.cargo\registry -> Cargo.lock
```
i think it's better to use target file (some Cargo.toml in exonum's case?) and also cache `target` dir.

exonum-dictionary.txt
```diff
@@ -130,5 +130,6 @@ uuids
 validator
 validator's
 validators
+veyor
```
DarkEld3r: I propose to rename the badge title to "App**v**eyor Build Status" (it isn't visible anyway), so we wouldn't need to add such strange "word" to the exceptions.

PR #636, opened 2018-04-18
Seperate message

vldm:
Seperates messages from macro code. 
Creates structure `Message` that can encapsulate any message.

Update:
After some of update tests starting fail, need to fix before merge

TODO:
- [ ] Fix sandbox tests.
- [ ] Update services again.


exonum/src/messages/mod.rs
```diff
@@ -26,19 +22,20 @@ use crypto::PublicKey;
...
 
-// TODO: implement common methods for enum types (hash, raw, from_raw, verify)
-// TODO: use macro for implementing enums (ECR-166)
-
-/// Raw transaction type.
-pub type RawTransaction = RawMessage;
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct RawTransaction {
+    service_id: u16,
+    payload: Vec<u8>,
```
alekseysidorov: Why not `Cow`?

exonum/src/messages/raw.rs
```diff
@@ -14,294 +14,216 @@
...
+        let authorised_message = AuthorisedMessage::new(value, author)?;
+        let signature = Self::sign(authorised_message);
+
+        Ok(SignedMessage {
+            authorised_message,
+            signature,
+        })
+    }
+
+    pub(crate) fn verify_buffer(buffer: &[u8]) -> Result<SignedMessage, Error> {
```
alekseysidorov: Why not `from_buffer` or `from_buf`?

exonum/src/messages/raw.rs
```diff
@@ -14,294 +14,216 @@
...
+        Ok(message)
+    }
+
+    pub(crate) fn write_to_buffer(self, buf: &mut BytesMut) {
+        let full_buffer = bincode::serialize(&val, Bounded(MAX_MESSAGE_SIZE))?;
+        buf.write_u32::<LittleEndian>(buf.len() as u32);
+        buf.extend_from_slice(buf);
+    }
+
+    fn sign<T: Serialize>(val: &T) -> Signature {
```
alekseysidorov: Why you do not use `write_to_buffer` here?

exonum/src/messages/raw.rs
```diff
@@ -14,294 +14,216 @@
...
+        Self::verify::<AuthorisedMessage>(&message.authorised_message,
+                                          message.signature,
+                                          &message.authorised_message.author)?;
+        Ok(message)
+    }
+
+    pub(crate) fn write_to_buffer(self, buf: &mut BytesMut) {
+        let full_buffer = bincode::serialize(&val, Bounded(MAX_MESSAGE_SIZE))?;
+        buf.write_u32::<LittleEndian>(buf.len() as u32);
+        buf.extend_from_slice(buf);
```
alekseysidorov: I suppose that we should move message lenght to tokio codec level.

exonum/src/messages/authorisation.rs
```diff
@@ -0,0 +1,159 @@
...
+use messages::Message;
+use storage::StorageValue;
+
+use super::protocol::{Protocol, ProtocolMessage};
+use super::PROTOCOL_MAJOR_VERSION;
+
+use encoding::serialize::encode_hex;
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct AuthorisedMessage {
```
mvlabat: Better to name it as `AuthorizedMessage` and change everywhere else 's' to 'z' in this word )

mvlabat: ~~AuthorizedMezzage~~

exonum/examples/explorer.rs
```diff
@@ -30,11 +32,11 @@ use blockchain::{consensus_keys, create_block, create_blockchain, CreateWallet,
 mod blockchain;
 
 /// Creates a transaction for the mempool.
-pub fn mempool_transaction() -> Box<Transaction> {
+pub fn mempool_transaction() -> Message<RawTransaction> {
     // Must be deterministic, so we are using consensus keys, which are generated from
     // a passphrase.
     let (pk_alex, key_alex) = consensus_keys();
-    CreateWallet::new(&pk_alex, "Alex", &key_alex).into()
+    Message::sign_tx(CreateWallet::new(&pk_alex, "Alex"), 0, (pk_alex, &key_alex))
```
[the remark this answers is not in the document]
vldm: `CreateWallet::new(&pk, "Alex").sign(pk, &sk)` is a good variant, but then we need aditional extension trait. 

Magic number is for service_id, previously in this code transaction without service was created. (This probably not a good practice)

exonum/src/node/mod.rs
```diff
@@ -563,15 +570,15 @@ impl NodeHandler {
...
         let peers: Vec<SocketAddr> = self.state
             .peers()
             .values()
             .map(|conn| conn.addr())
             .collect();
-
+        let message = message.into();
         for address in peers {
-            self.send_to_addr(&address, message);
+            self.send_to_addr(&address, message.clone());
```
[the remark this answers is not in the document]
vldm: Good point, we can investigate degradation, and probably put `SignedMessage` into `Arc`.

exonum/src/blockchain/transaction.rs
```diff
@@ -49,10 +100,10 @@ pub type TransactionResult = Result<(), TransactionError>;
 /// See also [the documentation page on transactions][doc:transactions].
 ///
 /// [doc:transactions]: https://exonum.com/doc/architecture/transactions/
-pub trait Transaction: Message + ExonumJson + 'static {
-    /// Verifies the internal consistency of the transaction. `verify` should usually include
-    /// checking the message signature (via [`verify_signature`]) and, possibly,
-    /// other internal constraints. `verify` has no access to the blockchain state;
+pub trait Transaction: ::std::fmt::Debug + Send + 'static + ::erased_serde::Serialize {
```
andreyluchin: Unless `Debug` is directly required for `Transaction` trait, I'd prefer removing it to keep trait as small as possible. `Debug` implementations should be enforced by `#[deny(missing_debug_implementation)]`, not by user interface.

exonum/src/events/codec.rs
```diff
@@ -40,12 +39,14 @@ impl MessagesCodec {
...
+    type Error = Error;
 
-    fn decode(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, io::Error> {
-        // Read header
-        if buf.len() < HEADER_LENGTH {
+    fn decode(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Error> {
+        const U32_LENGTH: usize = 4;
+
+        // Read size
+        if buf.len() < U32_LENGTH {
```
andreyluchin: I think using `std::mem::size_of<u32>` is better, especially since it is used once in this function.

exonum/src/messages/authorization.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    fn hash(&self) -> Hash {
+        hash(&::bincode::config()
+            .no_limit()
+            .serialize(self)
+            .expect("Expected serialize to work"))
+    }
+
+    fn verify<T: Serialize>(
```
andreyluchin: I'm just curious, I thought `SignedMessage` guarantees that it is validly signed, am I wrong? I think it would make sense to disallow invalidly signed messages by verifying signature in the constructor.

exonum/src/messages/helpers.rs
```diff
@@ -0,0 +1,87 @@
...
+impl HexTransaction {
+    pub(crate) fn serialize<S>(message: &Message<RawTransaction>, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: Serializer,
+    {
+        let hex_string = message.to_hex_string();
+        hex_string.serialize(serializer)
+    }
+
+    pub(crate) fn deserialize<'a, D>(deserializer: D) -> Result<Message<RawTransaction>, D::Error>
```
andreyluchin: Why did you choose not to move these methods into serde traits implementation? They are `pub(crate)` anyway

exonum/src/messages/mod.rs
```diff
@@ -13,255 +13,209 @@
 // limitations under the License.
 
 //! Consensus and other messages and related utilities.
+use std::fmt;
```
andreyluchin: I would like to see the state machine described in module level documentation here.

exonum/src/messages/authorization.rs
andreyluchin: and, of course, removing this method

exonum/src/messages/helpers.rs
vldm: Because it's a wrapper between our encoding_struct and serde, later this trait should be removed.

exonum/src/messages/authorization.rs
vldm: Check this method code, it's aready internal part of `SignedMessage` constructor. 
This method used to abstract `SignedMessage` from `crypto` module, and verification specific code. 

BTW: Its a private static method.

exonum/src/blockchain/transaction.rs
vldm: We discussed this before, and choose to require `Debug` from user.
This will be used with `:erased_serde::Serialize` to give end-user info about transaction.

PR #641, opened 2018-04-18
Reduced amount of copy-paste in messages module

popzxc:
Currently every function in `messages` module contain `match` with the similar contents.
I tried to reduce amount of boilerplate code and created two macros to generate required methods.

Pros: 
* Less code that look the same
* Less size of the module
* Easier to perform changes in enum (one place instead of many)

Cons:
* Probably less readable
* Can't generate functions other than simple method call of inner message.


popzxc:
Renamed macros as suggested

DarkEld3r:
It seems that people don't like (such) macros. :upside_down_face:

PR #642, opened 2018-04-18
Reject processed transactions [ECR-1251]

popzxc:
With only one node used, it was possible to execute the same transaction multiple times.
To prevent this, additional check was added into NodeHandler::handle_tx.

However, client gets response with tx_hash even for rejected transaction (and this behavior persists even when 2+ nodes are used).
I see no simple way to change it since NodeHandler::handle_tx doesn't return Result and thus not able to tell that transaction was rejected.

### DIFF
--- a/exonum/src/node/consensus.rs
+++ b/exonum/src/node/consensus.rs
@@ -617,6 +617,16 @@ impl NodeHandler {
         let mut fork = self.blockchain.fork();
         {
             let mut schema = Schema::new(&mut fork);
+            if schema.transactions().contains(&hash)
+                && !schema.transactions_pool().contains(&hash)
+            {
+                error!(
+                    "Received already committed transaction, hash={:?}",
+                    hash
+                );
+                return;
+            }
+
             schema.add_transaction_into_pool(msg.raw().clone());
         }
         self.blockchain
--- a/exonum/src/node/consensus.rs
+++ b/exonum/src/node/consensus.rs
@@ -532,14 +532,14 @@ impl NodeHandler {
         }
     }
 
-    /// Checks if transaction is new and adds it into pool.
+    /// Checks if the transaction is new and adds it to the pool.
     fn handle_tx_inner(&mut self, msg: RawTransaction) -> Result<(), String> {
         let hash = msg.hash();
 
         profiler_span!("Make sure that it is new transaction", {
             let snapshot = self.blockchain.snapshot();
             if Schema::new(&snapshot).transactions().contains(&hash) {
-                let err = format!("Received already committed transaction, hash {:?}", hash);
+                let err = format!("Received already processed transaction, hash {:?}", hash);
                 return Err(err);
             }
         });
@@ -581,9 +581,10 @@ impl NodeHandler {
             }
         });
 
+        // We don't care about result, because situation when transaction received twice
+        // is normal for internal messages (transaction may be received from 2+ nodes).
         match self.handle_tx_inner(msg) {
-            Ok(_) => return,
-            Err(e) => error!("{}", e),
+            _ => return,
         }
     }
 
--- a/exonum/src/node/consensus.rs
+++ b/exonum/src/node/consensus.rs
@@ -617,13 +617,9 @@ impl NodeHandler {
         let mut fork = self.blockchain.fork();
         {
             let mut schema = Schema::new(&mut fork);
-            if schema.transactions().contains(&hash)
-                && !schema.transactions_pool().contains(&hash)
+            if schema.transactions().contains(&hash) && !schema.transactions_pool().contains(&hash)
             {
-                error!(
-                    "Received already committed transaction, hash={:?}",
-                    hash
-                );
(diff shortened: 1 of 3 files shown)


exonum/src/node/consensus.rs
```diff
@@ -617,6 +617,12 @@ impl NodeHandler {
         let mut fork = self.blockchain.fork();
         {
             let mut schema = Schema::new(&mut fork);
+            if schema.transactions().contains(&hash) && !schema.transactions_pool().contains(&hash)
```
vitvakatu: Shouldn't we change [that line](https://github.com/exonum/exonum/blob/master/exonum/src/node/consensus.rs#L554)?

vitvakatu: > `!schema.transactions_pool().contains(&hash)`

I don't think we need it here.

popzxc: Agreed. 
Also I want to note that this part was copy-pasted from `validate_block_transactions` method, so we should think about it too.

popzxc:
Sure, I'm already working on it.

popzxc:
I've added a new service that is able to test real API calls, and a corresponding test.
By default, timeouts in service are too big to be used in tests, so I used smaller values.
However, I'm not sure that chosen values are adequate, so please tell me if they should be changed.

exonum/src/node/consensus.rs
```diff
@@ -576,6 +559,35 @@ impl NodeHandler {
...
         }
+        Ok(())
+    }
+
+    /// Handles raw transaction. Transaction is ignored if it is already known, otherwise it is
+    /// added to the transactions pool.
+    #[cfg_attr(feature = "flame_profile", flame)]
+    pub fn handle_tx(&mut self, msg: RawTransaction) {
+        //trace!("Handle transaction");
+        let tx = {
```
slowli: Nitpick: braces seem to be superfluous in this case.

exonum/src/node/consensus.rs
```diff
@@ -576,6 +559,35 @@ impl NodeHandler {
...
             self.has_full_propose(hash, round);
         }
+        Ok(())
+    }
+
+    /// Handles raw transaction. Transaction is ignored if it is already known, otherwise it is
+    /// added to the transactions pool.
+    #[cfg_attr(feature = "flame_profile", flame)]
+    pub fn handle_tx(&mut self, msg: RawTransaction) {
+        //trace!("Handle transaction");
```
slowli: I think such commented code should be removed.

exonum/tests/balance_service.rs
```diff
@@ -0,0 +1,260 @@
...
+        fn state_hash(&self, _: &Snapshot) -> Vec<Hash> {
+            vec![]
+        }
+
+        fn tx_from_raw(&self, raw: RawTransaction) -> Result<Box<Transaction>, encoding::Error> {
+            let tx = BalanceTransactions::tx_from_raw(raw)?;
+            Ok(tx.into())
+        }
+
+        fn public_api_handler(&self, ctx: &ApiContext) -> Option<Box<Handler>> {
```
slowli: I'm not quite sure why this redefinition and `BalanceApi` are needed. As far as I see, the transactions in the test are sent directly through the MPSC channel, not HTTP API.

exonum/src/node/consensus.rs
popzxc: I didn't remove this on purpose. From my point of view it's kind of strange that handle_incoming_tx contains trace, and handle_tx has it commented.  Since they are similar, wouldn't it be more logical to either both have or both not have traces?

exonum/src/node/consensus.rs
popzxc: Sure, I'll remove it.

exonum/tests/balance_service.rs
popzxc: Seems like you're right. I'll remove it.

popzxc:
Regarding the main body of comment: I totally agree. I don't like non-deterministic multi-threaded test either. 
At first I've tried to create test based on sandbox actually. But currently seems like sandbox has more low-level infrastructure and operates mainly internal messages.
So, the question is: would such a test be consistent with others in sandbox?

When I started work on this test, I've asked team for guidance, and it was told that adaptation of Balance service is an acceptable option.

exonum/src/node/consensus.rs
```diff
@@ -532,35 +532,18 @@ impl NodeHandler {
...
-                    return;
-                }
-            }
-        };
 
         profiler_span!("Make sure that it is new transaction", {
             let snapshot = self.blockchain.snapshot();
             if Schema::new(&snapshot).transactions().contains(&hash) {
-                return;
+                let err = format!("Received already committed transaction, hash {:?}", hash);
```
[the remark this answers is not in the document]
popzxc: 1. Ok
2. For internal messages, I guess you're right. For public API calls, I think, duplicated requests are not OK. Maybe it will be better to leave this line as is, and just remove `error!()` call in `handle_tx()`?

PR #651, opened 2018-04-25
WIP: Fix handling of duplicated messages in propose [ECR-1247]

popzxc:
Currently if already known transaction is gotten in `State::add_propose`, call to `bail!` occurs.
It leads to discarding all others unknown transactions and thus nodes can't sync.
Result is infinite round timeout.

Changing `bail!` to `warn!` and simple ignoring of gotten transaction does fix the situation.

- [x] Fix error in `State::add_propose`.
- [ ] Write tests.

### DIFF
--- a/exonum/src/node/state.rs
+++ b/exonum/src/node/state.rs
@@ -816,10 +816,7 @@ impl State {
                 for hash in msg.transactions() {
                     if transactions.get(hash).is_some() {
                         if !transaction_pool.contains(hash) {
-                            bail!(
-                                "Received propose with already\
-                                 committed transaction"
-                            )
+                            warn!("Received propose with already committed transaction");
                         }
                     } else {
                         unknown_txs.insert(*hash);


exonum/src/node/state.rs
```diff
@@ -816,10 +816,7 @@ impl State {
                 for hash in msg.transactions() {
                     if transactions.get(hash).is_some() {
                         if !transaction_pool.contains(hash) {
-                            bail!(
-                                "Received propose with already\
-                                 committed transaction"
-                            )
+                            warn!("Received propose with already committed transaction");
```
vldm: Node should never send propose with transaction that committed before. This is byzantine behaviour. Tx order in `Block` is significant, we can't ignore part of transactions in `Propose`.

popzxc: That's exactly what I've asked today on the meeting about (probably in wrong words).
If this behavior is not acceptable, then I guess there is nothing to be fixed in this ticket.
Fix of ECR-1251 fixes this bug too:
Committed transaction can't get into pool -> committed transaction can't get into proposal.

So this PR and ticket may be closed, I guess.

PR #653, opened 2018-04-25
Implementation of `Field` trait for `chrono::Duration` structure [ECR-1273]

popzxc:


exonum/src/encoding/fields.rs
```diff
@@ -258,6 +258,64 @@ impl<'a> Field<'a> for DateTime<Utc> {
...
+        from: CheckedOffset,
+        to: CheckedOffset,
+        latest_segment: CheckedOffset,
+    ) -> Result {
+        debug_assert_eq!((to - from)?.unchecked_offset(), Self::field_size());
+        let from_unchecked = from.unchecked_offset() as usize;
+        let to_unchecked = to.unchecked_offset() as usize;
+
+        let secs = LittleEndian::read_i64(&buffer[from_unchecked..from_unchecked + 8]);
+        let nanos = LittleEndian::read_i32(&buffer[from_unchecked + 8..to_unchecked]);
```
slowli: Can `|nanos| > 10 ** 9` lead to some problems? For one, it seems that these values would fail a roundtrip from/to the serialized form. The same question about positive / negative `nanos` - they can lead to multiple presentations of the same duration (e.g., `4 s + 4e8 ns = 5 s - 6e8 ns`).

exonum/src/encoding/fields.rs
```diff
@@ -258,6 +258,64 @@ impl<'a> Field<'a> for DateTime<Utc> {
...
+
+    unsafe fn read(buffer: &'a [u8], from: Offset, to: Offset) -> Self {
+        let secs = LittleEndian::read_i64(&buffer[from as usize..from as usize + 8]);
+        let nanos = LittleEndian::read_i32(&buffer[from as usize + 8..to as usize]);
+
+        // Assuming that buffer was checked and Duration object can be constructed.
+        Duration::seconds(secs) + Duration::nanoseconds(i64::from(nanos))
+    }
+
+    fn write(&self, buffer: &mut Vec<u8>, from: Offset, to: Offset) {
```
slowli: Observation: This method seems to lead to a different presentation for negative durations than one used in `chrono` internally - it seems like `nanos` would be non-positive for these durations.

exonum/src/encoding/fields.rs
```diff
@@ -258,6 +258,64 @@ impl<'a> Field<'a> for DateTime<Utc> {
...
+        let secs = self.num_seconds();
+        let nanos_as_duration = *self - Duration::seconds(secs);
+        // Since we're working with only nanos, no overflow is expected here.
+        let nanos = nanos_as_duration.num_nanoseconds().unwrap() as i32;
+        LittleEndian::write_i64(
+            &mut buffer[from as usize..to as usize - mem::size_of_val(&nanos)],
```
slowli: Nitpick: I'd suggest to unify the use of slice indexes in all methods; `mem::size_of::<T>()` seems the best option to me.

exonum/src/encoding/fields.rs
vitvakatu: According to `chrono` sources negative nanos are impossible:
```rust
pub struct Duration {
    secs: i64,
    nanos: i32, // Always 0 <= nanos < NANOS_PER_SEC
}
```

slowli: Yeah, but I don't see how it is reflected in `read` or `write` (unless I'm mistaken, `write` actually produces non-positive `nanos` for negative durations; e.g., `-5.4s` would be written as `{ secs: -5, nanos: -400_000_000 }`).

popzxc: Well, you're right. But since |nanos| does not exceed 10 ** 9, I see no problem here:
object is splitted into seconds and nanoseconds, and then constructed back.
Handling of this process is up to Duration object, since it was decided to completely encapsulate internals (even `nanos_mod_sec` method is private).
The only solution I see is ещ completely repeat library actions, but this way is pretty much error-prone, becuase internals may change.
Should we really strictly follow internal implementation when authors don't want us to base on it?

slowli: I see a potential problem in that a single `Duration` can be represented in multiple ways (i.e., not survive a roundtrip); I've mentioned an example in another comment. If there was a well-specified canonical presentation of `Duration` enforced through `check` and adhered to by `write`, I would not care which one it was (e.g., non-positive `nanos` for negative `secs` would do just fine).

exonum/src/encoding/fields.rs
alekseysidorov: Due to rust abi is unstable `mem::size_of::<T>()` can be different on the various rust versions.

exonum/src/storage/values.rs
```diff
@@ -282,6 +283,25 @@ impl StorageValue for DateTime<Utc> {
...
+impl StorageValue for Duration {
+    fn into_bytes(self) -> Vec<u8> {
+        let mut buffer = vec![0; Duration::field_size() as usize];
+        let from: Offset = 0;
+        let to: Offset = Duration::field_size();
+        self.write(&mut buffer, from, to);
+        buffer
+    }
+
+    fn from_bytes(value: Cow<[u8]>) -> Self {
```
vitvakatu: Hmm... I think we need to reuse `Field::read` method here, don't we?

exonum/src/storage/values.rs
```diff
@@ -427,6 +447,22 @@ mod tests {
         }
     }
 
+    #[test]
+    fn storage_value_for_duration_round_trip() {
+        let durations = [
```
vitvakatu: What about negative durations?

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - IPv6 addressing is now supported. (#615)
 
+- `Field` trait implemented for `chrono::Duration` structure. (#653)
```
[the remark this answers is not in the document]
vitvakatu: Also it's worth mentioning that `StorageValue` has also been implemented.

exonum/src/storage/values.rs
```diff
@@ -282,6 +283,25 @@ impl StorageValue for DateTime<Utc> {
     }
 }
 
+/// Uses little-endian encoding.
+impl StorageValue for Duration {
```
vitvakatu: You need to check whether it actually possible to use `Duration` inside `encoding_struct` - I think we also need `ExonumJson` here?

exonum/src/encoding/fields.rs
```diff
@@ -239,25 +239,117 @@ impl<'a> Field<'a> for DateTime<Utc> {
...
+fn is_duration_representation_valid(secs: i64, nanos: i32) -> bool {
+    // Signs are checked to avoid multiple representations for same duration.
+    // Example: 4 s + 4e8 ns = 5 s - 6e8 ns.
+    if (secs < 0 && nanos > 0) || (secs > 0 && nanos < 0) {
+        return false;
+    }
+
+    // Absolute value of nanoseconds must less than 10 ** 9.
+    let nanos_per_sec = 1_000_000_000;
+    if nanos <= -nanos_per_sec || nanos >= nanos_per_sec {
```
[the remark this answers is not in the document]
popzxc: Yeah, I checked it at first. But I didn't like the following line:
>The absolute value of i32::min_value() cannot be represented as an i32, and attempting to calculate it will cause an overflow. This means that code in debug mode will trigger a panic on this case.
Since we are checking here that nanos are in acceptable range, we must think of I32_MIN value too.

`checked_abs` returns an Option, and use of it will lead to writing even more code.

Implementing an abs function that returns u32 for just one use seems overkill for me.

exonum/src/storage/values.rs
popzxc: Duration::min_value() is negative.

exonum/src/storage/values.rs
vitvakatu: We assume that data in storage is always valid, so unsafe should be fine here. @DarkEld3r requesting your opinion

DarkEld3r: I think it is OK for now. We are going to the revise unsafe code after changes in serialization anyway.

Issue #637, opened 2018-04-18
Rename methods in `consensus.rs`


popzxc:
Currently [consensus.rs](https://github.com/exonum/exonum/blob/master/exonum/src/node/consensus.rs) contains methods with non-intuitive names:

* `has_majority_prevotes`
* `has_majority_precommits`
* `has_full_propose`

It will be better to rename them to more consistent names, e.g. `handle_majority_prevotes`, `handle_majority_precommits`, `has_full_propose`.

DarkEld3r:
Closed by #655.

PR #655, opened 2018-04-26
Rename has_* methods in consensus.rs, update sandbox and comments (#637)

zeroqn:
Rename has_full_propose to handle_full_propose.
Rename has_majority_prevotes to handle_majority_prevotes.
Rename has_majority_precommits to handle_majority_precommits.

### DIFF
--- a/exonum/src/node/consensus.rs
+++ b/exonum/src/node/consensus.rs
@@ -148,7 +148,7 @@ impl NodeHandler {
                 self.request(RequestData::Transactions(hash), node);
             }
         } else {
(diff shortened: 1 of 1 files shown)


zeroqn:
I split the commit, now it only contains changes in node/consensus.rs. Should I create a new issue for sandbox/consensus.rs?

aleksuss:
I suppose it would be ok if you will present this changes in one commit. @DarkEld3r what do you think ?

vitvakatu:
We're using squash & merge strategy, so the number of commits isn't important at all

PR #657, opened 2018-04-27
Increased time of awaiting in balance service test

popzxc:


popzxc:
To check that current values of timeout/await are better, I've launched this test ~200 times with script and got no failures.

PR #661, opened 2018-05-02
Add VoteAgainst to configuration service

### LINKED ISSUE #634: Add "VoteAgainst" transaction to the configuration service
It is possible to [vote](https://github.com/exonum/exonum/blob/master/services/configuration/src/transactions.rs#L63) _for_ some configuration, but it isn't possible to vote _against_ it. `VoteAgainst` transaction should be added.

zeroqn:
MaybeVote is converted to generic struct. The rest of implementation is straight forward.

closes #634

### DIFF
--- a/services/configuration/src/schema.rs
+++ b/services/configuration/src/schema.rs
@@ -125,6 +125,18 @@ impl From<MaybeVote> for Option<VotingDecision> {
     }
 }
 
+impl From<Vote> for MaybeVote {
+    fn from(vote: Vote) -> MaybeVote {
+        MaybeVote::some(VotingDecision::Vote(vote))
+    }
+}
+
+impl From<VoteAgainst> for MaybeVote {
+    fn from(vote_against: VoteAgainst) -> MaybeVote {
+        MaybeVote::some(VotingDecision::VoteAgainst(vote_against))
+    }
+}
+
 impl ::std::ops::Deref for MaybeVote {
     type Target = Option<VotingDecision>;
 
@@ -267,12 +279,9 @@ mod tests {
         let vote = Vote::new(&pubkey, &Hash::new([1; 32]), &key);
         assert_eq!(
             vote.clone().into_bytes(),
-            MaybeVote::some(VotingDecision::Vote(vote.clone())).into_bytes()
+            MaybeVote::from(vote.clone()).into_bytes()
         );
-        assert_eq!(
-            vote.hash(),
-            MaybeVote::some(VotingDecision::Vote(vote.clone())).hash()
-        );
+        assert_eq!(vote.hash(), MaybeVote::from(vote.clone()).hash());
 
         let db = MemoryDB::new();
         let mut fork = db.fork();
@@ -292,7 +301,7 @@ mod tests {
             assert_eq!(
                 stored_vote,
                 if i == 1 {
-                    MaybeVote::some(VotingDecision::Vote(vote.clone()))
+                    MaybeVote::from(vote.clone())
                 } else {
                     MaybeVote::none()
                 }
@@ -303,7 +312,7 @@ mod tests {
         let new_merkle_root = {
             let mut fork = db.fork();
             let mut index: ProofListIndex<_, MaybeVote> = ProofListIndex::new("index", &mut fork);
-            index.set(2, MaybeVote::some(VotingDecision::Vote(vote.clone())));
+            index.set(2, MaybeVote::from(vote.clone()));
             index.set(2, MaybeVote::none());
             index.merkle_root()
         };
--- a/services/configuration/src/schema.rs
+++ b/services/configuration/src/schema.rs
@@ -60,9 +60,9 @@ lazy_static! {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(tag = "vote_for", rename_all = "lowercase")]
 pub enum VotingDecision {
-    /// `Vote` transaction variant
+    /// `Vote` transaction variant.
     Yea(Vote),
-    /// `VoteAgainst` transaction variant
+    /// `VoteAgainst` transaction variant.
     Nay(VoteAgainst),
 }
 
--- a/services/configuration/src/schema.rs
+++ b/services/configuration/src/schema.rs
@@ -57,7 +57,7 @@ lazy_static! {
 
 /// A enum used to represent different kinds of vote, `Vote` and `VoteAgainst` transactions.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-#[serde(tag = "vote_for")]
+#[serde(tag = "vote_for", rename_all = "lowercase")]
 pub enum VotingDecision {
     /// `Vote` transaction variant
     Yea(Vote),
--- a/services/configuration/src/schema.rs
+++ b/services/configuration/src/schema.rs
@@ -91,11 +91,12 @@ impl StorageValue for VotingDecision {
     }
 }
 
-/// A functional equivalent to `Option<T>` used to store votes in the service schema.
+/// A functional equivalent to `Option<VotingDecision>` used to store votes in the service
+/// schema.
 ///
 /// # Notes
 ///
-/// The `None` variant of the type is represented by a `T` with all bytes set to zero.
+/// The `None` variant of the type is represented by a `Vote` with all bytes set to zero.
 #[derive(Clone, Debug, PartialEq)]
 pub struct MaybeVote(Option<VotingDecision>);
 
--- a/services/configuration/src/schema.rs
+++ b/services/configuration/src/schema.rs
@@ -323,8 +323,6 @@ mod tests {
 
     #[test]
     fn test_serialization_of_voting_decision() {
-        const VALIDATORS: usize = 5;
-
         let (pubkey, key) = crypto::gen_keypair();
         let vote = Vote::new(&pubkey, &Hash::new([1; 32]), &key);
         let vote_against = VoteAgainst::new(&pubkey, &Hash::new([1; 32]), &key);
@@ -341,34 +339,5 @@ mod tests {
             vote_against.hash(),
             VotingDecision::Nay(vote_against.clone()).hash()
         );
-
-        let db = MemoryDB::new();
-        let mut fork = db.fork();
-        let merkle_root = {
-            let mut index: ProofListIndex<_, VotingDecision> =
-                ProofListIndex::new("index", &mut fork);
-            for _ in 0..VALIDATORS {
-                index.push(VotingDecision::Yea(NO_VOTE.clone()));
-            }
-            index.set(1, VotingDecision::Yea(vote.clone()));
-            index.set(2, VotingDecision::Nay(vote_against.clone()));
-            index.merkle_root()
-        };
-        db.merge(fork.into_patch()).unwrap();
-
-        let snapshot = db.snapshot();
-        let index: ProofListIndex<_, MaybeVote> = ProofListIndex::new("index", &snapshot);
-        assert_eq!(index.get(1).unwrap(), MaybeVote::from(vote.clone()));
-        assert_eq!(index.get(2).unwrap(), MaybeVote::from(vote_against.clone()));
-
-        let new_merkle_root = {
-            let mut fork = db.fork();
-            let mut index: ProofListIndex<_, VotingDecision> =
-                ProofListIndex::new("index", &mut fork);
-            index.set(3, VotingDecision::Nay(vote_against.clone()));
-            index.set(3, VotingDecision::Yea(NO_VOTE.clone()));
-            index.merkle_root()
-        };
-        assert_eq!(merkle_root, new_merkle_root);
     }
 }
--- a/services/configuration/src/schema.rs
+++ b/services/configuration/src/schema.rs
@@ -19,7 +19,7 @@ use exonum::storage::{Fork, ProofListIndex, ProofMapIndex, Snapshot, StorageValu
 
 use std::borrow::Cow;
 
-use transactions::{Propose, Vote};
+use transactions::{Propose, Vote, VoteAgainst};
 
 // Defines `&str` constants with given name and value.
 macro_rules! define_names {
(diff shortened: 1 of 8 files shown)


slowli:
It's subjective, but I'd prefer to store `Vote`s and `VoteAgainst`s for a specific proposal in a single index, e.g., by introducing an enum for voting outcomes:

```rust
pub enum VotingDecision {
    Vote(Vote),
    VoteAgainst(VoteAgainst),
}
// MaybeVote ~ Option<VoteDecision>
```

This would ensure by design that there cannot be a `Vote` and `VoteAgainst` by a single validator.

The described change could be backward-compatible: as `VoteAgainst` has a `message_id` differing from `Vote`, it could be discriminated by it.

zeroqn:
I agree that using same index is better. I will try.

zeroqn:
Update the implementation. VoteAgainst and Vote transaction are now saved into same ProofListIndex.

services/configuration/src/schema.rs
```diff
@@ -52,15 +52,70 @@ lazy_static! {
...
         &Signature::zero(),
     ).into_bytes();
+    static ref VOTE_DECISION_BYTES_MARKER: Vec<u8> = b"consent".to_vec();
+    static ref VOTE_AGAINST_DECISION_BYTES_MARKER: Vec<u8> = b"dissent".to_vec();
 }
 
-/// A functional equivalent to `Option<Vote>` used to store votes in the service schema.
+/// A enum used to represent different kinds of vote, `Vote` and `VoteAgainst` transactions.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum VotingDecision {
```
slowli: To make the change backward-compatible, it would be necessary to change [the enum representation by `serde`](https://serde.rs/enum-representations.html). Although untagged technically can work, I suggest using internal tagging, i.e., `#[serde(tag = "type")]`. Also, I'm not quite sure about variant names: should they be renamed, and if so, how. A possible suggestion is `"yea" | "nay"` (used in parliaments in English-speaking countries; see [Wikipedia](https://en.wikipedia.org/wiki/Division_of_the_assembly)).

services/configuration/src/schema.rs
```diff
@@ -52,15 +52,70 @@ lazy_static! {
...
+        VOTE_AGAINST_DECISION_BYTES_MARKER.clone()
+    }
+}
+
+/// A functional equivalent to `Option<T>` used to store votes in the service schema.
 ///
 /// # Notes
 ///
-/// The `None` variant of the type is represented by a `Vote` with all bytes set to zero.
+/// The `None` variant of the type is represented by a `T` with all bytes set to zero.
```
slowli: This change is incorrect; the previous statement ("...by a `Vote`...") remains correct.

services/configuration/src/tests/api.rs
```diff
@@ -286,6 +293,49 @@ fn test_votes_for_propose() {
...
+        .collect::<Vec<_>>();
+    testkit.create_block_with_transactions(tx_dissenting_votes);
+    let response = api.votes_for_propose(&new_cfg.hash())
+        .expect("Dissenting votes for config is absent");
+    for entry in response.into_iter().take(testkit.majority_count()) {
+        let tx = entry.expect("VoteAgainst for config is absent");
+        assert!(
+            Schema::new(&testkit.snapshot())
+                .transactions()
+                .contains(&tx.hash(),),
```
slowli: Nitpick: comma `,` is obsolete here.

services/configuration/src/tests/api.rs
```diff
@@ -286,6 +293,49 @@ fn test_votes_for_propose() {
...
+        .validators()
+        .iter()
+        .map(|validator| new_tx_config_vote_against(validator, cfg_proposal_hash))
+        .map(to_boxed)
+        .collect::<Vec<_>>();
+    testkit.create_block_with_transactions(tx_dissenting_votes);
+    let response = api.votes_for_propose(&new_cfg.hash())
+        .expect("Dissenting votes for config is absent");
+    for entry in response.into_iter().take(testkit.majority_count()) {
+        let tx = entry.expect("VoteAgainst for config is absent");
```
slowli: I'd suggest to check that `entry` corresponds to the `VoteAgainst` variant.

services/configuration/src/tests/api.rs
```diff
@@ -463,3 +513,25 @@ fn test_post_vote_tx() {
...
+    let api = testkit.api();
+    // Commits the following configuration.
+    let new_cfg = {
+        let mut cfg = testkit.configuration_change_proposal();
+        cfg.set_actual_from(Height(10));
+        cfg.set_service_config("message", "First config change");
+        cfg.stored_configuration().clone()
+    };
+    let tx = new_tx_config_propose(&testkit.network().validators()[0], new_cfg.clone());
+    testkit.create_block_with_transactions(txvec![tx]);
```
slowli: Nitpick: it's possible to use `create_block_with_transaction(tx)` here.

services/configuration/src/transactions.rs
```diff
@@ -223,12 +247,18 @@ impl Transaction for Propose {
     }
 }
 
-impl Vote {
-    /// Checks context-dependent conditions for a `Vote` transaction.
+trait TxVoteUtil: Clone {
```
slowli: IMO, it could make more sense to define an enum like this

```rust
enum VotingDecisionRef<'a> {
    Vote(&'a Vote),
    VoteAgainst(&'a VoteAgainst),
}
```

and implement methods for it:

```rust
impl<'a> VotingDecisionRef<'a> {
    fn sender(&self) -> &PublicKey {
        match *self {
            VotingDecisionRef::Vote(vote) => vote.from(),
            VotingDecisionRef::VoteAgainst(vote) => vote.from(),
        }
    }

    fn cfg_hash(&self) -> &Hash { /* ... */ }

    fn precheck(&self, snapshot: &Snapshot) -> Result<StoredConfiguration, ServiceError>
    { /* ... */ }

    fn save(&self, fork: &mut Fork) { /* ... */ }
}

impl<'a> From<&'a Vote> VotingDecisionRef<'a> { /* trivial */ }
impl<'a> From<&'a VoteAgainst> VotingDecisionRef<'a> { /* trivial */ }
```

Then it could be used in the transaction code as follows:

```rust
let vote = VotingDecisionRef::from(self);
vote.precheck(fork.as_ref()).map_err(/* ... */)?;
vote.save(fork);
// ...
```

services/configuration/src/transactions.rs
```diff
@@ -321,3 +369,38 @@ impl Transaction for Vote {
...
+
+    fn execute(&self, fork: &mut Fork) -> ExecutionResult {
+        let _parsed_config = self.precheck(fork.as_ref()).map_err(|err| {
+            error!("Discarding dissenting vote {:?}: {}", self, err);
+            err
+        })?;
+
+        self.save(fork);
+        trace!(
+            "Put VoteAgainst:{:?} to corresponding cfg dissenting_votes_by_config_hash table",
```
slowli: Nitpick: the text is no longer accurate.

services/configuration/src/schema.rs
```diff
@@ -220,19 +288,36 @@ mod tests {
         let (pubkey, key) = crypto::gen_keypair();
```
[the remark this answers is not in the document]
zeroqn: Done.

services/configuration/src/schema.rs
```diff
@@ -100,16 +163,21 @@ impl CryptoHash for MaybeVote {
...
+            Some(VotingDecision::VoteAgainst(vote_against)) => vote_against.into_decision_bytes(),
             None => NO_VOTE_BYTES.clone(),
         }
     }
 
     fn from_bytes(bytes: Cow<[u8]>) -> Self {
         if NO_VOTE_BYTES.as_slice().eq(bytes.as_ref()) {
             MaybeVote::none()
+        } else if Vote::has_decision_marker(&bytes) {
+            MaybeVote::some(VotingDecision::Vote(Vote::from_decision_bytes(bytes)))
```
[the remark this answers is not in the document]
zeroqn: Done.

services/configuration/src/schema.rs
zeroqn: if we change variant names to "yea" and "nay", maybe we can use "vote_for" for tag name?
Or use "consent" and "dissent"?

services/configuration/src/schema.rs
zeroqn: Done.

services/configuration/src/tests/api.rs
zeroqn: Done

services/configuration/src/tests/api.rs
zeroqn: I saw that create_block_with_transactions() is used everywhere in this file, so want to be consistent with rest of code.

services/configuration/src/transactions.rs
zeroqn: Done

services/configuration/src/transactions.rs
zeroqn: Agree, this is better.

Done.

services/configuration/src/tests/api.rs
zeroqn: Done

services/configuration/src/tests/api.rs
zeroqn: Done

services/configuration/src/schema.rs
zeroqn: Rename to Yea and Nay, use "vote_for" for internal tag.

Done

zeroqn:
All done.

services/configuration/src/transactions.rs
```diff
@@ -321,3 +388,25 @@ impl Transaction for Vote {
...
 }
+
+impl Transaction for VoteAgainst {
+    fn verify(&self) -> bool {
+        self.verify_signature(self.from())
+    }
+
+    fn execute(&self, fork: &mut Fork) -> ExecutionResult {
+        let vote_against = VotingDecisionRef::from(self);
+        let _parsed_config = vote_against.precheck(fork.as_ref()).map_err(|err| {
```
slowli: Nitpick: you can just remove `let _parsed_config = ` part.

services/configuration/src/schema.rs
```diff
@@ -254,10 +314,64 @@ mod tests {
...
+        );
+        assert_eq!(
+            vote_against.hash(),
+            VotingDecision::Nay(vote_against.clone()).hash()
+        );
+
+        let db = MemoryDB::new();
+        let mut fork = db.fork();
+        let merkle_root = {
+            let mut index: ProofListIndex<_, VotingDecision> =
```
slowli: I don't quite understand the reasoning to test `ProofListIndex<_, VotingDecision>`, given that there are no such indexes in the schema (but `ProofListIndex<_, MaybeVote>` are). In general, this part of the test in the current form may be redundant: unlike the previous test, which tests compatibility between 2 different implementations of the schema, this one actually doesn't.

It still may make sense to test compatibility by comparing `ProofListIndex<_, Vote>` (which is equivalent to the previous `MaybeVote` implementation if some elements are set to `NO_VOTE`) and `ProofListIndex<_, MaybeVote>` (i.e., the new implementation). It should be possible to read from `ProofListIndex<_, MaybeVote>`, which was written as `ProofListIndex<_, Vote>`, and this conversion should not change the `merkle_root` of the index.

services/configuration/src/schema.rs
```diff
@@ -54,13 +55,51 @@ lazy_static! {
...
+impl StorageValue for VotingDecision {
+    fn into_bytes(self) -> Vec<u8> {
+        match self {
+            VotingDecision::Yea(vote) => vote.into_bytes(),
+            VotingDecision::Nay(vote_against) => vote_against.into_bytes(),
+        }
+    }
+
+    fn from_bytes(bytes: Cow<[u8]>) -> Self {
+        let raw_msg = RawMessage::from_vec(bytes.to_vec());
```
slowli: I don't quite like the cloning here, but it seems that there is no way around. cc @DarkEld3r, @vldm - maybe, there is?

services/configuration/src/schema.rs
```diff
@@ -54,13 +55,51 @@ lazy_static! {
     ).into_bytes();
 }
 
-/// A functional equivalent to `Option<Vote>` used to store votes in the service schema.
+/// A enum used to represent different kinds of vote, `Vote` and `VoteAgainst` transactions.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(tag = "vote_for")]
```
slowli: I'd also add [`#[serde(rename_all = "lowercase")]` attribute](https://serde.rs/container-attrs.html#serderenameall--); lower-case discriminators are more common in JSON.

services/configuration/src/tests/api.rs
```diff
@@ -286,6 +293,55 @@ fn test_votes_for_propose() {
...
+    testkit.create_block_with_transactions(tx_dissenting_votes);
+    let vote_against_variant = discriminant(&VotingDecision::Nay(new_tx_config_vote_against(
+        &testkit.network().validators()[0],
+        cfg_proposal_hash,
+    )));
+    let response = api.votes_for_propose(&new_cfg.hash())
+        .expect("Dissenting votes for config is absent");
+    for entry in response.into_iter().take(testkit.majority_count()) {
+        let tx = entry.expect("VoteAgainst for config is absent");
+        assert_eq!(
```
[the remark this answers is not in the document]
zeroqn: Done

services/configuration/src/transactions.rs
```diff
@@ -71,6 +71,30 @@ transactions! {
             /// See [crate docs](index.html) for more details on how the hash is calculated.
             cfg_hash: &Hash,
         }
+
+        /// VoteAgainst for the new configuration.
+        ///
+        /// # Notes
+        ///
+        /// The stored version of the transaction has a special variant corresponding to absence
+        /// of a vote. See [f`MaybeVote`] for details.
```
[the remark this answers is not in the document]
zeroqn: Done

services/configuration/src/transactions.rs
zeroqn: Done

services/configuration/src/schema.rs
slowli: You don't need to push `VoteAgainst` into the index - `VoteAgainst` was absent before this PR. The idea is to check that functionality that *was* present (i.e., `Vote` / `NO_VOTE`) hasn't been broken by this PR.

zeroqn: If so, that is already covered by test_serialization_of_maybe_vote(). I think ProofListIndex test can be removed.

services/configuration/src/schema.rs
```diff
@@ -69,21 +108,41 @@ impl MaybeVote {
...
-impl From<MaybeVote> for Option<Vote> {
-    fn from(value: MaybeVote) -> Option<Vote> {
-        value.0
+impl From<VoteAgainst> for MaybeVote {
+    fn from(vote_against: VoteAgainst) -> MaybeVote {
+        MaybeVote::some(VotingDecision::Nay(vote_against))
     }
 }
 
 impl ::std::ops::Deref for MaybeVote {
```
vitvakatu: nitpick: let's `use std::ops::Deref` and avoid `::` at the beginning :upside_down_face: 
It might be considered unnecessary requirement, but our code style policy is pretty strict

services/configuration/src/tests/api.rs
```diff
@@ -286,6 +293,50 @@ fn test_votes_for_propose() {
...
 }
 
+#[test]
+fn test_dissenting_votes_for_propose() {
+    use schema::VotingDecision;
+
+    let mut testkit: TestKit = TestKit::configuration_default();
+    let api = testkit.api();
+    // Apply a new configuration.
+    let cfg_change_height = Height(5);
```
vitvakatu: nitpick: I don't think it's necessary to use additional `cfg_change_height` variable here - you can use `set_actual_from(Height(5))`

services/configuration/src/tests/mod.rs
```diff
@@ -332,8 +338,32 @@ fn test_discard_vote_for_absent_propose() {
...
-    assert!(!votes.contains(&Some(illegal_vote)));
+    assert!(votes.contains(&Some(VotingDecision::Yea(legal_vote))));
+    assert!(!votes.contains(&Some(VotingDecision::Yea(illegal_vote))));
+}
+
+#[test]
+fn test_vote_against_for_propose() {
+    let mut testkit: TestKit = TestKit::configuration_default();
+
+    let cfg_change_height = Height(5);
```
vitvakatu: Same thing here

services/configuration/src/schema.rs
zeroqn: Done

services/configuration/src/tests/api.rs
zeroqn: Done

services/configuration/src/tests/mod.rs
zeroqn: Done

services/configuration/src/schema.rs
```diff
@@ -54,13 +56,51 @@ lazy_static! {
     ).into_bytes();
 }
 
-/// A functional equivalent to `Option<Vote>` used to store votes in the service schema.
+/// A enum used to represent different kinds of vote, `Vote` and `VoteAgainst` transactions.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(tag = "vote_for", rename_all = "lowercase")]
+pub enum VotingDecision {
+    /// `Vote` transaction variant
```
DarkEld3r: Please add a dot at the end of the comment.

services/configuration/src/schema.rs
```diff
@@ -54,13 +56,51 @@ lazy_static! {
...
+    Yea(Vote),
+    /// `VoteAgainst` transaction variant
```
DarkEld3r: And here.

services/configuration/src/schema.rs
DarkEld3r: I don't have any better ideas, either. 🙃

services/configuration/src/schema.rs
zeroqn: Done

services/configuration/src/schema.rs
zeroqn: Done

PR #662, opened 2018-05-02
Set fixed cspell version

vldm:


slowli:
Suggestion: maybe, use `package.json` / `package-lock.json` to specify `cspell` and `markdownlint` versions, like in the [`exonum-doc` repo](https://github.com/exonum/exonum-doc/blob/master/package.json)?

PR #664, opened 2018-05-03
Decompose `BlockResponse` [ECR-46]

SerhiiOryshych:
TODO:
- [x] Add sandbox test.
- [x] With invalid `precommits`
- [x] With invalid `tx_order`


exonum/src/node/consensus.rs
```diff
@@ -556,9 +550,16 @@ impl NodeHandler {
...
         // Go to handle full propose if we get last transaction
         for (hash, round) in full_proposes {
-            self.remove_request(&RequestData::Transactions(hash));
+            self.remove_request(&RequestData::TransactionsForPropose(hash));
             self.handle_full_propose(hash, round);
         }
+
+        let full_blocks = self.state.check_incomplete_blocks(hash);
+        // Go to handle full block if we get last transaction
+        for (hash, block) in full_blocks {
```
slowli: I'm *so* confused by this code. Can there be multiple blocks committed here? If yes, how does this work? If no, why use a cycle?

exonum/src/node/consensus.rs
```diff
@@ -770,6 +771,22 @@ impl NodeHandler {
                     ).raw()
                         .clone()
                 }
+                RequestData::TransactionsForBlock(ref block_hash) => {
+                    let txs: Vec<_> = self.state
+                        .incomplete_block(block_hash)
+                        .unwrap()
+                        .unknown_txs()
```
slowli: What's the guarantee that transactions are still unknown at this point? (I understand that this question pertains to `ProposeState` as well.)

exonum/src/node/state.rs
```diff
@@ -690,6 +722,11 @@ impl State {
         self.blocks.get(hash)
     }
 
+    /// Returns incomplete block with the specified hash.
+    pub fn incomplete_block(&self, hash: &Hash) -> Option<&IncompleteBlock> {
```
slowli: Can there be several incomplete blocks? Do they correspond to different heights? (The answer seems to be no, given that the `incomplete_blocks` container is cleared on transition to a new height.) I thought that an incomplete block is pushed into the structure only if its `BlockResponse` is verified to be authorized by a supermajority of validators; i.e., multiple blocks indicate Byzantine collusion.

exonum/src/node/state.rs
```diff
@@ -752,6 +790,25 @@ impl State {
...
     }
 
+    /// Checks whether some blocks are waiting for this transaction.
+    /// Returns a list of blocks that don't contain unknown transactions.
+    ///
+    /// Transaction is ignored if the following criteria are fulfilled:
+    ///
+    /// - transaction isn't contained in unknown transaction list of any blocks
+    /// - transaction isn't a part of block
+    pub fn check_incomplete_blocks(&mut self, tx_hash: Hash) -> Vec<(Hash, IncompleteBlock)> {
```
slowli: Method name is misleading, as it not only checks, but also mutates the state. I'd suggest a name like `remove_unknown_transaction`.

exonum/src/node/state.rs
```diff
@@ -863,6 +920,38 @@ impl State {
         }
     }
 
+    /// Finds unknown transactions in the block.
+    pub fn unknown_block_txs(
```
slowli: The method name is misleading, as it mutates the state. I'd suggest something like `create_incomplete_block`.

exonum/src/node/state.rs
```diff
@@ -863,6 +920,38 @@ impl State {
...
+        txs: &MapIndex<&&Snapshot, Hash, RawMessage>,
+        txs_pool: &KeySetIndex<&&Snapshot, Hash>,
+    ) -> Result<&IncompleteBlock, failure::Error> {
+        match self.incomplete_blocks.entry(msg.block().hash()) {
+            Entry::Occupied(..) => bail!("Block already found"),
+            Entry::Vacant(e) => {
+                let mut unknown_txs = HashSet::new();
+                for hash in msg.transactions() {
+                    if txs.get(hash).is_some() {
+                        if !txs_pool.contains(hash) {
```
slowli: I don't quite understand the reasoning to use `bail` here. If the `BlockResponse` is authenticated properly (which seems to be the case when the method is called from `handle_block`), then this should be a fundamental error.

exonum/src/node/state.rs
```diff
@@ -754,6 +797,23 @@ impl State {
         full_proposes
     }
 
+    /// Checks whether some blocks are waiting for this transaction.
```
slowli: This comment seemingly doesn't correspond to the method.

exonum/src/node/state.rs
```diff
@@ -754,6 +797,23 @@ impl State {
...
+    ///
+    /// Transaction is ignored if the following criteria are fulfilled:
+    ///
+    /// - transaction isn't contained in unknown transaction list of any blocks
+    /// - transaction isn't a part of block
+    pub fn remove_unknown_transaction(&mut self, tx_hash: Hash) -> Option<IncompleteBlock> {
+        if let Some(ref mut incomplete_block) = self.incomplete_block {
+            incomplete_block.unknown_txs.remove(&tx_hash);
+            if incomplete_block.unknown_txs.is_empty() {
+                return Some(incomplete_block.clone());
```
slowli: Is cloning really required here? That is, can `self.incomplete_block` be used by other code after it is returned from this method?

exonum/src/node/state.rs
```diff
@@ -865,6 +925,43 @@ impl State {
...
     }
 
+    /// Finds unknown transactions in the block.
+    pub fn create_incomplete_block(
+        &mut self,
+        msg: &BlockResponse,
+        txs: &MapIndex<&&Snapshot, Hash, RawMessage>,
+        txs_pool: &KeySetIndex<&&Snapshot, Hash>,
+    ) -> Result<bool, failure::Error> {
+        if let Some(ref incomplete_block) = self.incomplete_block {
```
slowli: Could it make sense to move this check to `handle_block` before heavyweight operations (e.g., signature verification), and ensure that `self.incomplete_block` is `None` here?

exonum/src/messages/protocol.rs
```diff
@@ -361,3 +362,15 @@ messages! {
         height: Height,
     }
 }
+
+impl BlockResponse {
+    /// Verify Merkle root of transactions in the block.
+    pub fn verify_tx_hash(&self) -> bool {
+        let db = MemoryDB::new();
```
DarkEld3r: I'm a little bit concerned about creating `MemoryDB` here, but probably it is OK.

exonum/src/node/consensus.rs
```diff
@@ -311,6 +297,35 @@ impl NodeHandler {
...
 
+    /// Executes and commits block. This function is called when node has full block information.
+    pub fn handle_full_block(&mut self, msg: &BlockResponse) {
+        let block = msg.block();
+        let block_hash = block.hash();
+
+        if self.state.block(&block_hash).is_none() {
+            let (block_hash, patch) =
+                self.create_block(block.proposer_id(), block.height(), msg.transactions());
+            // Verify block_hash
```
DarkEld3r: Please add a dot here. :upside_down_face:

exonum/src/node/state.rs
```diff
@@ -239,7 +251,8 @@ impl RequestData {
         #![cfg_attr(feature = "cargo-clippy", allow(match_same_arms))]
         let ms = match *self {
             RequestData::Propose(..) => PROPOSE_REQUEST_TIMEOUT,
-            RequestData::Transactions(..) => TRANSACTIONS_REQUEST_TIMEOUT,
+            RequestData::TransactionsForPropose(..) => TRANSACTIONS_REQUEST_TIMEOUT,
```
popzxc: Maybe `RequestData::TransactionsForPropose(..) | RequestData::TransactionsForBlock => TRANSACTIONS_REQUEST_TIMEOUT` to highlight the fact that those timeouts are the same?

exonum/src/node/state.rs
```diff
@@ -851,6 +909,35 @@ impl State {
         }
     }
 
+    /// Finds unknown transactions in the block.
+    pub fn create_incomplete_block(
```
popzxc: Since this function can panic, please add a note about it to the docs (like it's done in `add_prevote`). 
BTW, should it really panic? Maybe `error!(...)` will be enough here?

exonum/src/node/state.rs
```diff
@@ -851,6 +909,35 @@ impl State {
...
+                    )
+                }
+            } else {
+                unknown_txs.insert(*hash);
+            }
+        }
+
+        self.incomplete_block = Some(IncompleteBlock {
+            msg: msg.clone(),
+            unknown_txs: unknown_txs.clone(),
```
popzxc: As I can see, we are cloning `unknown_txs` here just to check that it's not empty below. It can easily be avoided, e.g.:
```
let response = Ok(!unknown_txs.is_empty());
self.incomplete_block = Some(IncompleteBlock {
    msg: msg.clone(),
    unknown_txs: unknown_txs,
});
response
```

exonum/src/node/state.rs
```diff
@@ -851,6 +910,36 @@ impl State {
...
+            }
+        }
+
+        let response = Ok(!unknown_txs.is_empty());
+        self.incomplete_block = Some(IncompleteBlock {
+            msg: msg.clone(),
+            unknown_txs,
+        });
+
+        response
```
slowli: Nit: can be expressed more succinctly with `self.incomplete_block.has_unknown_txs()`. Alternatively, you may return `&IncompleteBlock` instead of `bool`; IMO, it would be more understandable semantically.

exonum/src/node/consensus.rs
```diff
@@ -550,11 +565,18 @@ impl NodeHandler {
...
+        // Go to handle full propose if we get last transaction.
         for (hash, round) in full_proposes {
-            self.remove_request(&RequestData::Transactions(hash));
+            self.remove_request(&RequestData::ProposeTransactions(hash));
             self.handle_full_propose(hash, round);
         }
+
+        let full_block = self.state.remove_unknown_transaction(hash);
+        // Go to handle full block if we get last transaction
+        if full_block.is_some() {
```
slowli: `if let` could be better here; it allows to eliminate `full_block.unwrap()` later.

exonum/src/node/consensus.rs
```diff
@@ -767,6 +789,21 @@ impl NodeHandler {
                     ).raw()
                         .clone()
                 }
+                RequestData::BlockTransactions => {
+                    let txs: Vec<_> = match self.state.incomplete_block() {
+                        Some(incomplete_block) => {
+                            incomplete_block.unknown_txs().iter().cloned().collect()
+                        }
+                        None => Vec::new(),
```
slowli: Does it make sense to make a `TransactionsRequest` with empty tx list?

exonum/src/node/state.rs
slowli: Sorry for the confusion :sweat:  If I don't miss anything, the situation can occur only if a qualified majority of validators decided to include some transaction in the blockchain twice. This is a reason for panic, imo; we don't have a valid way to arrive to this point and don't know how to deal with such a transaction.

exonum/src/messages/protocol.rs
vldm: @SerhiiOryshych  can we create task about removing `MemoryDb`?

exonum/src/sandbox/consensus.rs
```diff
@@ -3156,6 +3254,164 @@ fn handle_block_response_tx_in_pool() {
...
+        VALIDATOR_2,
+        HEIGHT_ONE,
+        ROUND_ONE,
+        &propose.hash(),
+        &block1.hash(),
+        sandbox.time().into(),
+        sandbox.s(VALIDATOR_2),
+    );
+    // Precommit with invalid block hash.
+    let precommit_3 = Precommit::new(
```
vldm: Can you name this variable as `precommit_for_other_block` for example, or other more verbose name?

exonum/src/node/consensus.rs
```diff
@@ -245,33 +223,41 @@ impl NodeHandler {
...
         if self.state.block(&block_hash).is_none() {
-            // Verify transactions
-            let tx_hashes = if let Some(res) = self.validate_block_transactions(msg) {
-                self.blockchain
-                    .merge(res.1)
-                    .expect("Unable to save transaction to persistent pool.");
-                res.0
-            } else {
+            if !msg.verify_tx_hash() {
+                error!("Received block has wrong root hash, msg={:?}", msg);
```
vldm: "root hash" is strange message, can we rewrite it as "received block with invalid tx_hashes" or something more DevOps friendly?

exonum/src/node/consensus.rs
```diff
@@ -141,101 +141,86 @@ impl NodeHandler {
...
 
-        trace!("Handle block");
-
         let block = msg.block();
         let block_hash = block.hash();
 
         // TODO add block with greater height to queue (ECR-171)
         if self.state.height() != block.height() {
-            return;
+            return Err(format!("Received block hash another height, msg={:?}", msg));
```
vldm: fix msg

exonum/src/node/consensus.rs
```diff
@@ -311,6 +285,35 @@ impl NodeHandler {
         }
     }
 
+    /// Executes and commits block. This function is called when node has full block information.
+    pub fn handle_full_block(&mut self, msg: &BlockResponse) {
```
popzxc: Please document `panic!(..)` here too.

exonum/src/node/consensus.rs
```diff
@@ -141,137 +141,111 @@ impl NodeHandler {
...
-        Some((tx_hashes, fork.into_patch()))
-    }
-
-    /// Handles the `Block` message. For details see the message documentation.
-    // TODO write helper function which returns Result (ECR-123)
-    #[cfg_attr(feature = "flame_profile", flame)]
-    pub fn handle_block(&mut self, msg: &BlockResponse) {
-        // Request are sent to us
+    fn validate_block_response(&self, msg: &BlockResponse) -> Result<(), String> {
+        // Request are sent to us.
```
popzxc: Either `Request is` or `Requests are`.
And I suggest to make comment more verbose, e.g. `Check that request is sent to us` (or remove it completely, the next line is pretty self-explanatory).

SerhiiOryshych:
@slowli @DarkEld3r Created a separate task for testing these changes: https://jira.bf.local/browse/ECR-1625. Of course it must be done before the next release.

exonum/src/node/consensus.rs
```diff
@@ -767,6 +780,21 @@ impl NodeHandler {
                     ).raw()
                         .clone()
                 }
+                RequestData::BlockTransactions => {
+                    let txs: Vec<_> = match self.state.incomplete_block() {
```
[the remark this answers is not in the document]
DarkEld3r: @slowli Personally I like reduced visual noise. Additionally I suppose that Clippy will produce warnings at "non-ergonomic match expressions" (which can be suppressed of course).

PR #665, opened 2018-05-03
Fixed doc typo

alexander-irbis:
¯\_(ツ)_/¯


exonum/src/encoding/error.rs
```diff
@@ -72,7 +72,7 @@ pub enum Error {
         /// value that was parsed as size
         value: Offset,
     },
-    /// `RawMessage` is to short
+    /// `RawMessage` is too short
```
andreyluchin: Shouldn't this end with a dot? 🙃

PR #667, opened 2018-05-04
Add execute method in Service trait [ECR-651]

aleksuss:


CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `Field`, `CryptoHash`, `StorageValue` and `ExonumJson` traits have been
   implemented for `chrono::Duration` structure. (#653)
 
+- Added a method `execute` in `Service` trait. (#667)
```
vitvakatu: `execute method has been added to the Service trait`

exonum/src/blockchain/mod.rs
```diff
@@ -541,6 +546,25 @@ impl Blockchain {
     }
 }
 
+fn service_execute(service: &Service, mut fork: &mut Fork) {
```
vitvakatu: Do you really need `mut` before `fork` here?

exonum/src/blockchain/tests.rs
```diff
@@ -402,6 +506,40 @@ mod memorydb_tests {
         let mut blockchain = create_blockchain(path);
         super::handling_tx_panic_storage_error(&mut blockchain);
     }
+
+    #[test]
+    fn test_service_execute() {
+        let dir = TempDir::new(super::gen_tempdir_name().as_str()).unwrap();
```
vitvakatu: I think it's better to move the body of these tests into a separate function.

CHANGELOG.md
aleksuss: As for me the past simple tense is more appropriate here.

vitvakatu: We're using present perfect everywhere else though

exonum/src/blockchain/mod.rs
aleksuss: Really

exonum/src/blockchain/service.rs
```diff
@@ -172,6 +172,10 @@ pub trait Service: Send + Sync + 'static {
         Value::Null
     }
 
+    /// A service execution. This method is invoked for each service after execution
+    /// of all transactions in the mempool but before `handle_commit` handler.
```
slowli: - I think an example is needed to explain possible use cases.
- Please document the order in which `execute` are invoked for services. (Ideally, there should be a couple of tests checking this order.)
- "execution of all transactions in the mempool" - do you mean "...in a block"?

exonum/src/blockchain/tests.rs
```diff
@@ -361,6 +361,96 @@ mod transactions_tests {
...
+    fn tx_from_raw(&self, _raw: RawTransaction) -> Result<Box<Transaction>, MessageError> {
+        unimplemented!()
+    }
+
+    fn execute(&self, _fork: &mut Fork) {
+        panic!(Error::new("42"));
+    }
+}
+
+fn assert_service_execute(blockchain: &Blockchain, db: &mut Box<Database>) {
```
slowli: It could make sense to check that `state_hash` can be updated by `Service::execute` - seems like a possible place of regressions in the future.

exonum/src/blockchain/mod.rs
```diff
@@ -541,6 +546,25 @@ impl Blockchain {
...
+    fork.checkpoint();
+    match panic::catch_unwind(panic::AssertUnwindSafe(|| service.execute(&mut fork))) {
```
slowli: Continuing @vitvakatu's question: what if `fork` is passed here instead of `&mut fork` (seems kind of redundant)?

exonum/src/blockchain/mod.rs
```diff
@@ -271,6 +271,11 @@ impl Blockchain {
                     .expect("Transaction not found in the database.");
             }
 
+            // Invoke execute method for all services.
+            self.service_map
+                .values()
+                .for_each(|service| service_execute(service.as_ref(), &mut fork));
```
slowli: From `std` docs, it seems that `for` cycle could be preferable here.

exonum/src/blockchain/tests.rs
```diff
@@ -361,6 +361,96 @@ mod transactions_tests {
...
+    fn state_hash(&self, _snapshot: &Snapshot) -> Vec<Hash> {
+        vec![]
+    }
+
+    fn tx_from_raw(&self, _raw: RawTransaction) -> Result<Box<Transaction>, MessageError> {
+        unimplemented!()
+    }
+
+    fn execute(&self, _fork: &mut Fork) {
+        panic!("42");
```
slowli: If I understand the logic of tests correctly, this method should write to the index and *then* panic; then corresponding tests would actually check that the corresponding changes are actually rolled back.

exonum/src/blockchain/tests.rs
```diff
@@ -361,6 +361,96 @@ mod transactions_tests {
...
+    let index = ListIndex::new(IDX_NAME, &snapshot);
+    assert_eq!(index.len(), 1);
+    assert_eq!(index.get(0), Some(1));
+}
+
+fn assert_service_execute_panic(blockchain: &Blockchain, db: &mut Box<Database>) {
+    let (_, patch) = blockchain.create_patch(ValidatorId::zero(), Height::zero(), &[]);
+    let _ = db.merge(patch);
+    let snapshot = db.snapshot();
+    let index: ListIndex<&Box<Snapshot>, u32> = ListIndex::new(IDX_NAME, &snapshot);
```
slowli: Nitpick: I think you can use `ListIndex<_, u32>` for brevity.

exonum/src/blockchain/service.rs
aleksuss: @slowli I don't understand why the invoke order is important for us here. We expect that each service will work with own indices and own data, don't we? And a modifying indices from other services it's a bad practice ? At least now.

exonum/src/blockchain/mod.rs
alekseysidorov: But `iter().for_each()` in many cases can be transparently replaced by `par_iter().for_each()` from the `rayon`.

exonum/src/blockchain/service.rs
alekseysidorov: I think that you should add comment about the unspecified `Service:execute` order.

exonum/src/blockchain/service.rs
```diff
@@ -172,6 +172,12 @@ pub trait Service: Send + Sync + 'static {
         Value::Null
     }
 
+    /// A service execution. This method is invoked for each service after execution
+    /// of all transactions in the block but before `handle_commit` handler. The order of
+    /// invoking `execute` method for every service depends on the service ID. `execute` for
+    /// the service with the smallest ID is invoked first up to the largest one.
```
slowli: Nitpick: To be completely clear, I'd add the following sentence:

> Effectively, this means that services should not rely on a particular ordering of `Service::execute` invocations.

exonum/src/blockchain/service.rs
```diff
@@ -172,6 +172,12 @@ pub trait Service: Send + Sync + 'static {
         Value::Null
     }
 
+    /// A service execution. This method is invoked for each service after execution
+    /// of all transactions in the block but before `handle_commit` handler. The order of
```
slowli: Nitpick: I'd add a paragraph break after "This method is invoked..." to keep the first paragraph (which is used as summary in `Service` implementations) reasonably short.

exonum/src/blockchain/tests.rs
```diff
@@ -361,20 +361,110 @@ mod transactions_tests {
...
+    }
+
+    fn execute(&self, _fork: &mut Fork) {
+        panic!(Error::new("42"));
+    }
+}
+
+fn assert_service_execute(blockchain: &Blockchain, db: &mut Box<Database>) {
+    let (_, patch) = blockchain.create_patch(ValidatorId::zero(), Height::zero(), &[]);
+    let _ = db.merge(patch);
```
slowli: Nitpick: shouldn't `db.merge(patch).unwrap()` work better here?

exonum/src/blockchain/tests.rs
slowli: > and what about assert!(index.is_empty()); at line 451 ?

The index should still be empty because of rollback, shouldn't it?

exonum/src/blockchain/tests.rs
aleksuss: Temp directory destroys when `dir` out of scope.

exonum/src/blockchain/tests.rs
aleksuss: Exactly

exonum/src/blockchain/tests.rs
vitvakatu: You can return it from init function

aleksuss: Yes, I can. But what is a profit ?
You propose change 
`let dir = TempDir::new(super::gen_tempdir_name().as_str()).unwrap();`
for
```rust
fn create_temp_dir() -> TempDir {
    TempDir::new(super::gen_tempdir_name().as_str()).unwrap()
}
...
let dir = create_temp_dir();
```
???

alekseysidorov:
It seems that during the genesis block creation service can modify blockchain database twice: in `initialize` method and later in `execute`.
That sounds wrong for me, we need to investigate proper business logic for such case.

DarkEld3r:
@alekseysidorov 
Personally, I don't see a huge problem here, but we can simply don't call `execute` method during genesis block creation.

PR #673, opened 2018-05-07
Verifying transactions in a thread pool. [ECR-225]

andreyluchin:
### First attempt of parallelizing transaction verification
Note that this is still a prototype in an uncompilable state.

Workflow looks something like this:

1. Thread pool is created at node launch and loaded with a `for_each` future for a pool channel.
2. Transaction is received by the node.
3. `InternalRequest::VerifyTx(tx)` is send asynchronously to verify transaction.
4. Once request has been received by `InternalPart` the transaction is, again asynchronously, sent to the Thread Pool.
5. Once the Thread Pool receives the transaction and successfully verifies it, it composes and sends `InternalEvent::TxVerified` event.
6. Event is handled by `NodeHandler` - adds transaction to the transaction pool.

### New dependency
A new dependency is introduced - `tokio_threadpool`. This a subject of change during the current evolution of the `tokio` and `future` crates.

### Thoughts on its efficiency
While working on this task I found some problems with this approach that make me doubt that parallelization is going to be efficient.
#### Problem 1
Sending a request, sending to thread pool, handling `TxVerified` event - all happen asynchronously in a single thread tokio runtime. Because of this, it seems that a big portion of transaction being in the verification process is spent waiting for these async tasks to get executed by the given runtime. In an ideal world little to no time of the main thread should be spent on waiting on such things. However, I didn't find a way to make that real in the current Exonum's async architecture.
#### Problem 2
If the Thread Pool becomes full, sending a `VerifyTx` request becomes blocking. This means that the main thread will wait roughly T/N (+ synchronization time) where T is the time it takes for tx to get verified and N is number of worker threads in the Thread Pool. This means that there should be at least 2 threads in the Pool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Now nodes will switch to `min_propose_timeout` for block proposal timeout
   faster if they receive more than `propose_timeout_threshold` transactions
   during `max_propose_timeout`. (#844)
+### Internal Improvements
+
+#### exonum
+- Transactions are now verified in a thread pool. (#673)
 
 - Custom log formatting (along with `colored` and `term` dependencies) has been
   removed in favor of `env_logger`. (#857).
@@ -297,7 +301,6 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Maintenance CLI command for node management has been added. Currently the only
   supported command is `clear-cache` which clears node message cache. (#676)
-- Transactions are now verified in a thread pool. (#673)
 
 - `StoredConfiguration` validation has been extended with `txs_block_limit`
   parameter check. (#690)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   removed in favor of `env_logger`. (#857).
 
 - Several dependencies have been updated. (#861)
+
 - Transactions are now verified in a thread pool. Thread pool size is set to
   optimal value by default (CPU count) or can be configured manually. (#673)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 ### Internal Improvements
 
 #### exonum
+
 - Transactions are now verified in a thread pool. (#673)
 
 - Custom log formatting (along with `colored` and `term` dependencies) has been
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 #### exonum
 
 - Bug with pool size overflow has been fixed. (#853)
+- Transactions (signature) verification benchmark has been fixed. (#673)
+- Node no longer panics when transaction pool has a lot of transactions and
+  consensus is at round 0. (#673)
 
 ### Internal Improvements
 
@@ -61,16 +64,13 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Now nodes will switch to `min_propose_timeout` for block proposal timeout
   faster if they receive more than `propose_timeout_threshold` transactions
   during `max_propose_timeout`. (#844)
-### Internal Improvements
-
-#### exonum
-
-- Transactions are now verified in a thread pool. (#673)
 
 - Custom log formatting (along with `colored` and `term` dependencies) has been
   removed in favor of `env_logger`. (#857).
 
 - Several dependencies have been updated. (#861)
+- Transactions are now verified in a thread pool. Thread pool size is set to
+  optimal value by default (CPU count) or can be configured manually (#673)
 
 ## 0.9.1 - 2018-08-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -297,6 +297,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Maintenance CLI command for node management has been added. Currently the only
   supported command is `clear-cache` which clears node message cache. (#676)
+- Transactions are now verified in a thread pool. (#673)
 
 - `StoredConfiguration` validation has been extended with `txs_block_limit`
   parameter check. (#690)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 #### exonum
 
 - Bug with pool size overflow has been fixed. (#853)
+
 - Transactions (signature) verification benchmark has been fixed. (#673)
+
 - Node no longer panics when transaction pool has a lot of transactions and
   consensus is at round 0. (#673)
 
@@ -70,7 +72,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Several dependencies have been updated. (#861)
 - Transactions are now verified in a thread pool. Thread pool size is set to
-  optimal value by default (CPU count) or can be configured manually (#673)
+  optimal value by default (CPU count) or can be configured manually. (#673)
 
 ## 0.9.1 - 2018-08-02
 
--- a/exonum/Cargo.toml
+++ b/exonum/Cargo.toml
@@ -45,6 +45,7 @@ chrono = { version = "=0.4.5", features = ["serde"] }
 uuid = { version = "=0.6.5", features = ["serde"] }
 snow = "=0.3.1"
 rust_decimal = "=0.9.1"
+tokio-threadpool = "0.1.2"
 
 exonum_rocksdb = "0.7.4"
 exonum_sodiumoxide = { version = "0.0.20", optional = true }
--- a/exonum/Cargo.toml
+++ b/exonum/Cargo.toml
@@ -39,13 +39,13 @@ tokio-core = "=0.1.17"
 tokio-io = "=0.1.7"
 tokio-retry = "=0.2.0"
 tokio-timer = "=0.1.2"
+tokio-threadpool = "=0.1.5"
 failure = "0.1.2"
 os_info = "1.0.1"
 chrono = { version = "=0.4.5", features = ["serde"] }
 uuid = { version = "=0.6.5", features = ["serde"] }
 snow = "=0.3.1"
 rust_decimal = "=0.9.1"
-tokio-threadpool = "0.1.5"
 
 exonum_rocksdb = "0.7.4"
 exonum_sodiumoxide = { version = "0.0.20", optional = true }
--- a/exonum/Cargo.toml
+++ b/exonum/Cargo.toml
@@ -40,6 +40,7 @@ tokio-io = "=0.1.7"
 tokio-retry = "=0.2.0"
 tokio-timer = "=0.1.2"
 tokio-threadpool = "=0.1.5"
+tokio-executor = "=0.1.3"
 failure = "0.1.2"
 os_info = "1.0.1"
 chrono = { version = "=0.4.5", features = ["serde"] }
--- a/exonum/Cargo.toml
+++ b/exonum/Cargo.toml
@@ -45,7 +45,7 @@ chrono = { version = "=0.4.5", features = ["serde"] }
 uuid = { version = "=0.6.5", features = ["serde"] }
 snow = "=0.3.1"
 rust_decimal = "=0.9.1"
-tokio-threadpool = "0.1.2"
+tokio-threadpool = "0.1.5"
 
 exonum_rocksdb = "0.7.4"
 exonum_sodiumoxide = { version = "0.0.20", optional = true }
--- a/exonum/benches/criterion/transactions.rs
+++ b/exonum/benches/criterion/transactions.rs
@@ -30,8 +30,7 @@ use exonum::{
         Blockchain, ExecutionResult, GenesisConfig, Service, SharedNodeState, Transaction,
         TransactionSet, ValidatorKeys,
     },
-    crypto::{self, Hash, PublicKey},
-    encoding,
+    crypto::{self, Hash, PublicKey}, encoding,
     events::{
         error::other_error, Event, EventHandler, HandlerPart, InternalEvent, InternalPart,
         NetworkEvent,
@@ -203,8 +202,7 @@ impl TransactionsBenchmarkRunner {
         };
         // Emulates transactions from the network.
         let socket_addr = handler_part.handler.inner().system_state.listen_address();
-        let transactions = self
-            .transactions
+        let transactions = self.transactions
             .into_iter()
             .map(|raw| NetworkEvent::MessageReceived(socket_addr, raw))
             .collect::<Vec<_>>();
--- a/exonum/benches/criterion/transactions.rs
+++ b/exonum/benches/criterion/transactions.rs
@@ -31,7 +31,7 @@ use exonum::{
         TransactionSet, ValidatorKeys,
     },
     crypto::{self, Hash, PublicKey}, encoding,
-    events::{error::other_error, Event, EventHandler, HandlerPart, NetworkEvent},
+    events::{error::other_error, Event, EventHandler, HandlerPart, NetworkEvent, InternalEvent},
     messages::{Message, RawTransaction},
     node::{
         ApiSender, Configuration, ConnectList, DefaultSystemState, ListenerConfig, NodeApiConfig,
--- a/exonum/benches/criterion/transactions.rs
+++ b/exonum/benches/criterion/transactions.rs
@@ -108,7 +108,7 @@ impl TransactionsHandler {
 
 impl EventHandler for TransactionsHandler {
     fn handle_event(&mut self, event: Event) {
-        let is_transaction = if let Event::Network(NetworkEvent::MessageReceived(..)) = event {
+        let is_transaction = if let Event::Internal(InternalEvent::TxVerified(_)) = event {
             true
         } else {
             false
--- a/exonum/benches/criterion/transactions.rs
+++ b/exonum/benches/criterion/transactions.rs
@@ -199,6 +199,7 @@ impl TransactionsBenchmarkRunner {
         let internal_part = InternalPart {
             internal_tx: self.channel.internal_events.0,
             internal_requests_rx: self.channel.internal_requests.1,
+            thread_pool_size: None,
         };
         // Emulates transactions from the network.
         let socket_addr = handler_part.handler.inner().system_state.listen_address();
@@ -255,6 +256,7 @@ impl TransactionsBenchmarkRunner {
             mempool: Default::default(),
             services_configs: Default::default(),
             database: Default::default(),
+            thread_pool_size: Default::default(),
         }
     }
 }
--- a/exonum/benches/criterion/transactions.rs
+++ b/exonum/benches/criterion/transactions.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-const TRANSACTIONS_COUNT: usize = 40_000;
+const TRANSACTIONS_COUNT: usize = 100_000;
 const SAMPLE_SIZE: usize = 3;
 
 use criterion::{
@@ -30,8 +30,12 @@ use exonum::{
         Blockchain, ExecutionResult, GenesisConfig, Service, SharedNodeState, Transaction,
         TransactionSet, ValidatorKeys,
     },
-    crypto::{self, Hash, PublicKey}, encoding,
-    events::{error::other_error, Event, EventHandler, HandlerPart, NetworkEvent, InternalEvent},
+    crypto::{self, Hash, PublicKey},
+    encoding,
+    events::{
+        error::other_error, Event, EventHandler, HandlerPart, InternalEvent, InternalPart,
+        NetworkEvent,
+    },
     messages::{Message, RawTransaction},
     node::{
         ApiSender, Configuration, ConnectList, DefaultSystemState, ListenerConfig, NodeApiConfig,
@@ -193,20 +197,27 @@ impl TransactionsBenchmarkRunner {
             network_rx: self.channel.network_events.1,
             api_rx: self.channel.api_requests.1,
         };
+        let internal_part = InternalPart {
+            internal_tx: self.channel.internal_events.0,
+            internal_requests_rx: self.channel.internal_requests.1,
+        };
         // Emulates transactions from the network.
         let socket_addr = handler_part.handler.inner().system_state.listen_address();
-        let transactions = self.transactions
+        let transactions = self
+            .transactions
             .into_iter()
             .map(|raw| NetworkEvent::MessageReceived(socket_addr, raw))
             .collect::<Vec<_>>();
         let network_thread = thread::spawn(move || {
-            Core::new()?
-                .run(
-                    tx_sender
-                        .send_all(stream::iter_ok(transactions))
-                        .map(drop)
-                        .map_err(drop),
-                )
+            let mut core = Core::new()?;
+            let handle = core.handle();
+            core.handle().spawn(
+                tx_sender
+                    .send_all(stream::iter_ok(transactions))
+                    .map(drop)
+                    .map_err(drop),
+            );
+            core.run(internal_part.run(handle))
                 .map_err(|_| other_error("An error in the `Network` thread occurred"))
         });
         // Drops unused channels.
@@ -215,7 +226,10 @@ impl TransactionsBenchmarkRunner {
         let mut core = Core::new()?;
         core.run(handler_part.run())
             .map_err(|_| other_error("An error in the `Handler` thread occurred"))?;
-        network_thread.join().unwrap()
+        // Drops internal part.
+        drop(self.channel.internal_requests.0);
+        network_thread.join().unwrap()?;
+        Ok(())
     }
 
     fn node_config() -> NodeConfig {
--- a/exonum/src/events/internal.rs
+++ b/exonum/src/events/internal.rs
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// spell-checker:ignore threadpool
-
 use futures::{self, future, sync::mpsc, Future, Sink, Stream};
 use tokio_core::reactor::{Handle, Timeout};
 use tokio_executor::SpawnError;
--- a/exonum/src/events/internal.rs
+++ b/exonum/src/events/internal.rs
@@ -94,7 +94,7 @@ impl InternalPart {
                     InternalRequest::VerifyTx(tx) => {
                         if !txs_in_verification.insert(tx.raw().hash()) {
                             let f = future::ok::<(), ()>(());
-                            rto_box(f)
+                            to_box(f)
                         } else {
                             let f = futures::lazy(move || {
                                 pool_tx
--- a/exonum/src/events/internal.rs
+++ b/exonum/src/events/internal.rs
@@ -47,7 +47,7 @@ impl InternalPart {
                         .clone()
                         .send(InternalEvent::TxVerified(tx))
                         .wait()
-                        .map_err(|_| panic!("Cannot send tx to thread pool."));
+                        .expect("Cannot send tx to the thread pool.");
                 }
                 Ok(())
             })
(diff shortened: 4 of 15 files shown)


exonum/src/events/internal.rs
```diff
@@ -70,6 +91,20 @@ impl InternalPart {
                         }).map_err(|_| panic!("Can't execute shutdown"));
                         to_box(f)
                     }
+                    InternalRequest::VerifyTx(tx) => {
+                        if !txs_in_verification.insert(tx.raw().hash()) {
+                            let f = future::ok::<(), ()>(());
+                            to_box(f)
```
andreyluchin: This a hack around the fact that i have to return a future.

exonum/src/events/mod.rs
```diff
@@ -47,13 +48,17 @@ pub enum InternalEvent {
     Timeout(NodeTimeout),
     /// Shutdown the node.
     Shutdown,
+    /// Transaction has been successfully verified.
+    TxVerified(Box<Transaction>),
 }
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
```
andreyluchin: Is `PartialEq` and `Eq` vital here? Should it stay removed or should I implement it for `Box<Transaction>`?

aleksuss:
Don't forget about CHANGELOG 😄

exonum/src/events/internal.rs
```diff
@@ -70,6 +91,20 @@ impl InternalPart {
                         }).map_err(|_| panic!("Can't execute shutdown"));
                         to_box(f)
                     }
+                    InternalRequest::VerifyTx(tx) => {
+                        if !txs_in_verification.insert(tx.raw().hash()) {
```
popzxc: IHMO straight order will be more intuitive: at first process situation, when we can insert transaction, then when we can't.
Is this negation required here?

andreyluchin: You're right, it's not really required

andreyluchin:
Currently working on adding a new `thread_pool_size` config. After that PR is no longer considered WIP.

exonum/src/events/internal.rs
```diff
@@ -30,10 +36,32 @@ pub struct InternalPart {
...
+        let (pool_tx, pool_rx) = mpsc::channel::<Box<Transaction>>(0);
+        let internal_tx = self.internal_tx.clone();
+        thread_pool.spawn(futures::lazy(move || {
+            pool_rx.for_each(move |tx| {
+                if tx.verify() {
+                    internal_tx
+                        .clone()
+                        .send(InternalEvent::TxVerified(tx))
+                        .wait()
+                        .expect("Cannot send TxVerified event.");
```
alekseysidorov: We doesn't use unwrap in futures, so can you throw error through futures chain?

exonum/src/events/internal.rs
```diff
@@ -70,6 +98,19 @@ impl InternalPart {
                         }).map_err(|_| panic!("Can't execute shutdown"));
                         to_box(f)
                     }
+                    InternalRequest::VerifyTx(tx) => {
+                        if txs_in_verification.insert(tx.raw().hash()) {
+                            let f = futures::lazy(move || {
+                                pool_tx.send(tx).map(drop).map_err(into_other)
+                            }).map_err(|_| {
+                                panic!("Can't send tx for verification to the thread pool")
```
alekseysidorov: Please replace `panic` by `log_error`.

exonum/src/events/internal.rs
andreyluchin: Sorry for my limited Futures knowledge, but what futures chain are you talking about? The motivation behind what I did there was to make `send` call synchronous, which is why I chose to use `expect` there. Besides, the future is not passed on anywhere.

exonum/src/events/internal.rs
```diff
@@ -12,15 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use futures::{self, Future, Sink, Stream};
+// spell-checker:ignore threadpool
```
ozkriff: It seems to be a common term, why not to move it into the dictionary?

exonum/Cargo.toml
```diff
@@ -46,6 +46,7 @@ chrono = { version = "=0.4.4", features = ["serde"] }
 uuid = { version = "=0.6.5", features = ["serde"] }
 snow = "=0.2.1"
 rust_decimal = "=0.9.1"
+tokio-threadpool = "0.1.5"
```
alekseysidorov: Please use pinned versions.

exonum/src/node/consensus.rs
```diff
@@ -578,16 +578,18 @@ impl NodeHandler {
...
+    /// Handles an already verified transaction.
+    pub fn handle_verified_tx(&mut self, tx: &dyn Transaction) {
         // We don't care about result, because situation when transaction received twice
         // is normal for internal messages (transaction may be received from 2+ nodes).
-        let _ = self.handle_tx_inner(msg);
+        let _ = self.handle_tx_inner(tx.raw().clone());
     }
 
     /// Handles raw transactions.
+    #[cfg_attr(feature = "flame_profile", flame)]
```
alekseysidorov: Did not we remove the `flamer` dependency?

exonum/src/node/consensus.rs
```diff
@@ -578,16 +578,18 @@ impl NodeHandler {
...
         };
 
-        if !tx.verify() {
-            return;
-        }
+        self.execute_later(InternalRequest::VerifyTx(tx));
+    }
 
+    /// Handles an already verified transaction.
+    pub fn handle_verified_tx(&mut self, tx: &dyn Transaction) {
```
alekseysidorov: I wonder that here transaction should pass by value, but transaction code doesn't support that.

exonum/src/events/internal.rs
```diff
@@ -71,6 +76,23 @@ impl InternalPart {
                         }).map_err(|_| panic!("Can't execute shutdown"));
                         to_box(f)
                     }
+                    InternalRequest::VerifyTx(tx) => {
+                        let internal_tx = internal_tx.clone();
+                        let thread_pool = Rc::clone(&thread_pool);
+                        let f = future::lazy(move || {
+                            thread_pool.spawn(future::lazy(move || {
```
slowli: Are we OK with a possible panic here? It seems that using `thread_pool.sender().spawn(_)`could be safer.

exonum/src/events/internal.rs
```diff
@@ -71,6 +76,23 @@ impl InternalPart {
                         }).map_err(|_| panic!("Can't execute shutdown"));
                         to_box(f)
                     }
+                    InternalRequest::VerifyTx(tx) => {
+                        let internal_tx = internal_tx.clone();
```
slowli: Cloning an mpsc sender to send a single message looks inefficient (although it may be not, profiling could be helpful). I wonder if it's possible and makes sense to use `thread_local` crate here (or upgrade to `futures >0.1`).

exonum/src/node/consensus.rs
```diff
@@ -578,16 +578,18 @@ impl NodeHandler {
...
         // We don't care about result, because situation when transaction received twice
```
slowli: Do I understand this comment correctly that a signature for the same transaction may be verified multiple times if a transaction is received from multiple nodes? If so, this seems like an area for future improvement; we want to ignore known transactions right away.

exonum/src/events/internal.rs
```diff
@@ -31,6 +34,8 @@ pub struct InternalPart {
 
 impl InternalPart {
     pub fn run(self, handle: Handle) -> Box<dyn Future<Item = (), Error = io::Error>> {
+        let thread_pool = Rc::new(ThreadPoolBuilder::new().build());
```
slowli: Do we want to provide some settings for the thread pool?

andreyluchin:
# Update
1. Benchmarks are now fixed after being broken by consensus changes.
2. As pointed out by @slowli `thread_pool.spawn()` could panic either when it is at capacity or when it is shutdown. `thread_pool.spawn()` is now replaced with `'thread_pool.sender().spawn()` and errors are now properly handled.

exonum/src/node/consensus.rs
```diff
@@ -553,7 +553,7 @@ impl NodeHandler {
             .merge(fork.into_patch())
             .expect("Unable to save transaction to persistent pool.");
 
-        if self.state.is_leader() {
+        if self.state.is_leader() && self.state.round() != Round::zero() {
```
andreyluchin: The problem I encountered:
1. Node starts with 40_000 transactions in pool for verification.
2. Because of this, `maybe_add_propose_timeout` is triggered.
3. Since round = 0, panic occurs when trying to get `round.previous()`.

I decided that `self.maybe_add_propose_timeout()` deserved an additional check.

ilammy: Mmm... Digging through the callee tree of `NodeHandler::maybe_add_propose_timeout()` I see that `Round::previous()` is called only from `NodeHandler::round_start_time()`. I guess it may be better to teach `round_start_time()` to handle this situation correctly (just return `height_start_time()` I think) rather than tacking this ad-hoc fix on here.

andreyluchin: That's a very good point and needs to be discussed. I agree that fix proposed by me is very ad-hoc'y. But I think the problem is much deeper: reckless attitude towards error handling in Exonum. We have a ton of `assert`s and `panic!`s all around the non-test code which could be replaced by `Option` or `Result` (`Round::previous` is a great example of this). If I were to move this fix somewhere else, I would do so by making `Round::previous` return `Option` rather than teaching `round_start_time()` to do things it is not responsible for.

andreyluchin: @vldm I know you are not a fan of `panic`s, what do you think of this situation?

exonum/src/events/internal.rs
```diff
@@ -71,6 +82,41 @@ impl InternalPart {
                         }).map_err(|_| panic!("Can't execute shutdown"));
                         to_box(f)
                     }
+                    InternalRequest::VerifyTx(tx) => {
+                        let internal_tx = internal_tx.clone();
+                        let tx = Arc::new(tx);
+                        let thread_pool = Rc::clone(&thread_pool);
+
+                        let f = future::loop_fn(Err(SpawnError::at_capacity()), move |status| {
```
vldm: I didn't like  this retry mechanism. It's unreadable.

can we just  rewrite it as:
```rust
thread_pool.sender().spawn(future::lazy(move || {
                                            if tx.verify() {
                                                internal_tx
                                                    .wait()
                                                    .send(InternalEvent::TxVerified(
                                                        tx.raw().clone(),
                                                    ))
                                                    .expect("Cannot send TxVerified event.");
                                            }
                                            Ok(())
                                        })).expect("cannot verify new transaction);

```

Based on source of tokio-threadpool it will return `SpawnError::at_capacity()` error, only if `num_futures` overflow `MAX_FUTURES` which is `usize::MAX >> 2` probably this is a good value for us, and we really unexpected so many messages. 

@slowli @andreyluchin https://github.com/tokio-rs/tokio/blob/master/tokio-threadpool/src/sender.rs#L99 https://github.com/tokio-rs/tokio/blob/master/tokio-threadpool/src/pool/state.rs#L30

exonum/src/node/consensus.rs
vldm: Only genesis height starts from zero round, other started from first.

Maybe `State` should initialize with first round?
@alekseysidorov @slowli

exonum/src/events/internal.rs
andreyluchin: I don't want to forgo gracious error handling in favor of readability. To be fair, whole `internal.rs` is unreadable. I would much rather add some comments to make things clearer.

exonum/src/events/internal.rs
```diff
@@ -71,6 +82,41 @@ impl InternalPart {
                         }).map_err(|_| panic!("Can't execute shutdown"));
                         to_box(f)
                     }
+                    InternalRequest::VerifyTx(tx) => {
+                        let internal_tx = internal_tx.clone();
+                        let tx = Arc::new(tx);
```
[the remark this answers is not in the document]
andreyluchin: `tx` has to be cloned to be moved in a `FnMut` closure introduced below. I did not find another way to do that because `tx` is a trait object `Box<Transaction>` and has to remain so.

exonum/src/node/consensus.rs
```diff
@@ -585,13 +585,14 @@ impl NodeHandler {
...
         };
 
-        if !tx.verify() {
-            return;
-        }
+        self.execute_later(InternalRequest::VerifyTx(tx));
+    }
 
+    /// Handles an already verified transaction.
+    pub fn handle_verified_tx(&mut self, tx: RawTransaction) {
```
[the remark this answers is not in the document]
andreyluchin: It doesn't just call other method, it also ignores a potential error it returns. I can't think of a good way to do that without a wrapper method, if you have any suggestions I will happily rewrite that.

exonum/src/node/consensus.rs
slowli: @andreyluchin I'd not classify all `assert`s / `panic`s as necessarily reckless; in fact, I think the current codebase may *lack* them in consensus and other safety-critical code. If used properly, `assert`s verify invariants or (if `debug_assert` is used) to signal them to the developers. OTOH, we shouldn't `panic` when an error can be expected.

Re: 0th round. If I understand correctly, `Round::zero()` is used in code mostly as an ad-hoc replacement for `None`, allowing to declare some fields (e.g., the locked round of the node, or the latest known round of peers) as `Round` instead of `Option<Round>`. I'm not sure why this is the case, but I'd argue that this "optimization" hurts code maintainability.

exonum/src/node/consensus.rs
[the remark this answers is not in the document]
slowli: It's possible to use `self.handle_tx_inner(tx).ok()`. [`Result::ok`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) converts the result into an `Option`.

exonum/src/node/consensus.rs
dvush: I think that returning height_start_time for Round(0) may cause confusion. 
Maybe round should be replaced to store inner value as NonZeroU32 since 0 is logically invalid value for round number.
For lock round Option<Round> may be used to indicate that there can be no lock.

andreyluchin: https://exonum.com/doc/advanced/consensus/specification/#rounds
>Rounds are numbered from 1

This does indeed seem like a case for using placeholder values instead of `Option`. I believe we had (or still) have similar thing for IP address where `0.0.0.0` indicates its absence.

Anyway, fixing this is out of the scope of this PR for sure. But I would definitely create an issue for reworking `Round` if I get @DarkEld3r approval.

viktorbezhenar:
@alekseysidorov @slowli @vldm @ilammy @pavel-mukhanov @aleksuss @DarkEld3r @andreyluchin 

Guys, I see lots of communication here, and looks like most of the questions are resolved, so that, could you please Approve/request other changes here if needed?

What I know so far, that there is a separate benchmark gathering task is needed (requested by @alekseysidorov - https://jira.bf.local/browse/ECR-2160 - the task is to run pybench on the branch with implemented in this PR thread pool)

Is there anything else left that blocks you from accepting this PR?

Thank You!

aleksuss:
@andreyluchin resolve conflicts please

aleksuss:
@andreyluchin resolve conflicts please.
P.S. Sorry for closing this PR. This happened by accident.

exonum/src/node/mod.rs
```diff
@@ -938,8 +943,12 @@ impl Node {
         let network_thread = thread::spawn(move || {
             let mut core = Core::new()?;
             let handle = core.handle();
+
+            let thread_pool = ThreadPool::new();
```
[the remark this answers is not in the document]
slowli: Yeah, I haven't implemented thread pool config in my patch.

slowli:
Please don't merge, my changes have broken timeout processing. I'm trying to find a fix that looks "good", but if I don't find it, I'm going to revert back most changes I've made in `InternalPart`.

exonum/src/events/internal.rs
```diff
@@ -30,53 +33,156 @@ pub struct InternalPart {
...
+        future::lazy(move || {
+            if tx.verify() {
+                let send_event = internal_tx
+                    .send(InternalEvent::TxVerified(tx.raw().clone()))
+                    .map(drop)
+                    .map_err(|e| {
+                        panic!(
+                            "error sending verified transaction \
+                             to internal events pipe: {:?}",
+                            e
```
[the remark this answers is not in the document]
andreyluchin: This error would indicate that `Internal` channel receiver is dropped which is catastrophical and cannot be handled in Exonum. Thus the panic

exonum/src/events/mod.rs
```diff
@@ -41,21 +43,25 @@ pub type SyncSender<T> = Wait<Sender<T>>;
...
 pub enum InternalRequest {
     Timeout(TimeoutRequest),
     JumpToRound(Height, Round),
     Shutdown,
+    /// Async request to verify a transaction in the thread pool.
+    VerifyTx(Box<dyn Transaction>),
```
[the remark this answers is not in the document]
andreyluchin: It was initially `Box` for both, than I changed `TxVerified` to contain `RawTransactions` due to borrowing limitations in the Future hell. I did not research the possibility of making both of them `RawTransactions` so thanks for pointing that out, definitely will do.

CHANGELOG.md
```diff
@@ -66,6 +71,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   removed in favor of `env_logger`. (#857).
 
 - Several dependencies have been updated. (#861, #865, #871)
+- Several dependencies have been updated. (#861)
```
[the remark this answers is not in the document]
andreyluchin: Will do

slowli:
I've got the following solution:

```patch
--- exonum/src/events/internal.rs	(date 1534339578000)
+++ exonum/src/events/internal.rs	(date 1534348617024)
@@ -33,6 +33,10 @@
 }
 
 impl InternalPart {
+    /// The maximum number of concurrent timeouts that can be pending
+    /// at any moment of time.
+    const TIMEOUTS_BUFFER_SIZE: usize = 16;
+
     fn verify_transaction(
         tx: Box<dyn Transaction>,
         internal_tx: mpsc::Sender<InternalEvent>,
@@ -79,13 +83,14 @@
                 }
                 req => Some(req),
             })
-            .and_then(move |request| match request {
+            .map(move |request| match request {
                 InternalRequest::Timeout(TimeoutRequest(time, timeout)) => {
                     let duration = time.duration_since(SystemTime::now())
                         .unwrap_or_else(|_| Duration::from_millis(0));
+
                     let fut = Timeout::new(duration, &handle)
                         .expect("Unable to create timeout")
-                        .map(move |_| InternalEvent::Timeout(timeout));
+                        .map(|()| InternalEvent::Timeout(timeout));
 
                     Either::A(fut)
                 }
@@ -102,6 +107,7 @@
 
                 InternalRequest::VerifyTx(..) => unreachable!(),
             })
+            .buffer_unordered(Self::TIMEOUTS_BUFFER_SIZE)
             .forward(self.internal_tx.sink_map_err(into_other))
             .map(drop)
     }
```

Looks neat to me; `buffer_unordered` allows to process timeouts in the order they expire (which was the problem previously). **However,** it only works if we have a definite upper bound on the number of simultaneous pending timeouts; and I'm not sure we have one - it seems like we can have an arbitrary amount of pending `Request` timeouts at any given time. If this is the case, the only type of solution I see is to replace `buffer_unordered()` and the following calls to spawning the corresponding futures with `handle.spawn()`:

```rust
self.internal_requests_rx
    // other code
    .map(move |event| {
        let internal_tx = internal_tx.clone();

        let send_event = event
            .map_err(log_error)
            .and_then(|evt| internal_tx.send(evt).map(drop).map_err(log_error));
        handle.spawn(send_event);
    })
    .for_each(Ok)
```

exonum/src/events/internal.rs
```diff
@@ -30,53 +33,156 @@ pub struct InternalPart {
...
+                    let evt = InternalEvent::JumpToRound(height, round);
+                    Either::B(future::ok(evt))
+                }
+
+                InternalRequest::Shutdown => {
+                    let evt = InternalEvent::Shutdown;
+                    Either::B(future::ok(evt))
+                }
+
+                InternalRequest::VerifyTx(..) => unreachable!(),
```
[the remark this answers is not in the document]
slowli: - The question of taste, I suppose; I dislike monolithic chunks of code.
- Processing `VerifyTx` does not result in a future with `Item = InternalEvent` (well, it does, but it is processed on a separate executor), while all other `InternalRequest` variants do.

exonum/src/events/mod.rs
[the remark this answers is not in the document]
slowli: `verify()` is a method in `Transaction`, so verification of `RawTransaction`s is impossible as of now. This is going to change once the separation of transactions and messages is implemented.

exonum/src/events/internal.rs
```diff
@@ -30,53 +29,157 @@ pub struct InternalPart {
...
+            if tx.verify() {
+                let event = future::ok(InternalEvent::TxVerified(tx.raw().clone()));
+                Either::A(Self::send_event(event, internal_tx))
+            } else {
+                Either::B(future::ok(()))
+            }
+        })
+    }
+
+    pub fn run<E>(self, handle: Handle, verify_executor: E) -> impl Future<Item = (), Error = ()>
```
aleksuss: Add description to the public method pls

exonum/src/events/mod.rs
```diff
@@ -41,21 +43,25 @@ pub type SyncSender<T> = Wait<Sender<T>>;
...
     Timeout(NodeTimeout),
     /// Shutdown the node.
     Shutdown,
+    /// Transaction has been successfully verified.
+    TxVerified(RawTransaction),
 }
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
 pub enum InternalRequest {
```
aleksuss: Same here. Add description to the public enum.

slowli:
I've investigated a little bit a possibility to batch transaction verification:

```rust
fn bench_verify_transactions_thread_pool(b: &mut Bencher, &size: &usize) {
    use futures::{future, sync::mpsc, Stream};
    //use std::time::Duration;

    let transactions = gen_transactions(TRANSACTIONS_COUNT, size);
    let thread_pool = ThreadPoolBuilder::new().build();

    b.iter(|| {
        let (sender, receiver) = mpsc::channel(TRANSACTIONS_COUNT / 100);

        for chunk in transactions.chunks(100) {
            let sender = sender.clone();
            let chunk = chunk.to_vec();

            let future = future::lazy(move || {
                // 110 us is the approximate time to verify a single message on my PC
                // thread::sleep(Duration::from_micros(110 * 100));

                let messages: Vec<_> = chunk.iter()
                    .filter(|tx| tx.verify())
                    .map(|tx| tx.raw().clone())
                    .collect();

                sender.send(messages).map(drop).map_err(|_| panic!("cannot send"))
            });

            thread_pool.spawn(future);
        }

        drop(sender);
        let count = receiver.fold(0, |count, messages| Ok(count + messages.len()))
            .wait()
            .unwrap();
        assert_eq!(count, TRANSACTIONS_COUNT);
    })
}
```

It seems marginally better (~5-10% than a comparable code without batching), possibly due to wasting less resources on `mpsc` channel operations.

Interestingly, with `thread::sleep()` emulation the performance grows linearly with the number of threads, as one would expect; but with real verification, it's not (at least on my PC; the performance increases ~25% per added thread). So it may be the case that signature verification uses resources shared among CPU cores, or it is already parallelized somehow. Just guessing here.

exonum/benches/criterion/transactions.rs
```diff
@@ -12,33 +12,28 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-const TRANSACTIONS_COUNT: usize = 40_000;
-const SAMPLE_SIZE: usize = 3;
+const TRANSACTIONS_COUNT: usize = 1_000;
```
vldm: why we need this significant decrease of the transaction count?

exonum/benches/criterion/transactions.rs
```diff
@@ -264,27 +230,21 @@ fn bench_verify_transactions_simple(b: &mut Bencher, &size: &usize) {
...
+        core.run(verifier.send_all(&transactions)).unwrap();
+    });
+    verifier.join();
 }
 
 pub fn bench_verify_transactions(c: &mut Criterion) {
     crypto::init();
 
-    let parameters = (7..12).map(|i| pow(2, i)).collect::<Vec<_>>();
+    let parameters = (7..12).map(|i| 1 << i).collect::<Vec<_>>();
```
vldm: Yes, we know that previously in 90's bit shifting was faster than taking exponent of two, but can you revert your change in favour of readability

slowli: This was me ✋  I think `1 << i` is no less readable than `pow(2, i)`, and I'm not even a C/C++ guy. Strictly personal opinion though; I won't insist on it.

exonum/benches/criterion/transactions.rs
slowli: This is my change ✋ Unless there are reasonable arguments for huge transaction count (i.e., unless we _expect_ the timing growth with the transaction could to be super-linear or sub-linear), I see no particular reason to use it. The transaction count allowing for reasonable number of samples, and correspondingly higher accuracy could be more important. More specifically, I'd be against decreasing the sample size to `< 10`, which corresponds to `TRANSACTIONS_COUNT ~ 5_000` for the reasonable running time (~30s / bench on my laptop).

andreyluchin:
As by request of reviewers, benchmark results:
(dual-core i5 processor)
```
transactions/simple/size/128
                        time:   [65.316 ms 65.660 ms 66.203 ms]
                        thrpt:  [15.105 Kelem/s 15.230 Kelem/s 15.310 Kelem/s]
                        change: [-97.501% -97.474% -97.434%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high severe
transactions/simple/size/256
                        time:   [65.470 ms 65.675 ms 65.967 ms]
                        thrpt:  [15.159 Kelem/s 15.226 Kelem/s 15.274 Kelem/s]
                        change: [-97.643% -97.618% -97.597%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 20 measurements (15.00%)
  3 (15.00%) high severe
transactions/simple/size/512
                        time:   [66.314 ms 66.450 ms 66.598 ms]
                        thrpt:  [15.015 Kelem/s 15.049 Kelem/s 15.080 Kelem/s]
                        change: [-97.514% -97.497% -97.475%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) high mild
  1 (5.00%) high severe
transactions/simple/size/1024
                        time:   [67.724 ms 68.201 ms 68.879 ms]
                        thrpt:  [14.518 Kelem/s 14.663 Kelem/s 14.766 Kelem/s]
                        change: [-97.745% -97.671% -97.569%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) high mild
  1 (5.00%) high severe
transactions/simple/size/2048
                        time:   [71.661 ms 73.569 ms 77.230 ms]
                        thrpt:  [12.948 Kelem/s 13.593 Kelem/s 13.955 Kelem/s]
                        change: [-97.507% -97.399% -97.270%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) high mild
  1 (5.00%) high severe

transactions/event_loop/size/128
                        time:   [33.385 ms 33.571 ms 33.732 ms]
                        thrpt:  [29.645 Kelem/s 29.788 Kelem/s 29.954 Kelem/s]
                        change: [-99.534% -99.525% -99.516%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 20 measurements (15.00%)
  1 (5.00%) low mild
  2 (10.00%) high mild
transactions/event_loop/size/256
                        time:   [33.782 ms 33.988 ms 34.272 ms]
                        thrpt:  [29.178 Kelem/s 29.422 Kelem/s 29.602 Kelem/s]
                        change: [-99.558% -99.546% -99.528%] (p = 0.00 < 0.05)
                        Performance has improved.
transactions/event_loop/size/512
                        time:   [34.171 ms 34.316 ms 34.522 ms]
                        thrpt:  [28.967 Kelem/s 29.141 Kelem/s 29.264 Kelem/s]
                        change: [-99.536% -99.524% -99.511%] (p = 0.00 < 0.05)
                        Performance has improved.
transactions/event_loop/size/1024
                        time:   [34.946 ms 35.166 ms 35.420 ms]
                        thrpt:  [28.233 Kelem/s 28.437 Kelem/s 28.615 Kelem/s]
                        change: [-99.554% -99.542% -99.531%] (p = 0.00 < 0.05)
                        Performance has improved.
transactions/event_loop/size/2048
                        time:   [42.818 ms 46.612 ms 52.350 ms]
                        thrpt:  [19.102 Kelem/s 21.454 Kelem/s 23.355 Kelem/s]
                        change: [-99.457% -99.389% -99.307%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high severe
```

exonum/benches/criterion/transactions.rs
vldm: Ok, thanks, can we add some docs about this constant, that nobody will ask a question in future?

alekseysidorov: In  my opinion large amount of transaction helps to decrease influence of implicit  overheads.

slowli: @alekseysidorov Would you be more satisfied with the following settings?

```rust
const TRANSACTIONS_COUNT: usize = 4_000;
const SAMPLE_SIZE: usize = 10;
```

I'm not sure decreasing sample size further will do us good; the decrease from 20 to 10 increases st.dev. of results ~2 times if I understand the measuring methodology correctly.

To be frank, I don't think the benchmark in the present form has significant non-linear overheads (it *may* have had them when it included writing transactions to the DB), but I didn't investigate this question.

PR #674, opened 2018-05-07
Fix deserialize for proofs [ECR-1388]

alekseysidorov:
Currently deserialize for MapProofs are broken.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Fixed bug with shutdown requests handling. (#666)
 
+- Fixed deserialization of the `MapProof` data structure. (#674)
+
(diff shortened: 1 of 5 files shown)


slowli:
My fault :disappointed: Fortunately, the fix seems to be quite easy:

```patch
--- a/exonum/src/storage/proof_map_index/proof.rs
+++ b/exonum/src/storage/proof_map_index/proof.rs
@@ -81,7 +81,11 @@ impl<'de> Deserialize<'de> for ProofPath {
                     }
                 }

-                Ok(ProofPath::new(&bytes).prefix(len as u16))
+                Ok(if len == 8 * KEY_SIZE {
+                    ProofPath::new(&bytes)
+                } else {
+                    ProofPath::new(&bytes).prefix(len as u16)
+                })
             }
         }
```

Also, I'd be wary of implementing `PartialEq` for `MapProof`: the proofs with different order of `entries` are still technically correct and logically the same, for example.

PR #675, opened 2018-05-07
Core private API CORS support [ECR-1261]

VladYermakov:
Added `CorsMiddleware` to `create_private_api_handler` by `create_public_api_handler`'s example. 
Now method `create_private_api_handler` takes `config: &NodeApiConfig` parameter.
Do this changes are enough to support CORS in private API and do we need to test it and how?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum
 
+- Private API now support CORS. (#675)
+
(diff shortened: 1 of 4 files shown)


DENMROOT:
+1 for @slowli thoughts on separate Private API Cors config params - it make sense.
@VladYermakov - we are testing it through dashboard frontend which is started on separate IP:PORT - allowing us to see CORS in action

VladYermakov:
@slowli @DENMROOT Now we have separate parameters `PUBLIC_ALLOW_ORIGIN` and `PRIVATE_ALLOW_ORIGIN` for CORS settings in `NodeApiConfig`.

PR #680, opened 2018-05-10
Fixed calculation of max byzantine nodes amount [ECR-1280]

popzxc:
Initial problem: with only 3 nodes started, if two of them gets disconnected and one restores after some time, it can't reach working node round.

The reason: incorrect max byzantine nodes amount calculation. It was calculated as (1/3) of all validators, while it should have been (1/3)-1.

After this fix, reconnected node can reach actual round without any problem.

---

I've tried to write a test for this behavior, but didn't succeed. As far as I can see, sandbox doesn't provide functionality to emulate node disconnect. Is there any way to emulate such a behavior?

### DIFF
--- a/exonum/src/node/state.rs
+++ b/exonum/src/node/state.rs
@@ -586,7 +586,10 @@ impl State {
             }
             *known_round = round;
         }
-        let max_byzantine_count = self.validators().len() / 3;
+
+        // At max we can have (1/3) - 1 byzantine nodes.
+        // (1/3) is calculated via rounded up integer division.
+        let max_byzantine_count = (self.validators().len() + 2) / 3 - 1;
         if self.validators_rounds.len() <= max_byzantine_count {
             trace!("Count of validators, lower then max byzantine count.");
             return None;


pavel-mukhanov:
@popzxc maybe it's possible to emulate such behaviour with `events::tests::TestHandler`

popzxc:
@pavel-mukhanov As far as I can see, it doesn't provide consensus-related interfaces. And for me it seems counter-intuitive to test consensus in events module.

exonum/src/node/state.rs
```diff
@@ -586,7 +586,10 @@ impl State {
             }
             *known_round = round;
         }
-        let max_byzantine_count = self.validators().len() / 3;
+
+        // At max we can have (1/3) - 1 byzantine nodes.
```
slowli: Nitpick: I don't quite understand the arithmetic in the comment, although the code is correct. Suppose `N` is the total number of nodes, `f` is the number of Byzantine nodes. By the BFT security assumptions, `N >= 3*f + 1`, i.e., `f <= (N - 1)/3` (which, in turn, is equivalent to `f <= (N + 2)/3 - 1`, which is used in the code).

popzxc: Well, I tried to follow official [documentation](https://exonum.com/doc/advanced/consensus/specification/) which tells, that we should have +2/3 non-Byzantine nodes (which is equivalent to -1/3 Byzantine ones). So here I just tried to simplify it as (1/3) - 1. Probably, that's not quite correct indeed.

popzxc:
Regarding testing: I guess it's possible. The problem is we don't have existing unit tests for `State` yet and developing may take some time. I think it's a good idea to create a separate issue for it. 

About nitpicks: 
> IMO, naming mutating methods get_something is misleading; a more fitting name in this case would be update_validator_round.
Function itself does not update anything, so I guess `get_*` here is better, since this function is pure.

slowli:
@popzxc 

> Function itself does not update anything

This is wrong; see https://github.com/exonum/exonum/blob/0940ef123f4a8b1e020836d85038594eb09b6917/exonum/src/node/state.rs#L575-L587

The first and last lines modify an entry in `self.validators_rounds`.

popzxc:
Ah, I see. Then you're right, and more than that: I think it will be better to additionally highlight this. At a glance this code doesn't look like it's mutating something internal.

PR #681, opened 2018-05-11
Fix node round race [ECR-1280]

popzxc:
PR #680 didn't solved problem completely, since `add_round_timeout` wasn't called after `jump_to_round`.
This PR fixes it.

Also findings from the previous PR were fixed.

### DIFF
--- a/exonum/src/node/state.rs
+++ b/exonum/src/node/state.rs
@@ -568,9 +568,10 @@ impl State {
     }
 
     /// Updates known round for a validator and returns
-    /// a new actual round if at least one non byzantine validators are on a higher round.
+    /// a new actual round if at least one non byzantine validators is guaranteed to be on a higher round.
     /// Otherwise returns None.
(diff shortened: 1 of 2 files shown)


exonum/src/node/state.rs
```diff
@@ -568,9 +568,10 @@ impl State {
     }
 
     /// Updates known round for a validator and returns
-    /// a new actual round if at least one non byzantine validators are on a higher round.
+    /// a new actual round if at least one non byzantine validators is guaranteed to be on a higher round.
     /// Otherwise returns None.
-    pub fn get_actual_round(&mut self, id: ValidatorId, round: Round) -> Option<Round> {
+    pub fn update_validator_round(&mut self, id: ValidatorId, round: Round) -> Option<Round> {
```
vldm: btw: this method does not really "update" validator round

popzxc: It updates known round, as @slowli mentioned at the PR #680.

PR #683, opened 2018-05-15
Transfer benches to criterion-rs

vldm:
Using criterion-rs allows us to bench on stable rust, and collect better statistic proven data.


exonum/benches/criterion/block.rs
```diff
@@ -0,0 +1,235 @@
...
+        blockchain.merge(patch).unwrap();
+    }
+
+    let txs = prepare_txs(&mut blockchain, 100, 1000);
+
+    c.bench(
+        "timestamping",
+        Benchmark::new("timestamping", move |b| {
+            b.iter(|| execute_block(&blockchain, 100, &txs))
+        }).sample_size(16),
```
popzxc: Nitpick: Why 16 and not another number? I suggest to move out magic numbers to variables or constants with a brief explanation why this value has been chosen and not another. This comments applies also to other magic numbers usages.

exonum/benches/criterion/crypto.rs
```diff
@@ -0,0 +1,59 @@
...
+    let data = (0..count).map(|x| (x % 255) as u8).collect::<Vec<u8>>();
+    b.iter(|| hash(&data))
+}
+
+pub fn bench_crypto(c: &mut Criterion) {
+    ::exonum::crypto::init();
+
+    c.bench(
+        "hash",
+        ParameterizedBenchmark::new("hash", bench_hash, (6..16).map(|i| pow(2, i)))
```
popzxc: Nitpick: It's not obvious what `(6..16).map(|i| pow(2, i))` stands for.

popzxc:
In general LGTM.

vldm:
> We also have events::benches

@pavel-mukhanov 
We can transfer only benches with public api to criterion-rs

exonum/benches/criterion/block.rs
vldm: Because execute_block is not really "micro benchmark"
And execute it as regular benches, with 100 samples, lead to relatively big testing time (10 min on my mac).

This number does not contain any additional information, but is simply a compromise between long lead times and test quality.

popzxc: I guess such a comment will look just great in sources :smiley:

exonum/benches/criterion/crypto.rs
vldm: It's an argument for test, i am not sure is it should be obvious. We interesting about speed of methods with argument in between some size.

64 - is relatively small message
65536 - is relatively big

popzxc: Yep, I got that (after reading `critertion` source code mainly). What I meant is that it would be nice to add explaining remark to sources. I guess, it will make work with those lines in 2-3 months less confusing.

PR #690, opened 2018-05-18
Rename validate_configuration method

DarkEld3r:
Also `txs_block_limit` check has been added.

### DIFF
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -87,14 +87,28 @@ impl ConsensusConfig {
     /// Default value for max_message_len.
     pub const DEFAULT_MAX_MESSAGE_LEN: u32 = 1024 * 1024; // 1 MB
 
-    /// Checks if propose timeout is less than round timeout. Warns if fails.
+    /// Produces warnings if configuration contains non-optimal values.
+    ///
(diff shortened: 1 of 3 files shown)


exonum/src/blockchain/config.rs
```diff
@@ -87,16 +87,30 @@ impl ConsensusConfig {
...
     pub const DEFAULT_MAX_MESSAGE_LEN: u32 = 1024 * 1024; // 1 MB
 
-    /// Checks if propose timeout is less than round timeout. Warns if fails.
+    /// Produces warnings if configuration contains non-optimal values.
+    ///
+    /// Validation for logical correctness is performed in the `StoredConfiguration::try_deserialize`
+    /// method, but some values can decrease consensus performance.
     #[doc(hidden)]
-    pub fn validate_configuration(&self) {
+    pub fn warn_nonoptimal(&self) {
```
popzxc: IMHO `warn_nonoptimal` doesn't reflect the essence of this method. It doesn't simply warn, it warns only if something is wrong. Maybe something like `warn_if_nonoptimal`?

DarkEld3r:
Sorry for the mess, the test is finally fixed. :sweat_smile:

PR #694, opened 2018-05-18
Replace constant waiting by message (mpsc)

DarkEld3r:
Constant (200ms) timeout has been replaced by (somewhat) smarter behaviour: service signals about transactions processing using `mpsc`.

### DIFF
--- a/exonum/tests/balance_service.rs
+++ b/exonum/tests/balance_service.rs
@@ -85,7 +85,7 @@ pub mod contracts {
 }
 
 pub mod service {
-    use exonum::blockchain::{Service, Transaction, TransactionSet};
+    use exonum::blockchain::{Service, Transaction, TransactionSet, ServiceContext, Schema};
     use exonum::{encoding, messages::RawTransaction};
     use exonum::crypto::{gen_keypair, Hash};
     use exonum::storage::{Database, Fork, MemoryDB, Snapshot};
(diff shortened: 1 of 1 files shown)


DarkEld3r:
@popzxc I propose to preserve ECR-1410: ideally the test still should be rewritten using sandbox (probably sandbox should be deeply refactored too).

Additionally, CircleCI still failed:
```
thread '<unnamed>' panicked at 'Can't timeout', exonum/src/events/internal.rs:50:42
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Any', libcore/result.rs:945:5
test service::test_duplicated_transaction ... FAILED
```
:confused:

exonum/tests/balance_service.rs
```diff
@@ -159,7 +180,9 @@ pub mod service {
         api_tx.send(tx_copy).unwrap();
```
vldm: This test was about double tx execution, probably cloned tx was executed at second block?
But now test wait only one block.

May be we should write test as follow:
```
api_tx.send(tx_orig);
receiver.wait_for_message();
api_tx.send(tx_orig);
receiver.wait_for_message();
```

Additionally just after `api_tx.send`, but before tx was executed, there possible state where `height > 0`.

DarkEld3r:
Closed as outdated, should be fixed as part of ECR-1410 issue.

PR #701, opened 2018-05-23
Service metainformation endpoint [ECR-792]

VladYermakov:
Api now has `/v1/meta/` endpoint for metainformation about core and services


slowli:
@VladYermakov @DarkEld3r The proposed endpoint has a high degree of similarity with [the private `system/v1/network` endpoint](https://github.com/exonum/exonum/blob/6c9605e651d973c9735e0ec17c5b41d8d7d2533d/exonum/src/api/private/system.rs#L165-L172). Is this intersection intended? Maybe, reuse some code? Also, the problem description in Jira mentions returning service versions, which is not implemented as far as I can tell.

exonum/src/api/public/system.rs
```diff
@@ -25,6 +25,12 @@ struct MemPoolInfo {
     pub size: usize,
 }
 
+#[derive(Serialize, Deserialize)]
+struct MetaInfo<'a> {
```
DarkEld3r: I suppose this structure can be removed.

testkit/tests/system_api.rs
```diff
@@ -40,3 +42,14 @@ fn test_user_agent_info() {
...
     assert_eq!(info, expected);
 }
+
+#[test]
+fn test_network() {
+    let testkit = TestKitBuilder::validator().with_validators(2).create();
+    let api = testkit.api();
+    let info: NodeInfo = api.get_private(ApiKind::System, "/v1/network");
+
+    assert_eq!(info.core_version, option_env!("CARGO_PKG_VERSION").map(|ver| ver.to_owned()));
```
slowli: Late-to-the-party comment: As far as I understand, this assumes that the testkit version is equal to the core version, which doesn't *always* hold. I'd check `assert!(info.core_version.is_some())` instead.

PR #702, opened 2018-05-23
Noise handshake tests

pavel-mukhanov:
Check correctness of different steps of noise handshake implementation.
Due to selected handshake protocol(`Noise_XX`), we have two steps where we send 
messages to handshake listener.

To check we send some corrupted messages to handshake listener and check 
returned `Result`.

```
let res = send_handshake(&addr, HandshakeStep::One(1, EMPTY_MESSAGE));
assert!(res.is_err());

let res = send_handshake(&addr, HandshakeStep::Two(2, EMPTY_MESSAGE));
assert!(res.is_err());

let res = send_handshake(&addr, HandshakeStep::One(1, SMALL_MESSAGE));
...
```

On each step, the listener shouldn't panic.

### Remaining TODO:
- [x]  Get rid from `TcpStream` in tests. (Will do in the separate task)


exonum/src/events/noise/mod.rs
```diff
@@ -112,3 +114,174 @@ fn write_handshake_msg(
...
+
+    const EMPTY_MESSAGE: usize = 0;
+    const SMALL_MESSAGE: usize = NOISE_MIN_HANDSHAKE_MESSAGE_LENGTH - 1;
+    const BIG_MESSAGE: usize = NOISE_MAX_MESSAGE_LENGTH + 1;
+
+    impl Default for HandshakeParams {
+        fn default() -> Self {
+            let (public_key, secret_key) = gen_keypair_from_seed(&Seed::new([0; 32]));
+            HandshakeParams {
+                max_message_len: 1024,
```
vldm: `max_message_len` is smaller than `NOISE_MAX_MESSAGE_LENGTH`, is it ok?

exonum/src/events/noise/mod.rs
```diff
@@ -112,3 +114,174 @@ fn write_handshake_msg(
...
+        let addr: SocketAddr = "127.0.0.1:45002".parse().unwrap();
+        let addr2 = addr.clone();
+
+        // Wait until listener is initialized.
+        thread::sleep(Duration::from_millis(500));
+
+        thread::spawn(move || run_handshake_listener(&addr2));
+
+        let res = send_handshake(&addr, HandshakeStep::One(1, EMPTY_MESSAGE));
+        assert!(res.is_err());
```
vldm: The logic of send_handshake is too complicated, is it possible to check for a concrete error?

And this would be good if you will test `max_message_len` too. 
And probably `HandshakeStep::One` \  `HandshakeStep::Two` with valid message size should behave somehow too.

exonum/src/events/noise/mod.rs
```diff
@@ -112,3 +114,174 @@ fn write_handshake_msg(
...
+                HandshakeStep::Default => NoiseHandshake::send(&params, sock),
+                _ => noise_send_handshake_with_error(&params, sock, step),
+            })
+            .map(|_| ())
+            .map_err(into_other);
+
+        core.run(stream)
+    }
+
+    fn noise_send_handshake_with_error(
```
vldm: Could you briefly explain in comment what aspects this helpers functions adds.

I spent about 15 min to understand test logic.

exonum/src/events/noise/mod.rs
```diff
@@ -112,3 +114,174 @@ fn write_handshake_msg(
...
+        Default,
+        One(u8, usize),
+        Two(u8, usize),
+    }
+
+    const EMPTY_MESSAGE: usize = 0;
+    const SMALL_MESSAGE: usize = NOISE_MIN_HANDSHAKE_MESSAGE_LENGTH - 1;
+    const BIG_MESSAGE: usize = NOISE_MAX_MESSAGE_LENGTH + 1;
+
+    impl Default for HandshakeParams {
```
vldm: The `Default` trait implementation for external type inside unit test scope looking strange.
May be it would good to move it inside static method?  `HandshakeParams::default_test_params()` looks better for me. Or you can just use constructor inside `send_handshake`/ `run_handshake_listener` methods.

exonum/src/events/noise/mod.rs
```diff
@@ -112,3 +114,174 @@ fn write_handshake_msg(
...
+                        NoiseHandshake};
+    use events::error::{into_other, log_error};
+    use events::codec::MessagesCodec;
+    use crypto::{gen_keypair_from_seed, Seed};
+    use std::time::Duration;
+
+    #[derive(Debug, PartialEq, Copy, Clone)]
+    pub enum HandshakeStep {
+        Default,
+        One(u8, usize),
```
vldm: May be more verbose, `ClearText` \ `EncryptedText` variants would be better?
Or `BeforeHandshake` \ `AfterHandshake`.

exonum/src/events/noise/mod.rs
```diff
@@ -173,6 +174,9 @@ mod tests {
             run_handshake_listener(&addr2).unwrap();
         });
 
+        // Wait until listener is initialized.
+        thread::sleep(Duration::from_millis(500));
```
DarkEld3r: :cry:

exonum/src/events/noise/mod.rs
```diff
@@ -52,16 +52,17 @@ fn listen_handshake(stream: TcpStream, params: &HandshakeParams) -> HandshakeRes
...
-        write_handshake_msg(&mut noise)
-            .and_then(|(len, buf)| write(stream, &buf, len))
-            .and_then(|(stream, _msg)| read(stream))
-            .and_then(move |(stream, msg)| {
-                let _buf = noise.read_handshake_msg(&msg);
-                let noise = noise.into_transport_mode()?;
-                let framed = stream.framed(MessagesCodec::new(max_message_len, noise));
-                Ok(framed)
-            })
+        read_handshake_msg(&msg, &mut noise).and_then(move |_| {
```
slowli: Slightly off-topic: Would it be possible to refactor `read_handshake_msg` / `write_handshake_msg` so that they deal directly with streams? For now, they introduce boilerplate (`write` after `write_handshake_msg`, `read` before `read_handshake_msg` and are not really asynchronous.

exonum/src/events/noise/mod.rs
```diff
@@ -112,3 +114,190 @@ fn write_handshake_msg(
...
+            HandshakeParams {
+                max_message_len: 1024,
+                public_key,
+                secret_key,
+            }
+        }
+    }
+
+    #[test]
+    fn test_noise_normal_handshake() {
```
slowli: Does this test that the nodes can successfully send messages after the handshake? I feel this is the most important result for this test.

exonum/src/events/noise/mod.rs
```diff
@@ -112,3 +114,190 @@ fn write_handshake_msg(
...
+
+        let (sender, receiver) = mpsc::channel(0);
+        let receiver = add_timeout_millis(receiver, 500);
+        thread::spawn(move || run_handshake_listener(&addr2, sender));
+
+        // Use first handshake only to connect.
+        let _res = send_handshake(&addr, HandshakeStep::Default);
+        receiver.wait().next();
+
+        let res = send_handshake(&addr, HandshakeStep::One(1, EMPTY_MESSAGE));
```
slowli: Are you sure that multiple invocations of `send_handshake` won't mess up the other side of the channel? Resetting the channel after each handshake could be cleaner.

exonum/src/events/noise/mod.rs
```diff
@@ -112,3 +114,190 @@ fn write_handshake_msg(
...
+        receiver.wait().next();
+
+        let res = send_handshake(&addr, HandshakeStep::Default);
+        println!("Send handshake result {:?}", res);
+        assert!(res.is_ok());
+    }
+
+    #[test]
+    fn test_noise_handshake_errors() {
+        let addr: SocketAddr = "127.0.0.1:45002".parse().unwrap();
```
slowli: One more crazy idea: maybe, it makes sense to abstract (at least for tests' sake) `TcpStream` into `AsyncRead` / `AsyncWrite` in order to not bind to a real TCP socket where it's not necessary?

exonum/src/events/noise/mod.rs
pavel-mukhanov: Ok, because `max_message_len` is not relevant for the handshake.

exonum/src/events/noise/mod.rs
pavel-mukhanov: After refactoring this test is no longer needed, because sending messages after handshake are 
tested in `events::tests`.

exonum/src/events/noise/mod.rs
pavel-mukhanov: Good idea, but it's difficult to pass `noise` into the future because it needs to have static lifetime. It also doesn't implement the `Clone` trait. For example:

```rust
fn read_handshake_msg_stream(&self,
    stream: TcpStream,
    noise: &'static mut NoiseWrapper,
) -> Box<Future<Item = (usize, Vec<u8>), Error = io::Error>> {
    let res = read(stream).and_then(move |(stream, msg)| {
        noise.read_handshake_msg(&msg).map_err(|e| e.into())
    });

    Box::new(res)
}
```
Maybe it will be possible using `RefCell` of something, but I don't think it will be more readable.

exonum/src/events/noise/mod.rs
pavel-mukhanov: I removed this method after refactoring handshake.

exonum/src/events/noise/mod.rs
slowli: I've had [something like this](https://gist.github.com/slowli/cb3bd293744d5f97287490da791e2ff4) in mind (it compiles, but I skipped the tests). In particular, the code like

```rust
    let framed = channel.read_handshake_msg(stream)
        .and_then(|(stream, channel)| channel.write_handshake_msg(stream))
        .and_then(|(stream, channel)| channel.read_handshake_msg(stream))
        .and_then(|(stream, channel)| channel.finalize(stream));
```
looks much more discernible to me than the current implementation.

Now that I think about it, it may make sense to mock `Handshake`, not `HandshakeChannel` (i.e., make `read_handshake_msg` / `write_handshake_msg` internals of `NoiseHandshake`, and get rid of `HandshakeChannel` altogether).

pavel-mukhanov:
Mock `TcpStream` will be added in separate PR.

exonum/src/events/noise/mod.rs
```diff
@@ -40,80 +42,144 @@ pub struct HandshakeParams {
...
 }
 
-fn read(sock: TcpStream) -> Box<Future<Item = (TcpStream, Vec<u8>), Error = io::Error>> {
+fn read<S: AsyncRead + 'static>(sock: S) -> impl Future<Item = (S, Vec<u8>), Error = io::Error> {
     let buf = vec![0u8; HANDSHAKE_HEADER_LENGTH];
-    Box::new(
-        read_exact(sock, buf)
-            .and_then(|(stream, msg)| read_exact(stream, vec![0u8; msg[0] as usize])),
-    )
+    read_exact(sock, buf).and_then(|(stream, msg)| read_exact(stream, vec![0u8; msg[0] as usize]))
```
popzxc: It's not obvious what's going on here. Could you please add an explaining comment?

exonum/src/events/noise/mod.rs
```diff
@@ -40,80 +42,144 @@ pub struct HandshakeParams {
...
+fn write<S: AsyncWrite + 'static>(
+    sock: S,
     buf: &[u8],
     len: usize,
-) -> Box<Future<Item = (TcpStream, Vec<u8>), Error = io::Error>> {
+) -> impl Future<Item = (S, Vec<u8>), Error = io::Error> {
     let mut message = vec![0u8; HANDSHAKE_HEADER_LENGTH];
+
+    if len > NOISE_MAX_HANDSHAKE_MESSAGE_LENGTH {
+        return Either::A(err(io::Error::new(
```
popzxc: I don't really like `Either` implementation with `A` and `B` branches. Can we somehow use [one](https://docs.rs/futures/0.2.1/futures/future/enum.Either.html) with `Left` and `Right`?

exonum/src/events/noise/mod.rs
```diff
@@ -40,80 +42,144 @@ pub struct HandshakeParams {
     pub max_message_len: u32,
 }
 
+pub trait Handshake {
+    fn listen<S: AsyncRead + AsyncWrite + 'static>(self, stream: S) -> HandshakeResult<S>;
+
+    fn send<S: AsyncRead + AsyncWrite + 'static>(self, stream: S) -> HandshakeResult<S>;
+
+    fn read_handshake_msg<S: AsyncRead + 'static>(
```
slowli: I'm against putting any methods other than `listen` and `send`, to the `Handshake` interface; gives too much details about the implementation.

exonum/src/events/noise/tests.rs
```diff
@@ -84,3 +99,253 @@ fn test_converted_keys_handshake() {
...
+
+#[test]
+#[should_panic(expected = "WrongMessageLength(0)")]
+fn test_noise_handshake_errors_ee_empty() {
+    let addr: SocketAddr = "127.0.0.1:45003".parse().unwrap();
+    let step = HandshakeStep::EphemeralKeyExchange(1, EMPTY_MESSAGE);
+    let (sender_err, listener_err) = wait_for_handshake_result(&addr, step);
+
+    assert!(sender_err.is_err());
+    listener_err.unwrap()
```
popzxc: Maybe use [`assert_matches!(..)`](https://docs.rs/assert_matches/1.2.0/assert_matches/macro.assert_matches.html) here (and below)? 
It's already used in testkit, so I guess bringing that dev-dependency isn't a big evil.

exonum/src/events/noise/tests.rs
```diff
@@ -84,3 +99,253 @@ fn test_converted_keys_handshake() {
...
+// We need check result from both: sender and responder.
+fn wait_for_handshake_result(
+    addr: &SocketAddr,
+    step: HandshakeStep,
+) -> (IoResult<()>, IoResult<()>) {
+    let addr2 = addr.clone();
+
+    let (tx, rx) = mpsc::channel(0);
+    let (err_tx, err_rx) = mpsc::channel::<IoResult<()>>(0);
+    let rx = add_timeout_millis(rx, 500);
```
slowli: I'm not sure why this construction is required, as well as the "only to connect" handshake. It seems that the problem could be solved by `sleep`ing after starting the listener (there may exist other solutions, such as channel-based signaling).

exonum/src/events/noise/mod.rs
pavel-mukhanov: 👌right after we move to futures 0.2.

exonum/src/events/noise/mod.rs
```diff
@@ -40,80 +41,109 @@ pub struct HandshakeParams {
...
+        })
+    }
+
+    fn write_handshake_msg<S: AsyncWrite + 'static>(
+        mut self,
+        stream: S,
+    ) -> impl Future<Item = (S, Self), Error = io::Error> {
+        done(self.noise.write_handshake_msg())
+            .map_err(|e| e.into())
             .and_then(|(len, buf)| write(stream, &buf, len))
```
slowli: Is it possible to rewrite `noise.write_handshake_msg()` / `noise.write` / `write` in order for them to return (or accept, in case of `write`), `Vec<u8>` instead of a length-buffer tuple? Maybe, it's my problem not being as familiar with C-style programming, but imo this code style significantly complicates understanding of what's going on.

exonum/src/events/noise/mod.rs
```diff
@@ -40,80 +41,109 @@ pub struct HandshakeParams {
...
+        )));
+    }
 
-fn write_handshake_msg(
-    noise: &mut NoiseWrapper,
-) -> Box<Future<Item = (usize, Vec<u8>), Error = io::Error>> {
-    let res = noise.write_handshake_msg();
-    Box::new(done(res.map_err(|e| e.into())))
+    let mut message = vec![len as u8; HANDSHAKE_HEADER_LENGTH];
+    message.extend_from_slice(&buf[0..len]);
```
slowli: I wonder if this would work without reallocation. Are two invocations of `write_all` equivalent to a single invocation on the buffer concatenation?

exonum/src/events/noise/wrapper.rs
```diff
@@ -134,17 +138,13 @@ impl NoiseWrapper {
 
     fn read(&mut self, input: &[u8], len: usize) -> Result<(usize, Vec<u8>), NoiseError> {
```
slowli: (Slightly) off-topic: isn't it possible to guess the length of the output from the length of the input (like, the `input.len()` is the upper bound)? I'd like to see a signature here like

```rust
fn read(&mut self, input: &[u8]) -> Result<Vec<u8>, NoiseError>;
```

since this one looks too close to C for (at least my) comfort.

exonum/src/events/noise/wrapper.rs
```diff
@@ -24,7 +25,9 @@ use events::noise::HandshakeParams;
 pub const NOISE_MAX_MESSAGE_LENGTH: usize = 65_535;
 pub const TAG_LENGTH: usize = 16;
```
slowli: Off-topic: The code in `encrypt_msg` and `decrypt_msg` (which uses this constant, here is the link) is nearly incomprehensible to me. I think at the very least it should be augmented with asserts and comments, and tested.

exonum/src/events/noise/wrapper.rs
```diff
@@ -60,6 +63,12 @@ impl NoiseWrapper {
     }
 
     pub fn read_handshake_msg(&mut self, input: &[u8]) -> Result<(usize, Vec<u8>), NoiseError> {
+        if input.len() < NOISE_MIN_HANDSHAKE_MESSAGE_LENGTH
+            || input.len() > NOISE_MAX_MESSAGE_LENGTH
```
slowli: ...And this restriction seems to be unreachable by design if the method is invoked from `NoiseHandshake::read_handshake_msg`.

exonum/src/events/noise/mod.rs
```diff
@@ -40,80 +41,109 @@ pub struct HandshakeParams {
...
+        let framed = self.write_handshake_msg(stream)
+            .and_then(|(stream, handshake)| handshake.read_handshake_msg(stream))
+            .and_then(|(stream, handshake)| handshake.write_handshake_msg(stream))
+            .and_then(|(stream, handshake)| handshake.finalize(stream));
+        Box::new(framed)
+    }
 }
 
-fn read(sock: TcpStream) -> Box<Future<Item = (TcpStream, Vec<u8>), Error = io::Error>> {
+fn read<S: AsyncRead + 'static>(sock: S) -> impl Future<Item = (S, Vec<u8>), Error = io::Error> {
```
slowli: It could make sense to make the names more comprehensive for `read` and `write`, as they relate to Noise handshake messages specifically. Just an idea: maybe, create a newtype `HandshakeMessage(Vec<u8>)` and implement `read` and `write` as its methods?

exonum/src/events/noise/wrapper.rs
```diff
@@ -60,6 +63,12 @@ impl NoiseWrapper {
     }
 
     pub fn read_handshake_msg(&mut self, input: &[u8]) -> Result<(usize, Vec<u8>), NoiseError> {
+        if input.len() < NOISE_MIN_HANDSHAKE_MESSAGE_LENGTH
```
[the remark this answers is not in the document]
pavel-mukhanov: Unfortunately no, `snow` will panic is this case.

exonum/src/events/noise/wrapper.rs
pavel-mukhanov: Agree, I will add a task for that.

exonum/src/events/noise/mod.rs
```diff
@@ -40,80 +41,109 @@ pub struct HandshakeParams {
...
     buf: &[u8],
     len: usize,
-) -> Box<Future<Item = (TcpStream, Vec<u8>), Error = io::Error>> {
-    let mut message = vec![0u8; HANDSHAKE_HEADER_LENGTH];
-    LittleEndian::write_u16(&mut message, len as u16);
-    message.extend_from_slice(&buf[0..len]);
-    Box::new(write_all(sock, message))
-}
+) -> impl Future<Item = (S, Vec<u8>), Error = io::Error> {
+    if len > NOISE_MAX_HANDSHAKE_MESSAGE_LENGTH {
```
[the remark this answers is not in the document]
pavel-mukhanov: Yes, such message can only be received in tests. Will add the assertion.

exonum/src/events/noise/mod.rs
pavel-mukhanov: It's possible to use two `write_all` invocations.

pavel-mukhanov:
I fixed all test related comments, comments related to noise handshake implementation will be fixed in separate PR.

PR #703, opened 2018-05-23
Apply 1.26 formatting

DarkEld3r:


pavel-mukhanov:
LGTM, and what about `clippy` version update?

exonum/src/blockchain/schema.rs
```diff
@@ -48,27 +48,23 @@ define_names!(
     CONSENSUS_ROUND => "consensus_round";
 );
 
-encoding_struct! (
```
ozkriff: this indentation looks weird

ozkriff: maybe it wants {} braces?..

pavel-mukhanov: I think we should submit a bug to `rustfmt`. Currently, I didn't find any issues related to this in their repository.

DarkEld3r:
@pavel-mukhanov I prefer to fix Clippy in a separate pull request. 🙃

exonum/src/encoding/float.rs
```diff
@@ -369,12 +367,10 @@ mod tests {
     #[test]
     #[allow(dead_code)]
     fn test_f64_struct() {
-        encoding_struct!(
-            struct Msg {
-                single_float: F64,
-                vec: Vec<F64>,
-            }
-        );
+        encoding_struct!(struct Msg {
```
vitvakatu: Ouch, for what reason they started to format code inside macros? Looks really awful

exonum/src/blockchain/schema.rs
slowli: > maybe it wants {} braces?..

I think this may be the case; see [the formatting RFCs](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/statements.md#macros-in-statement-position):

> A macro use in statement position should use parentheses or square brackets as delimiters and should be terminated with a semi-colon.

So it's likely that the formatter thinks `encoding_struct!(...)` is a statement, not an item.

exonum/src/encoding/float.rs
DarkEld3r: Personally, I like the idea. :upside_down_face:

DarkEld3r:
`encoding_struct` formatting is fixed. :sweat_smile:

aleksuss:
@DarkEld3r I suppose you should to update clippy in this PR as well.

popzxc:
IMHO updating clippy in separate PR will cause some time interval when we have to use new fmt but old clippy which is kind of weird.

DarkEld3r:
@popzxc I didn't see a problem in that, but then I realized that we build benchmarks with nightly rust. I'm going to update Clippy as well.

DarkEld3r:
Finally fixed. :face_with_head_bandage:

PR #705, opened 2018-05-28
Snow wrapper for sodiumoxide [ECR-1372]

popzxc:


slowli:
Could you please add a test for DH implementation, like [this one](https://github.com/mcginty/snow/blob/2b07c4d04afe923ac4a1dd9a07dde719074f5141/src/wrappers/crypto_wrapper.rs#L379-L389)? I'm not sure that the current implementation is correct given that [DH is performed on `Curve25519`](https://download.libsodium.org/doc/advanced/scalar_multiplication.html#algorithm-details), which [is related](https://download.libsodium.org/doc/advanced/ed25519-curve25519.html) but distinct from the curve used for digital signatures (`Ed25519`). Also, I wonder if [the `kx` module from libsodium](https://download.libsodium.org/doc/key_exchange/) could be used with Noise - seems like a safer version of key exchange.

popzxc:
Well, I adapted this test and it passes (for now: at least on my machine).
What about usage of `kx`: I suggest to create a separate activity for this.

exonum/src/events/noise/sodium_resolver.rs
```diff
@@ -0,0 +1,244 @@
...
+            .expect("Can't construct public key for Dh25519");
+        let result =
+            sodium_curve25519::scalarmult(&self.privkey, &pubkey).expect("Can't calculate dh");
+
+        out[..32].copy_from_slice(&result[0..32]);
+    }
+}
+
+// Chacha20poly1305 cipher.
+
```
pavel-mukhanov: Useless space.

pavel-mukhanov:
Is it still work in progress?

popzxc:
Yep: I don't want to take this label until `exonum_sodiumoxide` is updated on `crates.io`.

slowli:
Sorry for nagging again: looking at libsodium code, the conversion between Ed25519 and Curve25519 *private* keys [looks trivial](https://github.com/jedisct1/libsodium/blob/569778b517496861a3880e0e690973bf08a52e08/src/libsodium/crypto_sign/ed25519/ref10/keypair.c#L73-L91), but one between *public* keys [does not](https://github.com/jedisct1/libsodium/blob/569778b517496861a3880e0e690973bf08a52e08/src/libsodium/crypto_sign/ed25519/ref10/keypair.c#L49-L71). So the problem may lie in setting a remote public key (which the previously added test doesn't check). To test, you can check that the key exchange results in a common shared secret:

```rust
// The code is a rough sketch

let (our_pk, our_sk) = crypto::gen_keypair_from_seed(...);
let (remote_pk, remote_sk) = crypto::gen_keypair_from_seed(...);

// shared secret on our side
let mut keypair: SodiumDh25519 = Default::default();
keypair.set(our_sk.as_ref());
let mut our_shared_secret = [0u8; 32];
keypair.dh(remote_pk.as_ref(), &mut our_shared_secret);

// shared secret on the remote side
let mut keypair: SodiumDh25519 = Default::default();
keypair.set(remote_sk.as_ref());
let mut remote_shared_secret = [0u8; 32];
keypair.dh(our_pk.as_ref(), &mut remote_shared_secret);

assert_eq!(our_shared_secret, remote_shared_secret);
```

popzxc:
@slowli Thanks for your nagging, it's pretty useful though :)
We've created a separate issue for that (ECR-1549) which will be completed in the next sprint.
In this PR we've decided to left things as it is, and the conversion will be added in another (with your test included).

exonum/src/events/noise/sodium_resolver.rs
```diff
@@ -0,0 +1,243 @@
...
+    fn fill_bytes(&mut self, out: &mut [u8]) {
+        let bytes: Vec<u8> = thread_rng().gen_iter::<u8>().take(out.len()).collect();
+        out.copy_from_slice(&bytes);
+    }
+}
+
+// Elliptic curve 25519.
+pub struct SodiumDh25519 {
+    privkey: sodium_curve25519::Scalar,
+    pubkey: sodium_curve25519::GroupElement,
```
vldm: Does `sodiumoxide` has no wrapers around `Curve25519` keys?
Can we implement it in `crypto` module?

exonum/src/events/noise/sodium_resolver.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+impl Default for SodiumDh25519 {
```
vldm: Why we need `default` implementation?

exonum/src/events/noise/sodium_resolver.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+}
+
+impl Dh for SodiumDh25519 {
+    fn name(&self) -> &'static str {
+        "25519"
+    }
+
+    fn pub_len(&self) -> usize {
+        32
```
vldm: Can we name this magic constant?

exonum/src/events/noise/sodium_resolver.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use hex::FromHex;
+
+    #[test]
+    fn test_curve25519() {
+        // Curve25519 test - draft-curves-10
```
vldm: Can you append more info, or reference link about this test?

exonum/src/events/noise/sodium_resolver.rs
popzxc: Those `Scalar` and `GroupElement` are exactly the wrappers for keys. You can find them in `sodiumoxide/crypto/scalarmult/curve25519.rs`.

exonum/src/events/noise/sodium_resolver.rs
popzxc: Should we? IMHO method name explains it pretty well. If we'll move out all the constants we will have something like `DH25519_PUBLIC_KEY_LENGTH`, `SHA256_HASH_LENGTH` etc, which won't add any new information. 

Another option: use constants right from `sodiumoxide`, but drawback is that we'll have to go to the another library if we'll need the value.

@DarkEld3r what do you think about it?

exonum/src/events/noise/sodium_resolver.rs
popzxc: If you mean a `struct` that contain both `Scalar` and `GroupElement`, then no. Then it can be implemented in `crypto`, yep.

vldm: `Scalar` and `GroupElement` are wrappers around `curve` primitives which in ECDH used as `SecretKey`\`PublicKey`. 
I am talking about abstraction `SecretKey`, `PublicKey` in submodule `x25519` in scope of `crypto.rs` (analog of existing `ed25519`)

exonum/src/events/noise/sodium_resolver.rs
vldm: Anyway using `[0;32]`( which is probably invalid `keypair`) is not a good practises as for me.

As i understand, `snow` init `DHProvider` with default value, and later sets the private key? Than we probably should make `pubkey`\`privkey` optional and `assert`s in methods is keys was really set.
@DarkEld3r ^

exonum/src/events/noise/sodium_resolver.rs
vldm: I am pretty sure we should have safe wrapper inside crypto/sodiumoxide which associative constant.

Than we can use `PublicKey::LENGTH_IN_BYTES` in code

exonum/src/events/noise/sodium_resolver.rs
popzxc: I based my implementation on `crypto_wrapper` from `snow`. AFAICS it also [uses](https://github.com/mcginty/snow/blob/master/src/wrappers/crypto_wrapper.rs) zeros.

Does it make sense to create some random valid key if it's not really usable until user sets the key manually?

exonum/src/events/noise/sodium_resolver.rs
popzxc: Got it.

exonum/src/events/noise/sodium_resolver.rs
vldm: >Does it make sense to create some random valid key if it's not really usable until user sets the key manually?

I don't like unspoken rules.
With external unstable library everything could break at any time, and if this (documented?) contract would change, than left it without `asserts` can lead a real debug hell in future.

May be it would good to add `assert_neq!(ZERO_KEY, key)` at method preludes?

pavel-mukhanov:
@slowli about `kx` module in `libsodium`. As I understand `noise` and `libsodium` key exchange algorithms does not differ very much. The key exchange in `libsodium` uses ECDH algorithm and hashes result shared keys using BLAKE2b. Noise in its turn also generates the shared key from ECDH and uses key derivation to produce AEAD cipher keys.

popzxc:
Now I fixed (I hope) all review comments from @slowli  and @vldm.
Except one thing: I decided not to encapsulate all direct calls to the `sodiumoxide` since it's a `SodiumResolver` and it's okay for it to work with wrapped library directly.

popzxc:
Seems like there are some problems with recently added tests. I'll merge a master branch and try to figure out the problem.

PR #707, opened 2018-05-29
Core Storage Versioning [ECR-1423] [ECR-1718]

andreyluchin:
This PR introduces basic Storage Versioning for core.

The version value is hardcoded in `storage::indexes_metadata::CORE_STORAGE_VERSION`.
It is stored in metadata index. This version is checked for existing databases, causes node to panic if invalid. If there is no existing database (first launch) the version is set automatically. This value is going to be changed manually upon the introduction of breaking changes on core's part.

~~A new dev-dependency is introduced: `assert_matches`. We already use this crate in configuration service and it seems reliable + it is a dev-dependency.~~

### Todo
- [x] ~~Make `Version` a dedicated type.~~ Combine `Version` functions and `VersionStatus` class. 
- [x] Create `storage_version` field in the configuration.
- [x] ~~Create `set-storage-version` command.~~

### New Todo (After reopen)
- [x] Remove `storage_version` field from configuration.
- [x] Store toml/json instead of integer as storage version (for future extension possibility).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum
 
+- Storage in exonum is now versioned. Old databases will not work with this
+  update. (#707)
+
 - `Iron` based web API has been replaced by the new implementation based
   on `actix-web`. (#727)
 
@@ -153,9 +156,6 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   moved to separate modules: `TimeSchema` to `exonum_time::schema`,
   `TimeProvider` and `MockTimeProvider` to `exonum_time::time_provider`,
   `ValidatorTime` to `exonum_time::api`. (#604)
-- Storage in exonum is now versioned. Old databases will not work with this update. (#707)
-- Storage in exonum is now versioned. Old databases will not work with this
-  update. (#707)
 
 ### New features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -153,6 +153,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   moved to separate modules: `TimeSchema` to `exonum_time::schema`,
   `TimeProvider` and `MockTimeProvider` to `exonum_time::time_provider`,
   `ValidatorTime` to `exonum_time::api`. (#604)
+- Storage in exonum is now versioned. Old databases will not work with this update. (#707)
 
 ### New features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   `TimeProvider` and `MockTimeProvider` to `exonum_time::time_provider`,
   `ValidatorTime` to `exonum_time::api`. (#604)
 - Storage in exonum is now versioned. Old databases will not work with this update. (#707)
+- Storage in exonum is now versioned. Old databases will not work with this
+  update. (#707)
 
 ### New features
 
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -222,7 +222,7 @@ impl StoredConfiguration {
             ));
         }
 
-        // Checkout storage_version.
+        // Check storage_version.
         if config.storage_version != storage::StorageVersion::current() {
             return Err(JsonError::custom(format!(
                 "storage_version({}) must be the same with storage version set in the core ({})",
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -59,6 +59,8 @@ pub struct StoredConfiguration {
     /// configuration will not take effect at all; the old configuration will
     /// remain actual.
     pub actual_from: Height,
+    /// Core storage version.
+    pub storage_version: u64,
     /// List of validators consensus and service public keys.
     pub validator_keys: Vec<ValidatorKeys>,
     /// Consensus algorithm parameters.
@@ -247,6 +249,7 @@ mod tests {
 
     use super::*;
     use crypto::{gen_keypair_from_seed, Seed};
+    use storage;
 
     // TOML doesn't support all rust types, but `StoredConfiguration` must be able to save as TOML.
     #[test]
@@ -262,6 +265,7 @@ mod tests {
         let toml_content = r#"
             previous_cfg_hash = "0000000000000000000000000000000000000000000000000000000000000000"
             actual_from = 42
+            storage_version = 0
 
             [[validator_keys]]
             consensus_key = "8a88e3dd7409f195fd52db2d3cba5d72ca6709bf1d94121bf3748801b40f6f5c"
@@ -351,6 +355,7 @@ mod tests {
             consensus: ConsensusConfig::default(),
             services: BTreeMap::new(),
             majority_count: None,
+            storage_version: storage::StorageVersion::current(),
         }
     }
 
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -224,11 +224,11 @@ impl StoredConfiguration {
 
         // Checkout storage_version.
         if config.storage_version != storage::StorageVersion::current() {
-            return Err(JsonError::curstom(
+            return Err(JsonError::custom(format!(
                 "storage_version({}) must be the same with storage version set in the core ({})",
                 config.storage_version,
                 storage::StorageVersion::current(),
-            ));
+            )));
         }
 
         Ok(config)
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -28,7 +28,7 @@ use std::collections::{BTreeMap, HashSet};
 
 use crypto::{hash, CryptoHash, Hash, PublicKey};
 use helpers::{Height, Milliseconds};
-use storage::{self, StorageValue};
+use storage::StorageValue;
 
 /// Public keys of a validator. Each validator has two public keys: the
 /// `consensus_key` is used for internal operations in the consensus process,
@@ -59,8 +59,6 @@ pub struct StoredConfiguration {
     /// configuration will not take effect at all; the old configuration will
     /// remain actual.
     pub actual_from: Height,
-    /// Core storage version.
-    pub storage_version: u64,
     /// List of validators consensus and service public keys.
     pub validator_keys: Vec<ValidatorKeys>,
     /// Consensus algorithm parameters.
@@ -222,15 +220,6 @@ impl StoredConfiguration {
             ));
         }
 
-        // Check storage_version.
-        if config.storage_version != storage::StorageVersion::current() {
-            return Err(JsonError::custom(format!(
-                "storage_version({}) must be the same with storage version set in the core ({})",
-                config.storage_version,
-                storage::StorageVersion::current(),
-            )));
-        }
-
         Ok(config)
     }
 }
@@ -258,7 +247,6 @@ mod tests {
 
     use super::*;
     use crypto::{gen_keypair_from_seed, Seed};
-    use storage;
 
     // TOML doesn't support all rust types, but `StoredConfiguration` must be able to save as TOML.
     #[test]
@@ -274,7 +262,6 @@ mod tests {
         let toml_content = r#"
             previous_cfg_hash = "0000000000000000000000000000000000000000000000000000000000000000"
             actual_from = 42
-            storage_version = 0
 
             [[validator_keys]]
             consensus_key = "8a88e3dd7409f195fd52db2d3cba5d72ca6709bf1d94121bf3748801b40f6f5c"
@@ -364,7 +351,6 @@ mod tests {
             consensus: ConsensusConfig::default(),
             services: BTreeMap::new(),
             majority_count: None,
-            storage_version: storage::StorageVersion::current(),
         }
     }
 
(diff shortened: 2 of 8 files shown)


andreyluchin:
Looks like despite tests being fixed, Travis still fails for some weird reason.
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

exonum/src/blockchain/mod.rs
```diff
@@ -183,12 +184,53 @@ impl Blockchain {
...
+            panic!("Could not set database version.")
+        }
+    }
+
+    /// Checks if storage version is supported.
+    ///
+    /// # Panics
+    ///
+    /// Panics if version is not supported or is not specified.
+    fn check_storage_version(&self) {
```
vldm: This method should probably be named with `assert` prefix.

andreyluchin: `assert_*` is a good name. Can you please elaborate on the `maintenance` command? I can only see one usage of it - setting storage version when there is none set. But this scenario is very limited and this command will only be useful when migrating to 0.8.

vitvakatu: I think migration to 0.8 is a good case for this command.

exonum/src/blockchain/mod.rs
```diff
@@ -183,12 +184,53 @@ impl Blockchain {
         let has_genesis_block = !Schema::new(&self.snapshot())
             .block_hashes_by_height()
             .is_empty();
-        if !has_genesis_block {
+        if has_genesis_block {
+            self.check_storage_version();
+        } else {
+            self.initialize_metadata();
```
slowli: Would it make sense to move `initialize_metadata` invocation into `create_genesis_block`? IMO, initialization logically belongs there.

exonum/src/storage/indexes_metadata.rs
```diff
@@ -154,8 +161,41 @@ pub fn assert_index_type(name: &str, index_type: IndexType, is_family: bool, vie
...
+            storage_ver: ver,
+        },
+        None => VersionStatus::Unspecified {
+            core_ver: CORE_STORAGE_VERSION,
+        },
+    }
+}
+
+// Sets storage version in metadata index and returns the version set.
+pub(crate) fn initialize_storage_version(view: &mut Fork) -> Version {
```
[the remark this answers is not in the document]
andreyluchin: The idea behind `VersionStatus` is to express the result of comparison of the value in the storage and the value in the core. I think it makes sense that we don't `write` `VersionStatus`, but rather we write the `Version`.

exonum/src/blockchain/mod.rs
andreyluchin: This got me wondering as well. My thoughts were that we write data to the storage with some specific version. Otherwise it would be that genesis block is written before the storage has a version.

exonum/src/storage/indexes_metadata.rs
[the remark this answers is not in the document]
slowli: It *could* make sense to combine functionality of both into one class, e.g., to provide stronger guarantees about the range of supported versions than `u64`:

```rust
pub enum Version {
    /// Initial version of the storage used in Exonum >=0.8.
    Version1,
    /*
    /// Hypothetical new version. Changes compared to version 1 may be
    /// documented here.
    Version2,
    */
    Unsupported(u64),
}
```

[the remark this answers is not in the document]
andreyluchin: I can definitely see that, I like the idea of making `Version` a dedicated type instead of `u64`. The only issue I have with this approach are the variant names (`Version1`, `Version2`, `Version3` seems hacky). Also, since this version is internal for the core, do we really need to document all these versions? For now I will settle with
```rust
pub(crate) enum Version {
    Supported(u64),
    Unsupported(u64),
    Unspecified,
}
```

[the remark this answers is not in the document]
slowli: More intelligent naming scheme may make sense, although I cannot come up with it right away. Specifying versions explicitly has the added benefit compared to `Supported(u64)` in that *all* supported versions are explicitly enumerated (there should be relatively few of them, I hope), so there is no chance to create something like `Version::Supported(100500)`. As for comments, I think they would be useful at least for core developers.

andreyluchin:
OK. As was discussed, `storage_version` field was added to the configuration to ensure that all nodes follow the same configuration. As of now, it is initialized with the same `storage::CORE_STORAGE_VERSION` constant, but besides that is not used anywhere. Push forward your ideas as to what should be done regarding configuration.

andreyluchin:
So `storage_version` field has been added to Stored Configuration. Since this version is determined by the core, it is set automatically without any user intervention. Let me know if you think that has to be changed.
This is what new stored configuration looks like:
```javascript
{
    "config": {
        "actual_from": 0,
        "consensus": {
            "max_message_len": 1048576,
            "max_propose_timeout": 200,
            "min_propose_timeout": 10,
            "peers_timeout": 10000,
            "propose_timeout_threshold": 500,
            "round_timeout": 3000,
            "status_timeout": 5000,
            "txs_block_limit": 1000
        },
        "majority_count": null,
        "previous_cfg_hash": "0000000000000000000000000000000000000000000000000000000000000000",
        "services": {
            "configuration": null
        },
        "storage_version": 0, // The one.
        "validator_keys": [
            {
                "consensus_key": "50c0175bf8a9cd678a92f40bad8816e503ddee1106e2539fae3c549962042292",
                "service_key": "339597347ebc8e192f34f914151f4ef1c94d45c29baa275e4d8ded1ba7ec6991"
            }
        ]
    },
    "hash": "8ef8ccdcd37e5b36da5447bdfceb79db44fa62080a532808dc514ecb98a135cf",
    "propose": null,
    "votes": null
}
```

andreyluchin:
While making a `set-storage-version` I realized that it is only going to work if the `configuration` service is running. This is because we chose to store `storage_version` in the configuration, so it is impossible to `set_storage-version` on a node in a network with non-empty blockchain or without `configuration` service running.

**There 3 things that can be done:**
1. Removing `storage_version` from configuration, introduce the `set-storage-version-command`.
2. Scrapping `set-storage-version` and keeping `storage_version` in the configuration.
3. Keeping both by implementing an algorithm which checks whether `configuration` service is running, setting a version in the storage and making a propose for a new configuration with new `storage_version`. (Though I may not understand this completely)

I am a fan of **2**nd way. 

If we choose to implement **3** we get a very limited mechanism for migrating to `0.9` (or whenever this is going to be released). How many existing Exonum blockchain networks run `configuration`? If not that much, is **3** much different from **2**?

The **1**st way is unpractical since we have decided that `storage_version` in the configuration is vital so that all nodes start with same storage version, which is important because serialized data representation (BTW this is where having separate serialization for communication and storage would help a lot).

My final decision though would be to go with **2** - __not adding `set-storage-version` command__  in this PR and then if it is decided to implement **3** (or something else) I will make a separate PR to avoid this one getting lost in time.

Anyway, let me know what you guys think.

exonum/src/blockchain/mod.rs
```diff
@@ -198,6 +238,7 @@ impl Blockchain {
             consensus: cfg.consensus,
             services: BTreeMap::new(),
             majority_count: None,
+            storage_version: storage::StorageVersion::current(),
```
slowli: I'm not sure this will work as intended retrospectively. For example, imagine a situation that an auditor with the current storage version 2 first connects to the network that was initialized with the storage version 1. Then again, this *may* be the intended behavior; frankly speaking, I have *very* approximate understanding what should happen during version transition, and how multiple versions can be supported during the blockchain lifetime. Is this documented anywhere? @DarkEld3r

exonum/src/storage/indexes_metadata.rs
```diff
@@ -154,8 +161,35 @@ pub fn assert_index_type(name: &str, index_type: IndexType, is_family: bool, vie
...
 
+#[derive(Debug, Clone)]
+pub(crate) enum StorageVersion {
+    Supported(VersionValue),
+    Unsupported(VersionValue),
+    Unspecified,
+}
+
+impl StorageVersion {
+    pub(crate) fn current() -> VersionValue {
```
slowli: Nit: I think it's unnecessary to use `pub(crate)` modifiers for methods in the type that is `pub(crate)` itself.

exonum/src/blockchain/config.rs
```diff
@@ -220,6 +222,15 @@ impl StoredConfiguration {
             ));
         }
 
+        // Check storage_version.
+        if config.storage_version != storage::StorageVersion::current() {
```
slowli: I'm not sure I follow this logic of version verification. Is it possible to change the storage version using the configuration service? If so, what should that signal? I'd imagine it would (indirectly) signal the full nodes to upgrade their software in order to support the new version; i.e., once a new config is activated, the node should stop if it does not support the new storage version (and perhaps a do some kind of migration otherwise before continuing to work). I'm not sure that this check reflects this kind of logic.

exonum/src/blockchain/mod.rs
```diff
@@ -183,12 +184,51 @@ impl Blockchain {
...
+            self.initialize_metadata();
             self.create_genesis_block(cfg)?;
         }
         Ok(())
     }
 
+    /// Initialized node-local metadata.
+    fn initialize_metadata(&mut self) {
+        let mut fork = self.db.fork();
+        storage::StorageVersion::write_current(&mut fork);
```
slowli: Nit: Is there a particular reason for using `StorageVersion` in the qualified form here?

andreyluchin: I chose `write_current` to be static because it does not actually require any state to do what it is supposed to :)

CHANGELOG.md
```diff
@@ -38,6 +38,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `ConsensusConfig::validate_configuration` method has been renamed to the
   `warn_if_nonoptimal`. (#690)
 
+- Storage in exonum is now versioned. Old databases will not work with this
```
DarkEld3r: This should be moved to the "unreleased" section.

exonum/src/blockchain/config.rs
```diff
@@ -220,6 +222,15 @@ impl StoredConfiguration {
             ));
         }
 
+        // Check storage_version.
```
DarkEld3r: Nitpick: I prefer to say "storage version" here. :upside_down_face:

exonum/src/blockchain/mod.rs
andreyluchin: This is definitely a concern for storage migration, but this out of scope for ECR-1423. To give you the sense of what are we trying to accomplish as of right now I want to quote the following lines from the task description:
> We need to store database version inside metadata table and panic if database contains unsupported version.

We agreed that **the migration** will be implemented *later*. For now we just have to implement a basic versioning mechanism. I can send you more information through a more private channel if you wish.

For now, the situation you described is, indeed, **intended behavior** (at least by me) - The blockchain runs only when all nodes have the same version.

exonum/src/blockchain/config.rs
```diff
@@ -59,6 +59,8 @@ pub struct StoredConfiguration {
     /// configuration will not take effect at all; the old configuration will
     /// remain actual.
     pub actual_from: Height,
+    /// Core storage version.
+    pub storage_version: u64,
```
DarkEld3r: I propose to remove version from the configuration: it looks like some rudiment, because storage version is our implementation detail and user probably should not be able to choose it.

exonum/src/blockchain/config.rs
DarkEld3r: I'm not sure that user should be able to initiate such migration. I think that it should be performed for example as part of migration from 1.0 to 2.0.

exonum/src/storage/indexes_metadata.rs
```diff
@@ -154,8 +161,35 @@ pub fn assert_index_type(name: &str, index_type: IndexType, is_family: bool, vie
     }
 }
 
+#[derive(Debug, Clone)]
+pub(crate) enum StorageVersion {
+    Supported(VersionValue),
+    Unsupported(VersionValue),
+    Unspecified,
```
DarkEld3r: I propose to remove `Unspecified` enum variant, because it is used only once. I propose to replace it with panic.

exonum/src/blockchain/mod.rs
```diff
@@ -172,7 +172,8 @@ impl Blockchain {
     ///
     /// # Panics
     ///
-    /// If the genesis block was not committed.
+    /// * If the genesis block was not committed.
+    /// * If storage version is not specified or not supported.
```
DarkEld3r: :confused:

andreyluchin:
Closing this PR due until design questions are solved internally.

exonum/src/storage/indexes_metadata.rs
```diff
@@ -14,16 +14,23 @@
...
     serialize::{json, WriteBufferWrapper}, CheckedOffset, Error as EncodingError, Field, Offset,
 };
 use storage::{base_index::BaseIndex, Fork, Snapshot, StorageValue};
 
 pub const INDEXES_METADATA_TABLE_NAME: &str = "__INDEXES_METADATA__";
 
+// Value of this constant is to be incremented manually
+// upon the introduction of breaking changes to the storage.
+const CORE_STORAGE_VERSION: StorageMetadata = StorageMetadata { version: 0 };
+const CORE_STORAGE_VERSION_KEY: &str = "__STORAGE_VERSION__";
```
DarkEld3r: I suppose constants should be renamed too.

PR #708, opened 2018-05-29
Add docker container for cryptocurrency-advanced example [ECR-937]

SerhiiOryshych:


examples/cryptocurrency-advanced/docker/launch.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+node_count=4
+start_peer_port=6331
+start_public_port=8200
+start_private_port=8091
+
+cd backend && mkdir example && cd example
+/root/.cargo/bin/exonum-cryptocurrency-advanced generate-template common.toml --validators-count 4
```
aleksuss: I suppose a good way is to create a variable that stores path `/root/.cargo/bin/exonum-cryptocurrency-advanced` and use it in the script.

examples/cryptocurrency-advanced/README.md
```diff
@@ -12,15 +12,32 @@ It implements most basic operations:
...
+### Using docker
+
+<!-- spell-checker:ignore serhiioryshych -->
+
+Simply run the following command to start the cryptocurrency service on 4 nodes
+on the local machine:
+
+```bash
+docker run -p 8280:8280 serhiioryshych/exonum-cryptocurrency-advanced-example
+```
```
aleksuss: Does `serhiioryshych` need to replace on `exonum` ?

examples/cryptocurrency-advanced/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:16.04
+
+ENV ROCKSDB_LIB_DIR=/usr/lib/x86_64-linux-gnu
+ENV SNAPPY_LIB_DIR=/usr/lib/x86_64-linux-gnu
+
+RUN apt-get update \
+  && apt-get install -y curl git gcc cpp \
+  libssl-dev pkg-config libsodium-dev libsnappy-dev librocksdb-dev
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable
```
slowli: IMO, versions for Rust and Node should be fixed to aid reproducibility.

examples/cryptocurrency-advanced/docker/launch.sh
```diff
@@ -0,0 +1,35 @@
...
+
+for i in $(seq 0 $((node_count - 1)))
+do
+  public_port=$((start_public_port + i))
+  $path_to_app run --node-config node_$((i + 1))_cfg.toml --db-path db$((i + 1)) --public-api-address 0.0.0.0:${public_port} &
+  echo "new node with ports: $public_port (public)"
+done
+
+cd ../../frontend
+npm install
```
slowli: I'd argue `npm install` and and `npm run build`  commands belong to the Docker script, not here.

examples/cryptocurrency-advanced/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash \
+  && apt-get install -y nodejs
+
+RUN mv /root/.cargo/bin/* /usr/bin
+
+WORKDIR /usr/src
+RUN git clone https://github.com/exonum/exonum.git \
```
slowli: To aid reproducibility, exonum repo could be cloned from a specific branch / tag (maybe, given as an `ENV` var). Also, it makes sense to delete the sources after building in order to reduce the size of the Docker image.

examples/cryptocurrency-advanced/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apt-get update \
+  && apt-get install -y curl git gcc cpp \
+  libssl-dev pkg-config libsodium-dev libsnappy-dev librocksdb-dev
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable
+
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash \
+  && apt-get install -y nodejs
+
+RUN mv /root/.cargo/bin/* /usr/bin
```
slowli: I'd suggest to combine with the `cargo ...` commands below. Also, I wonder if Rust is still needed after the installation of the cryptocurrency binary - it may make sense to remove it to reduce the Docker image bloat, or use a separate builder container, [like here](https://github.com/emk/rust-musl-builder/blob/master/examples/using-diesel/Dockerfile).

examples/cryptocurrency-advanced/docker/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+WORKDIR /usr/src
+RUN git clone https://github.com/exonum/exonum.git \
+  && cd exonum/examples/cryptocurrency-advanced/backend \
+  && cargo update && cargo install
+WORKDIR /usr/src/exonum/examples/cryptocurrency-advanced
+COPY launch.sh .
+
+ENTRYPOINT ["sh", "./launch.sh"]
```
slowli: The shebang at the beginning of the script suggests `bash`, but here you use `sh`. Just sayin'...

examples/cryptocurrency-advanced/docker/Dockerfile
vldm: Maybe it would be better to add CI tests to docker image rather than fix it to specific branch?

PR #713, opened 2018-05-30
Ignore test_duplicated_transaction

vldm:
Ignore unstable test


exonum/tests/balance_service.rs
```diff
@@ -130,6 +130,7 @@ pub mod service {
     }
 
     #[test]
+    #[ignore]
```
popzxc: Please add issue number.

PR #721, opened 2018-06-04
Apply 1.26.1 formatting and use import groups

DarkEld3r:
1.26.1 release brings some minor changes into formatting plus I made one more attempt to use import groups.

Personally I like using groups, even with somewhat strange (inconsistent) indentation, because it is equal to the previous formatting (in terms of indentation).


exonum/tests/node.rs
```diff
@@ -18,22 +18,21 @@ extern crate futures;
...
-use exonum::node::{ApiSender, ExternalMessage, Node};
-use exonum::storage::{Database, Fork, MemoryDB, Snapshot};
+use std::{sync::{Arc, Mutex},
+          thread::{self, JoinHandle},
+          time::Duration};
+
+use exonum::{blockchain::{Service, ServiceContext, Transaction},
+             crypto::Hash,
+             encoding::Error as EncodingError,
+             helpers,
```
andreyluchin: I hate the alignment spaces, they are pain in the butt to work with if your IDE does not support it. But I am all for keeping rustfmt up to date.

ozkriff: I'm not sure if this is `rustfmt`'s bug or not - I thought they must use block indent for everything now.

DarkEld3r: I don't like such alignment either, but it didn't bring anything new comparing to the previous formatting:
```rust
use exonum::blockchain::...;
use exonum::crypto::...;

use exonum::{blockchain::...,
             crypto::...};
```

ozkriff: I think that PR itself is okay and we must stick to default `rustfmt`'s style anyway, I just don't understand why visual indentation is used here - https://github.com/nrc/rfcs/blob/style-guide/style-guide/items.md#imports-use-statements - it should be a block indent :-\

DarkEld3r: @ozkriff 
Will you create an issue for this problem? :upside_down_face:

ozkriff: https://github.com/rust-lang-nursery/rustfmt/issues/2765

PR #724, opened 2018-06-06
Cleanup .gitignore entries

mvlabat:
Removed some redundant gitignore entries

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,22 +5,21 @@
 
 # cargo
 target
-*/target
 *.rs.bk
 Cargo.lock
 
 # fuzz testing
-exonum/fuzz/target
 exonum/fuzz/corpus
 exonum/fuzz/artifacts
 
+# node
+node_modules
+
 # node (for cspell)
-/node_modules
 /npm-debug.log
 /package-lock.json
 
 # node (for testkit example)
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -5,22 +5,21 @@
 
 # cargo
 target
-*/target
 *.rs.bk
```
ozkriff: I think that this is outdated as well - rustfmt doesn't create these anymore

.gitignore
```diff
@@ -5,22 +5,20 @@
...
-exonum/fuzz/target
 exonum/fuzz/corpus
 exonum/fuzz/artifacts
 
+# node
+node_modules
+
 # node (for cspell)
-/node_modules
 /npm-debug.log
```
slowli: It could make sense to make `npm-debug.log` directory-independent, same as `node_modules`.

.gitignore
```diff
@@ -5,22 +5,20 @@
...
 /package-lock.json
```
slowli: Off-topic: for reproducible builds, it may make sense to include `package-lock.json` (and `package.json`) into the git tree, like [in the documentation repo](https://github.com/exonum/exonum-doc/blob/master/package.json).

PR #726, opened 2018-06-07
Service name should be printed when writing errors to log [ECR-1471]

SergeiMal:
It'll make it easier to understand which error occurred and why


exonum/src/blockchain/mod.rs
```diff
@@ -354,21 +354,30 @@ impl Blockchain {
...
+            (
+                self.tx_from_raw(tx).or_else(|error| {
+                    Err(failure::err_msg(format!(
+                        "Service <{}>: {}, tx: {:?}",
+                        sn,
+                        error.description(),
+                        tx_hash
+                    )))
+                })?
+                ,
```
pavel-mukhanov: Hm, some weird indentation here.

PR #729, opened 2018-06-07
Warning Consensus Message [ECR-1266]

VladYermakov:
Now this warning message will be written when difference between heights is greater than 1.


DarkEld3r:
I think that we can mention this change in the changelog ("inner improvements" section).

VladYermakov:
@slowli what behavior is now expected for messages from previous height?

slowli:
Behavior should not be changed (unless this is done consciously, in which case the naming of the PR is misleading): Let `h = self.state.height()`. Then all messages with height `h' < h` must be ignored by the virtue of being useless to the consensus algorithm. My problem with the PR is that now height `h' = h - 1` doesn't seem to be ignored (while it should). Maybe I'm missing something and they *are* ignored with the help of some other checks? Would be happy to find out I'm wrong :slightly_smiling_face:

PR #733, opened 2018-06-08
Fixed panic at unwrap in configuration service [ECR-1642]

popzxc:


services/configuration/src/transactions.rs
```diff
@@ -291,9 +291,8 @@ impl<'a> VotingDecisionRef<'a> {
         if let Some(validator_id) = validator_index(snapshot, self.sender()) {
             let vote = schema
                 .votes_by_config_hash(self.cfg_hash())
-                .get(validator_id as u64)
-                .unwrap();
-            if vote.is_some() {
+                .get(validator_id as u64);
+            if vote.is_some() && vote.unwrap().is_some() {
```
vldm: maybe patten matching would look more clear `if let Some(Some(_)) `?

popzxc: IMHO it looks worse.

popzxc:
I understand that it doesn't show any useful case: it's just how I've discovered it :)
Regarding the test: you're right, I'll add it.

slowli:
Just to be clear: I propose to move the lines I've linked before accessing `votes_by_config_hash` instead of changing vote access logic. If the proposal exists, the votes for it are guaranteed to exist - [they are put into the index during proposal creation](https://github.com/exonum/exonum/blob/a14d3be6ea01f1fc0cbfca7a85922f927ca15b34/services/configuration/src/transactions.rs#L212-L214).

services/configuration/src/transactions.rs
```diff
@@ -410,3 +411,35 @@ impl Transaction for VoteAgainst {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::super::tests::{new_tx_config_vote, new_tx_config_vote_against};
+    use super::super::Service as ConfigurationService;
+    use super::{Hash, VotingDecisionRef};
```
DarkEld3r: Please use import groups here:
```rust
use super::{
    super::...,
    Hash, VotingDecisionRef,
};
```

ozkriff: Is this `super::super::` thing is a good idea at all? Maybe it's better to just start the path from crate's root?

services/configuration/src/transactions.rs
```diff
@@ -410,3 +411,35 @@ impl Transaction for VoteAgainst {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{super::tests::{new_tx_config_vote, new_tx_config_vote_against},
```
ozkriff: Is this `super::super::` thing is a good idea at all? Maybe it's better to just start the path from crate's root?

popzxc: Something weird happened to me :sweat:

services/configuration/src/transactions.rs
```diff
@@ -288,22 +288,23 @@ impl<'a> VotingDecisionRef<'a> {
...
         if let Some(validator_id) = validator_index(snapshot, self.sender()) {
             let vote = schema
                 .votes_by_config_hash(self.cfg_hash())
                 .get(validator_id as u64)
-                .unwrap();
+                .expect("Can't get vote for precheck");
+
             if vote.is_some() {
-                Err(AlreadyVoted)?;
+                return Err(AlreadyVoted);
```
slowli: Nit: Is there a particular reason you've changed this instance of erroneous return, but didn't change the one above, with `Err(AlreadyScheduled(_))`?

services/configuration/src/transactions.rs
```diff
@@ -410,3 +411,35 @@ impl Transaction for VoteAgainst {
...
+mod tests {
+    use exonum_testkit::{TestKit, TestKitBuilder};
+
+    use super::{Hash, VotingDecisionRef};
+    use errors::Error as ServiceError;
+    use tests::{new_tx_config_vote, new_tx_config_vote_against};
+    use Service as ConfigurationService;
+
+    #[test]
+    fn test_vote_without_propose() {
```
slowli: Nit: I'd prefer a more "high-level" test, with transactions being sent to the blockchain via `create_block*` and their status checked via explorer; this could be more stable across possible future revisions of transaction logic.

popzxc: I did this test low-level on purpose: high-level test will show that _something_ has broken, and this test in case of any regression will show the exact problem.
IMHO in that particular case the low-level test is better (although I can be wrong).

services/configuration/src/transactions.rs
popzxc: There are too much of similar entries in this file, so I guess I'll create a separate pull-request for this.

services/configuration/src/transactions.rs
slowli: I won't insist. Just a note: It's possible to find the exact error cause via the explorer API (or by looking at the transaction result directly) with something like

```rust
let bogus_tx = ...;
let block = testkit.create_block_with_transaction(tx);
assert_eq!(
    block[0].status().unwrap_err().error_type(),
    TransactionErrorType::Code(ErrorCode::UnknownConfigRef as u8),
);
```

PR #736, opened 2018-06-13
Add API endpoint to check consensus status [ECR-1649]

SergeiMal:


testkit/tests/system_api.rs
```diff
@@ -52,3 +52,14 @@ fn test_network() {
...
+#[test]
+fn test_consensus_status() {
+    let testkit = TestKitBuilder::validator().with_validators(2).create();
+    let api = testkit.api();
+    let info: ConsensusStatusInfo = api.get(ApiKind::System, "v1/consensus_status");
+    let expected = ConsensusStatusInfo {
+        status: false,
+    };
+    assert_eq!(info, expected);
+}
```
DarkEld3r: Please add a new line at the end of the file. :upside_down_face:

testkit/tests/system_api.rs
```diff
@@ -52,3 +52,14 @@ fn test_network() {
     assert_eq!(info.protocol_version, PROTOCOL_MAJOR_VERSION);
     assert!(info.services.is_empty());
 }
+
+#[test]
+fn test_consensus_status() {
+    let testkit = TestKitBuilder::validator().with_validators(2).create();
```
DarkEld3r: I propose to change the number of validators to 1 in order to make the test more meaningful.

exonum/src/api/public/system.rs
```diff
@@ -76,6 +83,15 @@ impl SystemApi {
         };
         router.get("/v1/user_agent", user_agent, "user_agent");
     }
+    fn consensus_status_info(self, router: &mut Router) {
```
DarkEld3r: Please separate methods with an empty line.

exonum/src/api/public/system.rs
```diff
@@ -19,6 +19,7 @@ use serde_json;
 use api::Api;
 use blockchain::{Blockchain, Schema, SharedNodeState};
 use helpers::user_agent;
+use std::net::SocketAddr;
```
DarkEld3r: Standard imports should be placed after third-party usings, but before inner ones.

exonum/src/api/public/system.rs
```diff
@@ -31,6 +32,12 @@ pub struct HealthCheckInfo {
     pub connectivity: bool,
 }
 
+#[doc(hidden)]
```
popzxc: Should this public `struct` really be hidden?

exonum/src/api/public/system.rs
```diff
@@ -31,6 +32,12 @@ pub struct HealthCheckInfo {
...
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+pub struct ConsensusStatusInfo {
+    pub status: bool,
```
popzxc: `status: true` and `status:false` are not really informative. Maybe use enum instead?

exonum/src/api/public/system.rs
popzxc: Missing empty line between methods.

exonum/src/blockchain/service.rs
```diff
@@ -410,6 +410,34 @@ impl SharedNodeState {
...
+    /// of a blockchain node.
+    pub fn majority_count(&self) -> usize {
+        self.state
+            .read()
+            .expect("Expected read lock.")
+            .majority_count
+    }
+    /// Updates internal majority count, from `State` of a blockchain node.
+    pub fn update_majority_count(&self, state: &State) {
+
```
popzxc: Was this empty line added on purpose?

exonum/src/blockchain/service.rs
```diff
@@ -410,6 +410,34 @@ impl SharedNodeState {
...
+    /// Returns the majority count from the current "State"
+    /// of a blockchain node.
+    pub fn majority_count(&self) -> usize {
+        self.state
+            .read()
+            .expect("Expected read lock.")
+            .majority_count
+    }
+    /// Updates internal majority count, from `State` of a blockchain node.
```
popzxc: Missing empty line between methods.

exonum/src/blockchain/service.rs
```diff
@@ -410,6 +410,34 @@ impl SharedNodeState {
...
+    pub fn update_majority_count(&self, state: &State) {
```
DarkEld3r: Is this function needed? You don't use it anywhere.

exonum/src/blockchain/service.rs
```diff
@@ -410,6 +410,34 @@ impl SharedNodeState {
...
+
+        let mut majority_count = state.majority_count();
+
+        let peers :Vec<SocketAddr> = self.peers_info()
+            .iter()
+            .map(|(s, _k)| (*s))
+            .collect();
+        if !peers.contains(&state.our_connect_message().addr()) {
```
popzxc: I'd expect an explaining comment here TBH.

exonum/src/api/public/system.rs
DarkEld3r: We are using `true` and `false` in `HealthCheckInfo` too.

popzxc: Maybe change it there too then? 😈

exonum/src/blockchain/service.rs
SergeiMal: it uses in node/basic.rs

DarkEld3r:
https://jira.bf.local/browse/ECR-1665

exonum/src/api/public/system.rs
DarkEld3r: https://jira.bf.local/browse/ECR-1665

testkit/tests/system_api.rs
```diff
@@ -52,3 +52,14 @@ fn test_network() {
     assert_eq!(info.protocol_version, PROTOCOL_MAJOR_VERSION);
     assert!(info.services.is_empty());
 }
+
+#[test]
+fn test_consensus_status_false() {
+    let testkit = TestKitBuilder::validator().with_validators(1).create();
```
popzxc: `with_validators(1)` is the default, so it could be skipped. (c)

testkit/tests/system_api.rs
```diff
@@ -52,3 +52,14 @@ fn test_network() {
     assert_eq!(info.protocol_version, PROTOCOL_MAJOR_VERSION);
     assert!(info.services.is_empty());
 }
+
+#[test]
+fn test_consensus_status_false() {
```
popzxc: `test_consensus_status_false` checks that `consensus_status` is true. Seems legit :upside_down_face:

exonum/src/api/public/system.rs
popzxc: Docs were added, but the attribute still persists. Is it ok?

exonum/src/node/basic.rs
```diff
@@ -220,6 +220,7 @@ impl NodeHandler {
     /// Node update internal `ApiState`.
     pub fn handle_update_api_state_timeout(&mut self) {
         self.api_state.update_node_state(&self.state);
+        self.api_state.update_majority_count(&self.state);
```
popzxc: Questionable: maybe make this method private and call it from the `update_node_state`? As far as I can see, those are should be called together.

SergeiMal: Yes, it make sense

testkit/tests/system_api.rs
SergeiMal: The same situation with test_healthcheck_connectivity_false.

popzxc: If I don't miss anything, tests look like the following:
```
#[test]
fn test_healthcheck_connectivity_false() {
    let testkit = TestKitBuilder::validator().with_validators(2).create();
    let api = testkit.api();
    let info: HealthCheckInfo = api.get(ApiKind::System, "v1/healthcheck");
    let expected = HealthCheckInfo {
        connectivity: false,
    };
    assert_eq!(info, expected);
}
```

```
#[test]
fn test_consensus_status_false() {
    let testkit = TestKitBuilder::validator().with_validators(1).create();
    let api = testkit.api();
    let info: ConsensusStatusInfo = api.get(ApiKind::System, "v1/consensus_status");
    let expected = ConsensusStatusInfo {
        status: true,
    };
    assert_eq!(info, expected);
}
```

`test_healthcheck_connectivity_false` checks that the `connectivity` field is false (which is perfectly consistent with the name).

`test_consensus_status_false` checks that the `status` field is true (which is kind of confusing).

Or maybe I miss something?

SergeiMal: healthcheck  with 2 validators must return connectivity = true,  but it failed. It seems some trouble with testkit, or testkit isn't suitable to perform these tests.

exonum/src/blockchain/service.rs
```diff
@@ -408,6 +409,32 @@ impl SharedNodeState {
...
+            .read()
+            .expect("Expected read lock.")
+            .majority_count
+    }
+
+    /// Updates internal majority count, from `State` of a blockchain node.
+    fn update_majority_count(&self, state: &State) {
+        let mut majority_count = state.majority_count();
+
+        let peers: Vec<SocketAddr> = self.peers_info().iter().map(|(s, _k)| (*s)).collect();
```
popzxc: For some reason previous comment is marked as outdated, but once again:
I think it'll be better to not use `collect` here. 
Something like this:
```
        let peers = self.peers_info();
        if peers
            .iter()
            .map(|(s, _k)| (*s))
            .find(|x| x == &state.our_connect_message().addr())
            .is_none()
        {
            majority_count -= 1;
        }
```

And an explaining comment, why are we decrementing majority count.

testkit/tests/system_api.rs
popzxc: Just to be clear: the only confusing thing is the name of the test which tells us that we are checking if status is false, and check for status to be true in the body.

exonum/src/blockchain/service.rs
SergeiMal: It will be removed at all in next commit

exonum/src/blockchain/service.rs
```diff
@@ -408,6 +409,25 @@ impl SharedNodeState {
...
+    /// of a blockchain node.
+    pub fn majority_count(&self) -> usize {
+        self.state
+            .read()
+            .expect("Expected read lock.")
+            .majority_count
+    }
+
+    /// Updates internal majority count, from `State` of a blockchain node.
+    fn update_majority_count(&self, state: &State) {
```
popzxc: Now, since this function doesn't anymore have complicated logic, I suggest to remove it. Logic itself can be moved to the `update_node_state`, and `update_node_state` can be rewritten like this:

```
    pub fn update_node_state(&self, state: &State) {
        let mut lock = self.state.write().expect("Expected write lock.");

        lock.peers_info.clear();
        lock.majority_count = state.majority_count();

        for (p, c) in state.peers().iter() {
             lock.peers_info.insert(c.addr(), *p);
        }
    }
```

so the lock will be acquired only once.
@DarkEld3r do you approve?

SergeiMal: It is good idea.

andreyluchin: I am not sure that this is better. `write` on `RwLock` is fairly cheap and I don't think it is worth making whole method atomic. Besides, other methods in this `impl` block acquire the lock each time and I see no good reason for changing that tradition. (IMHO)

popzxc: There is one reason: it's more readable. And until benches show that performance became worse, I don't think that we should think about it.
"Premature optimization is the root of all evil".

And for the future I hope that we'll rewrite other cases too.
(IMHO :upside_down_face: )

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `exonum::crypto::x25519` module to convert from Ed25519 keys to X25519 keys
   has been introduced. (#722)
+- Added a public `consensus_status` endpoint. (#736)
```
DarkEld3r: Please add an empty line between the changelog entries. Additionally, endpoint should probably be named as `v1/consensus_status`.

exonum/src/api/public/system.rs
```diff
@@ -31,6 +31,14 @@ pub struct HealthCheckInfo {
     pub connectivity: bool,
 }
 
+/// ConsensusStatusInfo shows the possibility to achieve the consensus between validators
+/// in current state
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
+pub struct ConsensusStatusInfo {
+    /// Consensus status: true - if consensus achieved, else - false
```
DarkEld3r: Please add a dot at the end of the comment.

PR #738, opened 2018-06-14
Added StorageKey implementation for Signature

wingedfox:
Added macro for crypto types
Updated Hash and PublicKey to implement StorageKey via macro


DarkEld3r:
Thank you for the pull request! 🎉 

Personally, I'm not fond of declaring tests through macro, especially in case the test body can be extracted into a function. Anyway, we have many places with similar code and I don't mind refactoring it afterwards.

wingedfox:
@DarkEld3r got it, thank you.

PR #743, opened 2018-06-18
Add API endpoint to check consensus status(Reopened) [ECR-1649]

SergeiMal:
Check for active validators count

### DIFF
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -424,17 +424,10 @@ impl SharedNodeState {
 
     /// Updates internal majority count, from `State` of a blockchain node.
     fn update_majority_count(&self, state: &State) {
-        let mut majority_count = state.majority_count();
-
-        let peers: Vec<SocketAddr> = self.peers_info().iter().map(|(s, _k)| (*s)).collect();
-        if !peers.contains(&state.our_connect_message().addr()) {
-            majority_count -= 1;
-        }
-
         self.state
             .write()
             .expect("Expected write lock.")
-            .majority_count = majority_count;
+            .majority_count = state.majority_count();
     }
 
     /// Returns a boolean value which indicates whether the node is enabled
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -323,6 +323,7 @@ pub struct ApiNodeState {
     // TODO: Update on event? (ECR-1632)
     peers_info: HashMap<SocketAddr, PublicKey>,
     is_enabled: bool,
+    majority_count: usize,
 }
 
 impl ApiNodeState {
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -410,6 +410,34 @@ impl SharedNodeState {
         }
     }
 
+
+    /// Returns the majority count from the current "State"
+    /// of a blockchain node.
+    pub fn majority_count(&self) -> usize {
+        self.state
+            .read()
+            .expect("Expected read lock.")
+            .majority_count
+    }
+    /// Updates internal majority count, from `State` of a blockchain node.
+    pub fn update_majority_count(&self, state: &State) {
+
+        let mut majority_count = state.majority_count();
+
+        let peers :Vec<SocketAddr> = self.peers_info()
+            .iter()
+            .map(|(s, _k)| (*s))
+            .collect();
+        if !peers.contains(&state.our_connect_message().addr()) {
+            majority_count -= 1;
+        }
+
+        self.state
+            .write()
+            .expect("Expected write lock.")
+            .majority_count = majority_count;
+    }
+
     /// Returns a boolean value which indicates whether the node is enabled
     /// or not.
     pub fn is_enabled(&self) -> bool {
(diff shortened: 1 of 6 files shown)


exonum/src/blockchain/service.rs
```diff
@@ -323,6 +323,9 @@ pub struct ApiNodeState {
     // TODO: Update on event? (ECR-1632)
     peers_info: HashMap<SocketAddr, PublicKey>,
     is_enabled: bool,
+    is_validator: bool,
+    majority_count:usize,
```
vitvakatu: missing space

exonum/src/blockchain/service.rs
```diff
@@ -395,19 +398,42 @@ impl SharedNodeState {
...
+            lock.peers_info.insert(c.addr(), *p);
         }
+        for v in state.validators().iter() {
+            lock.validators.push(*v);
+        }
+    }
+
+    /// Returns a boolean value which indicates whether the consensus is
+    /// Returns the majority count from the current "State"
+    /// of a blockchain node.
```
vitvakatu: Docs are messed up

exonum/src/blockchain/service.rs
```diff
@@ -395,19 +398,42 @@ impl SharedNodeState {
...
+    pub fn is_consensus(&self) -> bool {
+        let lock = self.state.read()
+            .expect("Expected read lock.");
+
+        let mut active_validators :usize = 0;
```
vitvakatu: wrong spacing

exonum/src/blockchain/service.rs
```diff
@@ -395,19 +398,42 @@ impl SharedNodeState {
...
+        for (_s, p) in lock.peers_info.iter() {
+            if lock.validators.iter()
+                .any(|x| x.consensus_key == *p) {
+                active_validators = active_validators.saturating_add(1);
```
vitvakatu: I like safety, but `active_validators` is `usize` and I don't think it's actually realistic to use `saturating_add`.

exonum/src/blockchain/service.rs
```diff
@@ -395,19 +398,42 @@ impl SharedNodeState {
...
+            }
+        }
+
+        if lock.is_validator {
+            // Peers list doesn't include current node address, so we have to increment its length.
+            // E.g. if we have 3 items in peers list, it means that we have 4 nodes overall.
+            active_validators = active_validators.saturating_add(1);
+        }
+
+        lock.majority_count > 0 && active_validators >= lock.majority_count
```
vitvakatu: Why do we need to check for `majority_count > 0`?

testkit/tests/system_api.rs
```diff
@@ -52,3 +53,12 @@ fn test_network() {
     assert_eq!(info.protocol_version, PROTOCOL_MAJOR_VERSION);
     assert!(info.services.is_empty());
 }
+
+#[test]
+fn test_consensus_status() {
```
vitvakatu: I suppose, one validator isn't the right test case here. And you need to check for `false` values too.

exonum/src/blockchain/service.rs
```diff
@@ -395,19 +398,42 @@ impl SharedNodeState {
...
+    /// Returns the majority count from the current "State"
+    /// of a blockchain node.
+    pub fn is_consensus(&self) -> bool {
+        let lock = self.state.read()
+            .expect("Expected read lock.");
+
+        let mut active_validators :usize = 0;
+        for (_s, p) in lock.peers_info.iter() {
+            if lock.validators.iter()
+                .any(|x| x.consensus_key == *p) {
```
popzxc: Personally I find expression `x.consensus_key == *p` very confusing. Is it possible to make it more readable?

exonum/src/blockchain/service.rs
```diff
@@ -395,19 +398,42 @@ impl SharedNodeState {
...
         }
+        for v in state.validators().iter() {
+            lock.validators.push(*v);
+        }
+    }
+
+    /// Returns a boolean value which indicates whether the consensus is
+    /// Returns the majority count from the current "State"
+    /// of a blockchain node.
+    pub fn is_consensus(&self) -> bool {
```
popzxc: Is it a good name?

popzxc:
General question: why is this PR was made from the already merged branch? 
Shouldn't it contain only one commit?

exonum/src/blockchain/service.rs
vitvakatu: I think the only way would be to rename `p` to something like `peer_key`

popzxc: And `x` to `peer`, I guess.

exonum/src/blockchain/service.rs
SergeiMal: At the beginning, just before the state uptaded, active_validators = 0 and majority_count = 0, and as result is_consensus will give true.

testkit/tests/system_api.rs
SergeiMal: It will be changed during ECR-1665

exonum/src/blockchain/service.rs
popzxc: I guess we should at least add a comment about it.

exonum/src/blockchain/service.rs
vitvakatu: `x` should be `validator`, I think

popzxc: Sure.

SergeiMal:
The PR closed. All changes (+ fixes according the comments ) will be committed with new separated PR

PR #744, opened 2018-06-19
Add API endpoint to check consensus status [ECR-1649]

SergeiMal:
Fixed reopened issue

### DIFF
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -414,15 +414,9 @@ impl SharedNodeState {
     pub fn consensus_status(&self) -> bool {
         let lock = self.state.read().expect("Expected read lock.");
 
-        let mut active_validators: usize = 0;
-        for peer_key in lock.peers_info.values() {
-            if lock.validators
-                .iter()
-                .any(|validator| validator.consensus_key == *peer_key)
-            {
-                active_validators += 1;
-            }
-        }
+        let mut active_validators = lock.peers_info.values().
+            filter(|peer_key| lock.validators.iter()
+                .any(|validator| validator.consensus_key == **peer_key)).count();
 
         if lock.is_validator {
             // Peers list doesn't include current node address, so we have to increment its length.
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -323,7 +323,9 @@ pub struct ApiNodeState {
     // TODO: Update on event? (ECR-1632)
     peers_info: HashMap<SocketAddr, PublicKey>,
     is_enabled: bool,
+    is_validator: bool,
     majority_count: usize,
+    validators: Vec<ValidatorKeys>,
 }
 
 impl ApiNodeState {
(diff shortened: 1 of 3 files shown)


exonum/src/blockchain/service.rs
```diff
@@ -400,19 +402,39 @@ impl SharedNodeState {
...
+        }
+
+        if lock.is_validator {
+            // Peers list doesn't include current node address, so we have to increment its length.
+            // E.g. if we have 3 items in peers list, it means that we have 4 nodes overall.
+            active_validators += 1;
+        }
+
+        // Just after Node is started (node status isn't updated) majority_count = 0,
+        // and as result is_consensus() will return true.
```
popzxc: I guess statement `and as result is_consensus() will return true` isn't quite correct.
I'd suggest `so we have to check that majority count is greater than 0.`.

exonum/src/blockchain/service.rs
```diff
@@ -400,19 +402,39 @@ impl SharedNodeState {
...
 
-    /// Returns the majority count from the current "State"
-    /// of a blockchain node.
-    pub fn majority_count(&self) -> usize {
-        self.state
-            .read()
-            .expect("Expected read lock.")
-            .majority_count
+    /// Returns a boolean value which indicates whether the consensus is
+    pub fn is_consensus(&self) -> bool {
```
popzxc: I still don't really like this name.

exonum/src/blockchain/service.rs
```diff
@@ -400,19 +402,39 @@ impl SharedNodeState {
...
         lock.peers_info.clear();
         lock.majority_count = state.majority_count();
+        lock.is_validator = state.is_validator();
+        lock.validators.clear();
 
-        for (p, c) in state.peers().iter() {
+        for (p, c) in state.peers() {
             lock.peers_info.insert(c.addr(), *p);
         }
+        for v in state.validators() {
```
popzxc: Maybe rewrite it to something like this? (then we won't have to clean validators list either)
```
lock.validators = state.validators().to_vec();
```

exonum/src/blockchain/service.rs
```diff
@@ -400,19 +402,37 @@ impl SharedNodeState {
...
     }
 
-    /// Returns the majority count from the current "State"
-    /// of a blockchain node.
-    pub fn majority_count(&self) -> usize {
-        self.state
-            .read()
-            .expect("Expected read lock.")
-            .majority_count
+    /// Returns a boolean value which indicates whether the consensus is
```
popzxc: `Whether the consensus is...`
A bit of intrigue? :)

SergeiMal: )) sorry "achieved" missed

exonum/src/blockchain/service.rs
```diff
@@ -400,19 +402,37 @@ impl SharedNodeState {
...
-        self.state
-            .read()
-            .expect("Expected read lock.")
-            .majority_count
+    /// Returns a boolean value which indicates whether the consensus is achieved.
+    pub fn consensus_status(&self) -> bool {
+        let lock = self.state.read().expect("Expected read lock.");
+
+        let mut active_validators: usize = 0;
+        for peer_key in lock.peers_info.values() {
```
DarkEld3r: I prefer to use `filter(...).count()` instead of loop.

testkit/tests/system_api.rs
```diff
@@ -55,10 +55,10 @@ fn test_network() {
 }
 
 #[test]
-fn test_consensus_status() {
+fn test_consensus_status_false() {
     let testkit = TestKitBuilder::validator().create();
     let api = testkit.api();
     let info: ConsensusStatusInfo = api.get(ApiKind::System, "v1/consensus_status");
-    let expected = ConsensusStatusInfo { status: true };
+    let expected = ConsensusStatusInfo { status: false };
```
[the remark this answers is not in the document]
SergeiMal: For now, this endpoint returns false in case one node. This is certainly correct behavior.
It going to be redesigned in context of ECR-1665.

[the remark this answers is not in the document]
popzxc: Ok, sure.

PR #745, opened 2018-06-19
Round-trip tests refactoring

DarkEld3r:


exonum/src/crypto/mod.rs
```diff
@@ -851,59 +851,64 @@ impl CryptoHash for Decimal {
...
     }
 
     #[test]
-    fn test_serialize_deserialize() {
-        let h = Hash::new([207; 32]);
-        let json_h = serde_json::to_string(&h).unwrap();
-        let h1 = serde_json::from_str(&json_h).unwrap();
-        assert_eq!(h, h1);
+    fn serialize_deserialize_hash() {
+        assert_serialize_deserialize(&Hash::new([207; 32]));
```
popzxc: Question: as far as I can see, all the tested hashes are filled with the same value (`0xCF` here). 
Do we have (and do we need) any tests to check that byte order is correct? As far as I can see, current tests will pass if somehow serialization will be done in little-endian, and deserialization in big-endian.

DarkEld3r: I don't want to rewrite tests in this pull request, but it is worth creating a separate issue.

exonum/src/storage/keys.rs
```diff
@@ -617,27 +610,32 @@ mod tests {
...
-        let mut buffer = [0u8; 16];
-        for value in values.iter() {
-            value.write(&mut buffer);
-            assert_eq!(*value, Decimal::read(&buffer));
+        assert_round_trip_eq(&decimals);
+    }
+
+    fn assert_round_trip_eq<T>(values: &[T])
+    where
+        T: StorageKey + Clone + PartialEq<<T as ToOwned>::Owned> + Debug,
```
ozkriff: Why is Clone required here?

DarkEld3r: It isn't. :upside_down_face:

PR #747, opened 2018-06-21
API endpoint consensus_enabled returns incorrect consensus status for validator node [ECR-1352]

VladYermakov:
API endpoint `consensus_enabled` returned` false` status at start because `SharedNodeState`'s constructor just calling `Default::default` and setting `is_enabled` by default to `false`.
So we need to set it to true manually or change the constructor.

### DIFF
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -330,7 +330,10 @@ pub struct ApiNodeState {
 
 impl ApiNodeState {
     fn new() -> ApiNodeState {
-        Self::default()
+        Self {
+            is_enabled: true,
+            ..Default::default()
+        }
     }
 }
 
--- a/exonum/src/node/mod.rs
+++ b/exonum/src/node/mod.rs
@@ -439,8 +439,6 @@ impl NodeHandler {
             system_state.current_time(),
         );
 
-        api_state.set_enabled(true);
-
         NodeHandler {
             blockchain,
             api_state,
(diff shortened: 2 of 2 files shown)


popzxc:
Hm.. Is the following case possible (I'm really not sure)?
Client for some reason wants to implement his own `Node` (e.g. he wants to write a brand new sandbox). Since `NodeHandler::new` takes `SharedNodeState` as an argument, client creates `SharedNodeState`, sets  `is_enabled` to `false` to explicitly disable consensus at the very beginning and constructs a `NodeHandler`. Then he discovers that consensus is enabled anyway and gets confused.

Sound kind of unrealistic, but if it's possible, I'd expect the following things:
1. `Default::default()` for `SharedNodeState` sets `is_enabled` to `true`.
2. `NodeHandler::new` checks `SharedNodeState::is_enabled` and sets its own `is_enabled` to the appropriate value.

VladYermakov:
@popzxc we cannot check the value because we don't know difference between default `false` and `false` set by client. So I think it's better to change the `SharedNodeState` constructor.

popzxc:
What I suggest is to set default value to `true`. Then, if the `true` is default, we don't care because consensus will be enabled as expected. But if the client will set it to `false`, we'll handle it as well.

popzxc:
It's better now, but still if someone'll change `SharedNodeState` to `false` and pass it to the `NodeHandler` we'll get incorrect behavior (`NodeHandler::is_enabled` will be `true`, and `SharedApiState::is_enabled` will be `false`).

I guess you should also add `self.is_enabled = api_state.is_enabled()` to the `NodeHandler::new(..)`.

VladYermakov:
@popzxc, hm, I got it. I'll change it a little bit later

PR #751, opened 2018-06-24
Clippy pedantic: Update Clippy to newest version [ECR-1621]

SergeiMal:
Update clippy to newest version
+ switch on clippy_pedantic lint

Now, our code contains a lot of places where the lints of pedantic group are trigered.
List of triggered lints at current time:
https://rust-lang-nursery.github.io/rust-clippy/v0.0.209/index.html#stutter
https://rust-lang-nursery.github.io/rust-clippy/v0.0.209/index.html#unseparated_literal_suffix
https://rust-lang-nursery.github.io/rust-clippy/v0.0.209/index.html#similar_names
https://rust-lang-nursery.github.io/rust-clippy/v0.0.209/index.html#if_not_else  
https://rust-lang-nursery.github.io/rust-clippy/v0.0.209/index.html#items_after_statements
https://rust-lang-nursery.github.io/rust-clippy/v0.0.209/index.html#doc_markdown

I propose to allow first 3 lints, they are quite useles. Butnext 3 lints need to be activated, and the souce code have to be changed according to the lints, I think. 

Examples of catched erroers:

```
##############################
1. stutter
error: item name ends with its containing module's name
  --> exonum/src/encoding/serialize/json.rs:52:1
   |
52 | / pub trait ExonumJson {
53 | |     /// write deserialized field in buffer on place.
54 | |     fn deserialize_field<B: WriteBufferWrapper>(
55 | |         value: &Value,
...  |
63 | |     fn serialize_field(&self) -> Result<Value, Box<Error + Send + Sync>>;
64 | | }
   | |_^
   |

##############################
2. unseparated_literal_suffix
error: integer type suffix should be separated by an underscore
   --> exonum/src/encoding/fields.rs:415:40
    |
415 |                     .copy_from_slice(&[0u8; SIZE_DIFF]);
    |                                        ^^^
    |

##############################
3. similar_names  
error: binding's name is too similar to existing binding
  |
note: lint level defined here
 --> exonum/src/lib.rs:24:44
  |
24| #![cfg_attr(feature = "cargo-clippy", deny(clippy_pedantic))]
  |                                            ^^^^^^^^^^^^^^^
  = note: #[deny(similar_names)] implied by #[deny(clippy_pedantic)]
note: existing binding defined here
 --> exonum/src/helpers/fabric/mod.rs:186:22
  |
18|     fn new_from_args(args: &[Argument], matches: &clap::ArgMatches) -> Context {
  |                      ^^^^
help: separate the discriminating character by an underscore like: `arg_0`


  error: binding's name is too similar to existing binding
  --> exonum/src/crypto/x25519.rs:51:9
   |
51 |     let sk_sod = SecretKeySodium::from_slice(&sk[..])?;
   |         ^^^^^^
   |
note: existing binding defined here
  --> exonum/src/crypto/x25519.rs:50:9
   |
50 |     let pk_sod = PublicKeySodium::from_slice(&pk[..])?;
   |         ^^^^^^
help: separate the discriminating character by an underscore like: `s_k_sod`
  --> exonum/src/crypto/x25519.rs:51:9
   |
51 |     let sk_sod = SecretKeySodium::from_slice(&sk[..])?;
   |         ^^^^^^
   
##############################
4. if_not_else
error: Unnecessary `!=` operation
   --> exonum/src/node/basic.rs:111:20
    |
111 |               } else if saved_message.addr() != message.addr() {
    |  ____________________^
112 | |                 error!("Received weird Connect message from {}", address);
113 | |                 return;
114 | |             } else {
115 | |                 need_connect = false;
116 | |             }
    | |_____________^
    |

    = help: change to `==` and swap the blocks of the if/else

##############################	
5. items_after_statements
 error: adding items after statements is confusing, since items exist from the start of the scope
  --> exonum/src/node/requests.rs:83:9
   |
83 | /         const EMPTY_RESPONSE_SIZE: u32 =
84 | |             (HEADER_LENGTH + SIGNATURE_LENGTH + 2 * PUBLIC_KEY_LENGTH + 8) as u32;
   | |__________________________________________________________________________________^
   |

##############################
6. doc_markdown
error: you should put `ConsensusStatusInfo` between ticks in the documentation
  --> exonum/src/api/public/system.rs:34:5
   |
34 | /// ConsensusStatusInfo shows the possibility to achieve the consensus between validators
   |     ^^^^^^^^^^^^^^^^^^^
   |
```


DarkEld3r:
Personally, I like the idea of separating literal suffixes by an underscore, but I agree that "slutter" and "similar_names" lints should be disabled.

exonum/src/lib.rs
```diff
@@ -21,6 +21,7 @@
 #![cfg_attr(feature = "flame_profile", feature(plugin, custom_attribute))]
 #![cfg_attr(feature = "flame_profile", plugin(exonum_flamer))]
 #![cfg_attr(feature = "long_benchmarks", feature(test))]
+#![cfg_attr(feature = "cargo-clippy", deny(unseparated_literal_suffix))]
```
DarkEld3r: I prefer to enable all pedantic warnings and disable some of them instead of enabling individual lints.

exonum/src/lib.rs
```diff
@@ -23,6 +23,11 @@
 #![cfg_attr(feature = "long_benchmarks", feature(test))]
 #![cfg_attr(feature = "cargo-clippy", deny(clippy_pedantic))]
 #![cfg_attr(feature = "cargo-clippy", allow(stutter, similar_names, items_after_statements))]
+#![cfg_attr(feature = "cargo-clippy", allow(use_self, cast_possible_truncation,
+                                            replace_consts, enum_glob_use, cast_sign_loss,
+                                            used_underscore_binding, cast_possible_wrap,
+                                            single_match_else, option_map_unwrap_or_else,
+                                            option_map_unwrap_or, filter_map))]
```
SergeiMal: The list of the lints:

use_self
cast_possible_truncation
replace_consts
enum_glob_use
cast_sign_loss
used_underscore_binding
cast_possible_wrap
single_match_else
option_map_unwrap_or_else
option_map_unwrap_or
filter_map

vitvakatu:
```diff
-    fn from(e: io::Error) -> ApiError {
+    fn from(e: io::Error) -> Self {
```
:smiley:

popzxc:
I guess current master should be merged.

exonum/src/crypto/x25519.rs
```diff
@@ -88,7 +88,7 @@ pub fn scalarmult_base(sc: &SecretKey) -> PublicKey {
 /// Converts Ed25519 public key to Curve25519 public key.
 ///
 /// See: [`into_x25519_keypair()`][1]
-/// [1]: fn.into_x25519_public_key.html
+/// [1]: `fn.into_x25519_public_key.html`
```
popzxc: Is it needed indeed? It's just a link.

exonum/src/encoding/segments.rs
```diff
@@ -197,6 +197,7 @@ impl<'a> SegmentField<'a> for RawMessage {
     }
 }
 
+#[cfg_attr(feature = "cargo-clippy", allow(use_self))]
```
popzxc: Why do we need it here?

exonum/src/encoding/serialize/json.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//#![cfg_attr(feature = "cargo-clippy", allow(use_self))]
```
popzxc: Commented line.

exonum/src/encoding/serialize/json.rs
```diff
@@ -316,6 +318,7 @@ impl<'a> ExonumJson for &'a [u8] {
     }
 }
 
+#[cfg_attr(feature = "cargo-clippy", allow(use_self))]
```
popzxc: Please here and below (if there are any more cases) either resolve the problem or document why `allow(use_self)` must be used.

exonum/src/encoding/serialize/json.rs
```diff
@@ -325,7 +328,7 @@ impl ExonumJson for Vec<RawMessage> {
     ) -> Result<(), Box<dyn Error>> {
         use messages::MessageBuffer;
         let bytes = value.as_array().ok_or("Can't cast json as array")?;
-        let mut vec: Vec<_> = Vec::new();
+        let mut vec: Self = Self::new();
```
popzxc: @SergeiMal Is clippy warning really triggered here? And why are you using `Self` when you set `use_self` to `allow` for this `impl`?

@DarkEld3r If so, seems like here `Self` usage is somewhat confusing. IMHO we should think if we need this lint at all. I like using `Self` in signatures, but this seems too much for me.

popzxc: Also there are some similar cases around the code.

exonum/src/lib.rs
```diff
@@ -18,6 +18,15 @@
 
 #![deny(missing_debug_implementations, missing_docs, unsafe_code, bare_trait_objects)]
 #![cfg_attr(feature = "long_benchmarks", feature(test))]
+//#![cfg_attr(feature = "cargo-clippy", deny(clippy_pedantic))]
```
popzxc: Commented code.

exonum/src/node/state.rs
```diff
@@ -805,6 +804,7 @@ impl State {
 
     /// Returns pre-votes for the specified round and propose hash.
     pub fn prevotes(&self, round: Round, propose_hash: Hash) -> &[Prevote] {
+        #[cfg_attr(feature = "cargo-clippy", allow(option_map_unwrap_or_else))]
```
popzxc: (here and below) Why? I'd expect any lint exception to be explained.

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -693,6 +680,19 @@ where
                                 .create();
                         }
                     }
+                } else {
+                    // Both children of `branch` do not fit
```
popzxc: Missing dot. And comment itself isn't helpful enough, I think.

exonum/src/encoding/serialize/json.rs
DarkEld3r: This code looks strange for me too.

exonum/src/storage/proof_map_index/proof.rs
vitvakatu: Please do not remove any comments from proof_map implementation. :upside_down_face: I doubt anyone understands this code completely.

popzxc: I want this comment to be more detailed. Maybe it'll be a good idea to create an issue to document this code well?

SergeiMal: I didn't change anything in the code, just make changes of if-else statement according to if_not_else lint.

exonum/src/crypto/x25519.rs
SergeiMal: It required by doc_markdown lint

exonum/src/storage/proof_map_index/proof.rs
popzxc: Ah, I didn't notice that. Then I'll hide this remark.

exonum/src/crypto/x25519.rs
popzxc: Then why some similar cases below aren't formatted this way?

exonum/src/encoding/segments.rs
SergeiMal: Because of:
```
error: unnecessary structure name repetition
   --> exonum\src\encoding\segments.rs:228:46
    |
228 |     fn extend_buffer(&self, mut buffer: &mut Vec<u8>) {
    |                                              ^^^^^^^ help: use the applicable keyword: `Self`
    |
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.211/index.html#use_self
```

popzxc: One more reason to disable this lint, as for me.

exonum/src/storage/values.rs
```diff
@@ -114,11 +114,11 @@ impl StorageValue for bool {
 }
 
 impl StorageValue for u8 {
-    fn into_bytes(self) -> Vec<u8> {
+    fn into_bytes(self) -> Vec<Self> {
```
SergeiMal: How do you think, is it make sense to use Self? Or restore u8? The same in other places with integral types.

popzxc: When we're using default types I stand for type name instead of `Self`. 
`Self` looks like something that belongs to us, and I won't expect it to be `Vec`, `u8`, `Result` or something similar.

But that's my opinion and other comrades may disagree though.

exonum/src/encoding/segments.rs
SergeiMal: I will remove use_self allowing and in another places too.

SergeiMal: I agree with popzxc, and will disable this one. But I propose to enable it time to time, and verify all sources to catch and fix exactly indisputable places.

exonum/src/storage/values.rs
DarkEld3r: For me it is somewhat strange to see thing like `Vec<Self>` or `HashMap<Self, i32>`. `Result<Self, _>` is OK though.

exonum/src/crypto/x25519.rs
SergeiMal: It is question to clippy ;)

SergeiMal: I guess it is the same situation like use_self

exonum/src/storage/values.rs
DarkEld3r: Additionally, in this situations `Vec<u8>` doesn't mean `Vec<Self>` at all, it is `Vec<byte>`.

DarkEld3r: https://github.com/rust-lang-nursery/rust-clippy/issues/2894

exonum/src/lib.rs
```diff
@@ -18,6 +18,15 @@
 
 #![deny(missing_debug_implementations, missing_docs, unsafe_code, bare_trait_objects)]
 #![cfg_attr(feature = "long_benchmarks", feature(test))]
+#![cfg_attr(feature = "cargo-clippy", deny(clippy_pedantic))]
+#![cfg_attr(
+    feature = "cargo-clippy",
+    allow(
+          stutter, similar_names, items_after_statements, use_self,
```
DarkEld3r: Why is the `items_after_statements` lint suppressed?

exonum/src/lib.rs
```diff
@@ -18,6 +18,15 @@
...
+          cast_possible_wrap, cast_possible_truncation, cast_sign_loss,
+          pub_enum_variant_names, single_match_else, option_map_unwrap_or_else
```
DarkEld3r: `pub_enum_variant_names` can be controversial (however I'm curious about how many such cases do we have), but why are `single_match_else` and `option_map_unwrap_or_else` suppressed?

exonum/src/storage/values.rs
```diff
@@ -244,7 +244,7 @@ impl StorageValue for RawMessage {
 }
 
 impl StorageValue for Vec<u8> {
-    fn into_bytes(self) -> Vec<u8> {
+    fn into_bytes(self) -> Self {
```
DarkEld3r: `Vec<u8>` must be used here.

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -152,9 +152,9 @@ pub enum ChildKind {
 }
 
 impl ::std::ops::Not for ChildKind {
-    type Output = ChildKind;
+    type Output = Self;
```
DarkEld3r: Not related to your changes, but lets add `use std::ops` and change line above to the `impl ops::Not for ChildKind`.

exonum/src/lib.rs
```diff
@@ -18,6 +18,15 @@
 
 #![deny(missing_debug_implementations, missing_docs, unsafe_code, bare_trait_objects)]
 #![cfg_attr(feature = "long_benchmarks", feature(test))]
+#![cfg_attr(feature = "cargo-clippy", deny(clippy_pedantic))]
+#![cfg_attr(
```
DarkEld3r: Lets add an explanation why are these lints are disabled. For example:
`slutter` and `similar_names` produce too much noise, `cast_possible_wrap`, `cast_possible_truncation` and `cast_sign_loss` don't give alternatives, `use_self` - can be enabled when https://github.com/rust-lang-nursery/rust-clippy/issues/2894 is fixed.

exonum/src/lib.rs
```diff
@@ -22,9 +22,19 @@
 #![cfg_attr(
     feature = "cargo-clippy",
     allow(
-          stutter, similar_names, items_after_statements, use_self,
+          items_after_statements, option_map_unwrap_or_else,
+          // Next cast.. lints don't give alternatives.
           cast_possible_wrap, cast_possible_truncation, cast_sign_loss,
-          pub_enum_variant_names, single_match_else, option_map_unwrap_or_else
+          // The lint does not work properly with desugaring and macro.
+          used_underscore_binding,
```
DarkEld3r: What king of desugaring? As for macros: I propose to suppress this lint locally (inside of macro or inside of modules with heavy macros usage).

exonum/src/lib.rs
```diff
@@ -18,6 +18,26 @@
 
 #![deny(missing_debug_implementations, missing_docs, unsafe_code, bare_trait_objects)]
 #![cfg_attr(feature = "long_benchmarks", feature(test))]
+#![cfg_attr(feature = "cargo-clippy", deny(clippy_pedantic))]
+#![cfg_attr(
+    feature = "cargo-clippy",
+    allow(
+          option_map_unwrap_or_else,
```
DarkEld3r: What with this lint?

exonum/src/lib.rs
```diff
@@ -18,6 +18,26 @@
...
+          // Next cast.. lints don't give alternatives.
+          cast_possible_wrap, cast_possible_truncation, cast_sign_loss,
+          // The lint does not work properly with desugaring and macro.
+          used_underscore_binding,
```
DarkEld3r: What king of desugaring? As for macros: I propose to suppress this lint locally (inside of the macro or inside of the modules with heavy macros usage).

exonum/src/blockchain/service.rs
```diff
@@ -520,6 +520,6 @@ impl SharedNodeState {
 
 impl<'a, S: Service> From<S> for Box<dyn Service + 'a> {
     fn from(s: S) -> Self {
-        Box::new(s) as Box<dyn Service>
+        Box::new(s) as Self
```
popzxc: @DarkEld3r what do you think about cases like this?

exonum/src/lib.rs
```diff
@@ -18,6 +18,26 @@
...
+          // filter(..).map(..) often looks more shorter and readable.
+          filter_map,
+          // Next lints produce too much noise.
+          stutter, similar_names,
+          // Variant name ends with the enum's name. Similar behavior to similar_names.
+          pub_enum_variant_names,
+          // Next lints allowed due to false possitive.
+          doc_markdown,
+          use_self, // can be enabled when rust-lang-nursery/rust-clippy#2894 is fixed.
```
DarkEld3r: Please move last comment before lint (and use the capital letter). :upside_down_face:

exonum/src/lib.rs
```diff
@@ -18,6 +18,26 @@
...
+#![cfg_attr(feature = "cargo-clippy", deny(clippy_pedantic))]
+#![cfg_attr(
+    feature = "cargo-clippy",
+    allow(
+          option_map_unwrap_or_else,
+          // Next cast.. lints don't give alternatives.
+          cast_possible_wrap, cast_possible_truncation, cast_sign_loss,
+          // The lint does not work properly with desugaring and macro.
+          used_underscore_binding,
+          // filter(..).map(..) often looks more shorter and readable.
```
popzxc: Please surround `filter(..).map(..)` with grave accent marks (`).

exonum/src/lib.rs
```diff
@@ -18,6 +18,26 @@
...
+          filter_map,
+          // Next lints produce too much noise.
```
popzxc: `/s/noise/false positives/`

exonum/src/lib.rs
```diff
@@ -18,6 +18,26 @@
 
 #![deny(missing_debug_implementations, missing_docs, unsafe_code, bare_trait_objects)]
 #![cfg_attr(feature = "long_benchmarks", feature(test))]
+#![cfg_attr(feature = "cargo-clippy", deny(clippy_pedantic))]
+#![cfg_attr(
+    feature = "cargo-clippy",
+    allow(
+          option_map_unwrap_or_else,
+          // Next cast.. lints don't give alternatives.
```
popzxc: I'd suggest replace "cast.." with \`cast_*\` or at least \`simple..\`. For me "cast.." looks like a punctuation part here.

exonum/src/lib.rs
popzxc: Please either move this line above (if you want to focus that this lint is disabled because of its incompleteness) or add `TODO` (if you want to highlight that this lint is temporary disabled).

exonum/src/blockchain/service.rs
DarkEld3r: Honestly, I don't have a strong opinion here. 🤔

ozkriff:
So, what is the formal consensus about `self` thing and is it documented somewhere?

popzxc:
I guess not. My opinion is simple: I like `Self`, but not when it's used within default types (except wrappers - `Option`, `Result`, `Mutex`, etc).
What is party politics I cannot tell though.

PR #752, opened 2018-06-24
Update Clippy to  v. 0.0.209 [ECR-1621]

SergeiMal:
Clippy version incriesed on Travis.
Also PR includes:
- 'unwrap_or_else(|| panic!(...))' used instead of 'expect' 
- Allow 'expect_fun_call' in case with second nested closure


3rdparty/profiler/src/lib.rs
```diff
@@ -88,15 +88,15 @@ impl ThreadFrame {
         let new_time = INTERRUPTED_TICKS.load(Ordering::SeqCst);
         if self.dumped_time < new_time {
             let name: String = SETTED_NAME.lock().unwrap().clone()
-                                .expect("Profiler: received interrupt without setted name.");
+                                .unwrap_or_else(|| panic!("Profiler: received interrupt without setted name."));
```
popzxc: @DarkEld3r, so what is our decision about `unwrap_or_else`? I still think that it looks ugly and more verbose.

ozkriff: maybe it would be better to run some microbenchmarks on both versions and see if it actually improves performance?

DarkEld3r: It would be interesting to run benchmarks, but I don't think that it will really change anything. I doubt that the difference will be significant.

exonum/src/helpers/types.rs
```diff
@@ -51,7 +51,7 @@ impl Height {
     /// let next_height = height.next();
     /// assert_eq!(11, next_height.0);
     /// ```
-    pub fn next(&self) -> Self {
+    pub fn next(self) -> Self {
```
SergeiMal: And what do you think about this changes. 
It catched by trivially_copy_pass_by_ref  with error: 'this argument is passed by reference, but would be more efficient if passed by value'

alexander-irbis: It is common to pass copy-types by value.

3rdparty/profiler/src/lib.rs
popzxc: Well, I guess it's because of [optimizations](https://doc.rust-lang.org/1.7.0/book/benchmark-tests.html#gotcha-optimizations).
Try to use `blackbox` or something.

popzxc: And also I'd expect something more realistic, e.g. benchmark that can either fail or not.

SergeiMal: Please see new results for next bench tests:
'''
 pub fn test_expect_bb( st: &str )
 {
     let _path = test::black_box(Some("Test")).expect(&format!("Unable to get  - {} ", st));
 }

 pub fn test_unwrap_or_else_bb(st: &str )
 {
     let _path = test::black_box(Some("Test")).unwrap_or_else(|| panic!("Unable to get  - {} ", st));
 }

 #[cfg(test)]
 mod tests {
     use super::*;
     use test::Bencher;

     #[bench]
     fn bench_expect(b: &mut Bencher) {
         b.iter(|| (0..1000).for_each(|_| test_expect_bb("test")));
     }

     #[bench]
     fn bench_unwrap_or_else(b: &mut Bencher) {
         b.iter(|| (0..1000).for_each(|_| test_unwrap_or_else_bb("test")));
     }
 }
'''

running 2 tests
test tests::bench_expect         ... bench:     100,821 ns/iter (+/- 12,645)
test tests::bench_unwrap_or_else ... bench:       1,334 ns/iter (+/- 59)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured; 0 filtered out

popzxc: Well, and if we'll bench some function which with equal probability will return either `Some("Test")` or `None` (not only `Some(..)`)?

SergeiMal: Unfortunally, the tests failed in None case.
I tried with

 pub fn test_expect_bb( st: &str, a: i32 )
 {
     let mut some_none = None;
     if a%2 > 0{
         some_none = Some("test");
     }
     let _path = test::black_box(some_none).expect(&format!("Unable to get  - {} ", st));
 }

     fn bench_expect(b: &mut Bencher) {
         b.iter(|| (0..1000).for_each(|a| test_expect_bb("test", test::black_box(a))));
     }

popzxc: Maybe [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) will help?

SergeiMal: Results with catch_unwrap do not differ so much between themselves:

running 2 tests
test tests::bench_expect         ... bench:   2,447,747 ns/iter (+/- 1,291,382)
test tests::bench_unwrap_or_else ... bench:   2,128,982 ns/iter (+/- 179,077)

popzxc: @DarkEld3r what do you think? 
Seems like `unwrap_or_else` is slightly faster, but is that difference significant?

SergeiMal: In my opinion, catch_unwind is very slow and neutralizes differ between expect(fn()) and unwrap_or_else( || ..).
So I changed the rule to generate the panic only one time of 50, and it gives more interesting result I think:

running 2 tests
test tests::bench_expect         ... bench:     206,515 ns/iter (+/- 30,707)
test tests::bench_unwrap_or_else ... bench:      97,094 ns/iter (+/- 13,507)

DarkEld3r: @popzxc 
I don't think that performance gain is worth it, on the other hand `.unwrap_or_else(|| panic!(...))` doesn't seem so terrible for me (now).

I'm definitely against such changes:
```diff
- .expect("Profiler: received interrupt without setted name.");
+ .unwrap_or_else(|| panic!("Profiler: received interrupt without setted name."));
```
There is no function call inside of `expect`.

exonum/src/helpers/fabric/details.rs
```diff
@@ -332,6 +332,7 @@ impl GenerateNodeConfig {
         let error_msg = &format!("Expected an ip address in {}: {:?}", PEER_ADDRESS, addr_str);
 
         let external_addr = addr_str.parse::<SocketAddr>().unwrap_or_else(|_| {
+            #[cfg_attr(feature = "cargo-clippy", allow(expect_fun_call))]
```
ozkriff: maybe I missed something, but this piece of code looks fishy. why are we pre-formatting this error msg before the closure and disabling the lint?

PR #753, opened 2018-06-25
Incorrect logic of "consensus_enabled" endpoint for auditor [ECR-1728]

VladYermakov:
Add `is_validator` field in `NodeHandler` struct.
Set `is_enabled` to `false` is node is auditor, and block its changing.
Question: how can we test this functionality?

### DIFF
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -29,7 +29,7 @@ use crypto::{Hash, PublicKey, SecretKey};
 use encoding::Error as MessageError;
 use helpers::{Height, Milliseconds, ValidatorId};
 use messages::RawTransaction;
-use node::{ApiSender, Node, State, TransactionSend};
+use node::{ApiSender, Node, NodeRole, State, TransactionSend};
 use storage::{Fork, Snapshot};
 
 /// A trait that describes the business logic of a certain service.
@@ -323,7 +323,7 @@ pub struct ApiNodeState {
     // TODO: Update on event? (ECR-1632)
     peers_info: HashMap<SocketAddr, PublicKey>,
     is_enabled: bool,
-    is_validator: bool,
+    node_role: NodeRole,
     majority_count: usize,
     validators: Vec<ValidatorKeys>,
 }
@@ -405,7 +405,10 @@ impl SharedNodeState {
 
         lock.peers_info.clear();
         lock.majority_count = state.majority_count();
-        lock.is_validator = state.is_validator();
+        lock.node_role = match state.validator_id() {
+            Some(validator_id) => NodeRole::Validator(validator_id),
+            None => NodeRole::Auditor,
+        };
         lock.validators = state.validators().to_vec();
 
         for (p, c) in state.peers() {
@@ -426,7 +429,7 @@ impl SharedNodeState {
             })
             .count();
 
-        if lock.is_validator {
+        if lock.node_role.is_validator() {
             // Peers list doesn't include current node address, so we have to increment its length.
             // E.g. if we have 3 items in peers list, it means that we have 4 nodes overall.
             active_validators += 1;
@@ -451,9 +454,9 @@ impl SharedNodeState {
         state.is_enabled = is_enabled;
     }
 
-    pub(crate) fn set_validator(&self, is_validator: bool) {
+    pub(crate) fn set_node_role(&self, role: NodeRole) {
         let mut state = self.state.write().expect("Expected read lock.");
-        state.is_validator = is_validator;
+        state.node_role = role;
     }
 
     /// Returns the value of the `state_update_timeout`.
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -451,6 +451,11 @@ impl SharedNodeState {
         state.is_enabled = is_enabled;
     }
 
+    pub(crate) fn set_validator(&self, is_validator: bool) {
+        let mut state = self.state.write().expect("Expected read lock.");
+        state.is_validator = is_validator;
+    }
+
     /// Returns the value of the `state_update_timeout`.
     pub fn state_update_timeout(&self) -> Milliseconds {
         self.state_update_timeout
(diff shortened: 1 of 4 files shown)


exonum/src/node/events.rs
```diff
@@ -77,6 +77,10 @@ impl NodeHandler {
                 self.connect(&address);
             }
             ExternalMessage::Enable(value) => {
+                if !self.is_validator {
+                    error!("Trying to enable auditor node");
```
vitvakatu: Error message is not clear for user.
"Trying to enable consensus, but the current node is auditor and can not affect consensus process."

exonum/src/node/mod.rs
```diff
@@ -439,7 +441,18 @@ impl NodeHandler {
             system_state.current_time(),
         );
 
-        let is_enabled = api_state.clone().is_enabled();
+        let mut is_enabled = api_state.clone().is_enabled();
+        let is_validator = validator_id.is_some();
+
+        if !is_validator {
+            if is_enabled {
+                error!("Provided enabled consensus for auditor node")
```
vitvakatu: Error message is not clear
"Consensus is enabled, but the current node is auditor"

exonum/src/node/mod.rs
```diff
@@ -439,7 +441,18 @@ impl NodeHandler {
...
+            }
+            is_enabled = false;
```
vitvakatu: `is_enabled` is unused

exonum/src/blockchain/service.rs
```diff
@@ -451,6 +451,11 @@ impl SharedNodeState {
         state.is_enabled = is_enabled;
     }
 
+    pub(crate) fn set_validator(&self, is_validator: bool) {
```
vitvakatu: Naming seems weird for me. Maybe `set_node_role`?

ozkriff: in that case, it should receive an enum

exonum/src/node/mod.rs
VladYermakov: No, it is used later in NodeHandler construction

exonum/src/blockchain/service.rs
VladYermakov: @ozkriff Is I got it right? `enum NodeRole { Validator, Auditor, }`
If yes, wouldn't it be overcomplicated?
Because we need to match this enum in `set_node_role` and before it to construct this enum we need to match bool value to this enum

vitvakatu: `is_validator` variable should be replaced with this enum as well

exonum/src/blockchain/service.rs
```diff
@@ -451,6 +454,11 @@ impl SharedNodeState {
         state.is_enabled = is_enabled;
     }
 
+    pub(crate) fn set_node_role(&self, role: NodeRole) {
+        let mut state = self.state.write().expect("Expected read lock.");
```
alekseysidorov: Maybe `write lock`?

exonum/src/node/mod.rs
```diff
@@ -449,6 +498,7 @@ impl NodeHandler {
             channel: sender,
             peer_discovery: config.peer_discovery,
             is_enabled,
+            node_role,
```
alekseysidorov: What happens with `node_role` if this node will become validator at runtime?

exonum/src/node/mod.rs
```diff
@@ -387,6 +389,39 @@ pub struct NodeSender {
...
+}
+
+impl Default for NodeRole {
+    fn default() -> Self {
+        NodeRole::Auditor
+    }
+}
+
+impl NodeRole {
+    /// Checks if node is validator
```
vitvakatu: Missing full stop at the end

exonum/src/node/events.rs
```diff
@@ -77,6 +77,10 @@ impl NodeHandler {
                 self.connect(&address);
             }
             ExternalMessage::Enable(value) => {
+                if self.node_role.is_auditor() {
+                    error!("Trying to enable consensus, but the current node is auditor and cannot affect consensus process");
```
popzxc: IMHO this message is too verbose. Wouldn't "Attempt to enable consensus for an auditor" be enough?

exonum/src/node/mod.rs
```diff
@@ -126,6 +126,8 @@ pub struct NodeHandler {
     pub peer_discovery: Vec<SocketAddr>,
     /// Does this node participate in the consensus?
     is_enabled: bool,
+    /// Is this node validator?
```
popzxc: I'd suggest simply "Node role.".

exonum/src/node/mod.rs
```diff
@@ -387,6 +389,39 @@ pub struct NodeSender {
     pub api_requests: SyncSender<ExternalMessage>,
 }
 
+///
+#[derive(Debug, Clone)]
+pub enum NodeRole {
+    ///
```
popzxc: Missing descriptions.

exonum/src/node/events.rs
vitvakatu: Any reason *not* to make message verbose?

popzxc: By "verbose" I mean "using or expressed in more words than are needed". If we are able to write message in less words with the same amount of information, I'd prefer to do it.

exonum/src/node/mod.rs
VladYermakov: Now: it just will forget to update `node_role` in `NodeHandler`, but will update `node_role` in `ApiState`.

exonum/src/node/mod.rs
```diff
@@ -439,7 +474,21 @@ impl NodeHandler {
             system_state.current_time(),
         );
 
-        let is_enabled = api_state.clone().is_enabled();
+        let mut is_enabled = api_state.clone().is_enabled();
+        let node_role = match validator_id {
```
vitvakatu: This match is used three times in code, what about moving it in  a method?

VladYermakov: Yeah, it should be better.

exonum/src/node/mod.rs
```diff
@@ -405,6 +405,14 @@ impl Default for NodeRole {
 }
 
 impl NodeRole {
+    /// Constructs new NodeRole by `validator_id`
```
vitvakatu: Add full stop at the end, please. And I think `from`, not `by`

PR #758, opened 2018-06-26
GET {system_base_path}/peers doesn't return list of incoming connections [ECR-1375]

VladYermakov:
Now we updating `incoming/outgoing_connections` in `SharedNodeState::update` together with `peers_info`.

### DIFF
--- a/exonum/src/node/state.rs
+++ b/exonum/src/node/state.rs
@@ -549,7 +549,7 @@ impl State {
         &self.peers
     }
 
-    /// Returns the addresses of known connections with their public keys.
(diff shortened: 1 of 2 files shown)


exonum/src/node/state.rs
```diff
@@ -549,6 +549,11 @@ impl State {
         &self.peers
     }
 
+    /// Returns the addresses of known connections with their public keys.
```
pavel-mukhanov: Comment is not clear, because public key identifies validator, not the connection.

PR #761, opened 2018-06-27
[Cryptocurrency-advanced] Transfer funds to yourself [ECR-1747]

VladYermakov:
Now when someone trying to send funds to itself error `Error::SendingFundsToYourself` will be displayed.

### DIFF
--- a/examples/cryptocurrency-advanced/backend/src/transactions.rs
+++ b/examples/cryptocurrency-advanced/backend/src/transactions.rs
@@ -47,12 +47,6 @@ pub enum Error {
     /// Can be emitted by `Transfer`.
     #[fail(display = "Insufficient currency amount")]
     InsufficientCurrencyAmount = 3,
-
-    /// Sending funds to yourself.
-    ///
-    /// Can be emitted by `Transfer`.
-    #[fail(display = "Sending funds to yourself")]
-    SendingFundsToYourself = 4,
 }
 
 impl From<Error> for ExecutionError {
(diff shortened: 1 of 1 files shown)


examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -47,6 +47,12 @@ pub enum Error {
     /// Can be emitted by `Transfer`.
     #[fail(display = "Insufficient currency amount")]
     InsufficientCurrencyAmount = 3,
+
+    /// Sending funds to yourself.
+    ///
+    /// Can be emitted by `Transfer`.
+    #[fail(display = "Sending funds to yourself")]
```
aleksuss: Would it better `An attempt to send funds to yourself` wouldn't it ?

examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -96,6 +102,10 @@ impl Transaction for Transfer {
         let hash = self.hash();
         let amount = self.amount();
 
+        if from == to {
+            Err(Error::SendingFundsToYourself)?
```
vitvakatu: Are you sure that we should check it in `execute`, not in `verify`? In basic cryptocurrency we do it in verify.

examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -91,7 +85,7 @@ transactions! {
 
 impl Transaction for Transfer {
     fn verify(&self) -> bool {
-        self.verify_signature(self.from())
+        (*self.from() != *self.to()) && self.verify_signature(self.from())
```
vitvakatu: Dereferencing should be redundant here (just `self.from() != self.to()`)

PR #763, opened 2018-06-27
Newer (1.27) formatting 

DarkEld3r:
- 1.27 formatting (better indentation).
- `![deny(missing_debug_implementations, missing_docs, unsafe_code, bare_trait_object)]` everywhere.

### DIFF
--- a/examples/cryptocurrency/src/lib.rs
+++ b/examples/cryptocurrency/src/lib.rs
@@ -23,7 +23,7 @@
 //! [docs]: https://exonum.com/doc/get-started/create-service
 //! [readme]: https://github.com/exonum/cryptocurrency#readme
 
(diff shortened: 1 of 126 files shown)


DarkEld3r:
It seems that `bare_trait_objects` must be allowed until `failure` is fixed.

DarkEld3r:
Finally. 🙃

examples/cryptocurrency/src/lib.rs
```diff
@@ -153,6 +155,8 @@ pub mod transactions {
 
 /// Contract errors.
 pub mod errors {
+    #![allow(bare_trait_objects)]
```
ozkriff: Maybe it would be better to leave s small explanation comment why this is allowed here (and with other allows)? With the link to ecr-1771?

DarkEld3r: Fixed. 👍

PR #756, opened 2018-06-26
Modular cryptographic backend

blackbeam:
This PR moves existing sodiumoxide-based cryptographic backend behind opt-out `sodiumoxide-crypto` feature. It also allows to use your own cryptographic library with exonum.

### Motivation
Currently there is no way to use other cryptographic libraries/algorithms with exonum. It is necessary for using local goverment certified libraries.

PR #764, opened 2018-06-27
ECR-1714 Some improvements of Noise Protocol Implementation

pavel-mukhanov:
### Changes:
- Use `Vec<u8>` instead of `(usize, &[u8])`.
- Remove redundant allocation in `write_handshake_msg`.
- Move `noise::read` and `noise::write` to `HandshakeRawMessage`.
- Update to `snow 0.2.1`. 

`HANDSHAKE_HEADER_LENGTH` has not been removed because it needed in #756.


exonum/src/events/noise/mod.rs
```diff
@@ -143,22 +143,29 @@ impl Handshake for NoiseHandshake {
...
-    buf: &[u8],
-    len: usize,
-) -> impl Future<Item = (S, Vec<u8>), Error = io::Error> {
-    debug_assert!(len < NOISE_MAX_HANDSHAKE_MESSAGE_LENGTH);
+    pub fn write<S: AsyncWrite + 'static>(
+        self,
+        sock: S,
+    ) -> impl Future<Item = (S, Vec<u8>), Error = io::Error> {
+        let len = self.0.len();
+        debug_assert!(len < NOISE_MAX_HANDSHAKE_MESSAGE_LENGTH);
```
slowli: Nit: I'd rename to `MAX_HANDSHAKE_MESSAGE_LENGTH` (similar to other constants) to avoid stuttering.

exonum/src/events/noise/mod.rs
```diff
@@ -143,22 +143,29 @@ impl Handshake for NoiseHandshake {
...
 
-    let mut message = vec![len as u8; HANDSHAKE_HEADER_LENGTH];
-    message.extend_from_slice(&buf[0..len]);
-    write_all(sock, message)
+        write_all(sock, vec![len as u8; HANDSHAKE_HEADER_LENGTH])
```
slowli: Kind reminder: the code in `read` implicitly fixes `HANDSHAKE_HEADER_LENGTH` as 1, so it could make sense to get rid of the constant and simplify the code accordingly.

exonum/src/events/noise/wrapper.rs
```diff
@@ -126,9 +126,9 @@ impl NoiseWrapper {
 
         msg.chunks(NOISE_MAX_MESSAGE_LENGTH - TAG_LENGTH)
             .for_each(|msg| {
```
slowli: Nit: `for` cycle seems more idiomatic Rust than `for_each`.

exonum/src/events/noise/wrapper.rs
```diff
@@ -126,9 +126,9 @@ impl NoiseWrapper {
...
-                let (written_bytes, written) = self.write(msg).unwrap();
+                let written = self.write(msg).unwrap();
```
slowli: - I'd add an `assert` or `debug_assert` that the length of `written` is as expected (`msg.len() + TAG_LENGTH`, if I understand correctly).
- Why is `unwrap()` OK here?

exonum/src/events/noise/wrapper.rs
```diff
@@ -105,7 +105,7 @@ impl NoiseWrapper {
                 msg.len()
             };
 
-            let (_, read_to) = self.read(msg, len_to_read).unwrap();
+            let read_to = self.read(msg, len_to_read).unwrap();
```
slowli: Like with writing, two suggestions:

- Explain why `unwrap()` is OK here.
- Check the length of the read message.

Also, if the length of `decoded_message` deterministically depends on the length of `data`, would it make sense to pre-allocate correspondingly?

exonum/src/events/noise/wrapper.rs
```diff
@@ -140,16 +140,16 @@ impl NoiseWrapper {
...
+        self.session.read_message(input, &mut buf)?;
+        Ok(buf)
     }
 
-    fn write(&mut self, msg: &[u8]) -> Result<(usize, Vec<u8>), NoiseError> {
+    fn write(&mut self, msg: &[u8]) -> Result<Vec<u8>, NoiseError> {
         let mut buf = vec![0u8; NOISE_MAX_MESSAGE_LENGTH];
         let len = self.session.write_message(msg, &mut buf)?;
-        Ok((len, buf))
+        Ok(buf[..len].to_vec())
```
slowli: Use `Vec::truncate`?

exonum/src/events/noise/wrapper.rs
vitvakatu: Don't think so, it's just a matter of taste.

slowli: Just looking [at Rust docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each):

> It's generally more idiomatic to use a `for` loop, but `for_each` may be more legible when processing items at the end of longer iterator chains.

exonum/src/events/noise/wrapper.rs
pavel-mukhanov: `unwrap` is not OK, I'll remove it.

exonum/src/events/noise/mod.rs
pavel-mukhanov: We have plans to use handshake messages with payload, therefore header length will be increased to 4(or maybe 8) bytes (https://github.com/blackbeam/exonum-1/blob/032cf1afc69b4bc77b381821db9c1b5cb359d73f/exonum/src/events/noise/mod.rs#L71). And so I decided to leave it.

exonum/src/events/noise/wrapper.rs
```diff
@@ -101,19 +99,26 @@ impl NoiseWrapper {
...
         let data = &data[NOISE_HEADER_LENGTH..];
-        let mut decoded_message = vec![0u8; 0];
 
-        data.chunks(NOISE_MAX_MESSAGE_LENGTH).for_each(|msg| {
-            let len_to_read = if msg.len() == NOISE_MAX_MESSAGE_LENGTH {
+        // Each message consists of the payload and 16 bytes(`TAG_LENGTH`)
+        // of AEAD authentification data. Therefore to calculate an actual message
+        // length we need to substract `TAG_LENGTH` multiplied by messages count
+        // from `data.len()`.
+        let len = data.len() - TAG_LENGTH * (data.len() / MAX_MESSAGE_LENGTH);
```
popzxc: I'd move this calculation to the separate method.

exonum/src/events/noise/wrapper.rs
```diff
@@ -126,35 +131,38 @@ impl NoiseWrapper {
...
-
-        msg.chunks(NOISE_MAX_MESSAGE_LENGTH - TAG_LENGTH)
-            .for_each(|msg| {
-                let (written_bytes, written) = self.write(msg).unwrap();
-                encoded_message.extend_from_slice(&written);
-                len += written_bytes;
-            });
+        // In case of encryption we need to add `TAG_LENGTH` multiplied by messages count to
+        // calculate actual message length.
+        let len = msg.len() + TAG_LENGTH * ((msg.len() / MAX_MESSAGE_LENGTH) + 1);
```
popzxc: I'd move this logic to the separate method.

exonum/src/events/noise/tests.rs
```diff
@@ -28,7 +28,10 @@ use crypto::{
     gen_keypair, gen_keypair_from_seed, x25519::into_x25519_keypair, Seed, PUBLIC_KEY_LENGTH,
 };
 use events::error::into_other;
-use events::noise::{write, Handshake, HandshakeParams, HandshakeResult, NoiseHandshake};
+use events::noise::{
```
DarkEld3r: It isn't related to your changes, but I have notices here and in the adjusted files that some imports are not grouped. Could you please fix it? :upside_down_face:

exonum/src/events/noise/wrapper.rs
```diff
@@ -126,35 +126,50 @@ impl NoiseWrapper {
     /// 4. Append all encrypted packets in corresponding order.
     /// 5. Write result message to `buf`
     pub fn encrypt_msg(&mut self, msg: &[u8], buf: &mut BytesMut) -> Result<Option<()>, io::Error> {
-        let mut len = 0_usize;
-        let mut encoded_message = vec![0_u8; 0];
+        let len = self.encrypted_msg_len(msg.len());
+        let mut encoded_message = Vec::with_capacity(len);
```
slowli: Nit: the message is *encrypted*, not encoded.

exonum/src/events/noise/wrapper.rs
```diff
@@ -101,19 +98,22 @@ impl NoiseWrapper {
...
         let data = &data[NOISE_HEADER_LENGTH..];
-        let mut decoded_message = vec![0_u8; 0];
 
-        data.chunks(NOISE_MAX_MESSAGE_LENGTH).for_each(|msg| {
-            let len_to_read = if msg.len() == NOISE_MAX_MESSAGE_LENGTH {
+        let len = self.decrypted_msg_len(data.len());
+        let mut decoded_message = Vec::with_capacity(len);
+
+        for msg in data.chunks(MAX_MESSAGE_LENGTH) {
+            let len_to_read = if msg.len() == MAX_MESSAGE_LENGTH {
```
slowli: Isn't tag applied to all messages? I'm not quite understanding the logic here.

exonum/src/events/noise/wrapper.rs
```diff
@@ -126,35 +126,50 @@ impl NoiseWrapper {
...
         let encoded_message = &encoded_message[0..len];
         msg_len_buf.extend_from_slice(encoded_message);
         buf.extend_from_slice(&msg_len_buf);
+
+        debug_assert_eq!(len, encoded_message.len());
         Ok(None)
     }
 
-    fn read(&mut self, input: &[u8], len: usize) -> Result<(usize, Vec<u8>), NoiseError> {
+    fn read(&mut self, input: &[u8], len: usize) -> Result<Vec<u8>, NoiseError> {
```
slowli: From the performance perspective, the most right way to do things would probably be to organize read as

```rust
fn read(&mut self, input: &[u8], output: &mut [u8]) -> Result<usize, NoiseError>
```

in order to be able to reuse the output buffer in `decrypt_msg`. At this point, it would probably be easier just to inline `self.session.read_message` :upside_down_face:

exonum/src/events/noise/wrapper.rs
```diff
@@ -126,35 +126,50 @@ impl NoiseWrapper {
...
+        buf.truncate(len);
+        Ok(buf)
+    }
+
+    // Each message consists of the payload and 16 bytes(`TAG_LENGTH`)
+    // of AEAD authentification data. Therefore to calculate an actual message
+    // length we need to substract `TAG_LENGTH` multiplied by messages count
+    // from `data.len()`.
+    fn decrypted_msg_len(&self, raw_message_len: usize) -> usize {
+        raw_message_len - TAG_LENGTH * (raw_message_len / MAX_MESSAGE_LENGTH)
```
slowli: I'm not sure the math is correct here (is it covered by tests?). If each Noise message contains a tag, then the formula would be

```rust
len - TAG_LENGTH * ceil(len / MAX_MESSAGE_LENGTH)
```

(Is there `ceil` for integers in stdlib?)

exonum/src/events/noise/wrapper.rs
```diff
@@ -126,35 +126,50 @@ impl NoiseWrapper {
...
+    }
+
+    // In case of encryption we need to add `TAG_LENGTH` multiplied by messages count to
+    // calculate actual message length.
+    fn encrypted_msg_len(&self, raw_message_len: usize) -> usize {
+        raw_message_len + TAG_LENGTH * ((raw_message_len / MAX_MESSAGE_LENGTH) + 1)
```
slowli: If each Noise message contains a tag, then the formula would be

```rust
len + TAG_LENGTH * (floor(len / (MAX_MESSAGE_LENGTH - TAG_LENGTH)) + 1)
```

(if I didn't screw the math, which *may* happen).

exonum/src/events/noise/mod.rs
slowli: But the code doesn't make sense with `HANDSHAKE_HEADER_LENGTH != 1`. I'm not sure it makes sense to generalize via `HANDSHAKE_HEADER_LENGTH` in general; you need to convert `len` to some type an write it (e.g., in little-endian).

exonum/src/events/noise/wrapper.rs
pavel-mukhanov: Your formula is right, 
```rust
raw_message_len - TAG_LENGTH * ((raw_message_len as f32 / MAX_MESSAGE_LENGTH as f32).ceil() as usize)
```
In the current implementation, we have an extra 16 bytes on the end of each message. But they are ignored because we specify length in a message header. I'll fix it in next PR. Tests also will be provided.

PR #767, opened 2018-06-29
Restore old error code behaviour [ECR-1799]

alekseysidorov:


exonum/src/api/backends/actix.rs
```diff
@@ -119,15 +121,22 @@ impl ExtendApiBackend for actix_web::Scope<ServiceApiState> {
 
 impl ResponseError for ApiError {
     fn error_response(&self) -> HttpResponse {
+        fn to_json(string: impl Display) -> serde_json::Value {
```
aleksuss: `dyn Display` ?

vitvakatu: Why not using https://docs.rs/serde_json/1.0.22/serde_json/fn.to_value.html ?

PR #772, opened 2018-07-03
Removed `iter_from` in `BaseIndex`

aleksuss:
Removed `iter_from` for getting rid of code duplication. `iter` has been become unified

### DIFF
--- a/exonum/src/storage/key_set_index.rs
+++ b/exonum/src/storage/key_set_index.rs
@@ -154,7 +154,7 @@ where
     /// ```
     pub fn iter(&self) -> KeySetIndexIter<K> {
         KeySetIndexIter {
-            base_iter: self.base.iter(&()),
+            base_iter: self.base.iter(&(), None as Option<&[u8]>),
         }
     }
 
@@ -177,7 +177,7 @@ where
     /// ```
     pub fn iter_from(&self, from: &K) -> KeySetIndexIter<K> {
         KeySetIndexIter {
-            base_iter: self.base.iter_from(&(), from),
+            base_iter: self.base.iter(&(), Some(from)),
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Breaking changes
 
- - A method `iter` in the `BaseIndex` module has been become unified and
-   a method `iter_from` has been removed. 
-   A method `iter` has changed a signature in trait `Snapshot`. (#772)
+- A method `iter` in the `BaseIndex` module has been become unified and
+  a method `iter_from` has been removed. 
+  A method `iter` has changed a signature in trait `Snapshot`. (#772)
 
 #### exonum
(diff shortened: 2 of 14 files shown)


exonum/src/storage/key_set_index.rs
```diff
@@ -154,7 +154,7 @@ where
     /// ```
     pub fn iter(&self) -> KeySetIndexIter<K> {
         KeySetIndexIter {
-            base_iter: self.base.iter(&()),
+            base_iter: self.base.iter(&(), None as Option<&[u8]>),
```
popzxc: As for me, syntax became less readable.

slowli:
On a semi-related note: I wonder if trying to extract a common interface of indexes would make sense. Something like

```rust
pub trait IndexTrait<K, V> {
     fn len(&self) -> usize;
     fn get(&self, key: &K) -> Option<V>;
     fn contains(&self, key: &K) -> bool;
     fn iter<'a>(&'a self, start_key: Option<&K>) -> Box<dyn Iterator<Item = (K, V)> + 'a>;
     // impl Iterator<...> would be better, but it's not supported for now
}
```

It could be useful with more declarative approach to indexes (like one I'm attempting to implement [in the ZeroMQ PoC](https://github.com/slowli/exonum/tree/zmq)):

```rust
impl IndexInfo {
    type Key: StorageKey; // not sure about bounds
    type Value: StorageValue;

    fn from_base<'a>(base: BaseIndex<&'a dyn Snapshot>)
        -> Box<dyn IndexTrait<Self::Key, Self::Value> + 'a>;
}

impl IndexInfo for Map<K, V> { /* ... */ }
```

Then, in addition to typed index getters in [`IndexAccess`](https://github.com/slowli/exonum/blob/75e3337c9b5e6c2ce649da66d37ad2749e3b5128/exonum/src/storage/meta/mod.rs#L518) and [`IndexAccessMut`](https://github.com/slowli/exonum/blob/75e3337c9b5e6c2ce649da66d37ad2749e3b5128/exonum/src/storage/meta/mod.rs#L552), it could be possible to provide something like

```rust
pub trait IndexAccess {
    fn boxed<'a, I>(&'a self, index: &I)
        -> Result<Box<dyn IndexTrait<K, V> + 'a>, Self::Error>
    where
        I: ToIndex,
        K == <I::Info as IndexInfo>::Key, // pseudocode, should be inlined
        V == <I::Info as IndexInfo>::Value,
}
```

I'm not completely sure whether such construction could make sense; e.g., it uses "slow" dynamic dispatch due to Rust's current restrictions, such as absence of parametric associated types. Just wanting to share my thoughts.

exonum/src/storage/key_set_index.rs
DarkEld3r: Is `as Option<&[u8]>` part really needed here?..

aleksuss: Unfortunately, yes

DarkEld3r:
@slowli As for me, extracting common interface is not very useful if it isn't required for something else, because different indexes have different logical usage. Although, it can make sense for some other goals.

PR #775, opened 2018-07-04
Make 'events' mod public but doc(hidden)

ozkriff:
Low-level access to events code is required for benchmarking (part of [ECR-1585]).

As `events` module is still effectively private, I'm disabling `missing_debug_implementations` and `missing_docs` lints.

### DIFF
--- a/exonum/src/events/mod.rs
+++ b/exonum/src/events/mod.rs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![allow(missing_debug_implementations, missing_docs)]
+
 pub use self::internal::InternalPart;
 pub use self::network::{NetworkConfiguration, NetworkEvent, NetworkPart, NetworkRequest};
 
--- a/exonum/src/events/noise/sodium_resolver.rs
+++ b/exonum/src/events/noise/sodium_resolver.rs
@@ -24,6 +24,12 @@ impl SodiumResolver {
     }
 }
 
+impl Default for SodiumResolver {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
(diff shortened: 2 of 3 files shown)


exonum/src/events/mod.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![allow(missing_debug_implementations, missing_docs)]
```
DarkEld3r: Is `missing_docs` really needed here? The module is "doc hidden" anyway. :thinking: 

Plus I'm thinking about allowing "missing debug implementations": why not just add `derive(Debug)` everywhere?

ozkriff: Yep, `missing_docs` barks even in `doc(hidden)` module.

About `derive(Debug)` - I tried, but there're some structs from `snow` that don't implement `Debug`.

popzxc: I guess you can create issue or even PR in snow. Maintainer reacts pretty fast.

ozkriff:
Oops. Fixing clippy warnings.

exonum/src/events/mod.rs
ozkriff: @popzxc will do :) But `allow(missing_debug_implementations)` will be removed in another PR

ozkriff:
Fixed. Ready for review again.

exonum/src/events/mod.rs
DarkEld3r: @ozkriff You are going to create an issue for removing `allow(missing_debug_implementations)`, right?

ozkriff: @DarkEld3r yep https://jira.bf.local/browse/ECR-1818

PR #776, opened 2018-07-04
Fix 'consensus_enabled' logic [ECR-1759]

ozkriff:
This flag must disable only consensus related events.

This PR re-enables `NetworkEvent`s and `ExternalMessage`s.

### DIFF
--- a/exonum/src/sandbox/consensus.rs
+++ b/exonum/src/sandbox/consensus.rs
@@ -203,7 +203,8 @@ fn test_disable_and_enable() {
 
     // A fail is expected here as the node is disabled.
     sandbox.assert_state(HEIGHT_TWO, ROUND_ONE);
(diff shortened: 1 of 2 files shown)


exonum/src/sandbox/consensus.rs
```diff
@@ -203,7 +203,8 @@ fn test_disable_and_enable() {
 
     // A fail is expected here as the node is disabled.
     sandbox.assert_state(HEIGHT_TWO, ROUND_ONE);
-    let result = try_add_one_height(&sandbox, &sandbox_state);
+    // TODO: use try_add_one_height (ECR-1817)
```
ozkriff: Looks like there's a bug in sandbox - it doesn't handle persistent mempool correctly. Created https://jira.bf.local/browse/ECR-1817 about that.

ozkriff:
Ready for review.

PR #777, opened 2018-07-04
Update config file on `ConnectList` update [ECR-1745]

popzxc:
In this PR a new entity was introduced: `ConfigManager` (maybe I'll rename it to `ConfigUpdater`).

Motivation: I wanted this feature to be detached from the core as much as possible, since direct work with file system in runtime isn't common thing for blockchain solutions.
Also I wanted to work with fs in the separate thread just to avoid any possible performance problems.

This PR isn't complete yet, and published just to discuss validity of proposed solution.

TODO list:

- [x] Create & integrate `ConfigManager`.
- [x] Add tests.
- [x] Replace `expect` with `error!(..)`.
- [x] Beautify code.
- [x] Update changelog.


andreyluchin:
Have we researched the possibility of using [tokio-fs](https://github.com/tokio-rs/tokio/tree/master/tokio-fs)? I think it is not a bad idea for file I/O operations to run asynchronously in a separate runtime and it shouldn't be too complicated.

popzxc:
I thought about it, but as for me it'll bring a new dependency without any notable profit.

pavel-mukhanov:
Some crazy idea: as far as we use `actix-web` maybe we can use `actix` instead of `mpsc::channel`. Or it's overkill in this case?

popzxc:
Yep, I thought about it too. But IMHO it'll be overkill indeed.
It's possible to make this thing more abstract and universal, but I don't think that too much flexibility is a good idea for something that may not be ever extended.
If we'll decide that we want to use similar concept for some other activities, I guess we'll be able to easily rewrite this code. But if we we will not, we may stuck with over-complicated construction that we've never asked for.

exonum/src/helpers/config.rs
```diff
@@ -66,3 +68,64 @@ fn do_save<T: Serialize>(value: &T, path: &Path) -> Result<(), Error> {
     file.write_all(value_toml.to_string().as_bytes())?;
     Ok(())
 }
+
+/// Structure that handles work with config file in runtime.
```
vitvakatu: `at runtime`?

exonum/src/helpers/config.rs
```diff
@@ -66,3 +68,64 @@ fn do_save<T: Serialize>(value: &T, path: &Path) -> Result<(), Error> {
...
+#[derive(Debug)]
+pub struct ConfigManager {
+    handle: thread::JoinHandle<()>,
+    tx: mpsc::Sender<ConfigRequest>,
+}
+
+/// Messages for ConfigReader.
```
vitvakatu: What is `ConfigReader`?

exonum/src/helpers/config.rs
```diff
@@ -66,3 +68,64 @@ fn do_save<T: Serialize>(value: &T, path: &Path) -> Result<(), Error> {
...
+#[derive(Debug)]
+pub enum ConfigRequest {
+    /// Request for connect list update in config file.
+    UpdateConnectList(ConnectListConfig),
+}
+
+impl ConfigManager {
+    /// Creates a new `ConfigManager` instance for the given path.
+    pub fn new(path: String) -> Self {
```
vitvakatu: It's better to use `P: AsRef<Path>` instead of `String` and `&str` for paths

exonum/src/helpers/config.rs
```diff
@@ -66,3 +68,64 @@ fn do_save<T: Serialize>(value: &T, path: &Path) -> Result<(), Error> {
...
+        });
+
+        ConfigManager { handle, tx }
+    }
+
+    /// Stores updated connect list in file system.
+    pub fn store_connect_list(&self, connect_list: ConnectListConfig) {
+        self.tx
+            .send(ConfigRequest::UpdateConnectList(connect_list))
+            .expect("Can't write");
```
vitvakatu: I know the possibility of error here, but let's use more descriptive message like `Can't  send message to ConfigManager thread`

exonum/src/helpers/fabric/details.rs
```diff
@@ -57,9 +57,12 @@ impl Run {
...
 
-    fn node_config(ctx: &Context) -> NodeConfig {
-        let path = ctx.arg::<String>(NODE_CONFIG_PATH)
-            .unwrap_or_else(|_| panic!("{} not found.", NODE_CONFIG_PATH));
+    fn node_config_path(ctx: &Context) -> String {
+        ctx.arg::<String>(NODE_CONFIG_PATH)
+            .unwrap_or_else(|_| panic!("{} not found.", NODE_CONFIG_PATH))
+    }
+
+    fn node_config(path: String) -> NodeConfig {
```
vitvakatu: `P: AsRef<Path>` is better

popzxc: We discussed this topic with @vitvakatu and decided to left `String` here, so interface of `Node::new` won't become uglier.

PR #781, opened 2018-07-06
Extend (slightly) entry index interface [ECR-1832]

DarkEld3r:


aleksuss:
What about to add some unittests ?

exonum/src/storage/entry.rs
```diff
@@ -172,4 +172,56 @@ where
...
+    /// index.set(10);
+    /// assert_eq!(Some(10), index.get());
+    ///
+    /// let value = index.swap(20);
+    /// assert_eq!(Some(10), value);
+    /// assert_eq!(Some(20), index.get());
+    /// ```
+    pub fn swap(&mut self, value: V) -> Option<V> {
+        let previous = self.get();
+        self.set(val);
```
aleksuss: `value` ?

DarkEld3r:
Fixed. :upside_down_face:

DarkEld3r:
I have created a separate issue for tests: https://jira.bf.local/browse/ECR-1840

PR #783, opened 2018-07-09
ECR-1774 Site and docs translations

boguslavsky:
Temporary reverted explorer API changes inside demos to not broke demos until release.


boguslavsky:
correct PR is here https://github.com/exonum/exonum/pull/784

PR #785, opened 2018-07-09
ECR-1835 Block can not be verified error

boguslavsky:
Temporary reverted explorer API changes inside demos to not broke demos until release.


vitvakatu:
I think we need to create a task to revert this PR later :upside_down_face:

vitvakatu:
:thinking: Does any other demo affected?

boguslavsky:
> 🤔 Does any other demo affected?
@vitvakatu nope, just timestamping and cryptocurrency advanced

PR #786, opened 2018-07-09
ECR-1782 Identify peer on responder's side

pavel-mukhanov:
Check that incoming connection corresponds to peer with a known public key.


exonum/src/events/network.rs
```diff
@@ -369,41 +377,51 @@ impl Listener {
...
-
-                    stream.for_each(move |raw| {
-                        let event = NetworkEvent::MessageReceived(addr, raw);
-                        network_tx.clone().send(event).map_err(into_other).map(drop)
-                    })
-                })
-                .map(|_| {
-                    // Ensure that holder lives until the stream ends.
-                    let _holder = holder;
+            let connection_handler = handshake
```
popzxc: All this construction with with inclusive `and_then`, logic, object creation looks very gargantuan. I find it really hard to support in the future. Is it possible to decompose this thing?

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
```diff
@@ -96,10 +96,19 @@ impl NoiseHandshake {
...
-    ) -> Result<Framed<S, MessagesCodec>, io::Error> {
+    ) -> Result<(Framed<S, MessagesCodec>, x25519::PublicKey), io::Error> {
         let noise = self.noise.into_transport_mode()?;
+        let remote_static_key = {
+            let rs = noise.session.get_remote_static().unwrap_or_else(|| {
+                // Panic because with selected handshake pattern we must have
+                // `remote_static_key` on final step of handshake.
+                panic!("Remote static key is not present!")
+            });
+            x25519::PublicKey::from_slice(rs).unwrap()
```
popzxc: Is `unwrap` in this line guaranteed to succeed?

exonum/src/node/basic.rs
```diff
@@ -71,6 +80,7 @@ impl NodeHandler {
     /// Handles the `Connect` message and connects to a peer as result.
     pub fn handle_connect(&mut self, message: Connect) {
         // TODO Add spam protection. (ECR-170)
+        // TODO: drop connection if checks has failed. (ECR-1837)
```
popzxc: `checks have` / `check has`.

exonum/src/events/network.rs
```diff
@@ -369,41 +377,51 @@ impl Listener {
...
+                                .flatten_stream();
+
+                            stream.for_each(move |raw| {
+                                let event = NetworkEvent::MessageReceived(addr, raw);
+                                network_tx.clone().send(event).map_err(into_other).map(drop)
+                            })
+                        })
+                        .map(|_| {
+                            // Ensure that holder lives until the stream ends.
+                            let _holder = holder;
```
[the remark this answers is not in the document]
pavel-mukhanov: Yes, but I'm not the author of this.

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
pavel-mukhanov: Yes, I removed it in separate PR.

exonum/src/events/noise/mod.rs
```diff
@@ -42,7 +43,8 @@ pub const MAX_MESSAGE_LENGTH: usize = 65_535;
 pub const TAG_LENGTH: usize = 16;
 pub const HEADER_LENGTH: usize = 4;
 
-type HandshakeResult<S> = Box<dyn Future<Item = Framed<S, MessagesCodec>, Error = io::Error>>;
+type HandshakeResult<S> =
+    Box<dyn Future<Item = (Framed<S, MessagesCodec>, x25519::PublicKey), Error = io::Error>>;
```
slowli: I don't like this change; it leaks implementation details about the handshake (e.g., this makes it impossible to create a no-op implementation, or an implementation based on other ciphersuites). OTOH, I'm not sure about a better implementation. Could it make sense to inject the authorization logic directly into `Handshake`, i.e., make it return an error if the connecting peer is ineligible?

exonum/src/events/network.rs
```diff
@@ -369,47 +380,78 @@ impl Listener {
...
+        let event = NetworkEvent::PeerConnected(info, connect);
+
+        let stream = network_tx
+            .clone()
+            .send(event)
+            .map_err(into_other)
+            .and_then(move |_| Ok(stream))
+            .flatten_stream();
+
+        stream.for_each(move |raw| {
```
slowli: Sorry for stupid question / off-topic: What happens if the `Connect` message is invalid? Are messages from the stream still processed with this `for_each`?

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
```diff
@@ -97,10 +97,20 @@ impl NoiseHandshake {
     pub fn finalize<S: AsyncRead + AsyncWrite + 'static>(
         self,
         stream: S,
-    ) -> Result<Framed<S, MessagesCodec>, io::Error> {
+    ) -> Result<(Framed<S, MessagesCodec>, x25519::PublicKey), io::Error> {
         let noise = self.noise.into_transport_mode()?;
+        let remote_static_key = {
+            let rs = noise.session.get_remote_static().unwrap_or_else(|| {
```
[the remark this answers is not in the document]
popzxc: New clippy requires `unwrap_or_else`, since it's lazy.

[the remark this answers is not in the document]
pavel-mukhanov: I prefer `expect` either. Clippy forbids it on `warn` level: https://rust-lang-nursery.github.io/rust-clippy/master/index.html#expect_fun_call. So I think we could use `expect`.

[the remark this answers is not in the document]
popzxc: We discussed this earlier [here](https://github.com/exonum/exonum/pull/752#pullrequestreview-131493028) (you have to show the thread to see the discussion).
I guess we should follow the accepted convention.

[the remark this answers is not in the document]
slowli: There is no function call in `expect` in this case though, it directly uses a statically allocated string. Does this still trigger clippy? I'd say this is a bug in clippy then, since the triggering code does not match the lint description.

exonum/src/events/network.rs
```diff
@@ -369,47 +380,78 @@ impl Listener {
...
-                    let _holder = holder;
+            let connection_handler = handshake
+                .listen(sock)
+                .and_then(move |(sock, key)| {
+                    let (_, stream) = sock.split();
+                    stream
+                        .into_future()
+                        .and_then(Ok)
+                        .map_err(|e| e.0)
+                        .and_then(move |(raw, stream)| Self::parse_connect_msg(raw, stream))
```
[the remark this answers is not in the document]
pavel-mukhanov: `addr` is moved.

exonum/src/events/noise/mod.rs
pavel-mukhanov: I thought about it and doesn't look like a better solution, because therefore we need to maintain and update `ConnectList` on the network's side.

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
```diff
@@ -97,15 +97,27 @@ impl NoiseHandshake {
     pub fn finalize<S: AsyncRead + AsyncWrite + 'static>(
         self,
         stream: S,
-    ) -> Result<Framed<S, MessagesCodec>, io::Error> {
+    ) -> Result<(Framed<S, MessagesCodec>, x25519::PublicKey), io::Error> {
         let noise = self.noise.into_transport_mode()?;
+        let remote_static_key = {
```
slowli: Nit: The separate scope for `rs` seems to be a bit superfluous; something like this would seemingly do fine:

```rust
let remote_static_key = noise
    .session
    .get_remote_static()
    .expect("Remote static key is not present!");
let remote_static_key = x25519::PublicKey::from_slice(remote_static_key)
    .expect("Remote static key is not valid x25519 key!");
```

(Admittedly, I haven't try to compile this code, so maybe I'm missing something.)

exonum/src/events/network.rs
```diff
@@ -369,47 +380,73 @@ impl Listener {
...
+                        .map_err(|e| e.0)
+                        .and_then(|(raw, stream)| (Self::parse_connect_msg(raw), Ok(stream)))
+                        .and_then(move |(connect, stream)| {
+                            trace!("Received handshake message={:?}", connect);
+
+                            Self::process_incoming_messages(
+                                stream,
+                                network_tx,
+                                connect,
+                                ConnectInfo {
```
slowli: Very minor nit: Maybe, it could make sense to rename local vars `addr` and `key` to `address` and `public_key`, respectively, in order to use the shorthand constructor here?

exonum/src/events/network.rs
```diff
@@ -369,47 +380,73 @@ impl Listener {
...
         Ok(Listener(to_box(server)))
     }
+
+    fn parse_connect_msg(raw: Option<RawMessage>) -> Result<Connect, io::Error> {
+        let raw = raw.ok_or_else(|| other_error("Incoming socket closed"))?;
+        let message = Any::from_raw(raw).map_err(into_other)?;
+        match message {
+            Any::Connect(connect) => Ok(connect),
+            other => Err(other_error(&format!(
+                "First message is not Connect, got={:?}",
```
slowli: Nit: I think expanding to "First message from a remote peer is not `Connect`" could be more helpful.

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
pavel-mukhanov: We cannot borrow `noise` in the same scope, because later we need to move it to the return value.

PR #790, opened 2018-07-10
Add new_in_family for Entry index

DarkEld3r:


exonum/src/storage/entry.rs
```diff
@@ -62,6 +64,38 @@ where
...
+    /// ```
+    /// use exonum::storage::{MemoryDB, Database, Entry};
+    ///
+    /// let db = MemoryDB::new();
+    /// let name = "name";
+    /// let index_id = vec![01];
+    /// let snapshot = db.snapshot();
+    /// let index: Entry<_, u8> = Entry::new_in_family(name, &index_id, &snapshot);
+    /// ```
+    pub fn new_in_family<S: AsRef<str>, I: StorageKey>(
```
dmitry-timofeev: But Y would you use families of entries, when you can use a map?

DarkEld3r: @dmitry-timofeev You are right, I need to rethink it.

PR #791, opened 2018-07-11
Improve Service#initialize documentation [ECR-1754]

SergeiMal:


exonum/src/blockchain/service.rs
```diff
@@ -169,9 +169,14 @@ pub trait Service: Send + Sync + 'static {
...
-    /// Initializes the information schema of the service
-    /// and generates an initial service configuration.
-    /// This method is called on genesis block creation.
+    /// This method is invoked for all deployed services during the blockchain initialization
+    /// on genesis block creation.
+    /// During the handling of the method the service able to make 2 things:
+    /// - store its own initial state to the storage (Fork)
+    /// - return an initial [global configuration][doc:global_cfg] of the service in the JSON
+    /// format, in case service has global configuration parameters. This configuration used
+    /// to create a genesis block.
```
vitvakatu: It's not clear, at what moment this global configuration is used in terms of service configuration process (TEMPLATE -> CONFIG -> FINALIZE -> RUN). Moreover, it's not clear what happens with parameters already specified by user? Will parameter be overwritten by this method?

exonum/src/blockchain/service.rs
```diff
@@ -169,9 +169,16 @@ pub trait Service: Send + Sync + 'static {
     /// [the `Service` example above](#examples).
     fn tx_from_raw(&self, raw: RawTransaction) -> Result<Box<dyn Transaction>, MessageError>;
 
-    /// Initializes the information schema of the service
-    /// and generates an initial service configuration.
-    /// This method is called on genesis block creation.
+    /// Invoked for all deployed services during the blockchain initialization
+    /// on genesis block creation at each time a node is started.
```
popzxc: "at" is redundant here.

exonum/src/blockchain/service.rs
```diff
@@ -169,9 +169,16 @@ pub trait Service: Send + Sync + 'static {
...
+    /// During the handling of the method the service able to perform the following activities:
```
popzxc: `service is able`

exonum/src/blockchain/service.rs
```diff
@@ -169,9 +169,16 @@ pub trait Service: Send + Sync + 'static {
...
+    /// - store its own initial state to the storage [`&mut Fork`]
+    /// - return an initial [global configuration][doc:global_cfg] of the service in the JSON
+    /// format, in case service has global configuration parameters. This configuration used
```
popzxc: `in case service` seems weird for me. Maybe just "if"?

popzxc: `This configuration is used`. This one was mentioned before, but for some reason wasn't fixed.

PR #793, opened 2018-07-11
Update changelog with Entry bugfix (new_in_family)

DarkEld3r:


DarkEld3r:
@ozkriff `#numbeer` format works in releases and in issues/pull requests. I suppose it is enough.

Issue #688, opened 2018-05-16
Log output should include milliseconds


DarkEld3r:
Currently time in logs [is formatted](https://github.com/exonum/exonum/blob/master/exonum/src/helpers/mod.rs#L107) using `DateTime::<Local>::from(time).to_rfc2822()`. Such output doesn't include milliseconds, which makes logs less useful.

#620 can be used as inspiration.

DarkEld3r:
Closed with #796.

PR #796, opened 2018-07-16
Include milliseconds in log timestamps (#688)

ilammy:
There is no readymade method for that so replace `to_rfc2822()` with an explicit format string which includes milliseconds.

In order to test these changes we add regexes (only for cfg(test)) because the timestamps are formatted according to local timezone which can differ on machines, so we can't really hardcode any expected values.

### DIFF
--- a/exonum/src/helpers/mod.rs
+++ b/exonum/src/helpers/mod.rs
@@ -22,7 +22,7 @@ pub mod user_agent;
 #[macro_use]
 pub mod metrics;
 
-use chrono::{DateTime, Local};
+use chrono::{DateTime, Utc};
 use colored::*;
 use env_logger::{Builder, Formatter};
 use log::{Level, Record, SetLoggerError};
@@ -110,7 +110,7 @@ fn has_colors() -> bool {
 }
 
 fn format_time(time: SystemTime) -> String {
-    DateTime::<Local>::from(time)
+    DateTime::<Utc>::from(time)
         .format(LOG_TIMESTAMP_FORMAT)
         .to_string()
 }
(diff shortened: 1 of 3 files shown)


exonum/src/helpers/mod.rs
```diff
@@ -151,3 +153,28 @@ fn format_log_record(buf: &mut Formatter, record: &Record) -> io::Result<()> {
...
+    use regex::Regex;
+
+    use std::time;
+
+    use super::*;
+
+    #[test]
+    fn time_formatting() {
+        // Thu,  1 Jan 1970 00:00:00.000 +0000
+        let r = Regex::new(
```
popzxc: Why compare against regex (and introduce a new dependency), when you can compare against expected stringp?

exonum/src/helpers/mod.rs
```diff
@@ -104,7 +104,9 @@ fn has_colors() -> bool {
 }
 
 fn format_time(time: SystemTime) -> String {
-    DateTime::<Local>::from(time).to_rfc2822()
+    DateTime::<Local>::from(time)
+        .format("%a, %e %b %Y %H:%M:%S%.3f %z")
```
popzxc: Maybe move this format string to a constant with comment describing what exactly it's doing?
With a brief look it looks confusing.

popzxc:
Regarding testing via regex. I don't really like it, since it's overkill and introduces a new dependency (and project is already builds pretty slow).

It's possible to use `Local` timezone only in production, and `Utc` for tests:
```
#[cfg(not(test))]
use chrono::Local as TimeOffset;
#[cfg(test)]
use chrono::Utc as TimeOffset;

...

fn format_time(time: SystemTime) -> String {
    DateTime::<TimeOffset>::from(time)
        .format("%a, %e %b %Y %H:%M:%S%.3f %z")
        .to_string()
}

...

#[test]
fn time_formatting() {
    assert_eq!(format_time(time::UNIX_EPOCH), "Thu,  1 Jan 1970 00:00:00.000 +0000");
}
```

though it looks a bit fishy. However, testing against expected value and not regex seems better choice to me.

@DarkEld3r  what do you think?

vitvakatu:
Btw, do we really need to use `Local` timezone? Why not `UTC` everywhere?

DarkEld3r:
I also prefer to use `UTC` everywhere.

PR #797, opened 2018-07-16
Consistent compact Display/Debug format for crypto types

ilammy:
It has been found strange that Display trait prints entire representations while Debug truncates them to 4 bytes. Furthermore, some Debug impls add ellipsis and some don't.

Update the macros that generate Debug impls so that they print the entire byte range. Add some new tests to ensure that Display impls are covered as well. Drop now-unused constant for display length limit.

### DIFF
--- a/exonum/src/crypto/crypto_lib/sodiumoxide/x25519.rs
+++ b/exonum/src/crypto/crypto_lib/sodiumoxide/x25519.rs
@@ -32,8 +32,6 @@ use crypto;
 pub const PUBLIC_KEY_LENGTH: usize = 32;
 /// Length of the secret Curve25519 key.
 pub const SECRET_KEY_LENGTH: usize = 32;
-/// The size to crop the string in debug messages.
-pub const BYTES_IN_DEBUG: usize = 4;
 
 /// Converts Ed25519 keys to Curve25519.
 ///
@@ -137,9 +135,7 @@ macro_rules! implement_x25519_type {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             write!(f, stringify!($name))?;
             write!(f, "(")?;
-            for i in &self.0[0..BYTES_IN_DEBUG] {
-                write!(f, "{:02X}", i)?
-            }
+            crypto::write_short_hex(f, &self.0[..])?;
             write!(f, ")")
         }
     }
--- a/exonum/src/crypto/macros.rs
+++ b/exonum/src/crypto/macros.rs
@@ -51,13 +51,6 @@ macro_rules! implement_public_crypto_wrapper {
         }
     }
 
-    impl FromStr for $name {
-        type Err = FromHexError;
-        fn from_str(s: &str) -> Result<Self, Self::Err> {
-            $name::from_hex(s)
-        }
-    }
-
     impl fmt::Debug for $name {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             write!(f, stringify!($name))?;
--- a/exonum/src/crypto/macros.rs
+++ b/exonum/src/crypto/macros.rs
@@ -62,16 +62,14 @@ macro_rules! implement_public_crypto_wrapper {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             write!(f, stringify!($name))?;
             write!(f, "(")?;
-            for i in &self[0..BYTES_IN_DEBUG] {
-                write!(f, "{:02X}", i)?
-            }
+            write_short_hex(f, &self[..])?;
             write!(f, ")")
         }
     }
 
     impl fmt::Display for $name {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-            f.write_str(&self.to_hex())
+            write_short_hex(f, &self[..])
         }
     }
     )
@@ -112,10 +110,8 @@ macro_rules! implement_private_crypto_wrapper {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             write!(f, stringify!($name))?;
             write!(f, "(")?;
-            for i in &self[0..BYTES_IN_DEBUG] {
-                write!(f, "{:02X}", i)?
-            }
-            write!(f, "...)")
+            write_short_hex(f, &self[..])?;
+            write!(f, ")")
         }
     }
 
--- a/exonum/src/crypto/mod.rs
+++ b/exonum/src/crypto/mod.rs
@@ -56,6 +56,16 @@ pub(crate) mod crypto_lib;
 /// The size to crop the string in debug messages.
 const BYTES_IN_DEBUG: usize = 4;
 
+fn write_short_hex(f: &mut fmt::Formatter, slice: &[u8]) -> fmt::Result {
+    for byte in slice.iter().take(BYTES_IN_DEBUG) {
+        write!(f, "{:02x}", byte)?;
+    }
+    if slice.len() > BYTES_IN_DEBUG {
+        write!(f, "...")?;
+    }
+    Ok(())
+}
+
 /// Signs a slice of bytes using the signer's secret key and returns the
 /// resulting `Signature`.
 ///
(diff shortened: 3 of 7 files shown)


exonum/src/crypto/crypto_lib/sodiumoxide/x25519.rs
```diff
@@ -137,7 +135,7 @@ macro_rules! implement_x25519_type {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             write!(f, stringify!($name))?;
             write!(f, "(")?;
-            for i in &self.0[0..BYTES_IN_DEBUG] {
+            for i in &self.0[..] {
```
vitvakatu: `[..]` is redundant here

exonum/src/crypto/macros.rs
```diff
@@ -62,7 +62,7 @@ macro_rules! implement_public_crypto_wrapper {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             write!(f, stringify!($name))?;
             write!(f, "(")?;
-            for i in &self[0..BYTES_IN_DEBUG] {
+            for i in &self[..] {
```
vitvakatu: Same here

exonum/src/crypto/macros.rs
```diff
@@ -112,10 +112,10 @@ macro_rules! implement_private_crypto_wrapper {
...
```
vitvakatu: And here

exonum/src/crypto/mod.rs
```diff
@@ -687,22 +684,39 @@ mod tests {
...
-        assert_eq!(format!("{:?}", &sk), "SecretKey(FFFFFFFF...)");
+        assert_eq!(format!("{:?}", &sk), "SecretKey(FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF)");
+    }
+
+    // Note that only public values have Display impl.
+    #[test]
+    fn display_format() {
+        // Check zero padding
+        let hash = Hash::new([1; HASH_SIZE]);
+        assert_eq!(format!("{}", &hash), "0101010101010101010101010101010101010101010101010101010101010101");
```
vitvakatu: Mmm... It seems invalid. `Display` should crop output to some amount of symbols then.

ilammy: [The docs] seem to suggest that `Display` should accurately represent the value as a string (i.e., so that one can write `from_str()` for that format). From this viewpoint we kinda should output the entire value there. But it can also be said that this is a use-case for `Serialize` trait, not Display.

[The docs]: https://doc.rust-lang.org/beta/std/fmt/index.html#fmtdisplay-vs-fmtdebug

ilammy:
@popzxc that makes sense. Maybe we should update the Display trait impls as well so to that all of them consistently output something like `01020304...` or `010203...FF`? If we ever need an accurate hex representation then we can introduce `to_string()` method which will output the full data.

exonum/src/crypto/mod.rs
ozkriff: Huh? `Display` trait is for human convenience only, it has no reasons to be an accurate representation.

vitvakatu:
`ToString::to_string` is auto-derived from `Display` trait, so it's bad idea to implement specific `to_string`.

ilammy:
Hm... I have just realized that [current types implement `FromStr`](https://github.com/ilammy/exonum/blob/501f26d6fe317c6f41d08f4a0260c036bcd7ed9a/exonum/src/crypto/macros.rs#L54) and it expects the string to be full hexadecimal representation. We also have Serde's FromHex, ToHex. I guess we'll have to remove `FromStr` if `Display` no longer outputs full representation.

exonum/src/crypto/crypto_lib/sodiumoxide/x25519.rs
ilammy: I remembered why I did this in this way: using `self.0` and `self` did not auto-convert to iterable slices for some types declared by these macros (while some others were fine).

I am going to write a single utility function for hex-formatting slices with ellipses and will do something about this redundancy.

ilammy: Okay... Here's the catch with [ed25519's types that we're using][1]. All types implement `Index<Output=[u8]>`, but only 'public' types (PublicKey and Signature) implement `AsRef<[u8]>`. So at least for 'private' types we'll have to use `&self[..]` or something to first convert them into a slice via range-indexing. Taking a reference works only for 'public' types.

[1]: https://sodiumoxide.github.io/sodiumoxide/sodiumoxide/crypto/sign/ed25519/index.html

ilammy:
> I guess we'll have to remove `FromStr` if `Display` no longer outputs full representation.

Now I'm not sure about this. [`FromStr`][2] is used all over our tests. It is also used by [`parse()`][1] method of strings. However, it does not seem to be tightly related to `Display` and formatting after all. So we _may_ keep it.

But I believe this is still a breaking change as

```rust
let h1 = Hash::from_str(
    "326c1da1a00b5b4c85929dac57f3c99ceea82ed2941173d879c57b8f21ae8c78"
);
let h2 = Hash::from_str(&h1.to_string());
assert_eq!(h1, h2);
```

will no longer work if this PR is merged.

[1]: https://doc.rust-lang.org/std/primitive.str.html#method.parse
[2]: https://doc.rust-lang.org/std/str/trait.FromStr.html

ilammy:
Okay, here comes v2 of this changeset.

* Display and Debug now use the same format for bytes:
   * lowercase (previously Display used lowercase and Debug used uppercase)
   * no more than 4 bytes, with ellipsis appended if truncated
* FromStr is still here (the test suite passes for me)
* We still use a bit weird `&self[..]` expression to convert into byte slice
* Seemingly unused BYTES_IN_DEBUG export has been dropped

vitvakatu:
Please do no rebase commits after gaining some review feedback. Use merge if you need to sync your PR and only append new commits, even if they doesn't look well organized

vitvakatu:
Let's keep it as it is now

ilammy:
Hm... My bad, I did not run the tests locally after the changes, but Travis CI has my back. That one test failure shows that `ToString` for `Hash` _is_ used somewhere and seems to be expected to return full representation. I'll investigate into it.

vldm:
1. PullRequest has misleading name, after merging commit would take pullrequest name.
2. Have you removed FromStr implementation for this types? Without this change this would a bit confusing that `FromStr::from_str(&hash.to_string())` produce error.

popzxc:
Please update `CHANGELOG.md`.

exonum/src/crypto/mod.rs
```diff
@@ -687,22 +697,39 @@ mod tests {
...
+        assert_eq!(format!("{:?}", &hash), "Hash(80808080...)");
         let sk = SecretKey::new([255; SECRET_KEY_LENGTH]);
-        assert_eq!(format!("{:?}", &sk), "SecretKey(FFFFFFFF...)");
+        assert_eq!(format!("{:?}", &sk), "SecretKey(ffffffff...)");
+    }
+
+    // Note that only public values have Display impl.
+    #[test]
+    fn display_format() {
+        // Check zero padding
```
DarkEld3r: Please add a dot at the end of the line.

exonum/src/crypto/mod.rs
```diff
@@ -687,22 +697,39 @@ mod tests {
...
+        let hash = Hash::new([1; HASH_SIZE]);
+        assert_eq!(format!("{}", &hash), "01010101...");
+
+        let pk = PublicKey::new([15; PUBLIC_KEY_LENGTH]);
+        assert_eq!(format!("{}", &pk), "0f0f0f0f...");
+        let signature = Signature::new([10; SIGNATURE_LENGTH]);
+        assert_eq!(format!("{}", &signature), "0a0a0a0a...");
+
+        // Check no padding
```
DarkEld3r: Same here.

PR #799, opened 2018-07-17
proofread changelog for 0.9 release

Maria-Nosyk:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Breaking Changes
 
-#### Exonum
+#### exonum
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,30 +5,30 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ## [Unreleased]
 
-### Breaking changes
+### Breaking Changes
 
-#### exonum
+#### Exonum
```
vitvakatu: It was intentional afaik, but I don't think it's worse now.

DarkEld3r: It was indeed intentional. 🙃 

Subsections are named the same as our "components". Otherwise we will have to rename `exonum-testkit`, `exonum-configuration`, etc. Personally I prefer consistent naming.

PR #801, opened 2018-07-18
WIP: Provide another view for mutating database without forks and patches

mvlabat:
Please, do not comment about method naming inconsistency and missing documentation, there's still work to be done. Tests will come as well.

All the other suggestions are welcome.


DarkEld3r:
Outdated.

PR #802, opened 2018-07-18
Optimize Merkle root verification for blocks [ECR-1622]

ilammy:
PR #664 refactored block verification but it uses ProofListIndex to compute Merkle root hash. [As suggested during code review](https://github.com/exonum/exonum/pull/664#discussion_r193049294), we can replace ProofListIndex and MemoryDB with direct computation of the root hash.

The tests cross-verify this implementation with ProofListIndex to ensure that they are consistent. However, I'm not sure whether we need this cross-verification (only checking hash values may be enough). Furthermore, maybe `merkle_root()` function can be moved to some more appropriate module. **Please advise on these two matters.**


exonum/src/helpers/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Merkle tree utilities.
+
+use crypto::{self, Hash, HashStream};
+
+/// Compute Merkle root hash for a given list of hashes.
```
vitvakatu: Computes

exonum/src/helpers/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// If `hashes` are empty then default (zero) value is returned.
```
vitvakatu: I think it's better to clarify that "\`Hash::zero\`" values is returned

exonum/src/helpers/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+pub fn root_hash<'a, I: IntoIterator<Item=&'a Hash>>(hashes: I) -> Hash {
+    let mut current_hashes: Vec<Hash> = hashes.into_iter().cloned().collect();
```
vitvakatu: Is `cloned` really needed here? Hashes implement `Copy` afaik

ilammy: It seems so. `Hash` _does_ implement `Copy`, but without `.cloned()` tacked into the pipeline `collect()` does not seem to be smart enough to perform the copy itself. Or maybe this is done on purpose by libstd devs so that copying is explicit.

```
error[E0277]: the trait bound `std::vec::Vec<crypto::Hash>: std::iter::FromIterator<&'a crypto::Hash>` is not satisfied
  --> exonum/src/helpers/merkle.rs:23:60
   |
23 |     let mut current_hashes: Vec<Hash> = hashes.into_iter().collect();
   |                                                            ^^^^^^^ a collection of type `std::vec::Vec<crypto::Hash>` cannot be built from an iterator over elements of type `&'a crypto::Hash`
   |
   = help: the trait `std::iter::FromIterator<&'a crypto::Hash>` is not implemented for `std::vec::Vec<crypto::Hash>`
```

We expect and receive an iterator over references to Hashes in `BlockResponse::verify_tx_hash()` as its `self.transactions()` method returns `&[Hash]`.

exonum/src/helpers/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+
+//! Merkle tree utilities.
+
+use crypto::{self, Hash, HashStream};
+
+/// Computes Merkle root hash for a given list of hashes.
+///
+/// If `hashes` are empty then `Hash::zero()` value is returned.
+pub fn root_hash<'a, I: IntoIterator<Item = &'a Hash>>(hashes: I) -> Hash {
+    let mut current_hashes: Vec<Hash> = hashes.into_iter().cloned().collect();
```
popzxc: Cloning here can be avoided, if the function signature will be changed:
```
pub fn root_hash(hashes: &[Hash]) -> Hash {
    if hashes.is_empty() {
        return Hash::zero();
    }
    let mut current_hashes = combine_hash_list(hashes);
    while current_hashes.len() > 1 {
        current_hashes = combine_hash_list(&current_hashes);
    }
    current_hashes[0]
}
```

What's the profit from taking `IntoIterator` as argument, if it's get collected in the first line? It could be collected as well on the callee side as well.
And with `&[Hash]` code looks easier as well.

exonum/src/helpers/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    fn hash_list(bytes: &[&[u8]]) -> Vec<Hash> {
+        bytes.iter().map(|chunk| crypto::hash(chunk)).collect()
+    }
+
+    #[test]
+    fn root_hash_single() {
+        assert_root_hash_eq(
+            &hash_list(&[b"1"]),
+            "6b86b273ff34fce19d6b804eff5a3f5747ada4eaa22f1d49c01e52ddb7875b4b",
```
popzxc: This module uses some "generic hash" and I think we should not test it against some hard-coded values. As for me, it's enough to test against `ProofListIndex`.

exonum/src/helpers/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+    use encoding::serialize::FromHex;
+    use storage::{Database, MemoryDB, ProofListIndex};
+
+    use super::*;
+
+    /// Cross-verify `root_hash()` with `ProofListIndex` against expected root hash value.
+    fn assert_root_hash_eq(hashes: &[Hash], expected: &str) {
+        let root_actual = root_hash(hashes);
+        let root_index = proof_list_index_root(hashes);
+        let root_expected = Hash::from_hex(expected).expect("hex hash");
```
popzxc: Nit: Please add more meaningful message to `expect` call.

popzxc:
And also please update the changelog.

exonum/src/helpers/merkle.rs
ilammy: I used `IntoIterator` as a way to generalize root hash computation, to not limit it to just slices. Though, I guess we can revert this change and keep it with slices until we really need it.

ilammy:
@popzxc, what should I mention in the changelog? Only addition of the new function or the performance optimization as well?

popzxc:
@ilammy I guess you should mention both introduction a new module+method and performance optimization. Both things can be added into "Internal Improvements" section.

exonum/src/helpers/merkle.rs
```diff
@@ -19,11 +19,11 @@ use crypto::{self, Hash, HashStream};
...
 ///
 /// If `hashes` are empty then `Hash::zero()` value is returned.
-pub fn root_hash<'a, I: IntoIterator<Item = &'a Hash>>(hashes: I) -> Hash {
-    let mut current_hashes: Vec<Hash> = hashes.into_iter().cloned().collect();
-    if current_hashes.is_empty() {
+pub fn root_hash(hashes: &[Hash]) -> Hash {
+    if hashes.is_empty() {
         return Hash::zero();
     }
+    let mut current_hashes = hashes.to_vec();
```
popzxc: Why do you need to convert hashes to vec? It's an additional space allocation that could be easily avoided.

exonum/src/helpers/merkle.rs
ilammy: As for in-place computation, I guess we can rewrite `combine_hash_list ` to be `fn(&mut Vec)` with in-place computations and zero additional Vec allocations. However, in this case we'd still need to clone the original slice. It is possible to (effectively) avoid copying by constructing the initial vector of the pre-combined half of the hashes, but I don't think it's worth the readability hit.

exonum/src/helpers/merkle.rs
ilammy: We need to have some storage to keep intermediate hashes, and its quite readable. Is there any obvious way to compute the root hash without any intermediate storage, stacks, etc. that I am not aware of? I'm sorry, I can't quite see how we can avoid intermediate storage completely.

popzxc: I'm not talking about every intermediate storage, just about initial one. If you'll write 
```
let mut current_hashes = combine_hash_list(hashes);
```
then you'll avoid the first allocation and the first `Vec` will contain only `N/2` hashes.

exonum/src/helpers/merkle.rs
```diff
@@ -0,0 +1,118 @@
...
+            if pair.len() == 2 {
+                hash_pair(&pair[0], &pair[1])
+            } else {
+                hash_one(&pair[0])
+            }
+        )
+        .collect()
+}
+
+fn hash_one(lhs: &Hash) -> Hash {
```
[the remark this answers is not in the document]
ilammy: Would it be okay to move `hash_{one,pair}` to `proof_list_index` module and export them as `pub(crate)` so that they are visible to `helpers::merkle`? Or it would be better to just move `root_hash()` from `helpers::merkle` to `proof_list_index`?

exonum/src/helpers/merkle.rs
ilammy: This won't be that straightforward with in-place implementation suggested by @slowli. It uses

```rust
fn combine_hash_list(hashes: &mut Vec<Hash>)
```

and avoids any additional `Vec<Hash>` allocations except for the initial one. (Currently we allocate a new Vec for each intermediate batch.)

I'm not sure that allocating half less storage is worth the readability hit as in that case we will need to effectively duplicate implementation of `combine_hash_list()`:

```rust
/// Initial version used for first computation and allocation.
fn combine_inital_hash_list(hashes: &[Hash]) -> Vec<Hash> {
    // current implementation
}

/// In-place version used for subsequent iterations.
fn combine_hash_list(hashes: &mut Vec<Hash>) {
    // in-place implementation which does effectively the same
}
```

Should we do it like that?

exonum/src/storage/proof_list_index/tests.rs
```diff
@@ -769,3 +769,47 @@ mod rocksdb_tests {
         super::same_merkle_root(db1, db2);
     }
 }
+
+mod root_hash_tests {
+    use crypto::{self, Hash};
+    use storage::{Database, MemoryDB, ProofListIndex};
+
+    /// Cross-verify `root_hash()` with `ProofListIndex` against expected root hash value.
+    fn assert_root_hash_correct(hashes: &[Hash]) {
```
dmitry-timofeev: Nit: I'd suggest moving the most interesting code here — tests — to the top, and utility methods — after their usages (i.e., after tests).

popzxc: Hm... I prefer the current situation: all the logic tests use at the top (so you'll know what's going on), and tests at the bottom (so you are prepared for them).

IMHO it's a bit of holy-war question :)

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -614,3 +618,33 @@ where
...
+            }
+            current_hashes[0]
+        }
+    }
+}
+
+fn combine_hash_list(hashes: &[Hash]) -> Vec<Hash> {
+    hashes
+        .chunks(2)
+        .map(|pair| {
```
slowli: Nit: you can use slice comprehension here:

```rust
match pair {
    [first, second] => hash_pair(first, second),
    [single] => hash_one(single),
    _ => unreachable!(),
}
```

DarkEld3r: I also think that this looks clearer than `if/else`.

PR #803, opened 2018-07-18
Exonum 0.9.0 release

DarkEld3r:


alekseysidorov:
Can you also update versions of pinned crates?

ozkriff:
> Can you also update versions of pinned crates?

Hmm, sounds like a risky thing to do right before the release to me.

PR #805, opened 2018-07-18
Fix missed external address in auditor configurations.

alekseysidorov:


vitvakatu:
:thinking: For some reason my comments are marked as resolved, but they are not.

ilammy:
@vldm

> This code below looks like we can't set listen_addr other than "broadcast" address. Which is probably bad security practices.

This is the address of the interface which will be used by node host to accept connections. Usually it will be set to the special value 'any' (all-zeros) to accept connections on any network interface. However, it may be set to a specific address if one wishes to accept connections originating only from a particular interface (e.g., use `127.0.0.1` or `::1` to allow connections only from localhost).

Usually non-router servers will have only one external interface, one loopback interface, and maybe some additional interfaces managed by Docker, virtual machines, etc. In this case using 'any' address is desirable and works fine. However, if the node server has multiple 'real' network interfaces (e.g., one for external Internet and one for company intranet) then it may be desirable to accept connections only from one of these networks.

I would not say this a security issue per se, but it would be nice to have the listening address configurable for some use cases.

PR #811, opened 2018-07-22
Drop connection when handshake has failed [ECR-1837]

pavel-mukhanov:
Move `ConnectList` and corresponding checks to the `network`.


slowli:
Just food for thought / my perspective: we have `Ed25519` consensus keys as identities, right? This, it could only make sense that the handshake procedure work with the identity of the peer on the level of the interface. Implementation details, such as mapping to x25519 keys, should ideally be incapsulated inside the handshake; it's not improbable to devise a handshake implemented differently (e.g., the old approach with unencrypted channels and signature-based connect messages; or the approach with TLS; or lightweight "certificates" of network keys), and in some cases such an implementation could be preferable. By cleanly separating the interface of the handshake from its implementation details, we can increase code reuse and leave space for future improvements.

viktorbezhenar:
@popzxc @DarkEld3r please approve if there is no other comments

PR #812, opened 2018-07-23
Restore cargo audit

DarkEld3r:


ozkriff:
Aaand it wants `Cargo.lock`. Hmm.

DarkEld3r:
@ozkriff I think that we don't need `cargo audit` if we are going to use [Dependabot](https://dependabot.com/rust.html):
> Dependabot checks your dependencies against the RustSec Advisory Database, and creates PRs immediately for any vulnerable top-level or subdependencies.

PR #826, opened 2018-07-24
Added possibility to use domain name as a peer's address

aleksuss:
Propose of this PR is to add possibility to use domain name as a peer's address


exonum/src/node/mod.rs
```diff
@@ -760,10 +762,28 @@ impl fmt::Debug for ApiSender {
...
+    D: Deserializer<'de>,
+{
+    let address_str: String = Deserialize::deserialize(value)?;
+    address_str
+        .to_socket_addrs()
+        .map_err(de::Error::custom)?
+        .next()
+        .ok_or_else(|| {
+            de::Error::custom(&format!(
+                "no one ip belongs to the hostname: {}",
```
popzxc: I'd suggest "Can't resolve IP for hostname:"

exonum/src/node/mod.rs
```diff
@@ -760,10 +762,28 @@ impl fmt::Debug for ApiSender {
     }
 }
 
+fn deserialize_socket_address<'de, D>(value: D) -> Result<SocketAddr, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let address_str: String = Deserialize::deserialize(value)?;
+    address_str
+        .to_socket_addrs()
```
popzxc: Question: Hmm.. Shouldn't we warn users that if they use hostnames, they should properly configure DNS and use only trusted (and not the same) servers? In a theoretical scenario, if some common DNS server will be hacked, and (1/3)+ nodes use this server, it may lead to a silent byzantine attack.

exonum/src/node/mod.rs
```diff
@@ -760,10 +762,28 @@ impl fmt::Debug for ApiSender {
...
+        .map_err(de::Error::custom)?
+        .next()
+        .ok_or_else(|| {
+            de::Error::custom(&format!(
+                "no one ip belongs to the hostname: {}",
+                address_str
+            ))
+        })
+}
```
ilammy: I'd say that hostname support should be more involved.

Usually when `getaddrinfo()` (deep within `to_socket_addrs()` implementation) returns more than one address then the application is expected to try them out until one of them connects. So in my opinion it would be better to do

```diff
 pub struct ConnectInfo {
      /// Peer address.
-     pub address: SocketAddr,
+     pub address: String,
```

and then handle hostnames in the connection code.

aleksuss: I thought about this and even have tried to do so. But such approach depends on `Connect` message wich plans to remove in future. Therefor for now we've decided to use simple approach. More intellegent and complex version will be done in another ticket.

exonum/src/node/mod.rs
aleksuss: This error message can be return from `to_socket_addrs` method. As I know.

exonum/src/node/mod.rs
aleksuss: What do you propose ? Just panic ? :-)

ilammy: I believe it is reasonable to warn the users about the possibility of DNS spoofing somewhere in the documentation on node configuration. But I do not think that we should do anything on nodes about that.

Proper configuration of DNS resolver (a system service) is responsibility of the administrator of node's machine. Using trusted DNS connections is in the same league as expecting that your ISP connects the node to the real Internet and does not reroute IP packets to compromised nodes.

popzxc: Yep, I was talking about some kind of disclaimer.

aleksuss: And what if user made a mistake in hostname ? How to inform him ?

CHANGELOG.md
```diff
@@ -11,8 +11,13 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
 - `NodePrivateConfig` fields have been renamed: `listen_addr` to `listen_address`
   and `external_addr` to `external_address`. (#809)
+  
+- `NodePublicConfig` `addr` field has been renamed to `address`. (#809)
 
-- `NodePublicConfig` `addr` field has been renamed to `address` (#809).
+### New features
+
+- Added possibility to use domain names instead of IP addresses as a peer's
+  address. `external_address` is not optional. (#826)
```
DarkEld3r: I think that non-optional `external_address` should be mentioned separately as breaking change (with "migration path").

examples/cryptocurrency/examples/demo.rs
```diff
@@ -46,7 +46,7 @@ fn node_config() -> NodeConfig {
         consensus_public_key,
         consensus_secret_key,
         genesis,
-        external_address: None,
+        external_address: "127.0.0.1:2000".parse().unwrap(),
```
DarkEld3r: I think `peer_address` can be reused here.

exonum/benches/criterion/transactions.rs
```diff
@@ -231,7 +231,7 @@ impl TransactionsBenchmarkRunner {
 
         NodeConfig {
             listen_address: peer_address,
-            external_address: Some(peer_address),
+            external_address: "127.0.0.1:2000".parse().unwrap(),
```
DarkEld3r: Same here.

exonum/src/helpers/fabric/maintenance.rs
```diff
@@ -114,9 +114,10 @@ impl Command for Maintenance {
             .arg::<String>(MAINTENANCE_ACTION_PATH)
             .unwrap_or_else(|_| panic!("{} not found.", MAINTENANCE_ACTION_PATH));
 
-        match action.as_ref() {
-            "clear-cache" => Self::clear_cache(&context),
-            _ => println!("Unsupported maintenance action: {}", action),
+        if action == "clear-cache" {
```
DarkEld3r: This change has been suggested by Clippy, right?
(I don't like it by the way. :upside_down_face: )

aleksuss: Exactly. Clippy is guilty

CHANGELOG.md
```diff
@@ -11,8 +11,15 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `NodePrivateConfig` fields have been renamed: `listen_addr` to `listen_address`
   and `external_addr` to `external_address`. (#809)
+  
+- `NodePublicConfig` `addr` field has been renamed to `address`. (#809)
 
-- `NodePublicConfig` `addr` field has been renamed to `address` (#809).
+- Config parameter `external_address` is not optional. (#826)
```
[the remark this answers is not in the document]
aleksuss: `listen_address` can be set to mask `0.0.0.0` (in case IP_V4) but `external_address` can't.

exonum/src/node/mod.rs
vldm: DNS spoofing, and other kind of traffic-redirection wouldn't lead to "byzantine attack", but can lead to DoS for exonum network.

My opinion, that user story should be next:
1) User sets peers address (it does not matter hostname or ip)
2) Current node trying to connect (trying to resolve hostname, checks pk validity) with some peer.
3) If current node fails to connect to some of peer (hostname/ip wrong or PK is differ), warn should apear in log.
4) If current node fails to connect to (majority_count - 1) peers, error should apear in log.

exonum/tests/testdata/config/config_domain.toml
```diff
@@ -0,0 +1,11 @@
...
+address = "127.0.0.1:6333"
+public_key = "16ef83ca4b231404daec6d07b24beb84d89c25944285d2e32a2dcf8f0f3eda72"
+
+[[peers]]
+address = "localhost:6333"
+public_key = "648e98a2405a40325d946bf8de6937795fe5c22ab095bca765a8b218e49ff5a3"
+
+[[peers]]
+address = "exonum.com:6333"
+public_key = "924625eb77b9ad21e76713e7ada715945fbf0a926698832e121484c797fcc58e"
```
vldm: Add new line at EOF

PR #830, opened 2018-07-30
Make pool count O(1) complexity. [ECR-977]

vldm:
Cache pool size in seperate index, this allows us to get pool size with O(1) complexity.


aleksuss:
This solution is not elegance as for me. What do you thinks about to create type which contains indexes `key_set_index` and `entry` inside itself ?

ilammy:
@aleksuss I'd guess that more elegant solution would imply some O(1) way to get entry count from `KeySetIndex` (or `BaseIndex` in general). Maybe, track the entry count in `BaseIndex::put()` and `BaseIndex::remove()`. But for that we'd need to teach `Fork` to tell us whether the changes actually add new entries (instead of updating existing ones), remove existing entries, etc. And this additional tracking will slow down all index operations a bit.

vldm:
@ilammy the one big problem is that it's not clear where `size` of idex should be stored. Probably `shadow tables` could help us with this.

slowli:
I principally agree with @ilammy in that number of entries in the index seems like a natural part of the general index interface. Some indexes have it already (e.g., `ListIndex` stores it in the "root" key). It seems that all indexes support updating this number in `O(1)` for single-element operations. I'm not sure whether it's possible / makes sense to implement the corresponding logic within `BaseIndex`; after all, some indexes contain auxiliary key entries (e.g., branch data for `Proof*Index`).

PR #831, opened 2018-07-30
Fix resolve ip address from localhost on windows os

aleksuss:


aleksuss:
@ilammy it lies in the fact that `localhost` domain name can be resolved into IPv4 or IPv6 as well. It depends on entry order for localhost in `hosts` file.

aleksuss:
@ilammy we don't need to update `ConnectInfo` in first case because in the config file we have ip `127.0.0.1` as address value and therefore we don't need to resolve it from domain name.

PR #838, opened 2018-07-31
Don't ignore files in sumbodules as dirs in cspell script [ECR-1851]

dvush:
Cspell ignores files in subdirectories.

Maybe all these globs should be replaced with `find -name "*.rs` with correct ignore?

### DIFF
--- a/cspell.sh
+++ b/cspell.sh
@@ -18,15 +18,5 @@
 
 set -e
 
-./node_modules/.bin/cspell sandbox/{src,examples,tests}/**/*.rs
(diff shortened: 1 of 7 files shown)


ozkriff:
Maybe it would be better to replace these manual paths with `find -name "*.rs" ......`?

dvush:
I think too. I just asked because maybe there are some .rs files that should be ignored and it is more convenient to have it like that.

ilammy:
I believe that it will be better to apply cspell to all *.rs files of the source tree, for the sake of future-compatibility. We can add comments like `// spell-checker:disable` to specific files if they cause any trouble.

PR #839, opened 2018-07-31
Replace consensus_enabled with network_enabled [ECR-1900]

VladYermakov:
- Now we have `is_network_enabled` field instead of `is_enabled`.
- Now if `is_network_enabled` set up to `false` node drops all connections, and reconnect to addresses in `connect_list` after `is_network_enabled` set up back to `true`.
- All actions now invoking without checking `is_network_enabled` because if network is disabled the node simply won't receive any requests.


vitvakatu:
I think we need tests to check described behaviour

exonum/src/node/events.rs
```diff
@@ -68,29 +68,31 @@ impl NodeHandler {
...
                     info!("The node is {} now", s);
-                    if self.is_enabled {
+                    if self.is_network_enabled {
+                        for address in self.state.connect_list().clone().peers.values() {
+                            self.connect(address);
+                        }
                         self.add_round_timeout();
+                    } else {
+                        for address in self.state.connections().clone().keys() {
+                            self.state.remove_peer_with_addr(address);
```
ivan-ochc: Seems that if we remove all peers here we will not be able to enable network again.

ivan-ochc: Seems that if we remove all peers here we will not be able to enable network again.

VladYermakov:
Need to remake using events. I will create new PR for this task.

PR #841, opened 2018-08-01
Fix bug with executing unknown propose. [ECR-2013]

alekseysidorov:


exonum/src/node/consensus.rs
```diff
@@ -270,7 +270,7 @@ impl NodeHandler {
         }
 
         // Commit propose
-        for (round, block_hash) in self.state.unknown_propose_with_precommits(&hash) {
+        for (round, block_hash) in self.state.take_unknown_propose_with_precommits(&hash) {
```
slowli: Off-topic, but I'm too curious: does the `commit` call in the cycle mean that several blocks can be committed within the cycle? How does it work? I think a comment explaining this stuff could be very helpful. If I understand correctly, this actually cannot happen because enough precommits can be collected for a single propose (as per the safety property of the consensus algorithm), which begs the question why use a vector in `state.unknown_proposes_with_precommits`.

exonum/src/node/state.rs
```diff
@@ -743,6 +743,7 @@ impl State {
         // TODO: Destruct/construct structure HeightState instead of call clear. (ECR-171)
         self.blocks.clear();
         self.proposes.clear();
+        self.unknown_proposes_with_precommits.clear();
```
slowli: Is it possible to write a test covering this change? (As far as I understand, it is this change that triggered the bug.)

exonum/src/node/consensus.rs
alekseysidorov: I suppose that only single propose can be committed as block on given height thus this cycle is redundant. 
But in my opinion this issue should be fixed during the consesus code refactoring.

exonum/src/node/state.rs
alekseysidorov: In theory this is possible but I suppose that this test case will be complex.

PR #842, opened 2018-08-01
Refactor sandbox constants [ECR-1947]

dvush:
Most of the constants are removed.


ozkriff:
Maybe "sandboxed" should be added to the main dictionary?

exonum/src/sandbox/consensus/invalid_message.rs
```diff
@@ -42,12 +45,12 @@ fn test_ignore_message_with_incorrect_signature() {
     let sandbox = timestamping_sandbox();
 
     let propose = Propose::new(
-        VALIDATOR_0,
-        HEIGHT_ZERO,
-        ROUND_ONE,
+        SANDBOXED_VALIDATOR_ID,
```
popzxc: As for me, `SANDBOXED_VALIDATOR_ID` looks weird near `ValidatorId(0)`. Why do we need to export it as a constant?

PR #843, opened 2018-08-01
 [ECR-1627] TODO cleanup pt.1

ponimas:
removal of the useless comment
using constant in the sandbox

### DIFF
--- a/exonum/src/sandbox/consensus/transactions.rs
+++ b/exonum/src/sandbox/consensus/transactions.rs
@@ -27,7 +27,7 @@ use messages::{
 use node::state::TRANSACTIONS_REQUEST_TIMEOUT;
 use sandbox::{
     config_updater::TxConfig, sandbox::timestamping_sandbox, sandbox_tests_helper::*,
-    timestamping::{TimestampTx, TimestampingTxGenerator},
+    timestamping::TimestampingTxGenerator,
 };
 
 /// idea of the test is to verify request transaction scenario: other node requests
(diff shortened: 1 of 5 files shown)


exonum/src/sandbox/sandbox_tests_helper.rs
```diff
@@ -48,6 +48,7 @@ pub const VALIDATOR_2: ValidatorId = ValidatorId(2);
 pub const VALIDATOR_3: ValidatorId = ValidatorId(3);
 pub const INCORRECT_VALIDATOR_ID: ValidatorId = ValidatorId(64_999);
 pub const PROPOSE_TIMEOUT: Milliseconds = 200;
+pub const DATA_SIZE: usize = 64;
```
dvush: Maybe this constant should be in timestamping file since it is used only for timestamping txs.

exonum/src/sandbox/consensus/transactions.rs
```diff
@@ -481,11 +481,11 @@ fn handle_tx_verify_signature() {
...
 
     // generate incorrect tx
-    let (public_key1, _) = gen_keypair();
-    let (_, secret_key2) = gen_keypair();
-
-    let data = vec![0; 64]; // TODO: Find the way how to get rid of hard-coded value. (ECR-1627)
-    let tx = TimestampTx::new(&public_key1, &data, &secret_key2);
+    let (public_key, _) = gen_keypair();
+    let (_, secret_key) = gen_keypair();
+    let tx = TimestampingTxGenerator::with_keypair(DATA_SIZE, (public_key, secret_key))
```
dvush: Maybe it would be nice to indicate that pub and secret keys are not compatible like it was before. Like pub_key, wrong_sec_key, what do you think?

PR #643, opened 2018-04-19
Propose timeout adjuster refactoring [ECR-717]

stanislav-tkach:
See changelog for details. :upside_down_face:

PR #844, opened 2018-08-01
Inject Proposes on transaction pressure [ECR-1441]

ilammy:
[Previously we have tweaked the timeout between **Propose** messages](https://github.com/exonum/exonum/pull/643) to adapt to current size of transaction queue. However, this timeout is fixed at the start of the round or at block creation time. If a node suddenly receives sufficiently large number of transactions during the timeout it will still wait for `max_propose_timeout()`.

Now we are going to check the transaction count every time we add a new transaction to the queue. If we have enough queued up then we schedule a quicker **Propose** timeout with `min_propose_timeout()`.

Previously scheduled timeouts will be handled later but they will be skipped due the checks in the beginning of `handle_propose_timeout()`.

Also note that we prevent expedited timeouts from being scheduled repeatedly. This is required as the transactions are not going anywhere out of the queue only because we schedule a timeout. We reset the flag when we actually handle the nearest timeout.

Threshold value now affects node behavior by causing **Propose** messages to be sent when new transactions are generated. Using threshold of zero in sandbox means that we'll send an extra Propose every time. Set it to non-zero value to avoid this unexpected behavior. Use `u32::MAX` rather than any hardcoded value like `10_000` to avoid surprises further.

### DIFF
--- a/exonum/src/node/consensus.rs
+++ b/exonum/src/node/consensus.rs
@@ -552,6 +552,8 @@ impl NodeHandler {
             .merge(fork.into_patch())
             .expect("Unable to save transaction to persistent pool.");
 
+        self.maybe_add_propose_timeout();
+
         let full_proposes = self.state.check_incomplete_proposes(hash);
         // Go to handle full propose if we get last transaction.
         for (hash, round) in full_proposes {
@@ -689,6 +691,7 @@ impl NodeHandler {
 
     /// Handles propose timeout. Node sends `Propose` and `Prevote` if it is a leader as result.
     pub fn handle_propose_timeout(&mut self, height: Height, round: Round) {
+        self.allow_expedited_propose = true;
         // TODO debug asserts (ECR-171)?
         if height != self.state.height() {
             // It is too late
--- a/exonum/src/node/consensus.rs
+++ b/exonum/src/node/consensus.rs
@@ -533,7 +533,8 @@ impl NodeHandler {
         }
     }
 
-    /// Checks if the transaction is new and adds it to the pool.
+    /// Checks if the transaction is new and adds it to the pool. This may trigger an expedited
+    /// `Propose` timeout on this node if transaction count in the pool goes over the threshold.
     fn handle_tx_inner(&mut self, msg: RawTransaction) -> Result<(), String> {
         let hash = msg.hash();
 
--- a/exonum/src/node/mod.rs
+++ b/exonum/src/node/mod.rs
@@ -126,6 +126,8 @@ pub struct NodeHandler {
     node_role: NodeRole,
     /// Configuration file manager.
     config_manager: Option<ConfigManager>,
+    /// Can we speed up Propose with transaction pressure?
+    allow_expedited_propose: bool,
 }
 
 /// Service configuration.
@@ -448,6 +450,7 @@ impl NodeHandler {
             is_enabled,
             node_role,
             config_manager,
+            allow_expedited_propose: true,
         }
     }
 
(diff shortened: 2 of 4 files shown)


ilammy:
I have tested this with cryptocurrency-advanced service using JS client to make sure that proposal injection is in fact triggered when many transactions are submitted during a short period of time. However, with default configuration I can't pump enough transactions there using `sendQueue` while regular concurrent `send` sometimes leads to node failures due to exhaustion of open file descriptors for connections. Though, it triggers fine if I set propose threshold to something like 10 and timeouts to 1–10 seconds.

Also, any pointer to how we can test this behavior automatically? I had to fixup sandbox configuration so that the new behavior does not break existing tests, but I guess it would be nice to add new tests to verify the new behavior. I'm going to look into it for some time.

exonum/src/node/mod.rs
```diff
@@ -640,6 +640,20 @@ impl NodeHandler {
         self.add_timeout(timeout, time);
     }
 
+    fn maybe_add_propose_timeout(&mut self) {
+        if self.need_faster_propose() && self.allow_expedited_propose {
+            info!("EXPEDITED PROPOSE TIMEOUT");
```
popzxc: I don't really like caps.

exonum/src/node/mod.rs
```diff
@@ -640,6 +640,20 @@ impl NodeHandler {
         self.add_timeout(timeout, time);
     }
 
+    fn maybe_add_propose_timeout(&mut self) {
+        if self.need_faster_propose() && self.allow_expedited_propose {
```
popzxc: I'd suggest to check for `self.allow_expedited_propose` at first, since rust does implement short circuit evaluation.

exonum/src/node/consensus.rs
```diff
@@ -689,6 +691,7 @@ impl NodeHandler {
 
     /// Handles propose timeout. Node sends `Propose` and `Prevote` if it is a leader as result.
```
popzxc: Shouldn't we mention that this function allows expedited propose now?

exonum/src/node/mod.rs
```diff
@@ -448,6 +450,7 @@ impl NodeHandler {
             is_enabled,
             node_role,
             config_manager,
+            allow_expedited_propose: true,
```
popzxc: Should it be enabled at the very beginning?

pavel-mukhanov: Why we never reset `allow_expedited_propose`?

popzxc: It is reseted in `maybe_add_propose_timeout`.

exonum/src/node/consensus.rs
ilammy: Mmm... I'm not sure this will be interesting for its users. Resetting a flag is just an implementation detail.

Maybe we should add comments to `handle_tx*` functions which tell that as a side effect of handling transactions we now may schedule a Propose timeout. But I'm not sure whether that is interesting enough as well.

exonum/src/node/mod.rs
ilammy: This flag allows us to schedule a Propose timeout with minimal delay once when we see that we have enough transactions bunched up in the pool. It is set to `false` when we actually schedule this timeout (in `maybe_add_propose_timeout`) to prevent additional 'expedited' timeouts from being scheduled by that function until we handle this one. The flag is reset in `handle_propose_timeout` when we handle the propose timeout and are ready for the next one. It is initialized to `true` by default so that we can schedule a faster timeout from the get go.

exonum/src/node/consensus.rs
ilammy: Should we mention this in changelog or this is minor enough?

popzxc: I'd prefer to mention this both in `changelog` and documentation. However, it's just my opinion. If you doubt, you can discuss it with the rest of the team :)

viktorbezhenar: @DarkEld3r wdyt?

viktorbezhenar:
@DarkEld3r merging?

exonum/src/node/consensus.rs
DarkEld3r: This comments are mostly for us, consensus (`node` module) should ideally be internal. This way the comments should be updated when it makes sense (I'm against too verbose comments).

PR #845, opened 2018-08-01
Update failure crate version to 0.1.2

aleksuss:

PR #846, opened 2018-08-02
Release 0.9.1 version(s)

DarkEld3r:
`0.9.1` hotfix including #841 and #845.


DarkEld3r:
@vitvakatu It is OK for the changelog to be broken, because this pull request will not actually be merged into the master.

It would probably be better to create a branch first and make a pull request targeting that branch. I will do this for the following releases. 🙃

aleksuss:
@vitvakatu this PR isn't going to be merged

PR #847, opened 2018-08-02
ECR-1627 TODO cleanup pt.2

ponimas:
Removed some outdated TODOS.
One of the sandbox tests updated.


exonum/src/sandbox/consensus/basic.rs
```diff
@@ -23,14 +23,16 @@ use std::collections::BTreeMap;
 use blockchain::{Blockchain, Schema};
 use crypto::{gen_keypair_from_seed, CryptoHash, Hash, Seed, HASH_SIZE, SEED_LENGTH};
 use helpers::{Height, Round};
-use messages::{Message, Precommit, Prevote, Propose, RawMessage, CONSENSUS};
+use messages::{Message, Precommit, Prevote, Propose, RawMessage, Status, CONSENSUS};
+
```
vitvakatu: Why empty line?

exonum/src/sandbox/consensus/basic.rs
```diff
@@ -23,14 +23,16 @@ use std::collections::BTreeMap;
...
 use sandbox::{
     sandbox::timestamping_sandbox, sandbox_tests_helper::*,
     timestamping::{TimestampingTxGenerator, TIMESTAMPING_SERVICE},
 };
 
 /// idea of the test is to verify that at certain periodic rounds we (`validator_0`) become a leader
 /// assumption: in some loops current node becomes a leader
+
```
vitvakatu: Why empty line?

exonum/src/sandbox/sandbox.rs
```diff
@@ -268,6 +267,26 @@ impl Sandbox {
...
+            let peers_request =
+                PeersRequest::from_raw(msg).expect("Incorrect message. PeersRequest was expected");
+
+            let id = self.addresses.iter().position(|&a| a == addr);
+            if let Some(id) = id {
+                assert_eq!(&self.p(ValidatorId(id as u16)), peers_request.to());
+            } else {
+                panic!("Sending PeersRequest to unknown peer {:?}", addr);
+            }
+            trace!("Sending {:?}", peers_request);
```
popzxc: Why do we need logging in sandbox?

ponimas: That wasn't my idea. If you grep - you'll find another traces in the sandbox' source and I don't think it's bad.
Could you please explain why you think this is "wrong"?

vitvakatu: Oh, actually, it's bad idea to use `trace` level. `info` should be better.

ponimas: @vitvakatu should I also raise logging level to `info` in `sandbox_tests_helper.rs`?

popzxc: In that particular case -- we have either `panic` (and `trace` will not be reached) or `trace`. Thus `trace` will be shown only if test is passed, and logging, why test did pass is a strange thing -- it should pass because it is expected to pass, isn't it?

ponimas: I don't log "why" test passed. I log what's going on in this particular moment. Just to be able to see that not in code but printed in the console. 
It won't affect **anything** so I don't understand why it worths discussion at all.

popzxc: I don't insist on anything, I just don't see any profit in this line. However, it does cost some CPU ticks, and if it doesn't help us (and I still think that it doesn't), I suppose we should not include it in the code.
Once again, it's no more than a question, if you're sure it's a useful thing, feel free to ignore my nit :)

vitvakatu: Just to elaborate - I don't think `trace` is a good idea because log output will be heavily mixed with a lot of log entries from rustc and cargo.

ponimas: @vitvakatu at least one can do `RUST_LOG=exonum=trace`, so I don't consider that a problem at all.

exonum/src/node/basic.rs
```diff
@@ -195,16 +195,15 @@ impl NodeHandler {
     pub fn handle_peer_exchange_timeout(&mut self) {
         if !self.state.peers().is_empty() {
             let to = self.state.peers().len();
-            let gen_peer_id = || -> usize {
+            let peer_id = {
```
aleksuss: Just a little nitpick.
I suppose a goal of this lambda is to avoid a calculation of `peer_id` if `self.state.peers().len() == 0`. In this case I guess we always have at least a one peer and this doesn't matter. But in general it's a good practice to use a lambda or a function reference in such cases.
P.S. By the way. What does say a clippy on this ?

ponimas: ooh, now I see the point...

exonum/src/sandbox/sandbox.rs
dvush: I found some method names in the sandbox confusing. (I am not talking about this PR).
For example here `trace!("Sending {:?}", peers_request);` make me think that this method is responsible for sending some messages, but in reality, messages are sent by NodeHandler and this function just asserts that this message was sent.
For instance, in the file node/basic.rs there is already tracing for this message  `trace!("Request peers from peer with addr {:?}", peer.addr());`
Please correct me if I am wrong.

ponimas: > I found some method names in the sandbox confusing.
I found all of them confusing.

PR #848, opened 2018-08-02
Timeout increase for consecutive rounds [ECR-1281]

dvush:
Here the size of the timeout increase is introduced as a parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `NoiseHandshake::finalize` now returns error if remote peer's public key is not
   in `ConnectList`. (#811)
 
+- Now nodes will switch to `min_propose_timeout` for block proposal timeout
+  faster if they receive more than `propose_timeout_threshold` transactions
+  during `max_propose_timeout`. (#844)
+
 ## 0.9.0 - 2018-07-19
 
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   Now timeout for round r is
   `first_round_timeout + (r-1)*round_timeout_increase`
    where `round_timeout_increase` is determined as a some percentage of
-  `first_round_timeout`. Value of this percentage is defined in constant inside
-  `ConsensusConfig` (#848)
+  `first_round_timeout`. Value of this percentage is defined
+  in `ConsensusConfig::TIMEOUT_LINEAR_INCREASE_PERCENT` constant (10%). (#848)
 
 ### New features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,12 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Config parameter `external_address` is now a required value. (#826)
 
-- Config parameter `round_timeout` has been removed.
-  `first_round_timeout` and `round_timeout_increase` should be used instead.
+- Config parameter `round_timeout` has been renamed to `first_round_timeout`.
   Now timeout for round r is
-  `first_round_timeout + (r-1)*round_timeout_increase`. (#848)
+  `first_round_timeout + (r-1)*round_timeout_increase`
+   where `round_timeout_increase` is determined as a some percentage of
+  `first_round_timeout`. Value of this percentage is defined in constant inside
+  `ConsensusConfig` (#848)
 
 ### New features
 
@@ -27,7 +29,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   addresses. (#826)
 
 - Now each consecutive round is longer than previous
-  by `round_timeout_increase`. (#848)
+  by some constant percentage of `first_round_timeout`. (#848)
 
 ### New Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,16 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Config parameter `external_address` is now a required value. (#826)
 
+- Config parameter `round_timeout` is removed.
+  `first_round_timeout` and `round_timeout_increase` should be used instead. (#848)
+
 ### New features
 
 - Added possibility to use domain names instead of IP addresses as a peer's
   addresses. (#826)
+
+- Now each consecutive round is longer than previous
+  by `round_timeout_increase`. (#848)
 
 ### New Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Config parameter `external_address` is now a required value. (#826)
 
-- Config parameter `round_timeout` is removed.
-  `first_round_timeout` and `round_timeout_increase` should be used instead. (#848)
+- Config parameter `round_timeout` has been removed.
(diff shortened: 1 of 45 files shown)


ilammy:
I believe this definitely warrants an entry in changelog.

exonum/src/blockchain/config.rs
```diff
@@ -89,16 +89,19 @@ pub struct StoredConfiguration {
...
     /// been committed during this period or not.
+    /// Each consecutive round will be longer then previous by round_timeout_increase.
     ///
     /// Note that rounds in Exonum
     /// do not have a defined end time. Nodes in a new round can
     /// continue to vote for proposals and process messages related to previous
     /// rounds.
-    pub round_timeout: Milliseconds,
+    pub first_round_timeout: Milliseconds,
+    /// Amount of time that is added to round timeout as round number increase.
```
vitvakatu: I think `Amount of time that is added to the round timeout each round` is better.

exonum/src/node/mod.rs
```diff
@@ -676,7 +681,9 @@ impl NodeHandler {
     /// Returns start time of the requested round.
     pub fn round_start_time(&self, round: Round) -> SystemTime {
         let previous_round: u64 = round.previous().into();
-        let ms = previous_round * self.round_timeout();
+        let ms = previous_round * self.first_round_timeout()
+            + (previous_round * previous_round.saturating_sub(1)) / 2
```
vitvakatu: Could you please clarify this formulae?

exonum/src/blockchain/config.rs
```diff
@@ -89,16 +89,19 @@ pub struct StoredConfiguration {
...
+    pub round_timeout_increase: Milliseconds,
```
vitvakatu: I don't think it should be constant value

exonum/src/blockchain/config.rs
```diff
@@ -89,16 +89,19 @@ pub struct StoredConfiguration {
...
+    /// time when second round starts, regardless of whether a new block has
     /// been committed during this period or not.
+    /// Each consecutive round will be longer then previous by round_timeout_increase.
     ///
     /// Note that rounds in Exonum
     /// do not have a defined end time. Nodes in a new round can
     /// continue to vote for proposals and process messages related to previous
     /// rounds.
-    pub round_timeout: Milliseconds,
+    pub first_round_timeout: Milliseconds,
```
vitvakatu: Why not `initial_round_timeout`?

exonum/src/node/mod.rs
dvush: It is the sum of the arithmetic progression. For round r cumulative increase in round start time due to increasing round timeout will be (r-1)(r-2)/2 * round_timeout_increase. I'll clarify it in the comment.

exonum/src/blockchain/config.rs
dvush: It is equivalent but I like first_* since it is more concrete.

dvush:
@vitvakatu 
>dismiss previous review
Why is it still blocked then?

exonum/src/sandbox/consensus/round_details.rs
```diff
@@ -1335,8 +1335,8 @@ fn handle_precommit_positive_scenario_commit_with_queued_precommit() {
     let sandbox = timestamping_sandbox();
     let sandbox_state = SandboxState::new();
 
-    let block_1_delay = 2 * sandbox.round_timeout() + PROPOSE_TIMEOUT + 1;
-    let block_2_delay = 2 * sandbox.round_timeout() + 2 * PROPOSE_TIMEOUT + 1;
+    let block_1_delay = 2 * sandbox.current_round_timeout() + PROPOSE_TIMEOUT + 1;
+    let block_2_delay = 2 * sandbox.current_round_timeout() + 2 * PROPOSE_TIMEOUT + 1;
```
ilammy: I think if round timeout is not fixed then we can't do `2 * sandbox.current_round_timeout()` anymore and have to be more elaborate. It may work with current configuration values, but may suddenly break if someone tweaks them.

dvush: I thought about it, but this parameter is not used by block builder anyway, should be revisited.

CHANGELOG.md
```diff
@@ -16,11 +16,17 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Config parameter `external_address` is now a required value. (#826)
 
+- Config parameter `round_timeout` is removed.
```
vitvakatu: We prefer to use present perfect tense in changelog.

CHANGELOG.md
```diff
@@ -16,11 +16,17 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+  `first_round_timeout` and `round_timeout_increase` should be used instead. (#848)
```
vitvakatu: I suggest to elaborate on it a little bit, just to make sure user will understand how to use new feature. And haven't you checked all the documentation? We need to update it too (perhaps in a different PRs)

exonum/src/blockchain/config.rs
```diff
@@ -89,16 +89,19 @@ pub struct StoredConfiguration {
...
+    /// Amount of time that is added to the round timeout each round
```
DarkEld3r: Please add a dot.

exonum/src/blockchain/config.rs
```diff
@@ -89,16 +89,19 @@ pub struct StoredConfiguration {
...
+    pub round_timeout_increase: Milliseconds,
```
DarkEld3r: Should this value be configurable? Can we simply use the percentage of `first_round_timeout`?

exonum/src/blockchain/config.rs
```diff
@@ -138,11 +141,11 @@ impl ConsensusConfig {
         const MIN_TXS_BLOCK_LIMIT: u32 = 100;
         const MAX_TXS_BLOCK_LIMIT: u32 = 10_000;
 
-        if self.round_timeout <= 2 * self.max_propose_timeout {
+        if self.first_round_timeout <= 2 * self.max_propose_timeout {
```
DarkEld3r: I suppose some validation of `round_timeout_increase` should be added too. Do you intentionally allow zero value?

CHANGELOG.md
```diff
@@ -16,16 +16,20 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
 - Config parameter `external_address` is now a required value. (#826)
 
-- Config parameter `round_timeout` is removed.
-  `first_round_timeout` and `round_timeout_increase` should be used instead. (#848)
+- Config parameter `round_timeout` has been renamed to `first_round_timeout`.
+  Now timeout for round r is
+  `first_round_timeout + (r-1)*round_timeout_increase`
+   where `round_timeout_increase` is determined as a some percentage of
+  `first_round_timeout`. Value of this percentage is defined in constant inside
+  `ConsensusConfig` (#848)
```
ilammy: Some last-minute nit: I'd reword this changelog entry to explicitly reference the `TIMEOUT_LINEAR_INCREASE_PERCENT` constant instead of 'some percentage' (it is public after all). We can also note its current value (10%) so that developers reading the changelog have an idea of its magnitude.

nemat-rakhmatov:
Hi all,

From the above conversations I can not quite get why this feature was implemented? Could someone shed the light into it? What were the requirements to this change?

Thanks.

DarkEld3r:
@nemat-rakhmatov 
Hello,

In Exonum we suppose that the network is partially synchronous (see https://exonum.com/blog/10-13-17-consensus-message-delivery-and-time/ for details): each message is delivered faster than unknown finite time T. As we don’t know this time T we should increase round timeout to have a stable BFT consensus. In future we are planning to implement adaptive round timeout adjustment (see https://exonum.com/blog/adjustable-timeouts-and-empty-blocks-2/).

nemat-rakhmatov:
Thanks for explanation (and links). Things getting more or less clear now.

PR #852, opened 2018-08-06
Fix expedited proposal reactivation [ECR-2040]

ilammy:
It turns out that `handle_propose_timeout()` calls way more common during network operation than I expected. We should actually allow expedited proposal to happen again only after we broadcast a real proposal to other nodes.

With this the nodes will no longer spam the log with bogus expedited timeout notifications. I should have found this earlier :( I remember now that someone in #844 has pointed out that this may not a good place to reset the flag.


exonum/src/node/consensus.rs
```diff
@@ -734,6 +733,8 @@ impl NodeHandler {
             trace!("Broadcast propose: {:?}", propose);
             self.broadcast(propose.raw());
 
+            self.allow_expedited_propose = true;
```
aleksuss: I'd name `allow_forced_propose` but never mind

PR #853, opened 2018-08-06
Fix pool overflow

vldm:
This pull request should fix overflow during persistent pool manipulation.

TODO:
- [x] Add more docs

### DIFF
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -481,6 +481,10 @@ impl Blockchain {
                 // Consensus messages cache is useful only during one height, so it should be
                 // cleared when a new height is achieved.
                 schema.consensus_messages_cache_mut().clear();
+                let txs_in_block = schema.last_block().tx_count();
+                let txs_count = schema.transactions_pool_len_index().get().unwrap_or(0);
+                debug_assert!(txs_count >= txs_in_block as u64);
+                schema.transactions_pool_len_index_mut().set(txs_count - txs_in_block as u64);
             }
             fork.into_patch()
         };
@@ -492,6 +496,7 @@ impl Blockchain {
             self.api_sender.clone(),
             self.fork(),
         );
+
         // Invokes `after_commit` for each service in order of their identifiers
         for service in self.service_map.values() {
             service.after_commit(&context);
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -483,8 +483,10 @@ impl Blockchain {
                 schema.consensus_messages_cache_mut().clear();
                 let txs_in_block = schema.last_block().tx_count();
                 let txs_count = schema.transactions_pool_len_index().get().unwrap_or(0);
-                debug_assert!(txs_count >= txs_in_block as u64);
-                schema.transactions_pool_len_index_mut().set(txs_count - txs_in_block as u64);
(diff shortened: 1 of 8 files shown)


exonum/src/blockchain/mod.rs
```diff
@@ -481,6 +481,10 @@ impl Blockchain {
                 // Consensus messages cache is useful only during one height, so it should be
                 // cleared when a new height is achieved.
                 schema.consensus_messages_cache_mut().clear();
+                let txs_in_block = schema.last_block().tx_count();
```
aleksuss: Actually I don't understand how transaction count from last block relates to size of pool after commit. I'd suggest that this value should be something like this:
```rust
let tx_pool_size_after_commit = {
       let txs_count = schema.transactions_pool_len_index().get().unwrap_or(0);
       if txs_count <= txs_block_limit {
              0
       else {
              txs_count - txs_block_limit
       }
};
```

ilammy:
I have verified these changes integrated with dynamic propose timeouts. They seem to work fine :+1: (despite my fears that they may be broken).

exonum/src/blockchain/mod.rs
ilammy: As I understand we assume here that all transactions included in the last block were present in our pool, so in order to get accurate pool entry count we should subtract the transaction count in the last block from our pool size cache.

However, it is possible that the last committed block originates from other node which had a different pool state? For example, our pool has [T1, T2], other node's pool has [T1, T2, T3, T4], the block includes [T1, T2, T3, T4]. In this case it seems we'll still have an overflow on line 487.

aleksuss: The question is: Which exactly last block do we get in this case ? I suppose it will be a block with current height but we need to know a size of block that will be generated after `self.merge(patch)?;` don't we ?

vldm: >The question is: Which exactly last block do we get in this case ? I suppose it will be a block with current height but we need to know a size of block that will be generated after self.merge(patch)?; don't we ?

We are working with Fork that contain's "new block". So in `fork` last_block is really block at current height. But in db\blockchain\node_state `last_block` is still outdated block.

>Actually I don't understand how transaction count from last block relates to size of pool after commit. I'd suggest that this value should be something like this:
```rust
let tx_pool_size_after_commit = {
       let txs_count = schema.transactions_pool_len_index().get().unwrap_or(0);
       if txs_count <= txs_block_limit {
              0
       else {
              txs_count - txs_block_limit
       }
};
```

I don't get it. If we get `txs_count > txs_block_limit` it's a bug, because we assume that block's tx is always in the pool.

exonum/src/blockchain/mod.rs
```diff
@@ -483,8 +483,10 @@ impl Blockchain {
                 schema.consensus_messages_cache_mut().clear();
                 let txs_in_block = schema.last_block().tx_count();
                 let txs_count = schema.transactions_pool_len_index().get().unwrap_or(0);
-                debug_assert!(txs_count >= txs_in_block as u64);
-                schema.transactions_pool_len_index_mut().set(txs_count - txs_in_block as u64);
+                debug_assert!(txs_count >= u64::from(txs_in_block));
```
ilammy: I'd say clippy is unreasonable at times. Maybe we should review `u32`, `u64`, `usize` usage for various 'lengths' to avoid such casts where possible.

PR #857, opened 2018-08-07
Remove logging utils in favor of env_logger [ECR-1709]

VladYermakov:
Removed `colored` and `term` crates. And changed log format to default for `env_logger`.
Now we have `WARN 2018-08-06T13:05:37Z: test_logger: Hello, world!` format for our logs with colored log level.


vitvakatu:
I remember we wanted milliseconds in the log output

exonum/src/helpers/mod.rs
```diff
@@ -22,31 +22,20 @@ pub mod user_agent;
...
-/// "Mon, 16 Jul 2018 13:37:18.594 +0100"
-const LOG_TIMESTAMP_FORMAT: &str = "%a, %e %b %Y %H:%M:%S%.3f %z";
-
 /// Performs the logger initialization.
 pub fn init_logger() -> Result<(), SetLoggerError> {
-    let mut builder = Builder::new();
-    builder.format(format_log_record);
+    let mut builder = Builder::from_default_env();
 
     if env::var("RUST_LOG").is_ok() {
```
ivan-ochc: Seems, that we don't need this check

PR #858, opened 2018-08-07
[ECR-1627] TODO cleanup pt.3

ponimas:
Outdated TODOS removed.
SandboxBuilder slightly improved.
Single node network test added.


exonum/src/sandbox/consensus/basic.rs
```diff
@@ -76,6 +76,19 @@ fn test_reach_one_height() {
     sandbox.assert_state(Height(2), Round(1));
 }
 
+/// idea of the test is to reach one height in the network with single validator
```
popzxc: Capital letter in the beginning and dot at the end :)

exonum/src/sandbox/consensus/basic.rs
```diff
@@ -76,6 +76,19 @@ fn test_reach_one_height() {
...
+#[test]
+fn test_one_validator() {
+    let sandbox = sandbox::timestamping_sandbox_builder()
+        .with_logger()
+        .with_validators(1)
```
popzxc: Isn't `with_validators(1)` default behavior?

exonum/src/sandbox/consensus/mod.rs
```diff
@@ -32,11 +32,8 @@ mod unsynchronized_message;
 // LOCK
 
 // - Send precommit when get lock   //covered in lock_to_propose_when_get_2_3_prevote_positive()
-// - if haven’t incompatible prevotes
+// - if haven't incompatible prevotes
```
popzxc: Maybe "If there are no incompatible prevotes"?

exonum/src/sandbox/consensus/basic.rs
ponimas: Nope.

exonum/src/sandbox/consensus/mod.rs
ponimas: Dunno. Haven't read it yet.

exonum/src/sandbox/consensus/basic.rs
ponimas: Oh

exonum/src/sandbox/consensus/basic.rs
popzxc: Hmm... For `TestKitBuilder` it's default, so as for me it's confusing. We have two structures with the same method but for one `with_validator(1)` is default, and for one it is not. I'm not sure if is is OK, but I think it is at least discussible.

ponimas: I'm sure it's ok.

popzxc: Ok then.

exonum/src/sandbox/sandbox.rs
```diff
@@ -765,21 +786,19 @@ fn gen_primitive_socket_addr(idx: u8) -> SocketAddr {
...
-    ];
-
-    let addresses: Vec<SocketAddr> = (1..5).map(gen_primitive_socket_addr).collect::<Vec<_>>();
+    let validators = (0..validators_count)
+        .map(|i| gen_keypair_from_seed(&Seed::new([i + 12; SEED_LENGTH])))
+        .collect::<Vec<_>>();
+
+    let service_keys = (0..validators_count)
+        .map(|i| gen_keypair_from_seed(&Seed::new([i + 20; SEED_LENGTH])))
+        .collect::<Vec<_>>();
```
ilammy: When we had only four validators the seed values were all different for them. However, now if we request more than eight validators then some keys may be generated using the same seeds:

* `i + 12` for `i == 8`
* `i + 20` for `i == 0`

I think that it is important to keep the seeds fixed so that tests are reproducible, but maybe we should avoid collisions. For example, with something like this:

```rust
let mut keys = (42..).map(|i| gen_keypair_from_seed(&Seed::new([i; SEED_LENGTH]));
let validators = keys.take(validators_count).collect::<Vec<_>>();
let service_keys = keys.take(validators_count).collect::<Vec<_>>();
```

exonum/src/sandbox/sandbox.rs
```diff
@@ -765,21 +786,19 @@ fn gen_primitive_socket_addr(idx: u8) -> SocketAddr {
...
+
+    let addresses: Vec<SocketAddr> = (1..=validators_count)
+        .map(gen_primitive_socket_addr)
+        .collect::<Vec<_>>();
```
ilammy: Here we can mention the type only once inside the turbofish.

exonum/src/sandbox/sandbox.rs
ilammy: Oh, that wouldn't work as is, but you get the idea.

I'm not sure that we need it though. It's just tests.

exonum/src/sandbox/sandbox.rs
```diff
@@ -743,13 +747,30 @@ impl SandboxBuilder {
...
+
+    pub fn with_logger(mut self) -> Self {
+        self.logger = true;
+        self
+    }
+
     pub fn build(self) -> Sandbox {
-        let mut sandbox = sandbox_with_services_uninitialized(self.services, self.consensus_config);
+        if self.logger {
+            let _ = env_logger::try_init();
```
vldm: Logger initialisation causes side effects: other test inside this process will get init logger as well.

ponimas: Yup. But that won't affect anything.
That's exactly why we use `try_init`

ponimas: Hm... I see, but is there a way to avoid that?

ilammy: Uh... Why do we need the logging to be configurable per-test? Maybe we can always initialize logging (probably will need to use `std::sync::Once` to do it only once) and then just fiddle with `RUST_LOG` environment variable when we run particular tests during debugging. I do not think that log output by default is valuable: if the test passes then I don't really care what happens there.

ponimas: @ilammy 
> Why do we need the logging to be configurable per-test?
Have no idea. It's the same scheme we use in TestKit. 
>  will need to use std::sync::Once to do it only once
`try_init` will init it only once anyway.

ilammy: Right! I remembered what would be nice to have.

For tests we should probably configure the logger to write output to **stdout** instead of stderr. This makes test log more readable as successful tests will not output anything other than "ok" while failing tests will dump the log along with panics and assertion failures.

ponimas: `2>&1` redirects stderr to stdout
`2>errrors.log` writes it to file

vldm: @ponimas there a big difference between "output log to stdout" and "redirect log to stdout".
You redirect log outside of current process.

Outputting log into stderr allows libtest to capturing, and only showing it on test fail.

So, i agrue with @liammy, can we:
1) init log using https://docs.rs/env_logger/0.5.12/env_logger/struct.Builder.html (with target Stdout)
2) initi it unconditionally and remove `with_logger` from sandbox builder 
3) probably use `Once` to init logger with builder

ponimas: @vldm  ok, now I see the point.
But have 2 questions left:
- Does that mean that we should do the same in the TestKit?
- I still can't get why the hell we should use `Once` and can't just use `try_init` and will be grateful for an explanation.

vldm: > Does that mean that we should do the same in the TestKit?
Probably yes, based on logging purposes in testkit @slowli 

> I still can't get why the hell we should use Once and can't just use try_init and will be grateful for an explanation.

https://docs.rs/env_logger/0.5.12/src/env_logger/lib.rs.html#725-727

`builder.init()` will fail at second initialisation

PR #865, opened 2018-08-09
Update actix and actix-web to 0.7.3 [ECR-2121]

ilammy:
* `Syn`/`Unsync` markers has been removed, they are not needed anymore.

* `actix::msgs::SystemExit` has been made private. now the system should be stopped by calling `System::stop()` which will send this message. Instead of passing SystemAddr just pass the System handle itself.

* `ws::start()` now accepts a reference to HttpRequest instead of instance.

* actix-web 0.7.3 has a regression (since 0.7.0) which breaks our tests. Apply a temporary workaround for it.

Obsoletes #814, #824.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Custom log formatting (along with `colored` and `term` dependencies) has been
   removed in favor of `env_logger`. (#857).
 
-- Several dependencies have been updated. (#861)
+- Several dependencies have been updated. (#861, #865)
 
 ## 0.9.1 - 2018-08-02
 
--- a/exonum/src/api/backends/actix.rs
+++ b/exonum/src/api/backends/actix.rs
@@ -19,7 +19,7 @@
 
 pub use actix_web::middleware::cors::Cors;
 
-use actix::{msgs::SystemExit, Addr, Arbiter, Syn, System};
+use actix::{Addr, System};
 use actix_web::{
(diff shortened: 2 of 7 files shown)


exonum/src/api/backends/actix.rs
```diff
@@ -108,6 +106,14 @@ impl ServiceApiBackend for ApiBuilder {
     }
 }
 
+// FIXME: workaround for a regression in actix-web 0.7.3 (ECR-2149)
```
DarkEld3r: We don't have such conventions, but personally I'm using "FIXME" to mark things that should be fixed _before_ creating pull request. I propose to change it into "TODO".

aleksuss:
@ilammy let's make clippy happy ?

ilammy:
Hm... Weird. The previous build was successful, it had the same versions of the toolchain and dependencies. I tried restarted the job, but the error persists (something with `mio` crate). I'll look into that.

CHANGELOG.md
```diff
@@ -65,7 +65,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Custom log formatting (along with `colored` and `term` dependencies) has been
   removed in favor of `env_logger`. (#857).
 
-- Several dependencies have been updated. (#861)
+- Several dependencies have been updated. (#861, #865)
```
aleksuss: @ilammy 
`actix` and `actix-web` dependencies have beet updated to 0.7.3 version (#861, #865) ?

ilammy: Yes, but I think this is a 'minor' update in the sense that it does not break Exonum's API, so it does not warrant any specific mention in the changelog. The link to the pull request should be enough.

PR #866, opened 2018-08-10
Remove connect messages [ECR-1452]

pavel-mukhanov:
Remove `connect` messages from the protocol. Add `PeersResponse` instead. 
Now, the handshake is handled by Noise protocol. As far as we have two connections between peers, `ConnectInfo` is transmitted as noise handshake payload.


exonum/src/events/network.rs
```diff
@@ -115,6 +116,28 @@ impl ConnectionsPool {
...
+                        peer,
+                        network_tx.clone(),
+                        &handle,
+                        &handshake_params,
+                    )
+            })
+            .map(|conn_tx| conn_fut(Ok(conn_tx).into_future()))
+    }
+
     fn connect_to_peer(
```
dvush: I am just curios what will happen if node will connect_to_peer where peer is itself(it can happen with peer request logic, for example). Node handler handles this situation well but I can't get how network part will handle it. When it will drop connection to itself especially if keepalive is set?

exonum/src/node/basic.rs
```diff
@@ -39,10 +40,9 @@ impl NodeHandler {
 
     /// Handles the `Connected` event. Node's `Connect` message is sent as response
```
dvush: Maybe it is an outdated comment.

exonum/src/events/network.rs
```diff
@@ -115,6 +116,31 @@ impl ConnectionsPool {
...
+    fn get_or_create(
+        &self,
+        peer: SocketAddr,
+        network_tx: &mpsc::Sender<NetworkEvent>,
+        handle: &Handle,
+        network_config: NetworkConfiguration,
+        handshake_params: &HandshakeParams,
+    ) -> Option<Box<dyn Future<Item = mpsc::Sender<RawMessage>, Error = io::Error>>> {
+        self.get(peer)
+            .map(|conn_tx| conn_fut(Ok(conn_tx).into_future()))
```
slowli: Off-topic: I find `conn_fut` name non-descriptive.

exonum/src/events/network.rs
```diff
@@ -290,31 +315,30 @@ impl RequestHandler {
...
+                            &network_tx,
+                            &handle,
+                            network_config,
+                            &handshake_params,
+                        );
+
+                        if let Some(conn_tx) = conn_tx {
+                            to_box(conn_tx)
+                        } else {
+                            to_box(Self::send_unable_connect_event(peer, &network_tx))
```
slowli: Does `send_unable_connect_event` really need to return something if the result is discarded anyway in all cases?

exonum/src/events/noise/mod.rs
```diff
@@ -45,11 +46,21 @@ pub const MAX_MESSAGE_LENGTH: usize = 65_535;
...
 
 pub trait Handshake {
-    fn listen<S: AsyncRead + AsyncWrite + 'static>(self, stream: S) -> HandshakeResult<S>;
-    fn send<S: AsyncRead + AsyncWrite + 'static>(self, stream: S) -> HandshakeResult<S>;
+    type Result;
+
+    fn listen<S: AsyncRead + AsyncWrite + 'static>(
+        self,
+        stream: S,
+    ) -> HandshakeResult<S, Self::Result>;
```
slowli: May be a personal opinion, but I'd prefer to return `HandshakeResult<S, Self>` here.

exonum/src/events/network.rs
```diff
@@ -393,21 +423,16 @@ impl Listener {
...
-                        .map_err(|e| e.0)
-                        .and_then(|(raw, stream)| (Self::parse_connect_msg(raw), Ok(stream)))
-                        .and_then(move |(connect, stream)| {
-                            trace!("Received handshake message={:?}", connect);
-
-                            Self::process_incoming_messages(stream, network_tx, connect, address)
-                        })
-                        .map(|_| {
+
+                    Self::process_incoming_messages(stream, network_tx, address, info.unwrap()).map(
```
slowli: `unwrap` seems inappropriate here.

exonum/src/events/noise/tests.rs
```diff
@@ -410,25 +423,27 @@ impl NoiseErrorHandshake {
 }
 
 impl Handshake for NoiseErrorHandshake {
-    fn listen<S>(self, stream: S) -> HandshakeResult<S>
+    type Result = Option<ConnectInfo>;
```
slowli: I'd say that `Option`al nature of the result could be better encapsulated into the error variants of `listen` / `send` results. Or is it a normal that the information cannot be obtained from the remote peer?

exonum/src/events/network.rs
pavel-mukhanov: Yes, `network_tx` is unsync sender and needs to be polled.

exonum/src/events/network.rs
pavel-mukhanov: In the current implementation, it will never be dropped. I'll fix that.

exonum/src/events/noise/tests.rs
pavel-mukhanov: It's normal for the sender to not receive information from responder during the handshake because we have two connections between peers. The sender receives it when responder creates a return connection.

exonum/src/events/network.rs
```diff
@@ -115,6 +116,31 @@ impl ConnectionsPool {
...
     }
 
+    fn get_or_create(
+        &self,
+        peer: SocketAddr,
+        network_tx: &mpsc::Sender<NetworkEvent>,
+        handle: &Handle,
+        network_config: NetworkConfiguration,
+        handshake_params: &HandshakeParams,
+    ) -> Option<Box<dyn Future<Item = mpsc::Sender<RawMessage>, Error = io::Error>>> {
```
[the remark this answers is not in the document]
pavel-mukhanov: I agree to make `connect_to_peer` async, will do in my next task which is to make a single connection between peers.

exonum/src/blockchain/mod.rs
```diff
@@ -507,7 +507,7 @@ impl Blockchain {
     }
 
     /// Saves the `Connect` message from a peer to the cache.
-    pub(crate) fn save_peer(&mut self, pubkey: &PublicKey, peer: Connect) {
+    pub fn save_peer(&mut self, pubkey: &PublicKey, peer: ConnectInfo) {
```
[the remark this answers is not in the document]
pavel-mukhanov: I agree we can use `ConfigManager` for this.

exonum/src/events/network.rs
slowli: Sorry for being obscure; I've meant something like

```rust
fn send_unable_connect_event(
    peer: SocketAddr,
    network_tx: mpsc::Sender<NetworkEvent>,
) -> Box<dyn Future<Item = (), Error = io::Error>> {
    let event = NetworkEvent::UnableConnectToPeer(peer);
    Box::new(
        network_tx
            .send(event)
            .map(drop)
            .map_err(|_| other_error("can't send network event")),
    )
}
```

Note `map(drop)`; we don't use the sender after we return it from the future.

exonum/src/node/mod.rs
```diff
@@ -794,6 +797,35 @@ pub struct ConnectInfo {
...
+        serde_json::to_vec(&self)
+    }
+
+    /// Tries to deserialize the given `value` into `ConnectInfo`.
+    pub fn try_deserialize(value: &[u8]) -> serde_json::Result<Self> {
+        serde_json::from_slice(value)
+    }
+}
+
+impl StorageValue for ConnectInfo {
```
alekseysidorov: You should implement `ConnectInfo` via `encoding_struct` and also implement some traits. See code for example.
https://github.com/alekseysidorov/exonum-btc-anchoring/blob/native-segwit/src/btc/macros.rs#L98

pavel-mukhanov: I tried, but `encoding_struct` doesn't work with `deserialize_with`.

alekseysidorov: Why you need to use `deserialize_with`?

pavel-mukhanov: We use `deserialize_with=deserialize_socket_address` to check validity of `SocketAddr`.

slowli: Off-topic: `deserialize_with` is used not to check validity, but rather to resolve hosts during deserialization. Which, IMO, looks like magic overdose (host resolution is non-deterministic and involves external global state); I'd be more satisfied with later resolution separated from deserialization. Btw, the present deserialization most probably does not survive a round-trip if used with the [non-human-readable](https://docs.rs/serde/1.0.72/serde/trait.Serializer.html#method.is_human_readable) / binary format.

slowli: To add to the previous comment: If I understand correctly, during serialization we _always_ record a resolved `SocketAddr` (I'm not saying it's necessarily correct in all cases), so during deserialization from storage we don't need custom resolution logic (= `deserialize_with`). The whole process looks significantly 🤦‍♂️ to me, so I'd suggest to at least explore the possibility to use choose the type for the peer address more smartly. As a possibility: use `{ host: String, port: u16 }` tuple (cf. [`ToSocketAddrs` implementation for `(&str, u16)`](https://doc.rust-lang.org/src/std/net/addr.rs.html#865-882)) and resolve it *after* deserialization.

pavel-mukhanov: You are right when deserializing from storage when don't need `deserialize_with`. But we also use `ConnectInfo` in config files and `deserialize_with` added only to resolve addresses when parsing config. It may be worth to use different `ConnectInfo` for storage and config files.

slowli: I'm not sure using a resolved address in stored `ConnectInfo` is the correct way to proceed in the long run:

- A host may be resolved to multiple addresses, some of which may not work for any particular node unpredictably.
- The resolution may change with time for a multitude of reasons.

aleksuss: @slowli Such approach for resolving addresses was implemented as a first step. We've planed to change it in the near future. More correct solution is to resolve addresses before connect to the nodes. But in this case we should change network messages cause they contain `SocketAddr` rather then `String` for address of peer record.

exonum/src/encoding/fields.rs
```diff
@@ -459,6 +460,52 @@ impl<'a> Field<'a> for SocketAddr {
...
+    }
+
+    unsafe fn read(buffer: &'a [u8], from: u32, to: u32) -> Self {
+        let public_key_offset = from + SocketAddr::field_size();
+
+        let address = SocketAddr::read(buffer, from, public_key_offset);
+        // Unwrap here is safe, because `from_slice` only fails if length is not equal to
+        // PUBLIC_KEY_LENGTH.
+        let public_key =
+            PublicKey::from_slice(&buffer[public_key_offset as usize..to as usize]).unwrap();
```
aleksuss: IMO `.unwrap()` looks a little bit weird here. Do you really sure it's ok ?

exonum/src/encoding/fields.rs
```diff
@@ -459,6 +460,56 @@ impl<'a> Field<'a> for SocketAddr {
...
+        let public_key_offset =
+            CheckedOffset::new(from.unchecked_offset() + SocketAddr::field_size());
+        let public_key_length = (to - public_key_offset)?.unchecked_offset() as usize;
+
+        if public_key_length != PUBLIC_KEY_LENGTH {
+            return Err(Error::IncorrectPublicKeyLength {
+                expected_length: PUBLIC_KEY_LENGTH,
+                actual_length: public_key_length,
+            });
+        }
```
ilammy: Is length check alone enough to guarantee that `PublicKey::from_slice()` in `read()` never fails? Can `PublicKey` work with literally any byte sequence if it has expected length?

exonum/src/events/codec.rs
```diff
@@ -149,8 +151,14 @@ mod test {
...
-        let mut params =
-            HandshakeParams::new(public_key, secret_key, SharedConnectList::default(), 1024);
+        let address: SocketAddr = "127.0.0.1:8000".parse().unwrap();
+        let mut params = HandshakeParams::new(
+            public_key,
+            secret_key,
+            SharedConnectList::default(),
+            1024,
+            address,
+        );
```
ilammy: I guess if we have this many parameters then it is time to have a builder for this struct... Though, `HandshakeParams::new()` seems to be used only twice, and in both cases it uses `SharedConnectList::default()` and `1024` as parameters. Maybe we can get rid of these and accept only the keys and the address.

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
```diff
@@ -140,25 +189,38 @@ impl NoiseHandshake {
...
-            .and_then(|(stream, handshake)| handshake.write_handshake_msg(stream))
-            .and_then(|(stream, handshake)| handshake.finalize(stream));
-        Box::new(framed)
+        match self.peers_exchange.clone() {
+            Some(peers_exchange) => {
+                let framed = self.write_handshake_msg(stream, &[])
+                    .and_then(|(stream, handshake)| handshake.read_handshake_msg(stream))
+                    .and_then(move |(stream, handshake, _)| {
+                        handshake.write_handshake_msg(stream, &peers_exchange.into_bytes())
+                    })
```
ilammy: This ugly formatting and `move` closure trigger me with inconsistency, but I don't see how this can be improved :(

exonum/src/node/basic.rs
```diff
@@ -18,31 +18,31 @@ use std::{error::Error, net::SocketAddr};
...
     pub fn handle_message(&mut self, raw: RawMessage) {
         match Any::from_raw(raw) {
-            Ok(Any::Connect(msg)) => self.handle_connect(msg),
             Ok(Any::Status(msg)) => self.handle_status(&msg),
             Ok(Any::Consensus(msg)) => self.handle_consensus(msg),
             Ok(Any::Request(msg)) => self.handle_request(msg),
             Ok(Any::Block(msg)) => self.handle_block(&msg),
             Ok(Any::Transaction(msg)) => self.handle_tx(msg),
             Ok(Any::TransactionsBatch(msg)) => self.handle_txs_batch(&msg),
+            Ok(Any::PeersExchange(msg)) => self.handle_peers_exchange(&msg),
```
ilammy: Maybe it's a bit late, but I think `Peers` or `PeerList` (or at least `ExchangePeers`) will be a better wording than `PeersExchange`. It would be in-line with other messages which are descriptive nouns of their content. This structure literally contains a "list of known peers". (And `PeerList` can't be misread for stupid jokes.)

exonum/src/node/mod.rs
ilammy: Do we have a task to track this issue with resolving?

CHANGELOG.md
```diff
@@ -22,6 +22,14 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Config parameter `external_address` is now a required value. (#826)
 
+- `Connect`-messages have been removed from the protocol. Now handshake between
+  peers are handled by noise protocol. (#866)
+
+- New message type `PeersExchange` has been added. Peers sends it as the
+  response to `PeersRequest`. (#866)
+
+### New features
```
ilammy: I think this line should be removed. There's already a new feature section below and the following item is definitely a breaking change.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -76,9 +76,9 @@ impl NoiseWrapper {
         self.read(input, MAX_MESSAGE_LENGTH)
     }
 
-    pub fn write_handshake_msg(&mut self) -> Result<Vec<u8>, NoiseError> {
+    pub fn write_handshake_msg(&mut self, msg: &[u8]) -> Result<Vec<u8>, NoiseError> {
         // Payload in handshake messages can be empty.
```
alekseysidorov: It seems that this comment is obsolete.

exonum/src/node/mod.rs
```diff
@@ -795,6 +793,35 @@ pub struct ConnectInfo {
     pub public_key: PublicKey,
 }
 
+impl ConnectInfo {
+    /// Tries to serialize the given `ConnectInfo` into a UTF-8 encoded JSON.
+    pub fn try_serialize(self) -> serde_json::Result<Vec<u8>> {
```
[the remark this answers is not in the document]
alekseysidorov: In my opinion `to_json` name is more appropriate for this method.

exonum/src/node/mod.rs
```diff
@@ -794,6 +797,35 @@ pub struct ConnectInfo {
...
+        serde_json::to_vec(&self)
+    }
+
+    /// Tries to deserialize the given `value` into `ConnectInfo`.
+    pub fn try_deserialize(value: &[u8]) -> serde_json::Result<Self> {
```
alekseysidorov: And `from_json` for this.

exonum/src/node/mod.rs
```diff
@@ -794,6 +797,35 @@ pub struct ConnectInfo {
...
+        serde_json::from_slice(value)
+    }
+}
+
+impl StorageValue for ConnectInfo {
+    fn into_bytes(self) -> Vec<u8> {
+        self.try_serialize().unwrap()
```
alekseysidorov: Why not just write as `Field`?

exonum/src/node/mod.rs
```diff
@@ -794,6 +797,35 @@ pub struct ConnectInfo {
...
+    }
+
+    fn from_bytes(value: Cow<[u8]>) -> Self {
+        Self::try_deserialize(value.as_ref()).unwrap()
```
alekseysidorov: Also here.

exonum/src/encoding/fields.rs
pavel-mukhanov: Yes, `sodiumoxide` implementation allows creating `PublicKey` from any byte sequence.

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
pavel-mukhanov: I don't like it either(

exonum/src/node/mod.rs
```diff
@@ -805,6 +809,38 @@ pub struct ConnectInfo {
...
+        self.write(&mut vec_bytes, 0, Self::field_size());
+        vec_bytes
+    }
+
+    /// Tries to deserialize the given `value` into `ConnectInfo`.
+    #[allow(unsafe_code)]
+    pub fn try_deserialize(value: &[u8]) -> Self {
+        unsafe { Self::read(&value, 0, Self::field_size()) }
+    }
+}
```
ilammy: This looks suspicious.

The methods are named with `try` prefix, but based on the return types they do not enable error handling. Usually `try_somethings()` return `Option` or `Result`. That's in contrast with regular `something()` which is expected to panic on error condition. Furthermore, serialization consumes `self` which is a bit unusual. These functions seem to be used only in `impl StorageValue`, I guess we can just inline them there and do not provide this functionality separately.

Also, the `unsafe` block raises some questions. Is it really safe? For example, what if the `value` slice is too short or contains corrupted data?

exonum/src/node/basic.rs
```diff
@@ -18,31 +18,31 @@ use std::{error::Error, net::SocketAddr};
...
-    /// if received `Connect` message is correct.
-    pub fn handle_connected(&mut self, address: &SocketAddr, connect: Connect) {
-        info!("Received Connect message from peer: {:?}", address);
-        // TODO: use `ConnectInfo` instead of connect-messages. (ECR-1452)
-        self.handle_connect(connect);
+    /// Handles the incoming connection. Node connects to the sender
+    /// if received `PeerList` is correct.
+    pub fn handle_connected(&mut self, peers_exchange: &PeerList) {
+        info!("Received PeerList  from peer: {:?}", peers_exchange);
+        self.handle_peers_exchange(&peers_exchange);
```
ilammy: I think various `peers_exchange` mentions in the code should be replaced with `peer_list` to match the new struct name.

exonum/src/node/mod.rs
```diff
@@ -805,6 +809,38 @@ pub struct ConnectInfo {
...
+    pub fn try_serialize(self) -> Vec<u8> {
+        let mut vec_bytes = vec![0_u8; Self::field_size() as usize];
+        self.write(&mut vec_bytes, 0, Self::field_size());
+        vec_bytes
+    }
+
+    /// Tries to deserialize the given `value` into `ConnectInfo`.
+    #[allow(unsafe_code)]
+    pub fn try_deserialize(value: &[u8]) -> Self {
+        unsafe { Self::read(&value, 0, Self::field_size()) }
```
aleksuss: `&` before value I suppose is redundant ?

exonum/src/node/mod.rs
```diff
@@ -809,34 +809,25 @@ pub struct ConnectInfo {
...
-
-impl StorageValue for ConnectInfo {
-    fn into_bytes(self) -> Vec<u8> {
-        self.try_serialize()
-    }
-
     fn from_bytes(value: Cow<[u8]>) -> Self {
-        Self::try_deserialize(value.as_ref())
+        assert_eq!(value.len(), Self::field_size() as usize);
+        unsafe { Self::read(value.as_ref(), 0, Self::field_size()) }
```
ilammy: I believe there are more things to validate here, aside from size. One of `ConnectInfo`'s fields is `SocketAddr` which is likely to have complex serialized structure that can easily be corrupted. It is used to hold IPv4 or IPv6 addresses, but what if the bytes are pure garbage and denote neither? Yes, the process is unlikely to segfault due to size assertion, but I guess networking code can start behaving in weird ways (or fail unexpectedly) if it encounters unintelligible `struct sockaddr`.

I guess we can roll with `unsafe` here if we give some pinky promise to never feed invalid data into this function, but I strongly prefer more explicit runtime checks over random crashes or undefined behavior resulting from `unsafe` blocks not being actually safe. Intended usage of `StorageValue` trait (serialize/deserialize values for blockchain storage) can act as such pinky promise, I guess, as you're supposed to use `from_bytes` only on slices resulting from `into_bytes`. And there is a precedent of `Duration` using `unsafe` for `StorageValue` implementation.

Thus I do not consider this serious enough to prevent the PR from being merged. If there is no low-effort way to avoid this unsafe block then go ahead. That's just my personal dislike of unmotivated `unsafe` blocks in cases where they should have been avoidable. Probably we should have a task to audit and clean up `unsafe` blocks at some point.

ilammy: > Probably we should have a task to audit and clean up unsafe blocks at some point.

And I should definitely learn to read before I write :man_facepalming: as we do have such tasks:

* [[ECR-174] Revise StorageKey interface](https://jira.bf.local/browse/ECR-174)
* [[ECR-641] Safety of StorageValue/StorageKey read methods](https://jira.bf.local/browse/ECR-641)
* [[ECR-800] Revise unsafe code](https://jira.bf.local/browse/ECR-800)

some of which even explicitly refer to the these exact issues of `StorageValue` trait.

So please go ahead with the merge, this particular `unsafe` block may be fixed later.

exonum/src/encoding/fields.rs
```diff
@@ -459,6 +460,56 @@ impl<'a> Field<'a> for SocketAddr {
     }
 }
 
+impl<'a> Field<'a> for ConnectInfo {
```
slowli: Not using `encoding_struct` increases the amount of code that is hardly analyzable, potentially leads to UB, and will get thrown out relatively soon. I'd advise to use `encoding_struct` and implement address resolving on the level of `ConnectListConfig` if we can't decouple resolve from deserialization right now.

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
```diff
@@ -143,35 +192,46 @@ impl NoiseHandshake {
 }
 
 impl Handshake for NoiseHandshake {
-    fn listen<S>(self, stream: S) -> HandshakeResult<S>
+    type Result = Vec<u8>;
```
slowli: Why is this type so abstract? Can't we return `SocketAddr` here? (For now; in general, we want to support domain names if I understand correctly.)

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
```diff
@@ -143,35 +192,46 @@ impl NoiseHandshake {
...
-        let framed = self.write_handshake_msg(stream)
-            .and_then(|(stream, handshake)| handshake.read_handshake_msg(stream))
-            .and_then(|(stream, handshake)| handshake.write_handshake_msg(stream))
-            .and_then(|(stream, handshake)| handshake.finalize(stream))
-            .map_err(move |e| {
-                e.context(format!("peer {} disconnected", peer_address))
-                    .into()
-            });
-        Box::new(framed)
+        match self.peer_list.clone() {
```
slowli: We've discussed it elsewhere: There is no reason to use a message here, a `SocketAddr` (or, potentially, domain name + port) should suffice.

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
```diff
@@ -143,35 +192,46 @@ impl NoiseHandshake {
...
+            Some(peer_list) => {
+                let framed = self.write_handshake_msg(stream, &[])
+                    .and_then(|(stream, handshake)| handshake.read_handshake_msg(stream))
+                    .and_then(move |(stream, handshake, _)| {
+                        handshake.write_handshake_msg(stream, &peer_list.into_bytes())
+                    })
+                    .and_then(|(stream, handshake)| handshake.finalize(stream, Vec::new()))
```
slowli: If we don't use the returned value in any case, I don't see why it should be returned in the interface. Also, if we return a socket address (potentially, with a domain name), I don't see convincing reasons why the result should be generic.

exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
```diff
@@ -44,20 +51,57 @@ impl HandshakeParams {
...
         }
     }
 
     pub fn set_remote_key(&mut self, remote_key: PublicKey) {
-        self.remote_key = Some(into_x25519_public_key(remote_key));
+        self.remote_key = Some(remote_key);
+        self.remote_key_x25519 = Some(into_x25519_public_key(remote_key));
+    }
+
+    pub fn peer_list(&self) -> Option<PeerList> {
```
slowli: See the comment about the struct fields. The very existence of this method tells me that `HandshakeParams` is designed incorrectly.

exonum/src/node/mod.rs
pavel-mukhanov: @ilammy yes, https://jira.bf.local/browse/ECR-2310

DarkEld3r:
Outdated.

PR #871, opened 2018-08-14
Drop tokio-timer dependency [ECR-2120]

ilammy:
@dependabot has been pestering us with update of tokio-timer from 0.1.2 to 0.2.5. However, the API of this crate has changed significantly since it has become a part of the main Tokio framework. My attempts at using just tokio-timer alone were not successful. It seems to work only when tightly integrated with tokio and tokio-core, so let's just use it via `tokio::util::FutureExt::deadline()` to set the deadlines we need. With that we can replace tokio-timer with tokio (which includes the timer).

Tokio is future-based framework so we'll need a `Core` to run the futures. Furthermore, futures and streams are not *that* well integrated so in order to extract an item from a stream with a deadline set we need to do some dancing around with putting a stream into a future and then getting it back. Unfortunately, with current API it is not possible to recover the stream when it times out. But we're panicking in that case anyway so that should be fine for tests.

Obsoletes #855

TODO:

- [x] Investigate [benchmark compilation failure](https://travis-ci.org/exonum/exonum/jobs/415797261)


exonum/src/events/tests.rs
```diff
@@ -45,19 +45,26 @@ impl TestHandler {
...
+            network_events_rx: Some(network_events_rx),
             network_requests_tx,
-            network_events_rx: receiver.wait(),
         }
     }
 
     pub fn wait_for_event(&mut self) -> Result<NetworkEvent, ()> {
-        let event = self.network_events_rx.next().unwrap()?;
-        Ok(event)
+        let rx = self.network_events_rx.take().expect("event stream gone");
```
slowli: It seems you can use [`Stream::by_ref()`](https://docs.rs/futures/0.1.23/futures/stream/trait.Stream.html#method.by_ref) here in order to not bother with wrapping the stream into an `Option<_>`:

```rust
let rx = self.network_events_rx.by_ref();
let future = rx.into_future()
    .deadline(Instant::now() + Duration::from_secs(30))
    .map_err(drop);

let mut core = Core::new().unwrap();
let (event, _) = core.run(future)?;
event.ok_or(())
```

ilammy: Awesome! Thank you.

This also solves the issue with retrieving the stream in case of timeout.

exonum/Cargo.toml
```diff
@@ -34,11 +34,11 @@ env_logger = "=0.5.12"
 atty = "=0.2.11"
 bytes = "=0.4.9"
 futures = "=0.1.23"
+tokio = "=0.1.7"
```
slowli: As `tokio` bundles most of the components together, could it make sense to use them through the `tokio` crate? I'm particularly interested in `tokio-core`, [which is deprecated](https://github.com/tokio-rs/tokio-core):

> This crate is scheduled for deprecation in favor of tokio.
>
> tokio-core is still actively maintained, but only bug fixes will be applied. All new feature development is happening in tokio.

Regarding [ECR-740](https://jira.bf.local/browse/ECR-740): I think it could be at least worth trying removing `tokio-core` from our deps; first, because it's deprecated; second, because `tokio` contains [`Runtime` interface](https://tokio-rs.github.io/tokio/tokio/runtime/index.html), which bundles together event loop, thread pool, task spawning, etc., which could make tasks like implementing a thread pool for transactions unnecessary _by design_ (unless we'd like to implement advanced stuff, something like GPU-accelerated verification).

PR #872, opened 2018-08-14
Add tests for "rebroadcast" endpoint [ECR-2130]

VladYermakov:
Added two tests:
- `sandbox::consensus::transactions::rebroadcast_transactions` which tests the logic of rebroadcast action logic
- `testkit::tests::system_api::rebroadcast` which tests endpoint logic


exonum/src/sandbox/sandbox.rs
```diff
@@ -243,6 +248,13 @@ impl Sandbox {
         self.inner.borrow_mut().handle_event(event);
     }
 
+    pub fn rebroadcast(&self) {
```
VladYermakov: Is it ok? Or maybe change this to `handle_event` or something similar with parameter of event?

ilammy: It'd keep it as is. I doubt a generic event handler will make the code more readable right now. `self.inner.borrow_mut().handle_event(...)` pattern is used only three times in the code and all of them are distinct.

Maybe we could rename it to `recv_rebroadcast()` or something like that so that it is visually different from `broadcast()` method. It's up to you.

exonum/src/sandbox/consensus/transactions.rs
```diff
@@ -232,6 +232,27 @@ fn duplicate_tx_in_pool() {
...
 
+#[test]
+fn rebroadcast_transactions() {
+    let sandbox = timestamping_sandbox();
+
+    let mut transactions = TimestampingTxGenerator::new(DATA_SIZE)
+        .take(5)
+        .collect::<Vec<_>>();
+
+    transactions.sort_by(|a, b| a.hash().cmp(&b.hash()));
```
vldm: Can we move this into separete sandbox_helper, i am sure we can replace sorting algorithm later.

exonum/src/sandbox/consensus/transactions.rs
```diff
@@ -232,6 +232,27 @@ fn duplicate_tx_in_pool() {
...
+
+    for tx in &transactions {
+        sandbox.recv(tx);
+    }
+
+    sandbox.recv_rebroadcast();
+
+    for tx in &transactions {
+        sandbox.broadcast(tx)
```
slowli: Do I understand correctly that this check enforces a particular order the transactions should be broadcast in? Ideally, I'd like to have a method to extract all broadcast messages (e.g., via an `Iterator`), and verify that they as a set correspond to the expected set. I suppose this is not critical, but would look good as an improvement.

PR #873, opened 2018-08-14
Add tests for NoiseWrapper::encrypt_msg/decrypt_msg [ECR-1712]

pavel-mukhanov:
Test to check `encrypt/decrypt` methods in `NoiseWrapper` has been added.


exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -143,13 +144,14 @@ impl NoiseWrapper {
         Ok(None)
     }
 
-    fn read(&mut self, input: &[u8], len: usize) -> Result<Vec<u8>, NoiseError> {
+    pub fn read(&mut self, input: &[u8], len: usize) -> Result<Vec<u8>, NoiseError> {
```
aleksuss: If we mark method as public then it's a good practice to add some docs to describe what does this method, isn't it ?

pavel-mukhanov: It's a typo, I forgot to revert this method to private.

exonum/src/events/noise/tests.rs
```diff
@@ -107,6 +112,84 @@ fn test_converted_keys_handshake() {
...
+
+    let res = h_i.encrypt_msg(message.as_ref(), &mut buffer_msg);
+    assert!(res.is_ok());
+
+    let len = LittleEndian::read_u32(&buffer_msg[..HEADER_LENGTH]) as usize;
+    let mut to_decrypt = vec![];
+    to_decrypt.extend_from_slice(&buffer_msg);
+
+    let res = h_r.decrypt_msg(len, &mut BytesMut::from(&to_decrypt[..]));
+    assert!(res.is_ok());
```
slowli: Nit: this assertion is redundant given the `unwrap()` on the next line.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -160,8 +162,12 @@ impl NoiseWrapper {
     // of AEAD authentication data. Therefore to calculate an actual message
     // length we need to subtract `TAG_LENGTH` multiplied by messages count
     // from `data.len()`.
+    //
+    // f32 precision is enough to calculate message length.
```
slowli: Cast to `f32` is not needed. `ceil` can be easily implemented for integer arguments.

exonum/src/events/noise/tests.rs
```diff
@@ -107,6 +112,84 @@ fn test_converted_keys_handshake() {
...
+    test_encrypt_decrypt(64);
+}
+
+#[test]
+fn test_noise_encrypt_decrypt_long_message() {
+    test_encrypt_decrypt(MAX_MESSAGE_LENGTH + 1);
+}
+
+#[test]
+fn test_noise_encrypt_decrypt_bogus_message() {
```
slowli: I wonder if it could make sense to test `MessagesCodec` here.

exonum/src/events/noise/tests.rs
```diff
@@ -107,6 +112,84 @@ fn test_converted_keys_handshake() {
...
+    let mut buffer_msg = BytesMut::with_capacity(msg_size);
+
+    let _res = h_i.encrypt_msg(&vec![0_u8; msg_size], &mut buffer_msg);
+
+    let len = LittleEndian::read_u32(&buffer_msg[..HEADER_LENGTH]) as usize;
+    let mut to_decrypt = vec![];
+    to_decrypt.extend_from_slice(&buffer_msg);
+
+    // Wrong length.
+    let res = h_r.decrypt_msg(len - 1, &mut BytesMut::from(&to_decrypt[..]));
```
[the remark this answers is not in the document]
pavel-mukhanov: Because it mutates noise `session` in `NoiseWrapper::read` method while reading the decrypting the message. But I agree to move chunking logic to codec, maybe not in this PR.

exonum/src/events/noise/tests.rs
pavel-mukhanov: `MessagesCodec` is tested for some wrong input in their own module. I think those tests combined with `encrypt_decrypt` tests introduced here is enough to cover codec functionality. But maybe I miss something.

exonum/src/events/noise/tests.rs
pavel-mukhanov: In my option is more clear to use `expect("...")`.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -160,14 +162,19 @@ impl NoiseWrapper {
...
+        debug_assert!(message_len > 0);
+        1 + (message_len - 1) / MAX_MESSAGE_LENGTH
     }
 
     // In case of encryption we need to add `TAG_LENGTH` multiplied by messages count to
     // calculate actual message length.
-    fn encrypted_msg_len(&self, raw_message_len: usize) -> usize {
-        raw_message_len + TAG_LENGTH * ((raw_message_len / MAX_MESSAGE_LENGTH) + 1)
+    fn encrypted_msg_len(raw_message_len: usize) -> usize {
+        raw_message_len + TAG_LENGTH * Self::tag_count(raw_message_len)
```
ilammy: A minor observation on the code layout. I tend to put utility functions (i.e., the ones without **self** argument) outside the **impl** so that I don't have to prefix them with `Self::` which I personally find weird.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -156,25 +158,30 @@ impl NoiseWrapper {
...
+// length we need to subtract `TAG_LENGTH` multiplied by messages count
+// from `data.len()`.
+fn decrypted_msg_len(raw_message_len: usize) -> usize {
+    raw_message_len - TAG_LENGTH * tag_count(raw_message_len)
+}
+
+// In case of encryption we need to add `TAG_LENGTH` multiplied by messages count to
+// calculate actual message length.
+fn encrypted_msg_len(raw_message_len: usize) -> usize {
+    raw_message_len + TAG_LENGTH * tag_count(raw_message_len)
```
slowli: This code may be incorrect (the previous one seems to be incorrect, too). As I've pointed out in #764, the correct formula is

```rust
len + TAG_LENGTH * (floor(len / (MAX_MESSAGE_LENGTH - TAG_LENGTH)) + 1)
```

This leads to errors in tests (see the other comment).

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -140,12 +141,13 @@ impl NoiseWrapper {
         buf.extend_from_slice(&msg_len_buf);
 
         debug_assert_eq!(len, encoded_message.len());
```
slowli: Maybe off-topic, but I think should be fixed here: this assert is useless because of `let encoded_message = &encrypted_message[0..len]` above (which is useless on its own right, too). The assertion should be performed immediately after forming the message. You can include writing the message length into the message initialization, too:

```rust
let len = encrypted_msg_len(msg.len());
let mut encrypted_message = vec![0; len + HEADER_LENGTH];
// It could be worth trying to research vectors with uninitialized values, or `BytesMut`.

LittleEndian::write_u32(&mut encrypted_message[..HEADER_LENGTH], len as u32);

const CHUNK_LEN: usize = MAX_MESSAGE_LENGTH - TAG_LENGTH;
for (i, msg) in msg.chunks(CHUNK_LEN).enumerate() {
    // ideally, it could be worth to avoid reallocations here
    // by making `self.write` accept `&mut [u8]`.
    let written = self.write(msg)?;

    let start = i * CHUNK_LEN;
    encrypted_message[start..(start + CHUNK_LEN)].copy_from_slice(&written);
}
```

exonum/src/events/noise/tests.rs
```diff
@@ -107,6 +112,84 @@ fn test_converted_keys_handshake() {
         .expect("Unable to transition session into transport mode");
 }
 
+#[test]
+fn noise_encrypt_decrypt_short_message() {
+    check_encrypt_decrypt_message(64);
```
slowli: Please add a test case with

```rust
check_encrypt_decrypt_message(MAX_MESSAGE_LENGTH - 
    SIGNATURE_LENGTH -
    HEADER_LENGTH -
    10)
```

(`10` can be replaced with another small integer). Such a test fails as of now because of incorrect message length calculation (see another line comment).

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -123,55 +128,60 @@ impl NoiseWrapper {
     /// 3. Result message: first 4 bytes is message length(`len').
     /// 4. Append all encrypted packets in corresponding order.
     /// 5. Write result message to `buf`
-    pub fn encrypt_msg(&mut self, msg: &[u8], buf: &mut BytesMut) -> Result<Option<()>, io::Error> {
-        let len = self.encrypted_msg_len(msg.len());
-        let mut encrypted_message = Vec::with_capacity(len);
+    pub fn encrypt_msg(&mut self, msg: &[u8], buf: &mut BytesMut) -> io::Result<()> {
```
slowli: FYI, I present a sketch of this method without additional allocations besides extending `buf`:

```rust
pub fn encrypt_msg(&mut self, msg: &[u8], buf: &mut BytesMut) -> io::Result<()> {
    const CHUNK_LENGTH: usize = MAX_MESSAGE_LENGTH - TAG_LENGTH;

    let message_len = encrypted_msg_len(msg.len()) + HEADER_LENGTH;
    buf.reserve(message_len);

    let start = buf.len();

    #[allow(unsafe_code)]
    unsafe {
        // We fill all bytes in the buffer in the cycle below.
        buf.set_len(start + message_len);
    }

    let encrypted_message = &mut buf[start..];

    LittleEndian::write_u32(
        &mut encrypted_message[..HEADER_LENGTH],
        (message_len - HEADER_LENGTH) as u32,
    );

    // Start of the last written chunk.
    let mut start = 0;
    // Length of the last written chunk.
    let mut len = HEADER_LENGTH;

    for (i, msg) in msg.chunks(CHUNK_LENGTH).enumerate() {
        debug_assert_eq!(start + len, HEADER_LENGTH + i * MAX_MESSAGE_LENGTH);
        start = HEADER_LENGTH + i * MAX_MESSAGE_LENGTH;
        len = self.write(msg, &mut encrypted_message[start..])?;
    }
    debug_assert_eq!(start + len, message_len);

    Ok(())
}
```

As it has unsafe code, it requires additional testing, so I'm against putting it in the code now. However, additional allocations clearly can be avoided.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -123,55 +128,60 @@ impl NoiseWrapper {
...
-        Builder::with_resolver(PARAMS.parse().unwrap(), Box::new(SodiumResolver::new()))
+// In case of encryption we need to add `TAG_LENGTH` multiplied by messages count to
+// calculate actual message length.
+fn encrypted_msg_len(raw_message_len: usize) -> usize {
+    raw_message_len
+        + TAG_LENGTH * (raw_message_len.div_floor(&(MAX_MESSAGE_LENGTH - TAG_LENGTH)) + 1)
+}
+
+fn div_ceil(lhs: usize, rhs: usize) -> usize {
+    match lhs.div_rem(&rhs) {
```
slowli: Nit: I wonder how the use of `div_rem` here compares to using built-in `%` and `/` operations. If I understand correctly, it's the only use of the `num` crate in the main code.

pavel-mukhanov: I think we can remove it.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
pavel-mukhanov: I like this idea to use only `buf` without additional allocations. Added TODO.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -123,55 +127,61 @@ impl NoiseWrapper {
...
-    // calculate actual message length.
-    fn encrypted_msg_len(&self, raw_message_len: usize) -> usize {
-        raw_message_len + TAG_LENGTH * ((raw_message_len / MAX_MESSAGE_LENGTH) + 1)
-    }
+// Each message consists of the payload and 16 bytes(`TAG_LENGTH`)
+// of AEAD authentication data. Therefore to calculate an actual message
+// length we need to subtract `TAG_LENGTH` multiplied by messages count
+// from `data.len()`.
+fn decrypted_msg_len(raw_message_len: usize) -> usize {
+    raw_message_len - TAG_LENGTH * (div_ceil(raw_message_len, MAX_MESSAGE_LENGTH))
```
slowli: Nit: obsolete parentheses around `div_ceil`.

CHANGELOG.md
```diff
@@ -59,6 +59,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Bug with pool size overflow has been fixed. (#853)
 
+- Bug in `NoiseWrapper::decrypt_msg` caused by wrong calculation of
+  encrypted and decrypted message sizes has been fixed. (#873)
```
DarkEld3r: Please add a new line before "Transactions (signature) verification...".

exonum/src/events/noise/tests.rs
```diff
@@ -25,19 +26,23 @@ use tokio_io::{AsyncRead, AsyncWrite};
 
 use std::{net::SocketAddr, thread, time::Duration};
 
+use bytes::BytesMut;
```
DarkEld3r: This is an external crate using, so it should be moved before std usings. Additionally, the empty line before `snow` using should be removed.

PR #876, opened 2018-08-16
Move crypto into separate crate [ECR-712]

VladYermakov:
Now crypto lives in separate crate.


exonum/benches/criterion/lib.rs
```diff
@@ -16,20 +16,21 @@
...
 #[macro_use]
 extern crate exonum;
+extern crate exonum_crypto as crypto;
 extern crate futures;
 extern crate num;
 extern crate rand;
 extern crate tempdir;
 extern crate tokio_core;
 
+use bench_crypto::bench_crypto;
```
VladYermakov: I had to rename this mod because we need to use exonum_crypto as crypto.
I'm thinking if we can rename it back.

exonum/src/blockchain/transaction.rs
```diff
@@ -37,7 +37,8 @@ const TRANSACTION_STATUS_PANIC: u16 = TRANSACTION_STATUS_OK + 1;
 pub type ExecutionResult = Result<(), ExecutionError>;
 /// Extended version of `ExecutionResult` (with additional values set exclusively by Exonum
 /// framework) that can be obtained through `Schema` `transaction_statuses` method.
-pub type TransactionResult = Result<(), TransactionError>;
+#[derive(Clone, Debug, PartialEq)]
+pub struct TransactionResult(pub Result<(), TransactionError>);
```
VladYermakov: I'm not happy to use newtype struct in this place, but we can't use `TransactionResult` in way we used it before because of need to `impl CryptoHash` on it.
TODO: think about how to improve it.

examples/cryptocurrency-advanced/backend/src/lib.rs
```diff
@@ -16,6 +16,7 @@
 
 #![deny(missing_debug_implementations, missing_docs, unsafe_code, bare_trait_objects)]
 
+extern crate exonum_crypto as crypto;
```
ilammy: I'd avoid renaming crates unless absolutely necessary (e.g., to avoid making gazillion of changes in existing codebase).

The crate _is_ called `exonum_crypto`, just as Testkit is called `exonum_testkit`. Plus we do have to fix the import list either way (as `crypto` is no longer a part of `exonum`), so we can simply use the proper name.

exonum/src/storage/hash.rs
```diff
@@ -32,8 +32,8 @@ impl<T: CryptoHash> UniqueHash for T {
...
-impl UniqueHash for Hash {
-    fn hash(&self) -> Self {
-        *self
-    }
-}
+//impl UniqueHash for Hash {
+//    fn hash(&self) -> Self {
+//        *self
+//    }
+//}
```
ilammy: Commented out code.

crypto/src/crypto_lib/sodiumoxide/x25519.rs
```diff
@@ -14,19 +14,22 @@
...
+use super::sodiumoxide::crypto::sign::ed25519::{
     convert_ed_keypair_to_curve25519, convert_ed_pk_to_curve25519, convert_ed_sk_to_curve25519,
     PublicKey as PublicKeySodium, SecretKey as SecretKeySodium,
 };
 
 use std::fmt;
 use std::ops::{Index, Range, RangeFrom, RangeFull, RangeTo};
 
-use crypto;
+use super::super::super::{
```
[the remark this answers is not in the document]
ilammy: `use` starts resolution from the crate root, can't we just remove the `super`s?

crypto/src/crypto_lib/sodiumoxide/x25519.rs
```diff
@@ -14,19 +14,22 @@
...
+    write_short_hex as crypto_write_short_hex, PublicKey as crypto_PublicKey,
```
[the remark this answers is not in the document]
ilammy: Maybe we could split definitions between modules somehow to avoid name conflicts.

crypto/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+readme = "README.md"
+license = "Apache-2.0"
+keywords = ["exonum", "cryptography", "library"]
+categories = ["cryptography", "hashing"]
+description = "An extensible framework for blockchain software projects."
+autobenches = false
+
+[dependencies]
+byteorder = "1.2.3"
+chrono = { version = "=0.4.5", features = ["serde"] }
```
alekseysidorov: It seems that serde feature here is redundant.

crypto/src/lib.rs
```diff
@@ -491,6 +506,12 @@ impl Default for Hash {
     }
 }
 
+impl CryptoHash for Hash {
+    fn hash(&self) -> Hash {
```
vldm: it's probably a bit "strange" CryptoHash implementation. (It didn't cryptohash value)

crypto/src/lib.rs
```diff
@@ -607,19 +628,17 @@ impl CryptoHash for DateTime<Utc> {
     }
 }
 
+// TODO: think about move it anywhere
 impl CryptoHash for Duration {
     fn hash(&self) -> Hash {
-        let mut buffer = vec![0; Self::field_size() as usize];
```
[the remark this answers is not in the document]
vldm: But now we get code duplication, @VladYermakov could you revert it back, and create saperate task for this change.

exonum/src/blockchain/transaction.rs
vldm: It's ok that `TransactionResult` is now a new type. 
But can you provide `Into<Result<(), TransactionError>> for TransactionResult ` `Asref` and the rest helper traits that could safe us from manual wrapping and unwrapping results?

crypto/src/lib.rs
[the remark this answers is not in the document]
VladYermakov: @vldm, no, I couldn't, because it used `Field`, that is in core and can't be used in `crypto`.

crypto/src/lib.rs
VladYermakov: Yeah, @vldm, agree with you. But the reason of this implementation was in
```
error[E0119]: conflicting implementations of trait `storage::hash::UniqueHash` for type `crypto::Hash`:
  --> exonum/src/storage/hash.rs:35:1
   |
29 | impl<T: CryptoHash> UniqueHash for T {
   | ------------------------------------ first implementation here
...
35 | impl UniqueHash for Hash {
   | ^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `crypto::Hash`
   |
   = note: upstream crates may add new impl of trait `crypto::CryptoHash` for type `crypto::Hash` in future versions
```
And I didn't get any better approach

vitvakatu: I've also found this weird and tried to change it. Without success though. Current solution looks not good, but maybe it's the best we can do.

exonum/src/blockchain/transaction.rs
VladYermakov: Ok, no one of this trait can be useful for us because of `Cannot move out of borrowed content` for `Into` and `AsRef`, so I'm leaving it as is.

ilammy: To make it really transparent we could use `Deref` into `Result`, but TBH that would be abuse of the trait originally meant for smart pointers. But I don't really see why we can't add

```rust
impl From<Result<(), TransactionError>> for TransactionResult {
    fn from(error: Result<(), TransactionError>) -> TransactionResult {
        TransactionResult(error)
    }
}
```

to make `?` and `into()` work.

crypto/src/lib.rs
```diff
@@ -367,10 +380,10 @@ implement_public_crypto_wrapper! {
 /// secret keys.
 ///
 /// ```
-/// use exonum::crypto;
+/// # extern crate exonum_crypto;
```
vitvakatu: :thinking: Why not `use exonum::crypto`? We reexported crate, shouldn't it work?

VladYermakov: I've decided not to use reexported `crypto` in examples of `crypto` itself, when we don't use exonum at all.
Why is that bad?

vitvakatu: Oh, okay, I see it know. I thought they're exonum examples

crypto/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+authors = ["The Exonum Team <exonum@bitfury.com>"]
+homepage = "https://exonum.com/"
+repository = "https://github.com/exonum/exonum"
+documentation = "https://docs.rs/exonum-crypto"
+readme = "README.md"
+license = "Apache-2.0"
+keywords = ["exonum", "cryptography", "library"]
+categories = ["cryptography", "hashing"]
+description = "Cryptography related types, constants, traits and functions. "
+autobenches = false
```
DarkEld3r: Is this really needed?

crypto/src/lib.rs
```diff
@@ -607,19 +628,17 @@ impl CryptoHash for DateTime<Utc> {
     }
 }
 
+// TODO: think about move it anywhere (ECR-2217)
```
DarkEld3r: Please add a dot at the end of the comment.

crypto/src/crypto_lib/sodiumoxide/x25519.rs
```diff
@@ -14,19 +14,21 @@
...
     PublicKey as PublicKeySodium, SecretKey as SecretKeySodium,
 };
 
 use std::fmt;
 use std::ops::{Index, Range, RangeFrom, RangeFull, RangeTo};
 
-use crypto;
+use write_short_hex;
+use PublicKey as crypto_PublicKey;
+use SecretKey as crypto_SecretKey;
```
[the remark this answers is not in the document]
VladYermakov: Because `x25519` has own `PublicKey/SecretKey`, and we need to reuse ones of crate for `into_x25519_keypair`.

PR #877, opened 2018-08-16
Update npm lock files

boguslavsky:
To fix known security vulnerabilities in 3rd party npm dependencies.


ilammy:
Please merge #818 as well while we're here with `package-lock.json` updates.

PR #879, opened 2018-08-16
Move events errors into failure.

vldm:
This pull-request transfer io::Error into failure::Error:

TODO:
- [ ] Update changelog.

### DIFF
--- a/exonum/src/events/codec.rs
+++ b/exonum/src/events/codec.rs
@@ -14,11 +14,9 @@
 
 use byteorder::{ByteOrder, LittleEndian};
 use bytes::BytesMut;
+use failure;
 use tokio_io::codec::{Decoder, Encoder};
 
-use std::io;
-
-use super::error::other_error;
 use events::noise::{NoiseWrapper, HEADER_LENGTH as NOISE_HEADER_LENGTH};
 use messages::{MessageBuffer, RawMessage, HEADER_LENGTH};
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum
 
+- `api::Error::InternalError` now contain `failure::Error` instead of `Box<::std::error::Error>`. (#879)
```
aleksuss: `contains` ?

exonum/src/events/codec.rs
```diff
@@ -59,25 +57,27 @@ impl Decoder for MessagesCodec {
         let mut buf = self.session.decrypt_msg(len, buf)?;
 
         if buf[0] != 0 {
-            return Err(other_error("Message first byte must be set to 0"));
+            bail!("Message first byte must be set to 0");
```
aleksuss: I'd say: `A first byte of the message must be set to 0`

exonum/src/events/noise/tests.rs
```diff
@@ -161,15 +162,11 @@ fn test_noise_handshake_errors_ee_empty() {
     ));
     let (_, listener_err) = wait_for_handshake_result(addr, &params, bogus_message, None);
 
-    assert!(
-        listener_err
-            .unwrap_err()
-            .description()
-            .contains("WrongMessageLength")
-    );
+    listener_err.unwrap()
```
pavel-mukhanov: Doesn't failure provides the description?

vldm: nope, and is there some reason why we need use it instead of should panic?

pavel-mukhanov: Now it's not necessary, but early versions of this tests checked two errors instead of one.

PR #892, opened 2018-08-21
Do not include node as its own peer in config [ECR-2153]

ilammy:
Update the **finalize** command to not include the node itself into its list of expected connection peers. This duplication does not affect the verifier network but confuses the users so get rid of it.

Note that we take only the public key into account. External addresses are likely to be the same if all nodes reside on different hosts with the same listening port. The case of a single-node network is also a bit special as you can see from `config01.toml`.

Finally, note that this affects only automatic configuration generator. We make no attempt to remove the node presence from a manually edited configuration and we do not enforce this restriction at runtime with the data structures.


aleksuss:
@ilammy LGTM, but add some description to CHANGELOG please.

PR #893, opened 2018-08-21
Improve error messages about peer disconnections [ECR-2154, ECR-2156]

ilammy:
Include the peer address into the error message so that we at least know who tries to connect to us.

Peer address is not readily available from the Noise session (as we have only IO stream of the TCP socket) so thread it as an additional argument.

### DIFF
--- a/exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
+++ b/exonum/src/events/noise/wrappers/sodium_wrapper/handshake.rs
@@ -152,7 +152,10 @@ impl Handshake for NoiseHandshake {
             .and_then(|(stream, handshake)| handshake.write_handshake_msg(stream))
             .and_then(|(stream, handshake)| handshake.read_handshake_msg(stream))
             .and_then(|(stream, handshake)| handshake.finalize(stream))
-            .map_err(move |e| e.context(format!("peer {} disconnected", peer_address)).into());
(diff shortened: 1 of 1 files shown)


exonum/src/events/network.rs
```diff
@@ -215,7 +215,7 @@ impl ConnectionsPool {
         if let Some(remote_public_key) = connect_list.find_key_by_address(&peer) {
             let mut handshake_params = handshake_params.clone();
             handshake_params.set_remote_key(remote_public_key);
-            NoiseHandshake::initiator(&handshake_params).send(stream)
+            NoiseHandshake::initiator(&handshake_params, peer).send(stream)
```
ilammy: Maybe we could put peer address into `handshake_params` to avoid changing function signature. But TBH this socket address has nothing to do with Noise handshake so I have some doubts about that.

DarkEld3r: I don't see problems in changing the function signature. You can add a comment why this parameter is needed if it bothers you.

ilammy:
Pulled updates from master, resolved conflicts with recently merged PR #474.

ilammy:
I have realized that `failure` crate allows us to provide better error messages with context. I've updated the code so that it logs better messages with more meaningful context. It resolves another issue as well. Updated PR title accordingly.

PR #894, opened 2018-08-22
Do not reconnect to validators not in connect list [ECR-2155]

ilammy:
Return value of `State::remove_peer_with_addr()` is used by `NodeHandler::remove_peer_with_addr()` to choose whether to try reconnecting to the remote verifier peer after the node has been disconnected from it for whatever reason.

Currently this includes disconnections caused by the fact that our node does not trust this particular verifier (when the verifier is present in the shared configuration but is missing from node's local peer list).

In this case we'll still try reconnecting but we will most likely fail again and produce yet another useless bunch of log lines. The local configuration will not change without node restart. The remote peer using an invalid public key is not a transient network error.

To avoid doing needless work, check if the validator is present in our list of trusted peers before attempting reconnection.

Note that the remote peer is also occasionally trying to connect to us. so if this was indeed some man-in-the-middle attempt then we'll still connect to the real peer once the attack is finished.


ilammy:
It turned out that there exists some controversy over the concept of 'untrusted validators'. There is an alternative solution for the original issue (log spam). Instead of closing connection to the validator not present in the connect list we could just add it to the connect list (given that we confirm that it is a _validator_ node). However, we should still close connections to unknown _auditor_ nodes (those who identify themselves with a key that's missing from both connect list and the validator list).

ilammy:
After some internal discussion we did not get to the consensus on whether we need to add unknown validators to the connect list. I propose that we keep the current behavior: node refuses to communicate with validators not present in connect list. We can address the change of connect list behavior in a different PR.

However, we will avoid using wording "untrusted validators" as they are not really untrusted. The code and PR's title have been updated accordingly. I have also refactored the code a bit to make reconnection conditions more explicit.

exonum/src/node/state.rs
```diff
@@ -574,16 +574,27 @@ impl State {
...
-    /// Removes a peer by the socket address.
-    pub fn remove_peer_with_addr(&mut self, addr: &SocketAddr) -> bool {
-        if let Some(pubkey) = self.connections.remove(addr) {
-            self.peers.remove(&pubkey);
-            return self.config
-                .validator_keys
-                .iter()
-                .any(|x| x.consensus_key == pubkey);
+    /// Removes a peer by the socket address. Returns Some public key of the peer if it was indeed
+    /// connected or None if there was no connection with given socket address.
```
DarkEld3r: I suppose `Some` and `None` should be placed inside backticks.

PR #896, opened 2018-08-22
Update npm dependencies in demos

boguslavsky:
Nothing major.


aleksuss:
@boguslavsky What do you think about to update CHANGELOG for this updates ?

PR #903, opened 2018-08-27
Proptest key/value set index compare to rust set [ECR-2042]

dvush:
Add proptest for key/value set index and simplify other tests.


exonum/tests/collections/main.rs
```diff
@@ -14,111 +14,50 @@
...
-                let len = list.len();
-                if len > 0 {
-                    list.truncate(size as usize % len);
-                }
-            }
-            ListAction::Set(idx, val) => {
-                let len = list.len();
-                if len > 0 {
-                    list[idx as usize % len] = val;
+macro_rules! proptest_compare_collections {
```
ilammy: It's a bit unhygienic to expect the macro expansion environment to contain `compare_collections()` and `generate_action()` functions with correct signatures.

And it seems that this macro could be replaced with a function generic over an impl of

```rust
trait CollectionComparison {
    type Index;
    type Reference;

    fn new_index(&str, &mut Fork) -> Self::Index;
    fn new_reference() -> Self::Reference;
    fn generate_actions() -> Box<dyn Strategy<Value = ListAction<i32>>>;
    fn compare_collections(&Self::Index, &Self::Reference) -> TestCaseResult;
};
```

or something. However, that's my personal preference of avoiding macros for non-syntax operations.

dvush: Thanks, and I agree.
But it looks like your proposal will be verbose to implement because there are two things that conflict.
1) If you use associated types for Index it is nice since you don't have to specify traits for types but then you can't satisfy lifetimes for index(afaik there are no HRTB for associated types see this https://github.com/rust-lang/rfcs/blob/master/text/1598-generic_associated_types.md)
2) If you use generics to solve lifetime issue with HRTB then you have to specify traits for Index type. It may be hard to do this since there are too many things that goes in different ways for different collections => it may be unreadable.

ilammy: Yeah... If using generics is more verbose and less readable then macros then I don't see any issue with using macros for the job. Lifetimes and type inference are hard.

Thank you for the explanation.

PR #916, opened 2018-08-29
Add seperate message module.

vldm:
TODO: 
- [x] Remove unimplemented
- [x] Check docs rendering
- [x] Remove comments
- [x] Add more docs (me: requesting suggestions from reviewers)


exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,660 @@
...
+    /// Return concrete `Message<T>`
+    ///
+    /// # Panics
+    ///
+    /// On serialization fail this method can panic.
+    pub fn concrete<T: ProtocolMessage>(message: T, author: PublicKey, secret_key: &SecretKey)
+                                   -> Message<T> {
+        T::try_from(Self::new(message, author, secret_key))
+            .expect("BUG: Newly created message matched not as transaction.")
+    }
```
ilammy: These back-and-forth conversions in `new()` and `concrete()` with expected panics bother me. Is there a better way to do this, or this is just quick-and-dirty implementation?

exonum/src/messages2/spec.rs
```diff
@@ -0,0 +1,82 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// A low-level versions of `transactions!` macro, which generates structs for messages,
+/// but does not require the messages to implement `Transaction`.
+#[macro_export]
+macro_rules! messages {
```
ilammy: While we're here... Do we still need this magic macro with the new messages?

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,660 @@
...
+/// `$SignedMessage` is a typename of `SignedMessage` from which $ProtocolName could be created;
+/// `$cls_num` and `$typ_num` is a constant which represent
+///     message class and message type respectively
+///
+/// Each `$MessageType` should implement `Clone` and `Debug`.
+///
+macro_rules! impl_protocol {
+    ($signed_message:ident => $proto_name:ident{
+        $($cls_num:expr => $cls:ident{
+            $( $typ:ident = $typ_num:expr),+ $(,)*
```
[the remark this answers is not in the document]
vldm: @ilammy type is reserved word in rust x(

exonum/src/messages2/spec.rs
vldm: We need it to keep compatible with old `transactions!` macro

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,660 @@
...
+            RawTransaction = 0,
+            Connect = 1,
+            Status = 2,
+        },
+        1 => Consensus {
+            Precommit = 0,
+            Propose = 1,
+            Prevote = 2,
+        },
+        3 => Responses {
```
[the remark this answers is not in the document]
vldm: Thank you for your finds, yes we can generate this constants implicitly, but we need to define it according to our specification. (Not according to order)
I will add more tests to this magic constants.

exonum/src/messages2/protocol.rs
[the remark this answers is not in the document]
ilammy: [`type` is, but `$type` isn't](
https://play.rust-lang.org/?gist=746109be49535f01f0ae45c09ca3df8c&version=stable&mode=debug&edition=2015) 🕶

exonum/src/messages2/protocol.rs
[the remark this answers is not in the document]
ilammy: We have discussed this offline and turned out some some time ago `transactions!` macro _was_ using implicit message numbering, but it turned out to cause various issues. Furthermore, the compiler actually _does produce_ warnings if class/message IDs are duplicated (due to **match** being used inside the macro). This should prevent this kind of typos from happening, so explicit numbering seems to be way we want to go here.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,147 @@
...
+        Signature::from_slice(&self.raw[sign_idx..]).expect("Couldn't read signature")
+    }
+
+    /// Return byte array representation of internal data.
+    pub fn raw(&self) -> &[u8] {
+        &self.raw
+    }
+
+    pub fn hash(&self) -> Hash {
+        hash(self.raw())
```
aleksuss: `&self.raw` ?
The reason is if we for example will remove method `pub fn raw(&self) -> &[u8]` we will break `pub fn hash(&self) -> Hash` as well.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,147 @@
...
+        if !crypto::verify(signature, &full_buffer, &public_key) {
+            bail!("Can't verify message.");
+        }
+        Ok(())
+    }
+}
+
+impl ToHex for SignedMessage {
+    fn write_hex<W: ::std::fmt::Write>(&self, w: &mut W) -> ::std::fmt::Result {
+        self.raw().write_hex(w)
```
aleksuss: The same here.

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+/// This struct used to transfer transaction in network.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct RawTransaction {
+    service_id: u16,
+    transaction_set: TransactionFromSet,
+}
+
+/// Concrete raw transaction payload inside `TransactionSet` linked with `message_id` in this set.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct TransactionFromSet {
```
aleksuss: `TransactionFromSet` is a strange structure name as for me. Can you please explain what is motivate you to chose such name.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,146 @@
...
+        if buffer.len() <= EMPTY_SIGNED_MESSAGE_SIZE {
+            bail!("Message too short message_len = {}", buffer.len())
+        }
+        let signed = SignedMessage { raw: buffer };
+        let sign_idx = signed.raw.len() - SIGNATURE_LENGTH;
+
+        let pk = signed.author();
+        let signature = signed.signature();
+
+        Self::verify(&signed.raw[0..sign_idx], &signature, &pk)?;
```
ilammy: I guess we could make this an instance method of `SignedMessage` to avoid unpacking public key and signature here.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,146 @@
...
+    #[cfg(test)]
+    pub(crate) fn unchecked_from_vec(buffer: Vec<u8>) -> Self {
+        SignedMessage { raw: buffer }
+    }
+    #[cfg(not(test))]
+    pub(in messages) fn unchecked_from_vec(buffer: Vec<u8>) -> Self {
+        SignedMessage { raw: buffer }
+    }
+
+    #[allow(unsafe_code)]
```
ilammy: This attribute is no longer necessary for `author()` and `signature()` methods.

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,670 @@
...
+use helpers::{Height, Round, ValidatorId};
+use storage::{Database, MemoryDB, ProofListIndex, StorageValue};
+
+#[doc(hidden)]
+/// `SignedMessage` size with zero bytes payload.
+pub const EMPTY_SIGNED_MESSAGE_SIZE: usize = PUBLIC_KEY_LENGTH + SIGNATURE_LENGTH + mem::size_of::<u8>()* 2;
+
+#[doc(hidden)]
+/// `Message<TransactionsResponse>` size without transactions inside.
+pub const TRANSACTION_RESPONSE_EMPTY_SIZE: usize = EMPTY_SIGNED_MESSAGE_SIZE + 40;
```
ilammy: I guess it would be nice to have this magic constant `40` explained as well.

exonum/src/messages2/authorization.rs
ilammy: Using a getter method has its merits too. If we change the implementation details and store messages in a more convoluted way then we may not have to manually update all the impls which use the getter as it will still return the necessary raw message bytes.

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub(in messages) fn new(payload: T, message: SignedMessage) -> Message<T> {
+        Message { payload, message }
+    }
+
+    /// Returns hash of full message.
+    pub fn hash(&self) -> Hash {
+        hash(self.message.raw())
+    }
+
+    /// Returns serialized buffer.
```
DarkEld3r: a serialized

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn serialize(self) -> Vec<u8> {
+        self.message.raw
+    }
+
+    /// Return link to inner.
```
DarkEld3r: To the inner what? Payload?

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub fn inner(&self) -> &T {
+        &self.payload
+    }
+
+    /// Return link to signed message.
```
DarkEld3r: ...a link to the signed...

What is "link"? A reference?

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    pub(crate) fn signed_message(&self) -> &SignedMessage {
+        &self.message
+    }
+
+    /// Returns public key of message creator.
```
DarkEld3r: ...of the message...

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,670 @@
...
+//!     * processing - how message is processed and result of the processing
+//!     * generation - in which cases message is generated
+
+use bit_vec::BitVec;
+use chrono::{DateTime, Utc};
+
+use std::borrow::Cow;
+use std::fmt::Debug;
+use std::net::SocketAddr;
+use std::mem;
```
DarkEld3r:
```
use std::{borrow::Cow, fmt::Debug, ...};
```

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,146 @@
...
+        Ok(signed)
+    }
+
+    #[cfg(test)]
+    pub(crate) fn unchecked_from_vec(buffer: Vec<u8>) -> Self {
+        SignedMessage { raw: buffer }
+    }
+    #[cfg(not(test))]
+    pub(in messages) fn unchecked_from_vec(buffer: Vec<u8>) -> Self {
+        SignedMessage { raw: buffer }
```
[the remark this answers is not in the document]
vldm: We need to allow usage from cfg(test) and (in messages) module.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,146 @@
...
+        buffer.push(tag);
+        buffer.extend_from_slice(value.as_ref());
+        let signature = Self::sign(&buffer, secret_key).expect("Couldn't form signature");
+        buffer.extend_from_slice(signature.as_ref());
+        SignedMessage { raw: buffer }
+    }
+
+    /// Create `SignedMessage` wrapper from raw buffer.
+    /// Checks binary format and signature.
+    pub fn verify_buffer(buffer: Vec<u8>) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
vldm: I prefer to notify users that this function `verify_buffer`.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+/// | - - - - - - - -| - - - - - - |
+/// | `0..32`  | author's PublicKey     |
+/// | `32`     | message class          |
+/// | `33`     | message type           |
+/// | `34..N`  | Payload                |
+/// | `N..N+64`| Signature                |
+///
+///
+/// Every creation of `SignedMessage` lead to signature verification, or data signing procedure,
+/// which can slowdown your code. Beware `SignedMessage` message, this procedure is not free.
```
vitvakatu: > Beware `SignedMessage` message, this procedure is not free.

I'm not sure I understand this sentence :thinking:

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+
+        let pk = signed.author();
+        let signature = signed.signature();
+
+        Self::verify(signed.data_without_signature(), &signature, &pk)?;
+
+        Ok(signed)
+    }
+
+    fn data_without_signature(&self) -> &[u8]{
```
vitvakatu: missing space before `{` (rustfmt ?!)

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+    ) -> Result<(), Error> {
+        if !crypto::verify(signature, &full_buffer, &public_key) {
+            bail!("Can't verify message.");
+        }
+        Ok(())
+    }
+}
+
+impl ToHex for SignedMessage {
+    fn write_hex<W: ::std::fmt::Write>(&self, w: &mut W) -> ::std::fmt::Result {
```
vitvakatu: Please avoid `::` at the beginning

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// Transaction raw buffer.
+/// This struct is used to transfer transactions in network.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct RawTransaction {
+    service_id: u16,
+    transaction_set: TransactionSetPart,
+}
+
+/// Concrete raw transaction transaction inside `TransactionSet`.
```
vitvakatu: There is no `TransactionSet`

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+/// This type used inner inside `transactions!` to transfer some set.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct TransactionSetPart {
+    message_id: u16,
+    payload: Vec<u8>,
+}
+
+impl TransactionSetPart {
+    /// Creates TransactionSetPart from unchecked raw data.
```
vitvakatu: Please add backticks

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub fn service_id(&self) -> u16 {
+        self.service_id
+    }
+
+}
+
+impl BinaryForm for RawTransaction {
+    fn serialize(&self) -> Result<Vec<u8>, encoding::Error> {
+        let mut buffer = Vec::new();
+        buffer.resize(2, 0);
```
vitvakatu: Why not creating `vec![0;2]`?

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+impl<T: ProtocolMessage> StorageValue for Message<T> {
+    fn into_bytes(self) -> Vec<u8> {
+        self.message.raw
+    }
+
+    fn from_bytes(value: Cow<[u8]>) -> Self {
+        let message = SignedMessage::unchecked_from_vec(value.into_owned());
+        //TODO: Remove additional deserialization
```
vitvakatu: Issue number?

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+            message_id,
+            payload,
+        })
+    }
+}
+
+/// Wrappers around pair of concrete message payload, and full message binary form.
+/// Internally binary form saves message lossless,
+/// this important for use in a scheme with
+/// non-canonical serialization, for example with a `Protobuf`.
```
ilammy: I'd rephrase this comment like this to describe the intent better:

```rust
/// Wraps a `Payload` together with corresponding `SignedMessage`.
///
/// Usually one wants to work with fully parsed messages (i.e., `Payload`). However, occasionally
/// we need to retransmit the message into the network or save its serialized form. We could
/// serialize the `Payload` back, but Protobuf does not have a canonical form so the resulting
/// payload may have different binary representation (thus invalidating the message signature).
///
/// So we use `Message` to keep the original byte buffer around with the parsed `Payload`.
```

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,709 @@
...
+/// `$ProtocolName` is an name of protocol which is described by this schema,
+///     enum with same name will be created. All message classes are encapsulate by this enum;
+/// `$MessageClass` is a module name which is designed to handle messages,
+///     enum with same name will be crated. All message within `MessageClass` should be unique;
+/// `$MessageType` is a concrete messages within some `$MessageClass`;
+/// `$SignedMessage` is a typename of `SignedMessage` from which $ProtocolName could be created;
+/// `$cls_num` and `$typ_num` is a constant which represent
+///     message class and message type respectively
+///
+/// Each `$MessageType` should implement `Clone` and `Debug`.
```
ilammy: These `$names` now do not refer to anything. I guess this should be rewritten to instruct the users in plain prose.

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,709 @@
...
+
+    }
+}
+
+impl Protocol {
+    /// Creates new protocol message.
+    ///
+    /// # Panics
+    ///
+    /// On serialization fail this method can panic.
```
ilammy: (First of all, correct grammar would be "this method can panic on serialization failure".)

What exactly are the conditions for a panic? Is the user expected to provide some 'correct' arguments to avoid the panic, or is this just a warning that serialization may fail for example due to memory allocation issues?

In the former case I would describe the specific expectations (e.g., that the secret and public keys must match). Otherwise I think we can omit the panic section if avoiding the panic is not something under caller's control.

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,709 @@
...
+        message: T,
+        author: PublicKey,
+        secret_key: &SecretKey,
+    ) -> Message<T> {
+        let value = message.serialize().expect("Couldn't serialize data.");
+        let (cls, typ) = T::message_type();
+        let signed = SignedMessage::new(cls, typ, value, author, secret_key);
+        T::into_message_from_parts(message, signed)
+    }
+    /// Checks buffer and return instance of `Protocol`.
```
ilammy: I guess we should also say that the buffer is parsed here.

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,660 @@
...
+            Consensus::Precommit(ref msg) => msg.round(),
+        }
+    }
+}
+
+impl<T: ProtocolMessage> Into<Protocol> for Message<T> {
+    fn into(self) -> Protocol {
+        ProtocolMessage::into_protocol(self)
+    }
+}
```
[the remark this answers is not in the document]
ilammy: The point stands.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+/// which can slowdown your code. Beware `SignedMessage` message, this procedure is not free.
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Ord, PartialOrd)]
+pub struct SignedMessage {
+    pub(in messages) raw: Vec<u8>,
+}
+
+impl SignedMessage {
+    pub(crate) fn new(
+        cls: u8,
```
alekseysidorov: What is the `cls`? Maybe this means `class`?

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+        tag: u8,
+        value: Vec<u8>,
+        author: PublicKey,
+        secret_key: &SecretKey,
+    ) -> SignedMessage {
+        let mut buffer = Vec::new();
```
alekseysidorov: Why not `with_capacity`? Benchmarks shows that preallocated vectors faster.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+        buffer.extend_from_slice(value.as_ref());
+        let signature = Self::sign(&buffer, secret_key).expect("Couldn't form signature");
+        buffer.extend_from_slice(signature.as_ref());
+        SignedMessage { raw: buffer }
+    }
+
+    /// Creates `SignedMessage` wrapper from the raw buffer.
+    /// Checks binary format and signature.
+    pub fn verify_buffer(buffer: Vec<u8>) -> Result<Self, Error> {
+        if buffer.len() <= EMPTY_SIGNED_MESSAGE_SIZE {
```
alekseysidorov: `ensure!` macro shorter.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+            bail!("Message too short message_len = {}", buffer.len())
+        }
+        let signed = SignedMessage { raw: buffer };
+
+        let pk = signed.author();
+        let signature = signed.signature();
+
+        Self::verify(signed.data_without_signature(), &signature, &pk)?;
```
alekseysidorov: Why not payload?

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    fn data_without_signature(&self) -> &[u8]{
+        debug_assert!(self.raw.len() > EMPTY_SIGNED_MESSAGE_SIZE);
+        let sign_idx = self.raw.len() - SIGNATURE_LENGTH;
+        &self.raw[0..sign_idx]
+    }
+
+    #[cfg(test)]
+    pub(crate) fn unchecked_from_vec(buffer: Vec<u8>) -> Self {
```
alekseysidorov: Naming is not enough clear, maybe `from_vec_unchecked` be better?

exonum/src/messages2/mod.rs
alekseysidorov: I agree, `TransactionSetPart` sounds a little bit weird.

exonum/src/messages2/mod.rs
alekseysidorov: Also magic number is not clear without context.

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,709 @@
...
+        /// Exonum node responses.
+        2 => Responses {
+            /// Information about transactions, that sent as response to `TransactionsRequest`.
+            TransactionsResponse = 0,
+            /// Information about block, that sent as response to `BlockRequest`.
+            BlockResponse = 1,
+        },
+        /// Exonum node requests.
+        3 => Requests {
+
```
alekseysidorov: Extra newline

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,709 @@
...
+                }
+
+                fn try_from(p: $protocol_name) -> Result<Message<Self>,Protocol> {
+                    match p {
+                        $protocol_name::$class($class::$type(s)) => Ok(s),
+                        p => Err(p)
+                    }
+                }
+
+                fn into_protocol(this: Message<Self>) -> Protocol {
```
alekseysidorov: Name sounds a little bit unclear. Why not just to implement `From` conversion?

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,709 @@
...
+        T::into_message_from_parts(message, signed)
+    }
+    /// Checks buffer and return instance of `Protocol`.
+    pub fn verify_buffer(buffer: Vec<u8>) -> Result<Protocol, failure::Error> {
+        let signed = SignedMessage::verify_buffer(buffer)?;
+        Self::deserialize(signed)
+    }
+
+    /// Trying to convert instance of `Protocol` into `Message<RawTransaction>`.
+    pub fn try_into_transaction(self) -> Result<Message<RawTransaction>, failure::Error> {
```
alekseysidorov: Why not just to use `RawTransaction::try_from`?

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,709 @@
...
+    {
+        let set: TransactionSetPart = transaction.into();
+        let raw_tx = RawTransaction::new(service_id, set);
+        Self::concrete(raw_tx, public_key, secret_key)
+    }
+}
+
+impl Requests {
+    /// Returns public key of the message recipient.
+    pub fn to(&self) -> PublicKey {
```
alekseysidorov: In my opinion `receiver` is more clear.

exonum/src/messages2/authorization.rs
vldm: Thank you for your review, but could you clarify why i should avoid it? Neither https://github.com/exonum/exonum/blob/master/CONTRIBUTING.md
nor rustfmt code stype didn't tell anything about this.


And i prefer to keep uses as local as possible.

exonum/src/messages2/mod.rs
vldm: @vitvakatu `TransactionSet` is a set inside `blockchain` mod https://github.com/exonum/exonum/blob/master/exonum/src/blockchain/transaction.rs#L361

What did you mean that there is no `TransactionSet`?

@alekseysidorov  what your suggestions about this type?
This is a DTO of `TransactionSet` instance.

exonum/src/messages2/mod.rs
vldm: @vitvakatu the reason was unclear allocation rules of `vec!` macro. (If macro allocate exactly N bytes, then every push lead to reallocation)

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+        SignedMessage { raw: buffer }
+    }
+
+    #[cfg(not(test))]
+    pub(in messages) fn unchecked_from_vec(buffer: Vec<u8>) -> Self {
+        SignedMessage { raw: buffer }
+    }
```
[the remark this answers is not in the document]
vldm: @ilammy Thank you for your argumentation. I remove `#[cfg(not(test))]` one.

>I'd also mark the unchecked_*() methods as unsafe to make their users think more closely whether they need them.

Calling this methods didn't lead to memory unsafety

exonum/src/messages2/authorization.rs
aleksuss: Agree with @alekseysidorov. We can use `class` cause it's not reserved keyword.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,145 @@
...
+        let sign_idx = self.raw.len() - SIGNATURE_LENGTH;
+        Signature::from_slice(&self.raw[sign_idx..]).expect("Couldn't read signature")
+    }
+
+    /// Return byte array representation of internal data.
+    pub fn raw(&self) -> &[u8] {
+        &self.raw
+    }
+
+    pub fn hash(&self) -> Hash {
```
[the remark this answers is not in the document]
vldm: This type is private btw, but i add brief docs, ty x)

exonum/src/messages2/protocol.rs
vldm: Why it's not clear to you?
This is an internal trait that provide all kind function for being a part of `Protocol`.

Without it we should everywhere add bound

`fn bla<T: ProtocolMessage>(t:T)`
vs
`fn bla<T>(t:T)
where T:ProtocolMessage,
Message<T>: Into<Protocol>
`

which probably is not clear.

Additionally, `Into` can be implemented to multiple type params, this mean you should explicitly define to what type you want to convert.

```
let t: Message<X> = ...
foo(t.into());
```

exonum/src/messages2/protocol.rs
vldm: I am sure, this question It's out of scope of current design.

exonum/src/messages2/protocol.rs
vldm: It's probably implementation specific. And user shouldn't care about this.

exonum/src/messages2/protocol.rs
vldm: @ilammy 
>What exactly are the conditions for a panic? Is the user expected to provide some 'correct' arguments to avoid the panic, or is this just a warning that serialization may fail for example due to memory allocation issues?

Both, based on serialisation implementation. 
For example our `exonum_encoding` never return error. It will panic on OOM internaly.
Serde json for example can return error based on internal conditions.

Probably service creator shouldn't ever use this methods. But we should notice core developers who probably will use it, that they should provide unfailable `BinaryForm::serialize` implementation.

exonum/src/messages2/authorization.rs
alekseysidorov: But you know minimal expected size and thus can minimize reallocations here.

exonum/src/messages2/authorization.rs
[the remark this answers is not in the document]
ilammy: Strictly speaking, it will not lead to unsafe memory operations, but it may, for example, cause `author()` and other getters to panic if `SignedMessage` is constructed from an invalid buffer. `std::char::from_u32_unchecked()` illustrates the same principle: while passing it invalid arguments does not immediately lead to a crash or undefined behavior, doing so breaks the contract expected from `char`.

I do not insist on `unsafe`, `unchecked` is enough of a warning. Just adding my two cents and some historical precedents.

exonum/src/messages2/authorization.rs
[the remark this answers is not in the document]
DarkEld3r: I still think that creating an instance is the main purpose of this function. Additionally, it seems strange for me to emphasize the _safe_ behavior. I prefer to name the functions `from_buffer` and `from_buffer_unchecked` (if we need a constructor without validation) similar to the standard library (for example `from_utf8` and `from_utf8_unchecked`).

exonum/src/messages2/authorization.rs
DarkEld3r: @vldm 
Check `CONTRIBUTING.md` again. 🙃

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,152 @@
+use failure::Error;
+use hex::{FromHex, ToHex};
+
+use std::fmt;
+
+use crypto::{
+    self, hash, Hash, PublicKey, SecretKey, Signature, PUBLIC_KEY_LENGTH, SIGNATURE_LENGTH,
+};
+
```
DarkEld3r: Please remove this empty line.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,152 @@
...
+/// | `0..32`   | author's public key     |
+/// | `32`      | message class           |
+/// | `33`      | message type            |
+/// | `34..N`   | payload                 |
+/// | `N..N+64` | signature               |
+///
+/// `SignedMessage` will verify the size of the buffer and the signature provided in it.
+/// This allows to keep the raw message buffer, but avoid verifying its signature again
+/// as every `SignedMessage` instance is guaranteed to have a correct signature.
+
```
DarkEld3r: I suppose this empty line should be removed.

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,153 @@
...
+        &self.raw[PUBLIC_KEY_LENGTH + 2..sign_idx]
+    }
+
+    /// Returns ed25519 signature for this message.
+    pub(in messages) fn signature(&self) -> Signature {
+        let sign_idx = self.raw.len() - SIGNATURE_LENGTH;
+        Signature::from_slice(&self.raw[sign_idx..]).expect("Couldn't read signature")
+    }
+
+    /// Return byte array representation of internal data.
```
DarkEld3r: **Returns...**

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub fn raw(&self) -> &[u8] {
+        &self.raw
+    }
+
+    /// Calculates a hash of inner data.
+    pub fn hash(&self) -> Hash {
+        hash(&self.raw)
+    }
+
+    /// Signs buffer with secret_key.
```
DarkEld3r: ...with the `secret_key`,

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// This method returns ed25519 signature.
+    fn sign(full_buffer: &[u8], secret_key: &SecretKey) -> Result<Signature, Error> {
+        let signature = crypto::sign(&full_buffer, secret_key);
+        Ok(signature)
+    }
+
+    /// Verify buffer integrity, and authenticate buffer.
```
DarkEld3r: Verifie**s**...

Additionally comma is needless.

exonum/src/messages2/helpers.rs
```diff
@@ -0,0 +1,43 @@
...
+pub trait BinaryForm: Sized {
+    /// Converts transaction into serialized form.
+    fn serialize(&self) -> Result<Vec<u8>, Error>;
+
+    /// Converts a serialized byte array into a transaction.
+    fn deserialize(buffer: &[u8]) -> Result<Self, Error>;
+}
+
+/// Uses `ToHex`/`FromHex` to serialize arbitrary type `T` as
+/// hex string rather than real Serde::serialize.
```
DarkEld3r: hexadecimal string?

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /// Converts `ServiceTransaction` back to raw data.
+    pub fn into_raw_parts(self) -> (u16, Vec<u8>) {
+        (self.transaction_id, self.payload)
+    }
+}
+
+impl RawTransaction {
+    /// Creates new instance of RawTransaction.
```
DarkEld3r: ...a new instance of `RawTransaction`.

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+        service_id: u16,
+        transaction_set: ServiceTransaction,
+    ) -> RawTransaction {
+        RawTransaction {
+            service_id,
+            transaction_set,
+        }
+    }
+
+    /// Returns user defined data that should be used for deserialization.
```
DarkEld3r: Returns the user...

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn transaction_set(self) -> ServiceTransaction {
+        self.transaction_set
+    }
+
+    /// Returns service_id specified for current transaction.
```
DarkEld3r: Returns `service_id` specified for the current transaction.

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+        }
+        let transaction_id = LittleEndian::read_u16(&buffer[0..2]);
+        let payload = buffer[2..].to_vec();
+        Ok(ServiceTransaction {
+            transaction_id,
+            payload,
+        })
+    }
+}
+/// Wraps a `Payload` together with corresponding `SignedMessage`.
```
DarkEld3r: Please add an extra empty line before this line.

+ the corresponding

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+pub struct Message<T> {
+    //TODO: inner T duplicate data in SignedMessage, we can use owning_ref,
+    //if our serialization format allows us (ECR-2315).
+    payload: T,
+    #[serde(with = "HexStringRepresentation")]
+    message: SignedMessage,
+}
+
+impl<T: ProtocolMessage> Message<T> {
+    /// Creates new instance of the message.
```
DarkEld3r: a new instance

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn hash(&self) -> Hash {
+        hash(self.message.raw())
+    }
+
+    /// Returns a serialized buffer.
+    pub fn serialize(self) -> Vec<u8> {
+        self.message.raw
+    }
+
+    /// Return reference to payload.
```
DarkEld3r: ...to the payload.

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn payload(&self) -> &T {
+        &self.payload
+    }
+
+    /// Return reference to signed message.
```
DarkEld3r: ...to the signed...

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("Transaction")
+            .field("message_id", &self.transaction_id)
+            .field("payload_len", &self.payload.len())
+            .finish()
+    }
+}
+
+impl<T> ToHex for Message<T> {
+    fn write_hex<W: ::std::fmt::Write>(&self, w: &mut W) -> ::std::fmt::Result {
```
DarkEld3r:
```diff
- fn write_hex<W: ::std::fmt::Write>(&self, w: &mut W) -> ::std::fmt::Result
+ fn write_hex<W: fmt::Write>(&self, w: &mut W) -> fmt::Result
```

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+        self.message.raw().write_hex(w)
+    }
+
+    fn write_hex_upper<W: ::std::fmt::Write>(&self, w: &mut W) -> ::std::fmt::Result {
```
DarkEld3r: Same here.

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+impl<T: ProtocolMessage> StorageValue for Message<T> {
+    fn into_bytes(self) -> Vec<u8> {
+        self.message.raw
+    }
+
+    fn from_bytes(value: Cow<[u8]>) -> Self {
+        let message = SignedMessage::from_vec_unchecked(value.into_owned());
+        //TODO: Remove additional deserialization [ECR-2315]
```
DarkEld3r: Please add a dot at the end of the comment.

Additionally I would love to see space between "//` and "TODO".

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,693 @@
...
+use chrono::{DateTime, Utc};
+use failure;
+
+use std::{borrow::Cow, fmt::Debug, mem, net::SocketAddr};
+
+use blockchain;
+use crypto::{CryptoHash, Hash, PublicKey, SecretKey, PUBLIC_KEY_LENGTH, SIGNATURE_LENGTH};
+use helpers::{Height, Round, ValidatorId};
+use storage::{Database, MemoryDB, ProofListIndex, StorageValue};
+
```
DarkEld3r: Needless empty line.

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,693 @@
...
+use super::{BinaryForm, Message, RawTransaction, ServiceTransaction, SignedMessage};
+
+#[doc(hidden)]
+/// `SignedMessage` size with zero bytes payload.
```
DarkEld3r: Please move comment before the attribute (for consistency).

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,693 @@
...
+pub const EMPTY_SIGNED_MESSAGE_SIZE: usize =
+    PUBLIC_KEY_LENGTH + SIGNATURE_LENGTH + mem::size_of::<u8>() * 2;
+
+#[doc(hidden)]
+/// `Message<TransactionsResponse>` size without transactions inside.
```
DarkEld3r: Same here.

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,693 @@
...
+    pub fn concrete<T: ProtocolMessage>(
+        message: T,
+        author: PublicKey,
+        secret_key: &SecretKey,
+    ) -> Message<T> {
+        let value = message.serialize().expect("Couldn't serialize data.");
+        let (cls, typ) = T::message_type();
+        let signed = SignedMessage::new(cls, typ, value, author, secret_key);
+        T::into_message_from_parts(message, signed)
+    }
```
DarkEld3r: Please add an empty line after this line.

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,693 @@
...
+    /// Checks buffer and return instance of `Protocol`.
+    pub fn from_raw_buffer(buffer: Vec<u8>) -> Result<Protocol, failure::Error> {
+        let signed = SignedMessage::from_raw_buffer(buffer)?;
+        Self::deserialize(signed)
+    }
+
+    /// Creates new raw transaction message.
```
DarkEld3r: ...a new...

exonum/src/messages2/protocol.rs
```diff
@@ -0,0 +1,693 @@
...
+}
+
+impl StorageValue for Protocol {
+    fn into_bytes(self) -> Vec<u8> {
+        self.signed_message().raw().to_vec()
+    }
+
+    fn from_bytes(value: Cow<[u8]>) -> Self {
+        let message = SignedMessage::from_vec_unchecked(value.into_owned());
+        //TODO: Remove additional deserialization [ECR-2315]
```
DarkEld3r: Please add a dot (and " ").

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+        let transaction_id = LittleEndian::read_u16(&buffer[0..2]);
+        let payload = buffer[2..].to_vec();
+        Ok(ServiceTransaction {
+            transaction_id,
+            payload,
+        })
+    }
+}
+///
```
DarkEld3r: Please remove `///` here. :upside_down_face:

exonum/src/messages2/authorization.rs
```diff
@@ -0,0 +1,153 @@
...
+impl SignedMessage {
+    /// Creates `SignedMessage` from parts.
+    pub(crate) fn new(
+        class: u8,
+        tag: u8,
+        value: Vec<u8>,
+        author: PublicKey,
+        secret_key: &SecretKey,
+    ) -> SignedMessage {
+        let mut buffer = Vec::with_capacity(2 + value.len() + PUBLIC_KEY_LENGTH + SIGNATURE_LENGTH);
```
aleksuss: @vldm is it possible to get rid off the magic number ?

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+/// Transaction raw buffer.
+/// This struct is used to transfer transactions in network.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct RawTransaction {
+    service_id: u16,
+    transaction_set: ServiceTransaction,
+}
+
+/// Concrete raw transaction transaction inside `TransactionSet`.
```
dmitry-timofeev: What is `TransactionSet` — I don't see one in the PR?

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+/// This type used inner inside `transactions!` to transfer some set.
```
dmitry-timofeev: Could "to transfer some set" be clarified — it doesn't seem to make sense.

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        transaction_set: ServiceTransaction,
+    ) -> RawTransaction {
+        RawTransaction {
+            service_id,
+            transaction_set,
+        }
+    }
+
+    /// Returns the user defined data that should be used for deserialization.
+    pub fn transaction_set(self) -> ServiceTransaction {
```
dmitry-timofeev: What is `transaction_set` (a set of what?)?

exonum/src/messages2/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+/// Version of the protocol. Different versions are incompatible.
+pub const PROTOCOL_MAJOR_VERSION: u8 = 1;
+
+/// Transaction raw buffer.
+/// This struct is used to transfer transactions in network.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct RawTransaction {
+    service_id: u16,
+    transaction_set: ServiceTransaction,
```
dmitry-timofeev: Why can't we have 
```
struct RawTransaction:
  service_id : u16
  tx_id: u16
  payload: bytes[]
```
?

exonum/src/messages2/mod.rs
vldm: it's a trait outside of pr, located in `blockchain/transaction.rs`

exonum/src/messages2/helpers.rs
```diff
@@ -0,0 +1,43 @@
...
+use serde::{de, ser, Deserialize, Deserializer, Serialize, Serializer};
+
+use std::fmt::Display;
+
+use encoding::Error;
+
+/// Helper trait to define serialization format.
+pub trait BinaryForm: Sized {
+    /// Converts transaction into serialized form.
+    fn serialize(&self) -> Result<Vec<u8>, Error>;
```
alekseysidorov: `serialize` can be confused with the `serde::serialize`, in my opinion `encode` and `decode` or `from_bytes` and `to_bytes` are more convenient.

exonum/src/messages2/authorization.rs
DarkEld3r: Personally, I prefer to see something like that:
```rust
let mut buffer = Vec::with_capacity(mem::size_of_val(&class) + mem::size_of_val(&tag) + PUBLIC_KEY_LENGTH + SIGNATURE_LENGTH);
```

exonum/src/messages2/mod.rs
dmitry-timofeev: So, given I have a `service_id` of a tx receiver, would an API to submit transaction look like this:

```
submitTransaction(receiver_id: u16, tx_parameters: ServiceTransaction) -> Hash
```

PR #917, opened 2018-08-29
Add EOF handling in MessageCodec [ECR-2240]

pavel-mukhanov:
Added `decode_eof` method to handle EOF in `MessageCodec`.


exonum/src/events/codec.rs
```diff
@@ -80,13 +79,30 @@ impl Decoder for MessagesCodec {
...
+        let raw = RawMessage::new(MessageBuffer::from_vec(data));
+        Ok(Some(raw))
+    }
+
+    fn decode_eof(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if let Some(frame) = try!(self.decode(buf)) {
+            Ok(Some(frame))
+        } else {
+            if !buf.is_empty() {
+                trace!("Bytes remaining in buffer after receiving EOF.")
```
ilammy: It would be nice to log the number of bytes remaining as it may sometimes give useful insights during debugging. E.g., when there is 1 byte left then one can suspect a trailing null byte or something, when there is a whole megabyte left in the buffer then something is definitely going wrong here.

exonum/src/events/codec.rs
```diff
@@ -147,6 +163,39 @@ mod test {
...
+    #[test]
+    #[should_panic(expected = "Received malicious message with wrong total_len")]
+    fn decode_message_wrong_length() {
+        let data = vec![0u8, 0, 0, 0, 0, 0, 32, 0, 0, 0];
+        let mut bytes: BytesMut = BytesMut::new();
+        let (ref mut responder, ref mut initiator) = create_encrypted_codecs();
+        let raw = RawMessage::new(MessageBuffer::from_vec(data.clone()));
+        initiator.encode(raw, &mut bytes).unwrap();
+
+        responder.decode(&mut bytes).unwrap();
```
ilammy: I'd rearrange the code blocks a bit to make the test flow easier to follow:

```rust
let (ref mut responder, ref mut initiator) = create_encrypted_codecs();

let data = vec![0u8, 0, 0, 0, 0, 0, 32, 0, 0, 0];
let raw = RawMessage::new(MessageBuffer::from_vec(data));

let mut bytes: BytesMut = BytesMut::new();
initiator.encode(raw, &mut bytes).expect("encoding message");

responder.decode(&mut bytes).expect("decoding message");
```

There are people who are not fans of empty lines and promote terser code (this is only my personal preference), but I think that at least variable declarations could be regrouped and moved closer to their usage sites.

exonum/src/events/codec.rs
```diff
@@ -147,6 +163,39 @@ mod test {
...
+        let data = vec![0u8, 0, 0, 0, 0, 0, 10, 0, 0, 0];
+        let mut bytes: BytesMut = BytesMut::new();
+        let (ref mut responder, ref mut initiator) = create_encrypted_codecs();
+        let raw = RawMessage::new(MessageBuffer::from_vec(data.clone()));
+        initiator.encode(raw, &mut bytes).unwrap();
+
+        match responder.decode_eof(&mut bytes.clone()) {
+            Ok(Some(ref r)) if r == &RawMessage::new(MessageBuffer::from_vec(data.clone())) => {}
+            _ => panic!("Wrong input"),
+        };
```
ilammy: I'd suggest to split this into a separate unit-test for the 'happy path' of `decode_eof()` and validate only EOF condition handling in this test.

exonum/src/events/codec.rs
```diff
@@ -80,13 +79,30 @@ impl Decoder for MessagesCodec {
...
+            );
+        }
+
+        let data = buf.split_to(total_len).to_vec();
+        let raw = RawMessage::new(MessageBuffer::from_vec(data));
+        Ok(Some(raw))
+    }
+
+    fn decode_eof(&mut self, buf: &mut BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if let Some(frame) = try!(self.decode(buf)) {
```
slowli: Nit: Judging by [docs](https://doc.rust-lang.org/std/macro.try.html), using `?` operator instead of `try!` is preferred.

Also, I personally would find the following implementation more obvious:

```rust
let maybe_message = self.decode(buf)?;
if maybe_message.is_none() && !buf.is_empty() {
    trace!(...);
}
maybe_message
```

exonum/src/events/codec.rs
```diff
@@ -80,13 +79,30 @@ impl Decoder for MessagesCodec {
...
+            Ok(Some(frame))
+        } else {
+            if !buf.is_empty() {
```
slowli: Would it make sense to empty the buffer here, or is it expected to be dropped soon anyway?

exonum/src/events/codec.rs
```diff
@@ -147,6 +163,39 @@ mod test {
...
+    #[test]
+    fn decode_message_eof() {
+        let data = vec![0u8, 0, 0, 0, 0, 0, 10, 0, 0, 0];
+        let mut bytes: BytesMut = BytesMut::new();
+        let (ref mut responder, ref mut initiator) = create_encrypted_codecs();
+        let raw = RawMessage::new(MessageBuffer::from_vec(data.clone()));
+        initiator.encode(raw, &mut bytes).unwrap();
+
+        match responder.decode_eof(&mut bytes.clone()) {
+            Ok(Some(ref r)) if r == &RawMessage::new(MessageBuffer::from_vec(data.clone())) => {}
```
slowli: Nit: If I understand correctly, you can use `AsRef<[u8]>` implementation for `RawMessage` here:

```rust
Ok(Some(ref message)) if *message.as_ref() == data => {}
```

exonum/src/events/codec.rs
```diff
@@ -147,6 +163,39 @@ mod test {
...
+            _ => panic!("Wrong input"),
+        };
+
+        // Emulate EOF behaviour.
+        bytes.truncate(1);
+        assert!(responder.decode_eof(&mut bytes).unwrap().is_none());
+
+        bytes.clear();
+        assert!(responder.decode_eof(&mut bytes).unwrap().is_none());
+    }
```
slowli: Nit: It could be interesting to test a case with several messages in the buffer when the EOF is received, although it seems to be unnecessary from the code coverage perspective.

exonum/src/events/codec.rs
```diff
@@ -80,13 +79,30 @@ impl Decoder for MessagesCodec {
             );
         }
 
-        // Read message
-        if buf.len() >= total_len {
-            let data = buf.split_to(total_len).to_vec();
-            let raw = RawMessage::new(MessageBuffer::from_vec(data));
-            return Ok(Some(raw));
+        if total_len > buf.len() {
```
slowli: This has been bothering me for a while: what if `total_len < buf.len()`, i.e., the message is undersized compared to its Noise wrapper? Do we accept such framing?

exonum/src/events/codec.rs
```diff
@@ -147,6 +163,39 @@ mod test {
...
+        let (ref mut responder, ref mut initiator) = create_encrypted_codecs();
+        let raw = RawMessage::new(MessageBuffer::from_vec(data.clone()));
+        initiator.encode(raw, &mut bytes).unwrap();
+
+        match responder.decode_eof(&mut bytes.clone()) {
+            Ok(Some(ref r)) if r == &RawMessage::new(MessageBuffer::from_vec(data.clone())) => {}
+            _ => panic!("Wrong input"),
+        };
+
+        // Emulate EOF behaviour.
```
slowli: Nit: "behaviour" is British spelling, the US one is "behavior".

exonum/src/events/codec.rs
pavel-mukhanov: After fixing `decrypted_msg_len` calculation we obtain the exact same message that we encrypted(without trailing zeros), so I think we can accept the only messages with total_len == buf.len().

exonum/src/events/codec.rs
pavel-mukhanov: I don't think it necessary, because `FramedRead` will be calling `decode_eof` until we reach partial message that we cannot decode.

exonum/src/events/codec.rs
pavel-mukhanov: Yes, it will be dropped.

exonum/src/events/codec.rs
pavel-mukhanov: But I slightly modified the test and now it covers multiple messages case.

exonum/src/events/codec.rs
pavel-mukhanov: I've rewritten the test, now it covers case with 2 messages left in the buffer.

PR #918, opened 2018-08-30
Add all_entries_unchecked method [ECR-1959]

vitvakatu:
Tiny but useful change.

# Why?

At EJB, we need to get raw proof data as fast as possible for sending it to the Java side.

# Is it safe from user perspective?

I'm not sure. One of possible solutions would be marking this as `doc(hidden)` or even behind feature gate.
I need to hear your opinions.

UPD: Also includes change from `Vec` to `impl Iterator` for most public methods in `MapProof` API.

### DIFF
--- a/exonum/src/storage/proof_map_index/proof.rs
+++ b/exonum/src/storage/proof_map_index/proof.rs
@@ -466,12 +466,10 @@ impl<K, V> MapProof<K, V> {
     /// Existing entries have `Some` value, non-existing have `None`.
     /// This method does not perform any integrity checks of the proof.
     pub fn all_entries_unchecked(&self) -> impl Iterator<Item = (&K, Option<&V>)> {
-        self.entries
-            .iter()
-            .map(|e| match e {
-                OptionalEntry::Missing { ref missing } => (missing, None),
-                OptionalEntry::KV { ref key, ref value } => (key, Some(value)),
-            })
+        self.entries.iter().map(|e| match e {
+            OptionalEntry::Missing { ref missing } => (missing, None),
+            OptionalEntry::KV { ref key, ref value } => (key, Some(value)),
+        })
     }
 }
 
@@ -599,30 +597,24 @@ where
 impl<K, V> CheckedMapProof<K, V> {
     /// Retrieves references to keys that the proof shows as missing from the map.
     pub fn missing_keys(&self) -> impl Iterator<Item = &K> {
-        self.entries
-            .iter()
-            .filter_map(|kv| match *kv {
-                (ref key, None) => Some(key),
-                _ => None,
-            })
+        self.entries.iter().filter_map(|kv| match *kv {
+            (ref key, None) => Some(key),
+            _ => None,
+        })
     }
 
     /// Retrieves references to key-value pairs that the proof shows as present in the map.
     pub fn entries(&self) -> impl Iterator<Item = (&K, &V)> {
-        self.entries
-            .iter()
-            .filter_map(|kv| match *kv {
-                (ref key, Some(ref value)) => Some((key, value)),
-                _ => None,
-            })
+        self.entries.iter().filter_map(|kv| match *kv {
+            (ref key, Some(ref value)) => Some((key, value)),
+            _ => None,
+        })
     }
 
     /// Retrieves references to existing and non-existing entries in the proof.
     /// Existing entries have `Some` value, non-existing have `None`.
     pub fn all_entries(&self) -> impl Iterator<Item = (&K, Option<&V>)> {
-        self.entries
-            .iter()
-            .map(|&(ref k, ref v)| (k, v.as_ref()))
+        self.entries.iter().map(|&(ref k, ref v)| (k, v.as_ref()))
     }
 
     /// Returns a hash of the map that this proof is constructed for.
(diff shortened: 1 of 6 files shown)


ilammy:
> Is it safe from user perspective?

I'd look at historical data. Are there any precedents of misuse of the `*_unchecked()` API? If there are then maybe we should look into a generalized solution for that problem. Otherwise I deem it acceptable to clearly communicate to the users that these functions do not perform some of the expected validations, and expect the developers to be intelligent enough to not misuse them. The functions follow the convention of `*_unchecked` suffix and have a corresponding line in their documentation comments. Maybe we could also label them `unsafe` for additional inconvenience and visibility, but that's as far as I personally would go.

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -460,6 +460,20 @@ impl<K, V> MapProof<K, V> {
     pub fn missing_keys_unchecked(&self) -> Vec<&K> {
         self.entries.iter().filter_map(|e| e.as_missing()).collect()
     }
+
+    /// Retrieves references to existing and non-existing entries in the proof.
+    ///
+    /// Existing entries have `Some` value, non-existing have `None`.
+    /// This method does not perform any integrity checks of the proof.
+    pub fn all_entries_unchecked(&self) -> Vec<(&K, Option<&V>)> {
```
slowli: I wonder if returning `impl Iterator<Item = (&K, Option<&V>)>` could cover more use cases. When other `unchecked` methods were written, `impl Trait` wasn't in stable Rust yet, so they used `Vec<_>` out of necessity. Also, it could make `missing_keys_unchecked()` method easily implementable as a filter:

```rust
pub fn missing_keys_unchecked(&self) -> impl Iterator<Item = &K> {
    self.all_entries_unchecked().filter_map(|pair| match pair {
        (key, None) => Some(key),
        _ => None,
    })
}
```

exonum/src/storage/proof_map_index/proof.rs
```diff
@@ -460,6 +460,20 @@ impl<K, V> MapProof<K, V> {
...
+        self.entries
+            .iter()
+            .map(|e| match e {
+                OptionalEntry::Missing { ref missing } => (missing, None),
+                &OptionalEntry::KV { ref key, ref value } => (key, Some(value)),
```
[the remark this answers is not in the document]
vitvakatu: Yep, totally missed that one!

vitvakatu:
Additional question - should I mention that in Changelog?

exonum/src/storage/proof_map_index/proof.rs
vitvakatu: Seems reasonable to me, but we have some issues with supporting latest Rust version in EJB. It will be fixed before we'll use Exonum 0.10 (where this PR will be landed), but right now I wouldn't be able to test the functionality. So I'll check it and maybe fix it.

aleksuss:
@vitvakatu A public API was chnged so I suppose YES
CC @DarkEld3r

exonum/src/storage/proof_map_index/proof.rs
slowli: Sure. Just to clarify things, `impl Trait` [was added in Rust 1.26](https://blog.rust-lang.org/2018/05/10/Rust-1.26.html), which seems to be the version used by EJB, at least judging [by the Travis config](https://github.com/exonum/exonum-java-binding/blob/5df785dcaae87d7fa9c4ab9832b9218dc41e39d8/.travis.yml#L20).

DarkEld3r:
I also think that this should be mentioned in the Changelog (in the "new features" section).

vitvakatu:
Changelog updated, API changed, tests might be falling at the moment.

exonum/src/storage/proof_map_index/tests.rs
```diff
@@ -384,7 +384,7 @@ fn check_map_multiproof<K, V>(
     let proof = proof.check().unwrap();
     assert_eq!(proof.merkle_root(), table.merkle_root());
     assert_eq!(missing_keys.iter().collect::<Vec<&_>>(), {
-        let mut actual_keys = proof.missing_keys();
+        let mut actual_keys = proof.missing_keys().collect::<Vec<_>>();
```
slowli: Partially off-topic: The code of `check_map_proof` is overly complicated, it could be simplified as

```rust
fn check_map_multiproof<K, V>(
    proof: MapProof<K, V>,
    keys: Vec<K>,
    table: &ProofMapIndex<&mut Fork, K, V>,
) where
    K: ProofMapKey + Eq + Ord + Debug,
    V: StorageValue + Eq + Ord + Debug,
{
    let mut entries: Vec<(K, V)> = Vec::new();
    let mut missing_keys: Vec<K> = Vec::new();

    for key in keys {
        if table.contains(&key) {
            let value = table.get(&key).unwrap();
            entries.push((key, value));
        } else {
            missing_keys.push(key);
        }
    }

    let proof = proof.check().unwrap();
    assert_eq!(proof.merkle_root(), table.merkle_root());
    assert_eq!(
        missing_keys.iter().collect::<BTreeSet<_>>(),
        proof.missing_keys().collect::<BTreeSet<_>>()
    );
    assert_eq!(
        entries
            .iter()
            .map(|&(ref k, ref v)| (k, v))
            .collect::<BTreeSet<_>>(),
        proof.entries().collect::<BTreeSet<_>>()
    );
}
```

vitvakatu: Not going to fix that in this PR. I've created ECR-2294.

PR #920, opened 2018-08-31
Refactor sandbox message creation.

vldm:
Abstract user from the process of creating messages. 
This change encapsulate message creating in helpers methods ( `create_x`). This allows implementing future refactors more fluently, for example
1. we can derive keys from `sandbox` structure to provide better user interface (`create_propose(validator_id...)` vs `create_propose(validator_id, pk, ....)`)
2. we can change message creation process and refactor only those methods (usefull for message header refactor, and serialisation)


mvlabat:
Am I alone here who doesn't see any positive effect, but just redundant methods in the Sandbox? (

PR #921, opened 2018-09-03
Fix websockets [ECR-2242]

mvlabat:
There are no tests yet :( We should work on test infrastructure to test websockets.

### DIFF
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -13,16 +13,16 @@
 // limitations under the License.
 
 //! API and corresponding utilities.
-
-pub use self::error::Error;
-pub use self::state::ServiceApiState;
-pub use self::with::{FutureResult, Immutable, Mutable, NamedWith, Result, With};
+pub use self::{
+    error::Error, state::ServiceApiState,
+    with::{FutureResult, Immutable, Mutable, NamedWith, Result, With},
+};
 
 use serde::{de::DeserializeOwned, Serialize};
 
 use std::{collections::BTreeMap, fmt};
 
-use self::backends::actix;
+use self::{backends::actix, node::public::ExplorerApi};
 use blockchain::{Blockchain, SharedNodeState};
 
 pub mod backends;
@@ -336,8 +336,6 @@ impl ApiAggregator {
         blockchain: &Blockchain,
         shared_node_state: SharedNodeState,
     ) -> ServiceApiBuilder {
-        use self::node::public::ExplorerApi;
-
         let mut builder = ServiceApiBuilder::new();
         let service_api_state = ServiceApiState::new(blockchain.clone());
         ExplorerApi::wire(builder.public_scope(), service_api_state, shared_node_state);
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -288,7 +288,7 @@ impl ApiAggregator {
         );
         inner.insert(
             "explorer".to_owned(),
-            Self::explorer_api(node_state.clone()),
+            Self::explorer_api(&blockchain, node_state.clone()),
         );
         // Adds services APIs.
         inner.extend(blockchain.service_map().iter().map(|(_, service)| {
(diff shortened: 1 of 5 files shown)


exonum/src/api/websocket.rs
```diff
@@ -86,8 +91,12 @@ impl Handler<Broadcast> for Server {
     type Result = ();
 
     fn handle(&mut self, Broadcast { block_hash }: Broadcast, _ctx: &mut Self::Context) {
+        let snapshot = self.service_api_state.snapshot();
+        let schema = Schema::new(snapshot);
+        let block_header = schema.blocks().get(&block_hash);
+        let block_header_json = serde_json::to_value(block_header).unwrap().to_string();
         for address in self.subscribers.values() {
-            let _ = address.do_send(Message(format!("Committed new block {:?}", block_hash)));
+            let _ = address.do_send(Message(block_header_json.clone()));
```
aleksuss: `.clone()` is confusing me a little. There is really no possibilities to get rid of it ?

exonum/src/node/consensus.rs
```diff
@@ -500,6 +498,9 @@ impl NodeHandler {
                 .new_height(&block_hash, self.system_state.current_time());
             (block_state.txs().len(), block_state.proposer_id())
         };
+
+        self.api_state.broadcast(&block_hash);
```
aleksuss: Just to clarify. Do I understand right that `.broadcast` executes asynchronously ?

mvlabat: That's right

exonum/src/api/websocket.rs
mvlabat: I'm not sure that's possible, as I can't move it on every loop iteration

andreyluchin:
Have a few things to address regarding websocket which do not relate to this PR (which does what it is supposed to do well).

The idea of introducing WebSocket-based communication to Exonum was born due to a very specific request from one of the internal teams to have mechanism to notify a client on block commit. At that point it was not clear how necessary would WebSocket communication be, so we went with minimal and easiest implementation to fill relevant requirement (notification on block commit).

Right now, I believe, it has become clearer that WebSocket communication with the client is necessary. This is why I ask the core team to consider the following idea along with use-cases I encountered in the wild:

## Extend Exonum Api to work with WebSockets
### Case 1: Blockchain Explorer
Front-end developers for Exonum services often need a notification mechanism for some fundamental consensus events. I would specifically emphasize on the following:
- Block Commit. - After subscribing client is notified on each block commit.
- Transaction Commit. - After supplying transaction hash and subscribing client is notified when transaction with corresponding hash is committed to the blockchain.
### Case 2: Websockets in Service API
Consider the case of a somewhat standard **Land Registry**. Database has thousands of `Land` objects each representing a land plot in real life. The front-end needs to load **all** land objects and display them as they are downloaded via API. With current Exonum capabilities, the situation would have to be solved by implementing an endpoint to retrieve every land plot.  This has a number of disadvantages:
- It is a blocking API call (takes a lot of time to send thousands of objects via single HTTP Response).
- It is **impossible** to render plots on the front-end dynamically, it is necessary to wait for all objects to download.

 The resolution to such problem is first to divide the list of objects into smaller chunks and send each chunk separately, but introduces a number of problems too:

- It is unclear how to divide the data, especially if it stored in an unordered `MapIndex`.
- The size of each chunk is often chosen randomly as it is hard to estimate the best one.
- Front-end logic becomes more complicated.
- Often necessary to create *helper* indexes to support data division.

This is a situation where WebSockets would shine.

- Would allow to establish an asynchronous (in contrast to HTTP) communication channel.
- Front-end c logic is straight forward (WebSocket is [supported](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications) by JS out of the box).
- Allows for dynamic rendering on each object downloaded.

exonum/src/api/mod.rs
```diff
@@ -332,9 +332,15 @@ impl ApiAggregator {
         self.inner.insert(prefix.into(), builder);
     }
 
-    fn explorer_api(shared_node_state: SharedNodeState) -> ServiceApiBuilder {
+    fn explorer_api(
+        blockchain: &Blockchain,
+        shared_node_state: SharedNodeState,
+    ) -> ServiceApiBuilder {
+        use self::node::public::ExplorerApi;
```
vitvakatu: We typically have all our imports at the top of the file, don't we?
cc @DarkEld3r

DarkEld3r: We don't have a strict rule about that, but I prefer imports at the top of the file because it is more manageable in terms of refactoring.

exonum/src/node/consensus.rs
vldm: I better prefer `notify` rather than `broadcast`

And why we didn't sending some "block info" instead of `block hash`, that's how we can remove `blockchain` dependency in `Actor` and make it more independent.

PR #925, opened 2018-09-04
Add screenshots to demos

boguslavsky:


aleksuss:
@boguslavsky could you please explain why ? Just for the sake of curiosity 😄

PR #926, opened 2018-09-04
Some refactoring to make network.rs more clear

pavel-mukhanov:
Moved `and_then` closures to separate messages.

### DIFF
--- a/exonum/src/events/network.rs
+++ b/exonum/src/events/network.rs
@@ -230,7 +230,9 @@ impl NetworkPart {
         handshake_params: &HandshakeParams,
     ) -> Box<dyn Future<Item = (), Error = failure::Error>> {
         let network_config = self.network_config;
-        // Cancellation token
+        // `cancel_sender` is converted to future when we receive
+        // `NetworkRequest::Shutdown` causing its being completed with error.
+        // After that completes `cancel_handler` and event loop stopped.
         let (cancel_sender, cancel_handler) = unsync::oneshot::channel();
 
         let request_handler = RequestHandler::from(
@@ -435,7 +437,7 @@ impl<'a> Listener<'a> {
         listener
             .incoming()
             .for_each(move |(sock, address)| {
-                let holder = Rc::downgrade(&incoming_connections_counter);
+                let holder = Rc::clone(&incoming_connections_counter);
                 // Check incoming connections count
                 let connections_count = Rc::weak_count(&incoming_connections_counter);
                 info!("connections_count {}", connections_count);
--- a/exonum/src/events/network.rs
+++ b/exonum/src/events/network.rs
@@ -444,7 +444,6 @@ impl Listener {
                 trace!("Received handshake message={:?}", connect);
                 Self::process_incoming_messages(stream, network_tx, connect, address)
             })
-
     }
 
     fn parse_connect_message(raw: Option<RawMessage>) -> Result<Connect, failure::Error> {
--- a/exonum/src/events/network.rs
+++ b/exonum/src/events/network.rs
@@ -25,7 +25,9 @@ use tokio_retry::{
     strategy::{jitter, FixedInterval}, Retry,
 };
 
-use std::{cell::RefCell, collections::HashMap, net::SocketAddr, rc::{Rc, Weak}, time::Duration};
+use std::{
+    cell::RefCell, collections::HashMap, net::SocketAddr, rc::{Rc, Weak}, time::Duration,
+};
 
 use super::{
     error::{log_error, result_ok}, to_box,
@@ -319,7 +321,7 @@ impl RequestHandler {
         RequestHandler(to_box(requests_handler))
     }
 
-    #[cfg_attr(feature="cargo-clippy", allow(too_many_arguments))]
+    #[cfg_attr(feature = "cargo-clippy", allow(too_many_arguments))]
     fn handle_send_message(
         peer: SocketAddr,
         msg: RawMessage,
--- a/exonum/src/events/network.rs
+++ b/exonum/src/events/network.rs
@@ -26,7 +26,7 @@ use tokio_retry::{
 };
 
 use std::{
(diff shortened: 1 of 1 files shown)


exonum/src/events/network.rs
```diff
@@ -425,7 +437,28 @@ impl Listener {
...
+            .into_future()
+            .map_err(|e| e.0)
+            .and_then(|(raw, stream)| (Self::parse_connect_message(raw), Ok(stream)))
+            .and_then(move |(connect, stream)| {
+                trace!("Received handshake message={:?}", connect);
+                Self::process_incoming_messages(stream, network_tx, connect, address)
+            })
+            .map(|_| {
+                // Ensure that holder lives until the stream ends.
+                let _holder = holder;
```
ilammy: By the way, does this actually do anything useful? I don't understand this part.

`holder` seems to be a weak pointer. What's the point of copying it into the closure owned by the future if weak pointers do not extend the lifetime of the owned object? Furthermore, it's `Weak<()>`, a weak pointer to the unit value.

exonum/src/events/network.rs
```diff
@@ -148,48 +150,31 @@ impl ConnectionsPool {
...
-                    network_config.tcp_keep_alive.map(Duration::from_millis);
-                sock.set_keepalive(duration)?;
-                Ok(sock)
-            })
-            .and_then(move |sock| {
-                Self::build_handshake_initiator(sock, &peer, &handshake_params)
-            })
-            // Connect socket with the outgoing channel
+            .and_then(move |socket| Self::configure_socket(socket, network_config))
+            .and_then(move |sock| Self::build_handshake_initiator(sock, &peer, &handshake_params))
```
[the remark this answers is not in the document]
DarkEld3r: I prefer `socket` by the way.

exonum/src/events/network.rs
```diff
@@ -346,6 +320,59 @@ impl RequestHandler {
             });
         RequestHandler(to_box(requests_handler))
     }
+
+    #[cfg_attr(feature = "cargo-clippy", allow(too_many_arguments))]
```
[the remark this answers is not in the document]
pavel-mukhanov: It's kind of chicken-egg problem to create `RequestHandler` we need to construct future, to construct future we need to pass these parameters into it.
```rust
struct RequestHandler(
    Box<dyn Future<Item = (), Error = failure::Error>>,
);

fn new(...) -> Self {
   let requests_handler = receiver
      .for_each(...)
   ...
   RequestHandler(to_box(requests_handler))
}
```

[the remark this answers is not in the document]
ilammy: I was thinking of doing a hack like this:

```rust
struct RequestHandler {
    foo: Bar,
    future: Option<Box<dyn Future<Item = (), Error = failure::Error>>>,
}

impl RequestHandlder {
    pub fn new(foo: Bar) -> Self {
        let mut handler = RequestHandler {
            foo,
            future: None,
        };
        handler.future = Some(to_box(handler.make_future()));
        handler
    }

    fn make_future(&self) -> impl Future<Item = (), Error = failure::Error> {
        // Here we can conveniently use `self` and don't have `to_box()` everything.
    }
}
```

but I wonder if it actually makes the code more readable.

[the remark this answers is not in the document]
pavel-mukhanov: Yeah, I used this approach, but removed `future` from fields.

exonum/src/events/network.rs
ilammy: Oh, I just noticed the comment explaining the refcount abuse. Yes, using an explicit integer counter is way easier to comprehend than this weak_count thing.

exonum/src/events/network.rs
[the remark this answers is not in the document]
ilammy: Nicely done!

exonum/src/events/network.rs
pavel-mukhanov: I forgot to decrement connections_counter, therefore integer counter is not enough, and I need something like AtomicUsize.

pavel-mukhanov: It seems that the solution with `Rc` is good because it decrements automatically after the connection is gone. And `holder` ensures that decrement doesn't happen earlier.

ilammy: Actually, the counter does not have to be atomic. Futures are handled on a single thread, and `Rc` uses non-atomic counter as well. The counter can be explicitly decremented when the future completes with something like this:

```diff
-                    .map(|_| {
-                        // Ensure that holder lives until the stream ends.
-                        let _holder = holder;
-                    })
+                    .then(move |result| {
+                        incoming_connections_count -= 1;
+                        result
+                    })
```

Though, you do have a point in that `Rc` does this (semi-)automatically and will probably behave better in case of a panic. But we still need to do this weird dance with making sure that `holder` is copied into the closure and kept around while the future is in progress.

Well, I just found it weird to use `Rc` here. I don't have anything against it.

pavel-mukhanov: Hm, let's try your approach.

ilammy: Though, I have some doubt for whether using a simple integer will suffice. I'd suppose that it will be moved... _somewhere_ alongside the future, but maybe the compiler will wiggle out and copy the counter which will be incorrect: there will be a copy for `new()` with zero in it, a copy (of zero) for the `ForEach` closure where it is incremented, and another copy of zero for the `Then` closure which will be decremented.

In that case `Rc` does not seem to be a bad choice. After all, `Rc` is used to denote shared ownership and we kinda share the counter between connections. Though, something like `RefCell` may be more semantically correct.

Also, we do not seem to have any tests which verify handling of `max_connections`. This has to be checked manually. If the new approach does not work then I guess we can leave the `Rc` as it, in the hope that it works :) It was there before, after all.

pavel-mukhanov: I tested manually, integer counter works incorrectly, so I reverted `Rc` again and I think we should stop at this point :sweat: Maybe I return to this issue in the next PR.

PR #929, opened 2018-09-06
Added a possibility to create `ServiceApiBuilder` with blockchain

aleksuss:
The improvement allows to update Exonum version > 0.8 in EJB.

### DIFF
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -263,7 +263,8 @@ impl ServiceApiBuilder {
 
     /// Returns an optional value to the PublicKey.
     pub fn public_key(&self) -> Option<PublicKey> {
-        self.blockchain().map(|blockchain| blockchain.service_keypair.0)
+        self.blockchain()
+            .map(|blockchain| blockchain.service_keypair.0)
     }
 }
 
--- a/exonum/src/api/mod.rs
+++ b/exonum/src/api/mod.rs
@@ -206,7 +206,7 @@ impl ServiceApiScope {
 ///     }
 /// }
 ///
-/// # let mut builder = ServiceApiBuilder::default();
+/// # let mut builder = ServiceApiBuilder::new();
 /// // Adds `MyApi` handlers to the corresponding builder.
 /// builder.public_scope()
 ///     .endpoint("v1/ping", MyApi::ping)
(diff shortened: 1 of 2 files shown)


exonum/src/api/mod.rs
```diff
@@ -218,14 +220,25 @@ impl ServiceApiScope {
 /// ```
 #[derive(Debug, Clone, Default)]
 pub struct ServiceApiBuilder {
+    blockchain: Option<Blockchain>,
```
vitvakatu: Any specific reason for `Option` here? Why not storing just `Blockchain`? (with `ServiceApiBuilder::new(b: Blockchain)`)

aleksuss: @vitvakatu Yes. The reason is to leave a method `new` as is.

exonum/src/api/mod.rs
```diff
@@ -237,6 +250,22 @@ impl ServiceApiBuilder {
...
+    /// Returns an optional reference to the ApiSender.
+    pub fn api_sender(&self) -> Option<&ApiSender> {
+        self.blockchain().map(|blockchain| &blockchain.api_sender)
+    }
+
+    /// Returns an optional value to the PublicKey.
+    pub fn public_key(&self) -> Option<PublicKey> {
+        self.blockchain()
+            .map(|blockchain| blockchain.service_keypair.0)
+    }
```
ilammy: I'd rephrase the documentation comments to explain when `Some` values are expected from these getters. For example, "Returns service public key of the blockchain if blockchain is set" or something like that. It's obvious from return types that the getters return optional values. Developers are more interested when they can do something with these values.

exonum/src/api/mod.rs
vitvakatu: Ah okay.
@DarkEld3r I think the we might need patch release from 0.9 with this change. It's not a requirement, but it would be much simpler and cleaner solution for EJB team. Is it possible? Otherwise we need to either use git dependency, or wait till 0.10.

exonum/src/api/mod.rs
vldm: @ilammy as for me state: "if blockchain is set" isn't more clear.

Probably we should add info, that this `Option` is safe to unwrap inside `wire` method?

aleksuss: @vldm I suppose it can be safe if `builder.blockchain().is_some()`

exonum/src/api/mod.rs
DarkEld3r: @vitvakatu  It is definitely possible. :+1:

PR #935, opened 2018-09-09
Add rules about imports

DarkEld3r:


DarkEld3r:
Fixed. 🙃

vldm:
This changes looks like unreasoned wishlist, there no clear rule, WHY you should avoid somethng.

When you introduce some rule, this rule should be accepted by all team members or atleast be argumented somehow.

ilammy:
Well, it does look like an abuse :thinking:

My personal preference is to use the unqualified names whenever possible. This lowers the amount of nesting `::` in the code which makes searching for definitions easier. When you see an unfamiliar `foo` in the code then you can simply Ctrl+F for `foo` in the file and you'll either find foo's local definition in this file, or an import statement for it. Using nested names complicates this process as you have no idea whether some module is external, or is another module defined in this file, what part of the name you should select, etc.

There are some exceptions, mostly for generic-looking names (e.g, `init()` functions, `Error` types). In cases like these I sometimes prefer to import the most specific module and use a single nesting level to disambiguate (e.g., `io::Error`, `grpc::init()`). This is as close to the ideal from the previous paragraph as we can get. An alternative would be to import with renaming (i.e, `use std::io::Error as IoError`) if this form is more common historically or leads to shorter code.

Some more insights can be found [in this extensive discussion](https://github.com/rust-lang-nursery/fmt-rfcs/issues/24).

PR #937, opened 2018-09-10
adding codecov.io coverage (ECR-758)

atrykush:
Adding codecov.io coverage, according to ECR-758


atrykush:
Reports are being generated

PR #941, opened 2018-09-12
Fix and improve documentation checks

ilammy:
This should fix Travis builds which are occasionally failing since #876 has been merged.

We fix the broken documentation links, make sure the problem will be noticed when we face it again, and do some cleanup of old workarounds. See commit messages for details.


aleksuss:
@ilammy Good job 👍

PR #942, opened 2018-09-13
Resolve domain name on connect. [ECR-2310]

dvush:
Now hostnames are resolved on connect,
resolved addresses are stored in the connect list of node.

Todo:
- [x] Try resolved addresses one after another.
- [x] Add tests.

New todo:
- [x] Fix tests in sandbox.
- [x] Changelog
- [x] Possibly simplify something inside network and peer info endpoint
- [x] Use `ConnectList`


exonum/src/blockchain/service.rs
```diff
@@ -423,8 +423,10 @@ impl SharedNodeState {
         lock.validators = state.validators().to_vec();
 
         for (p, c) in state.peers() {
-            lock.peers_info.insert(c.addr(), *p);
-            lock.outgoing_connections.insert(c.addr());
+            if let Some(addr) = state.get_resolved_peer_address(c.pub_addr()) {
```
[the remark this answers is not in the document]
dvush: Excuse me for maybe confusing naming, but get_resolved_peer_address does not resolve address, it just gets it from cached value in connect list.
There are resolve_and_cache_peer_address that is used for resolving names and it is called in NodeHandler::connect -> names are resolved before connect only.

exonum/src/node/connect_list.rs
```diff
@@ -45,31 +67,65 @@ impl ConnectList {
...
+            .find(|(_, a)| a.address.as_str() == address)
+            .map(|(p, _)| p)
+    }
+
+    /// Resolves network address and stores it in the `ConnectList`.
+    pub fn resolve_and_cache_peer_address(&mut self, address: &str) -> Option<SocketAddr> {
+        let key = *self.find_key_by_unresolved_address(address)?;
+        let entry = self.peers.get_mut(&key).unwrap();
+
+        if entry.resolved_cache.is_empty() {
```
aleksuss: What if not empty but record in DNS had changed ?

exonum/src/node/connect_list.rs
```diff
@@ -45,31 +67,65 @@ impl ConnectList {
...
+        self.peers
+            .iter()
+            .find(|(_, a)| a.address.as_str() == address)
+            .map(|(p, _)| p)
+    }
+
+    /// Resolves network address and stores it in the `ConnectList`.
+    pub fn resolve_and_cache_peer_address(&mut self, address: &str) -> Option<SocketAddr> {
+        let key = *self.find_key_by_unresolved_address(address)?;
+        let entry = self.peers.get_mut(&key).unwrap();
```
aleksuss: What do you think about changing`.unwrap()` for `?` ?

exonum/src/node/connect_list.rs
dvush: Thanks, fixed.

exonum/src/node/connect_list.rs
dvush: I think here it is ok since we use key that was found in this map, as well as using map[key] down the file should be ok too.

exonum/src/node/connect_list.rs
```diff
@@ -14,26 +14,48 @@
...
+/// Network address of the peer.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PeerAddress {
+    /// External address of the peer hostname:port.
+    pub address: String,
+    /// Currently used address resolution.
+    resolved: Option<SocketAddr>,
+    /// Backup address resolutions.
+    resolved_cache: Vec<SocketAddr>,
+}
```
ilammy: I'm not sure that we need to keep the socket addresses after the connection is successfully made. We only need one `SocketAddr` to work and others are effectively useless. If the connection fails at some point and we need to reconnect then we'd better retry resolving as well (e.g., to get a different server from a load-balancing DNS). So I believe we can simply store address strings instead of `PeerAddress` structs.

exonum/src/node/mod.rs
```diff
@@ -605,9 +607,15 @@ impl NodeHandler {
...
-    pub fn connect(&mut self, address: &SocketAddr) {
+    pub fn connect(&mut self, address: &str) {
         let connect = self.state.our_connect_message().clone();
-        self.send_to_addr(address, connect.raw());
+        let resolved = self.state.resolve_and_cache_peer_address(address);
+        if let Some(socket) = resolved {
+            self.send_to_addr(&socket, connect.raw());
+        } else {
+            info!("Failed to resolve hostname of peer: {}", address);
+            return;
```
ilammy: This does not look like a fatal error to me. We may try other addresses that we got from the resolver before giving up.

Also, I believe the log line should use `error!` or `warn!` instead of `info!`. And is it really okay to just log and ignore the error here? Do the callers expect the connection to happen asynchronously? Can they reasonably handle a connection error?

exonum/src/node/state.rs
```diff
@@ -570,18 +578,27 @@ impl State {
 
     /// Adds the public key, address, and `Connect` message of a validator.
     pub fn add_peer(&mut self, pubkey: PublicKey, msg: Connect) -> bool {
-        self.connections.insert(msg.addr(), pubkey);
+        // This code is temporary solution until `Connect` message is removed,
+        // we should use ip address of the incoming message here but we use resolved public address instead.
```
ilammy: This may use an issue reference (ECR-XXXX) so that we do not forget to remove 'temporary' workarounds.

exonum/src/events/tests.rs
```diff
@@ -247,12 +247,14 @@ fn test_network_handshake() {
...
+    connect_list.add(t1.connect_info.clone());
 
     let mut t2 = ConnectionParams::from_address(second);
-    connect_list.add(t2.connect_info);
+    connect_list.add(t2.connect_info.clone());
 
-    let connect_list = SharedConnectList::from_connect_list(connect_list);
+    let mut connect_list = SharedConnectList::from_connect_list(connect_list);
+    connect_list.resolve_and_cache_peer_address(&first.to_string());
+    connect_list.resolve_and_cache_peer_address(&second.to_string());
```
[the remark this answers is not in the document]
dvush: Addresses are resolved by node in the connect method. Here we have connect list test in isolation so we have this calls here.
There is only one place where address is resolved -> in node handler connect. You can't forget to resolve it, since input to those function is string address, if it fails to resolve then connect method fails.

exonum/src/helpers/fabric/details.rs
```diff
@@ -335,26 +335,18 @@ impl Command for GenerateCommonConfig {
...
-            SocketAddr::V6(_) => "::".parse().unwrap(),
-        };
-        let listen_address = SocketAddr::new(listen_ip, external_address.port());
+        let listen_port = address_str
+            .rsplitn(2, ':')
+            .next()
+            .and_then(|p| p.parse::<u16>().ok())
+            .unwrap_or(DEFAULT_EXONUM_LISTEN_PORT);
+
+        let listen_address = SocketAddr::new("0.0.0.0".parse().unwrap(), listen_port);
```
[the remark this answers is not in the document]
dvush: This code is just for config generation. In config file itself you can specify listen address as you wish.

exonum/src/node/connect_list.rs
dvush: The idea of resolved_cache is to try next ip address if previous resolved address from array was wrong. It is used to track which addresses was not tried before. 
Because of the structure of network the failure of resolve propagates like this: 

call to node connect -> resolve -> network connection pool try to establish connection to resolved address -> cannot connect to peer event -> reconnect(goto 1).

That is why information about previous resolves is stored here, so after this cycle we try different ip.

exonum/src/node/mod.rs
dvush: This else handles the case when we have 0 IPs from resolving address, we cannot try another for that reason.

examples/cryptocurrency/examples/demo.rs
```diff
@@ -37,16 +37,16 @@ fn node_config() -> NodeConfig {
         ..Default::default()
     };
 
-    let peer_address = "0.0.0.0:2000".parse().unwrap();
+    let peer_address = "0.0.0.0:2000";
 
     NodeConfig {
-        listen_address: peer_address,
+        listen_address: peer_address.parse().unwrap(),
```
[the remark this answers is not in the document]
dvush: Thanks, I did not think about it.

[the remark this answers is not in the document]
dvush: I think there are no good way of "resolving" names like [..]:enp1s0
In underlying structure you have u32 called scope_id for this kind of functionality and meaning of this mapping to devices afaik is implementation(OS) defined.

exonum/src/node/connect_list.rs
```diff
@@ -45,31 +67,68 @@ impl ConnectList {
...
+            .map(|i| i.collect())
+            .unwrap_or_default();
+
+        entry.resolved_cache.retain(|a| resolved_vec.contains(a));
+
+        if entry.resolved_cache.is_empty() {
+            entry.resolved_cache = resolved_vec;
+        }
+
+        let resolved = entry.resolved_cache.pop();
```
pavel-mukhanov: Why we need `pop` here?

dvush: It is used for removing address that we have tried already(in case hostname was resolved to multiple IPs)

exonum/src/helpers/fabric/details.rs
```diff
@@ -721,3 +766,97 @@ impl Command for GenerateTestnet {
...
+        ctx.set_arg(PEER_ADDRESS, external.to_string());
+        ctx.set_arg(LISTEN_ADDRESS, listen.to_string());
+        assert_eq!(
+            GenerateNodeConfig::addresses(&ctx),
+            (
+                format!("{}:{}", external, DEFAULT_EXONUM_LISTEN_PORT),
+                listen.parse().unwrap()
+            )
+        );
+    }
```
ilammy: I believe it would be better to test `GenerateNodeConfig::addresses()` in isolation and split this one test into several with descriptive names. But the current approach is fine too if no one else objects.

dvush: Who do you mean? 
Because I think that `GenerateNodeConfig::addresses()` is tested in isolation here. Only call to this function is tested.

dvush:
@ilammy 

Thanks for the comments.

> Should we wait for more changes regarding the address cache usage? (Avoiding repeated address resolution when that we discussed offline the other day.)

I think that it is better to live it like that. 
I believe that if listen address meaning will change and node will not be able to listen on it the worst thing that can happen is node will fail to listen on `listen address` and crash, which is fine since situation is exceptional.
On the other hand, handling this specific case will increase complexity and reduce typing(because we will have `String` instead of `SocketAddr` listen address)

exonum/src/helpers/fabric/details.rs
ilammy: Oh. I'm terribly sorry. I misread the diff :(

exonum/Cargo.toml
```diff
@@ -41,6 +41,7 @@ tokio-executor = "=0.1.4"
 tokio-io = "=0.1.8"
 tokio-retry = "=0.2.0"
 tokio-threadpool = "=0.1.6"
+tokio-dns-unofficial = "=0.4.0"
```
DarkEld3r: "Unofficial" looks promising. :upside_down_face:

ilammy: Yep. This should have been `tokio-dns-ietf-approved`. While still being pre-1.0 :rofl:

I think it only means that the crate is not maintained by Carl Lerche himself. I don't think that's an issue.

PR #945, opened 2018-09-17
One connection between peers [ECR-2129]

pavel-mukhanov:
Rework network to provide one connection between peers.

**Note:** as far as we don't have full requirments for the network protocol yet, this PR aims only resolve some bugs caused by single direction channels.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   calculations in Exonum Java Bindings, but can be used for debug purposes
   as well. (#918)
 
-- Now peers requires only one connection to exchange messages between
+- Now peers require only one connection to exchange messages between
   them. (#945)
 
 ### Bug Fixes
--- a/exonum/src/events/network.rs
+++ b/exonum/src/events/network.rs
@@ -16,6 +16,7 @@ use failure;
 use futures::{
     future, future::{err, Either}, sync::mpsc, unsync, Future, IntoFuture, Sink, Stream,
 };
+use tokio::net::{TcpListener, TcpStream};
 use tokio_codec::Framed;
 use tokio_core::reactor::Handle;
 
@@ -23,7 +24,7 @@ use tokio_retry::{
     strategy::{jitter, FixedInterval}, Retry,
 };
 
-use std::{collections::HashMap, net::SocketAddr, rc::Rc, time::Duration};
+use std::{cell::RefCell, collections::HashMap, net::SocketAddr, rc::Rc, time::Duration};
 
 use super::{error::log_error, to_box};
 use events::{
@@ -31,8 +32,6 @@ use events::{
 };
 use helpers::Milliseconds;
 use messages::{Any, Connect, Message, RawMessage};
-use std::cell::RefCell;
-use tokio::net::{TcpListener, TcpStream};
 
 const OUTGOING_CHANNEL_SIZE: usize = 10;
 
--- a/exonum/src/events/tests.rs
+++ b/exonum/src/events/tests.rs
@@ -17,7 +17,7 @@ use tokio::util::FutureExt;
 use tokio_core::reactor::Core;
 
 use std::{
-    net::SocketAddr, thread, time::{self, Duration},
+    net::SocketAddr, thread, time::{self, Duration, SystemTime},
 };
 
 use blockchain::ConsensusConfig;
@@ -29,7 +29,6 @@ use events::{
 use helpers::user_agent;
 use messages::{Connect, Message, MessageWriter, RawMessage};
 use node::{state::SharedConnectList, ConnectInfo, ConnectList, EventsPoolCapacity, NodeChannel};
-use std::time::SystemTime;
 
 #[derive(Debug)]
 pub struct TestHandler {


exonum/src/events/codec.rs
```diff
@@ -203,8 +202,7 @@ mod test {
 
     fn create_encrypted_codecs() -> (MessagesCodec, MessagesCodec) {
         let (public_key, secret_key) = gen_keypair_from_seed(&Seed::new([1; SEED_LENGTH]));
-        let mut params =
-            HandshakeParams::new(public_key, secret_key, SharedConnectList::default(), 1024);
+        let mut params = HandshakeParams::with_default_params(public_key, secret_key);
         params.set_remote_key(public_key);
```
aleksuss: @pavel-mukhanov Why can't we set `public_key` in the `with_default_params` method ?

exonum/src/node/basic.rs
```diff
@@ -127,6 +127,7 @@ impl NodeHandler {
         self.blockchain.save_peer(&public_key, message);
         if need_connect {
             info!("Send Connect message to {}", address);
+            //TODO: remove responding connect
```
ilammy: If we have a task for this it should be mentioned here.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -26,8 +26,8 @@ use std::fmt::{self, Error, Formatter};
 use super::{handshake::HandshakeParams, resolver::SodiumResolver};
 use events::noise::{error::NoiseError, HEADER_LENGTH, MAX_MESSAGE_LENGTH, TAG_LENGTH};
 
-pub const HANDSHAKE_HEADER_LENGTH: usize = 1;
-pub const MAX_HANDSHAKE_MESSAGE_LENGTH: usize = 255;
+pub const HANDSHAKE_HEADER_LENGTH: usize = 2;
+pub const MAX_HANDSHAKE_MESSAGE_LENGTH: usize = 65535;
```
ilammy: It seems that Noise does not support messages bigger than this. Maybe we could use a comment here in case somebody wants to expand the header length to 4 bytes some day.

exonum/src/events/noise/wrappers/sodium_wrapper/wrapper.rs
```diff
@@ -78,10 +78,9 @@ impl NoiseWrapper {
         Ok(buf)
     }
 
-    pub fn write_handshake_msg(&mut self) -> Result<Vec<u8>, NoiseError> {
-        // Payload in handshake messages can be empty.
+    pub fn write_handshake_msg(&mut self, msg: &[u8]) -> Result<Vec<u8>, NoiseError> {
         let mut buf = vec![0_u8; MAX_MESSAGE_LENGTH];
-        let len = self.write(&[], &mut buf)?;
+        let len = self.write(msg, &mut buf)?;
```
ilammy: I've reviewed the call stack and it seems that `noise` crate verifies message length and returns an error if `msg` is too big.

This is not a request for change, just a review remark in case anybody else is wondering whether we'll panic if it turns out that `msg.len() > buf.len()`, and whether we need a length check like in `read_handshake_msg()` in order to avoid a panic. It seems that we don't need them.

exonum/src/events/network.rs
```diff
@@ -160,376 +311,214 @@ impl ConnectionsPool {
...
-            network_tx,
-            handle,
-            handshake_params,
-            outgoing_connections: ConnectionsPool::new(),
+        if pool.contains(&address) {
+            to_box(Self::send_message(&pool, &address, &message))
+        } else if self.can_create_connections() {
+            to_box(self.create_new_connection(&address, message))
+        } else {
+            to_box(self.send_unable_connect_event(&address))
```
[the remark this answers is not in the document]
pavel-mukhanov: Good point, but based on your next comment I changed pool implementation to `Rc<RefCell>`, because message processing is handled successively.

exonum/src/events/network.rs
```diff
@@ -87,67 +86,219 @@ pub struct NetworkPart {
     pub network_tx: mpsc::Sender<NetworkEvent>,
 }
 
-#[derive(Debug, Default, Clone)]
-struct ConnectionsPool {
-    inner: Rc<RefCell<HashMap<SocketAddr, mpsc::Sender<RawMessage>>>>,
+#[derive(Clone, Debug)]
+struct ConnectionPool {
+    peers: Arc<RwLock<HashMap<SocketAddr, mpsc::Sender<RawMessage>>>>,
```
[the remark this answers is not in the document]
pavel-mukhanov: Thanks for pointing out, we can safely use `Rc<RefCell<_>>`.

exonum/src/events/network.rs
```diff
@@ -14,27 +14,25 @@
...
-use super::{
-    error::{log_error, result_ok}, to_box,
-};
+use super::{error::log_error, to_box};
 use events::{
     codec::MessagesCodec, error::into_failure, noise::{Handshake, HandshakeParams, NoiseHandshake},
 };
 use helpers::Milliseconds;
 use messages::{Any, Connect, Message, RawMessage};
+use std::cell::RefCell;
```
DarkEld3r: Please combine the `std` usings.

exonum/src/events/network.rs
```diff
@@ -14,27 +14,25 @@
...
+use tokio::net::{TcpListener, TcpStream};
```
DarkEld3r: Please move third-party usings before the `std` ones.

exonum/src/events/tests.rs
```diff
@@ -21,15 +21,15 @@ use std::{
...
-use env_logger;
+use crypto::{gen_keypair, gen_keypair_from_seed, PublicKey, SecretKey, Seed, SEED_LENGTH};
 use events::{
     error::log_error, network::{NetworkConfiguration, NetworkPart}, noise::HandshakeParams,
     NetworkEvent, NetworkRequest,
 };
 use helpers::user_agent;
 use messages::{Connect, Message, MessageWriter, RawMessage};
 use node::{state::SharedConnectList, ConnectInfo, ConnectList, EventsPoolCapacity, NodeChannel};
+use std::time::SystemTime;
```
DarkEld3r: Same here.

CHANGELOG.md
```diff
@@ -63,6 +63,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   calculations in Exonum Java Bindings, but can be used for debug purposes
   as well. (#918)
 
+- Now peers requires only one connection to exchange messages between
```
aleksuss: `require` ?

PR #948, opened 2018-09-19
Add master branch to 'safelist' [ECR-2387]

DarkEld3r:
This is needed to prevent Travis from building changes twice if a branch is pushed in addition to a pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: rust
 branches:
   only:
   - master
+  - /*release*/
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,10 @@ language: rust
 branches:
   only:
   - master
+  # Github release tags (for example "v0.9" or "v0.9.1").
   - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
-  - /.*-release/
+  # Branch names endings with "-release" (for example "0.9.0-release").
+  - /-release$/
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ language: rust
 branches:
   only:
   - master
-  - /*release*/
+  - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
+  - /.*-release/
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: rust
 
+branches:
+  only:
+  - master
+
 addons:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,10 @@
 language: rust
 
+branches:
+  only:
+  - master
+  - /*release*/
```
dvush: Maybe it should be .* ?

ilammy: Actually yes. Travis treats these entries as (Ruby) regular expressions. So if you want to include branches like `0-9-1-release` then we should use something like `/-release$/` to match all branches ending with `-release`.

I think we should also add `/^v\d+\.\d+(\.\d+)?(-\S*)?$/` to the list as suggested by @dmitry-timofeev and Travis docs. This is to make sure that Travis builds our release tags which follow semver naming convention.

dmitry-timofeev: I see, thanks. I added the ver regexp in our PR, just in case we ever have releases not in master (if that's not the case, it might be redundant as Travis is already set up to build `master`): https://github.com/exonum/exonum-java-binding/pull/484

DarkEld3r: Fixed. :upside_down_face:

.travis.yml
```diff
@@ -1,5 +1,11 @@
 language: rust
 
+branches:
+  only:
+  - master
+  - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
+  - /.*-release/
```
ilammy: We can even use a shorter `/-release/` regex. Ruby matches it anywhere in the string. This form allows releases like `0.10.0-release-rc1`. But maybe we want `/-release$/` instead to allow it only at the end.

Though, this is not critical and PR can be merged as it is now.

.travis.yml
```diff
@@ -1,5 +1,11 @@
 language: rust
 
+branches:
+  only:
+  - master
+  - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
```
mvlabat: I dunno if this a common thing, but reading such a regex puts me in trance a little bit :D
Shall we add a comment for this line maybe?

ilammy: What? You don't speak :camel:? :laughing: I guess `# release tags like v0.9, v0.9.2, v0.9.3-rc1` will be a good explanation with some examples.

mvlabat: Well, it took me some time to decode that, but I don't "speak" it, yeah )

DarkEld3r: I have added some comments. :upside_down_face:

PR #951, opened 2018-09-20
Change error msg when transaction execution fails. [ECR-1267]

dvush:


exonum/src/blockchain/mod.rs
```diff
@@ -300,7 +300,7 @@ impl Blockchain {
                 self.execute_transaction(*hash, height, index, &mut fork)
                     // Execution could fail if the transaction
                     // cannot be deserialized or it isn't in the pool.
-                    .expect("Transaction not found in the database.");
+                    .expect("Transaction execute error.");
```
ilammy: "Transaction _execution_ error" would be more accurate grammatically.

PR #954, opened 2018-09-21
Use PUBLIC_KEY_LENGTH constant instead of hardcoded integer

blackbeam:
This PR removes hardcoded length of a public key from `exonum/tests/config.rs`.


DarkEld3r:
@blackbeam Hello and thank you for the pull request!

PR #955, opened 2018-09-21
Add time information from precommits to the range requests [ECR-2407]

DarkEld3r:


DarkEld3r:
WIP because I'm trying to remove duplication in `blocks` and `blocks_with_time` methods.

exonum/src/api/node/public/explorer.rs
```diff
@@ -211,17 +278,55 @@ impl ExplorerApi {
...
             txs: inner.transaction_hashes().to_vec(),
+            time: median_precommits_time(&inner.precommits()),
         }
     }
 }
+
+fn median_precommits_time(precommits: &[Precommit]) -> DateTime<Utc> {
+    let mut times: Vec<_> = precommits.iter().map(|p| p.time()).collect();
+    times.sort();
+    times[times.len() / 2]
```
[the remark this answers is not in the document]
DarkEld3r: Additionally, `precommits` cannot be empty because this would mean that consensus is broken, but `debug_assert` will definitely make it clearer.

exonum/src/api/node/public/explorer.rs
```diff
@@ -211,17 +278,55 @@ impl ExplorerApi {
         );
         api_scope
             .endpoint("v1/blocks", Self::blocks)
+            .endpoint("v1/blocks_with_time", Self::blocks_with_time)
```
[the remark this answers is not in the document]
DarkEld3r: `BlocksRange` structure can be extended to holding optional times:
```rust
pub struct BlocksRange {
    pub range: Range<Height>,
    pub blocks: Vec<Block>,
    pub times: Vec<DateTime<Utc>>,
}
```
But I don't like this approach because these vectors don't necessary have the same size. In the current approach the type system guarantees it.

Still, it wouldn't be a breaking change for external api users.

DarkEld3r:
@ilammy Finally I decided to use an optional parameter instead of a separate request. Another advantage is a natural code deduplication. :upside_down_face:

exonum/src/api/node/public/explorer.rs
```diff
@@ -69,6 +74,9 @@ pub struct BlocksQuery {
     /// If true, then only non-empty blocks are returned. The default value is false.
     #[serde(default)]
     pub skip_empty_blocks: bool,
+    /// If true, then `BlocksRange`'s `times` field will contain median time from the
```
vitvakatu: Should be there `serde(default)`?

exonum/src/api/node/public/explorer.rs
```diff
@@ -124,11 +133,19 @@ impl ExplorerApi {
...
+
         let blocks: Vec<_> = blocks_iter
             .rev()
             .filter(|block| !query.skip_empty_blocks || !block.is_empty())
             .take(query.count)
-            .map(|block| block.into_header())
+            .map(|block| {
+                if query.add_blocks_time {
+                    times.push(median_precommits_time(&block.precommits()));
+                }
```
ilammy: (Code style nit. Feel free to disregard.)

I'd use `inspect` for populating `times` like this:

```rust
.inspect(|block| {
    if query.add_blocks_time {
        times.push(median_precommits_time(&block.precommits()));
    }
})
.map(|block| block.into_header())
```

IMO, this makes usage of iterator combinators more focused: `map` is used only for element conversion and `inspect` is used for side-effects.

exonum/src/api/node/public/explorer.rs
```diff
@@ -216,12 +238,20 @@ impl ExplorerApi {
...
             block: inner.header().clone(),
             precommits: inner.precommits().to_vec(),
             txs: inner.transaction_hashes().to_vec(),
+            time: median_precommits_time(&inner.precommits()),
         }
     }
 }
+
+fn median_precommits_time(precommits: &[Precommit]) -> DateTime<Utc> {
+    debug_assert!(precommits.is_empty(), "Precommits cannot be empty");
```
ilammy: Um... shouldn't it be the opposite?

```diff
-    debug_assert!(precommits.is_empty(), "Precommits cannot be empty");
+    debug_assert!(!precommits.is_empty(), "Precommits cannot be empty");
```

exonum/src/api/node/public/explorer.rs
DarkEld3r: I love such "style nits"! :+1:

PR #958, opened 2018-09-24
Fix max incoming connections limit check

dvush:

### DIFF
--- a/exonum/src/events/network.rs
+++ b/exonum/src/events/network.rs
@@ -229,7 +229,7 @@ impl NetworkHandler {
                 let holder = incoming_connections_counter.clone();
                 // Check incoming connections count
                 let connections_count = Rc::strong_count(&incoming_connections_counter) - 1;
-                if connections_count > incoming_connections_limit {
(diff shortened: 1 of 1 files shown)


exonum/src/events/network.rs
```diff
@@ -229,7 +229,7 @@ impl NetworkHandler {
                 let holder = incoming_connections_counter.clone();
                 // Check incoming connections count
                 let connections_count = Rc::strong_count(&incoming_connections_counter) - 1;
-                if connections_count > incoming_connections_limit {
+                if connections_count >= incoming_connections_limit {
```
ilammy: An alternative would be

```diff
-                let connections_count = Rc::strong_count(&incoming_connections_counter) - 1;
+                let connections_count = Rc::strong_count(&incoming_connections_counter);
                 if connections_count > incoming_connections_limit {
```

which is arguably more simple, but the current form is fine too.

The current decremented form even has its merits: `connection_count` really reflects the connection count, not including that additional reference held by the server's future.

PR #959, opened 2018-09-25
Fix /healthcheck and /peers endpoints [ECR-2396]

pavel-mukhanov:
Fix some bugs caused by the transition towards one connection between peers.

### DIFF
--- a/exonum/src/blockchain/service.rs
+++ b/exonum/src/blockchain/service.rs
@@ -428,12 +428,10 @@ impl SharedNodeState {
             lock.peers_info.insert(c.addr(), *p);
         }
 
-        info!("state.connections {:?}", state.connections());
-
         for (address, c) in state.connections() {
             let connect_info = ConnectInfo {
                 address: *address,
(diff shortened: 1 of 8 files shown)


exonum/src/blockchain/service.rs
```diff
@@ -27,8 +27,10 @@ use api::{websocket, ServiceApiBuilder};
 use blockchain::{ConsensusConfig, Schema, StoredConfiguration, ValidatorKeys};
 use crypto::{Hash, PublicKey, SecretKey};
 use encoding::Error as MessageError;
+use events::network::ConnectionType;
 use helpers::{Height, Milliseconds, ValidatorId};
 use messages::RawTransaction;
+use node::ConnectInfo;
```
DarkEld3r:
```diff
- use node::ConnectInfo;
- use node::{ApiSender, NodeRole, State, TransactionSend};
+ use node::{ApiSender, ConnectInfo, NodeRole, State, TransactionSend};
```

PR #970, opened 2018-09-28
Add tests for ConnectList [ECR-53]

dvush:


exonum/src/events/network.rs
```diff
@@ -310,38 +310,46 @@ impl NetworkHandler {
...
-                    );
-                    pool.add(&key, conn_addr.clone(), sender_tx);
-                    let connection = Connection::new(handle, socket, receiver_rx, conn_addr);
-                    to_box(Self::handle_connection(connection, message, &network_tx))
-                }
-            })
-            .map(drop)
+            let (sender_tx, receiver_rx) = mpsc::channel::<RawMessage>(OUTGOING_CHANNEL_SIZE);
+            let pool = self.pool.clone();
+            to_box(
```
pavel-mukhanov: You can use `futures::Either` instead of `to_box`.

exonum/src/events/network.rs
```diff
@@ -310,38 +310,46 @@ impl NetworkHandler {
...
+                        }
+                    })
+                    .map(drop),
+            )
+        } else {
+            Either::B(err(format_err!(
+                "Trying to connect to peer not from ConnectList key={}",
+                key
+            )))
+        }
```
ilammy: Personally, I prefer early returns to using `if let` or `match` blocks when there is a lot of code in them. By the time one gets to the `else` branch with error handling they already forget what clause this `else` is for and what situation is handled there.

So instead of

```rust
if let Some(thing) = thing {
    // 30 lines of dense code
} else {
    Err(format!("error"))
}
```

I find that something like

```rust
let thing = match thing {
    Some(thing) => thing,
    None => return Err(format!("error")),
};

// however many lines of code interacting with thing
```

reads much better because you immediately get the `None` case out of the way. And nesting depth is a bit better.

The same can go for wrapping constructors like you may see here with `Either::A`. For me

```rust
let connection = Retry::spawn(strategy, action)
    .map_err(into_failure)
    // ...
    .map(drop);

Either::A(connection)
```

reads better than

```rust
Either::A(
    Retry::spawn(strategy, action)
        .map_err(into_failure)
        // ...
        .map(drop)
)
```

because the nesting is shallower (which – as a side effect – leaves rustfmt more freedom), and you immediately know where to look for the return type.

(These are totally my personal preferences though. I just wanted to share some ideas, no action is necessary in this PR, but you may consider these approaches.)

dvush: I agree about nesting.
But I found early returns to be hard to compile for some reason in this kind of code. I think there are some kind of issue with type inference when you do this(for example if function should return "impl Future" and you use early return), I am sure that is is fixable with some casting but when I was trying to make this code compile I did what you see here and then forgot.
BTW, Clippy often asks to rewrite match of Option to "if let".

DarkEld3r: I also agree about nesting in general, but I'm not so sure about "wrapping constructors". :upside_down_face:

PR #971, opened 2018-09-28
Add cryptocurrency example.

vldm:

### DIFF
--- a/examples/cryptocurrency/tests/api.rs
+++ b/examples/cryptocurrency/tests/api.rs
@@ -25,11 +25,10 @@ extern crate exonum_cryptocurrency as cryptocurrency;
 extern crate exonum_testkit;
 #[macro_use]
 extern crate serde_json;
-extern crate hex;
 
 use exonum::{
     api::{self, node::public::explorer::TransactionQuery},
-    crypto::{self, Hash, PublicKey, SecretKey}, messages::{Message, RawTransaction},
+    crypto::{self, Hash, PublicKey, SecretKey}, messages::{self, Message, RawTransaction},
 };
 use exonum_testkit::{ApiKind, TestKit, TestKitApi, TestKitBuilder};
 
@@ -232,7 +231,7 @@ impl CryptocurrencyApi {
         let (pubkey, key) = crypto::gen_keypair();
         // Create a pre-signed transaction
         let tx = TxCreateWallet::sign(name, &pubkey, &key);
-        let data = hex::encode(tx.clone().serialize());
+        let data = messages::to_hex_string(&tx);
         let tx_info: serde_json::Value = self.inner
             .public(ApiKind::Explorer)
             .query(&json!({ "tx_body": data }))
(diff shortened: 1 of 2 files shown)


examples/cryptocurrency/tests/api.rs
```diff
@@ -227,28 +228,29 @@ impl CryptocurrencyApi {
...
     /// Note that the transaction is not immediately added to the blockchain, but rather is put
     /// to the pool of unconfirmed transactions.
-    fn create_wallet(&self, name: &str) -> (TxCreateWallet, SecretKey) {
+    fn create_wallet(&self, name: &str) -> (Message<RawTransaction>, SecretKey) {
         let (pubkey, key) = crypto::gen_keypair();
         // Create a pre-signed transaction
-        let tx = TxCreateWallet::new(&pubkey, name, &key);
-
+        let tx = TxCreateWallet::sign(name, &pubkey, &key);
+        let data = hex::encode(tx.clone().serialize());
```
ilammy: I remember we have some utilities and traits related to hexadecimal conversions in `exonum::encoding`. For one thing, I think we even reexport `hex::encode` as `exonum::encoding::serialize::encode_hex`.

examples/cryptocurrency/src/lib.rs
```diff
@@ -151,6 +145,26 @@ pub mod transactions {
             }
         }
     }
+
+    impl TxCreateWallet {
+        #[doc(hidden)]
```
ilammy: Why hide it?

examples/cryptocurrency/src/lib.rs
```diff
@@ -212,21 +230,17 @@ pub mod contracts {
...
-        /// signature.
-        fn verify(&self) -> bool {
-            self.verify_signature(self.pub_key())
-        }
-
         /// If a wallet with the specified public key is not registered, then creates a new wallet
         /// with the specified public key and name, and an initial balance of 100.
         /// Otherwise, performs no op.
-        fn execute(&self, view: &mut Fork) -> ExecutionResult {
+        fn execute(&self, mut tc: TransactionContext) -> ExecutionResult {
```
aleksuss: I'd prefer to rename `tc` variable into `context`.

examples/cryptocurrency/src/lib.rs
vldm: This is only "test" methods used for backward compatibility with old test code.

examples/cryptocurrency/tests/api.rs
vldm: Yeap we can, but encoding is module of our serialisation, and i am not sure we should reexport items that does not belong to this concrete serialisation.

What do you think? @DarkEld3r @ilammy

ilammy: Well... but we _do_ send these transactions to Exonum. Therefore it is expected that transactions should be encoded in some Exonum-specific way (lowercase base16 in this case). So it would not be unusual for Exonum to provide some readymade tools for encoding.

I don't have any strong opinion on this matter. I'm totally fine with using `hex` crate for base16 encoding if the web API requires the request to be JSONs of the form `{ "tx_body": "7472616e73616374696f6e20626f647920627974657320617320612068657861646563696d616c20737472696e6720616b6120626173653136" }` Any library or handwritten function would do fine, I guess. I just wondered whether we already have something specifically meant for this use case. Because [messages seem to use `hex` internally](https://github.com/exonum/exonum/blob/e98cb28ce21e3237d1b63f45c535a342a1b20a50/exonum/src/messages/mod.rs#L218-L237) it may be a good fit.

PR #973, opened 2018-10-01
Fix error message in `ApiSender::send` [ECR-706]

SergeiMal:

### DIFF
--- a/exonum/src/node/mod.rs
+++ b/exonum/src/node/mod.rs
@@ -752,7 +752,7 @@ impl ApiSender {
 impl TransactionSend for ApiSender {
     fn send(&self, tx: Box<dyn Transaction>) -> Result<(), failure::Error> {
         if !tx.verify() {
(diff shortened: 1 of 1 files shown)


exonum/src/node/mod.rs
```diff
@@ -752,7 +752,7 @@ impl ApiSender {
 impl TransactionSend for ApiSender {
     fn send(&self, tx: Box<dyn Transaction>) -> Result<(), failure::Error> {
         if !tx.verify() {
-            bail!("Unable to verify transaction");
+            bail!("Invalid transaction");
```
aleksuss: Let me ask Why ? 😄

DarkEld3r: I think I got the idea: we are actually able to the verify transaction, but because this method returns `false` - the transaction is considered invalid. But I would expand description a little bit. For example, "Invalid transaction (verify() returns false)".

PR #974, opened 2018-10-01
Fix network disconnect handling. [ECR-2310]

dvush:


exonum/src/events/network.rs
```diff
@@ -14,7 +14,8 @@
 
 use failure;
 use futures::{
-    future, future::{err, Either}, sync::mpsc, unsync, Future, IntoFuture, Sink, Stream,
+    future, future::{err, Either}, stream::{SplitSink, SplitStream}, sync::mpsc, unsync, Future,
```
DarkEld3r:
```diff
- future, future::{err, Either},
+ future::{self, err, Either},
```
:upside_down_face:

exonum/src/events/network.rs
```diff
@@ -489,28 +563,7 @@ impl NetworkHandler {
...
-                    .map(drop),
-            )
-        } else {
-            Box::new(err(format_err!(
-                "Attempt to connect to disconnect with peer with key {:?} which \
-                 is not in the connection pool",
-                peer
-            )))
-        }
+        self.pool.len() < self.network_config.max_outgoing_connections
```
ilammy: I remember using `<` vs `<=` here has some history. Is this change intentional?

exonum/src/events/network.rs
```diff
@@ -337,9 +368,19 @@ impl NetworkHandler {
...
+                                receiver_rx,
+                                conn_addr,
+                                *message.pub_key(),
+                            );
+                            to_box(Self::handle_connection(
+                                connection,
+                                message,
+                                pool,
+                                &network_tx,
+                            ))
```
ilammy: All these
[ｖ
ｅ
ｒ
ｔ
ｉ
ｃ
ａ
ｌ](https://webimg.ccpgamescdn.com/zncvdb47ti8o/45K7khNcBaKMsyC8SCMGES/a88230d4ce62b115139ed9a38d21d58b/EVEOnline_ships_Naglfar.jpg)
argument lists everywhere leave me with a feeling that some sort of a _context_ could be of use here. So that we can bundle these things and pass them around as a whole. I don't know, we could put them in `self` or some explicit struct.

This is not a nit. You don't have to change anything in this PR. The situation is far from dire in this file. That's just a thing that we might think about in background. If there is some natural, obvious way to group these values then we might use it.

exonum/src/events/network.rs
dvush: Yes, here it should be this way.

PR #980, opened 2018-10-03
Refactor time service to new messages module

vldm:

### DIFF
--- a/services/time/examples/simple_service.rs
+++ b/services/time/examples/simple_service.rs
@@ -99,9 +99,9 @@ impl TxMarker {
 }
 
 impl Transaction for TxMarker {
-    fn execute(&self, mut tc: TransactionContext) -> ExecutionResult {
-        let author = tc.author();
-        let view = tc.fork();
+    fn execute(&self, mut context: TransactionContext) -> ExecutionResult {
+        let author = context.author();
+        let view = context.fork();
         let time = TimeSchema::new(&view).time().get();
         match time {
             Some(current_time) if current_time <= self.time() => {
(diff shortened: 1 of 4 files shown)


services/time/examples/simple_service.rs
```diff
@@ -71,28 +73,35 @@ impl<'a> MarkerSchema<&'a mut Fork> {
...
-    fn verify(&self) -> bool {
-        self.verify_signature(self.from())
+impl TxMarker {
+    fn sign(
+        mark: i32,
+        time: DateTime<Utc>,
+        pk: &PublicKey,
+        sc: &SecretKey,
+    ) -> Message<RawTransaction> {
+        Protocol::sign_transaction(TxMarker::new(mark, time), SERVICE_ID, *pk, sc)
```
ilammy: If we have a vertical argument list then we can as well use proper variable names like `public_key` instead of `pk`. Though, it's a one-line function so it may be okay to keep them short. Just be consistent with **k**'s and **c**'s, please :)

On a side note, why public key needs to be dereferenced but secret key doesn't? That's also a bit inconsistent.

(Ditto for `TxTime`.)

services/time/examples/simple_service.rs
```diff
@@ -71,28 +73,35 @@ impl<'a> MarkerSchema<&'a mut Fork> {
...
     }
+}
 
-    fn execute(&self, view: &mut Fork) -> ExecutionResult {
+impl Transaction for TxMarker {
+    fn execute(&self, mut tc: TransactionContext) -> ExecutionResult {
```
ilammy: I will be resident @aleksuss here to ask for a rename into `context`, for the sake of consistency.

(Ditto for `TxTime`.)

services/time/examples/simple_service.rs
vldm: This method used for backward compatibility with old signature. Thats why Pk is dereferenced (in new messages it's used by value).

PR #982, opened 2018-10-03
Add new check for outgoing connections [ECR-2434]

dvush:

### DIFF
--- a/exonum/src/events/network.rs
+++ b/exonum/src/events/network.rs
@@ -374,8 +374,8 @@ impl NetworkHandler {
                     })
                     .and_then(move |(socket, raw)| (Ok(socket), Self::parse_connect_msg(Some(raw))))
                     .and_then(move |(socket, message)| {
-                        let connection_limit = pool.count_outgoing() >= max_connections;
-                        if pool.contains(message.pub_key()) || connection_limit {
+                        let connection_limit_reached = pool.count_outgoing() >= max_connections;
+                        if pool.contains(message.pub_key()) || connection_limit_reached {
                             Box::new(future::ok(()))
                         } else {
                             let conn_addr = ConnectedPeerAddr::Out(
--- a/exonum/src/events/network.rs
+++ b/exonum/src/events/network.rs
@@ -52,13 +52,6 @@ impl ConnectedPeerAddr {
             ConnectedPeerAddr::Out(_, _) => false,
         }
     }
-
-    pub fn is_outgoing(&self) -> bool {
-        match self {
-            ConnectedPeerAddr::In(_) => false,
-            ConnectedPeerAddr::Out(_, _) => true,
-        }
-    }
 }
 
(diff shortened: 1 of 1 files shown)


exonum/src/events/network.rs
```diff
@@ -45,6 +45,22 @@ pub enum ConnectedPeerAddr {
...
+impl ConnectedPeerAddr {
+    pub fn is_incoming(&self) -> bool {
+        match self {
+            ConnectedPeerAddr::In(_) => true,
+            ConnectedPeerAddr::Out(_, _) => false,
+        }
+    }
+
+    pub fn is_outgoing(&self) -> bool {
+        match self {
```
pavel-mukhanov: Maybe `!self.is_incoming()` is better.

exonum/src/events/network.rs
```diff
@@ -360,7 +381,8 @@ impl NetworkHandler {
                     })
                     .and_then(move |(socket, raw)| (Ok(socket), Self::parse_connect_msg(Some(raw))))
                     .and_then(move |(socket, message)| {
-                        if pool.contains(message.pub_key()) {
+                        let connection_limit = pool.count_outgoing() >= max_connections;
```
ilammy: These comparisons using `>=` here vs `<` in `can_create_connections()` slightly concern me. These different lines agree because `!(a < b)` ≡ `(a >= b)`, but someone some day may forget about it, change one line but not the other, and introduce an inconsistency.

I'd rewrite it to using `<` in both cases (e.g., `!connections_available`). Or even introduce a method on `ConnectionPool` but that'd be a bit overkill, I guess.

exonum/src/events/network.rs
dvush: Maybe it is ok here, because this method is not used.  
But in general  pedantic clippy does not allow to use `if !x {} else {}` it requires you to reshuffle it and it is part of the standard library style where you have for ex. `is_some()` and `is_none()` for `Option`

exonum/src/events/network.rs
dvush: I think I'll leave sign like this here since the same convention is used in the listener method where incoming connections are accepted. But I'll rename variable to clarify for future which sign should be used here.

exonum/src/events/network.rs
```diff
@@ -113,8 +122,12 @@ impl ConnectionPool {
         }
     }
 
-    fn len(&self) -> usize {
-        self.peers.borrow().len()
+    fn count_outgoing(&self) -> usize {
+        let peers = self.peers.borrow();
+        peers
+            .iter()
+            .filter(|(_, e)| !e.address.is_incoming())
```
ilammy: I wonder why have you removed `is_outgoing` in order to use `!is_incoming()` here. It would be more direct to leave `is_outgoing` and remove `is_incoming` instead.

PR #987, opened 2018-10-04
Update changelog with 0.9.2 and 0.9.3 releases

DarkEld3r:


ozkriff:
> `X` codecov/project — 93.15% (-0.01%) compared to 3ab7f39

Why CHANGELOG is affecting the coverage? O.o

DarkEld3r:
@ozkriff I'm not sure, but as far as I can see, caverage isn't very accurate and sometimes questionable. For example: https://codecov.io/gh/exonum/exonum/src/master/exonum/src/messages/protocol.rs

PR #998, opened 2018-10-10
remove outdated dependency added by chance

boguslavsky:


examples/timestamping/frontend/package.json
```diff
@@ -21,7 +21,6 @@
   "dependencies": {
     "axios": "^0.18.0",
     "big-integer": "^1.6.36",
-    "boostrap": "^2.0.0",
```
aleksuss: `bootstrap` ?

boguslavsky: @aleksuss yes, `bootstrap` is already in dependency list, see below

PR #1004, opened 2018-10-11
Implement SegmentField for Option<T: Field>

blackbeam:
The motivation is to allow this:

```rust
encoding_struct! (
    struct Foo {
        bar: Option<T>,
    }
);
```

From the data perspective `Option<T>` is just a subset of `Vec<T>` with additional limitation (len <= 1) so i see no reason to not to have it. However this does not mean, that there is no such reason 😄

As in case of `Vec`'s implementation, fields like `Option<Option<T>>` will not compile which is good since such fields are not directly representable in JSON.


aleksuss:
@blackbeam fix conflicts please

PR #1019, opened 2018-10-30
Rename folders containing demo services [EAN-238]

MartaDubno:
The folders containing demo services have been renamed to give a better idea of what they contain. I've tried to track any links to the folders and change them as well.


DarkEld3r:
I'm not sure about such changes. Originally directories were named the same as services (in the corresponding `Cargo.toml` files). I suppose it is OK to make changes in the documentation (and README files), but I'm not sure that files and directories should actually be renamed. Perhaps, `cryptocurrency-advanced` isn't the perfect name, but neither is `demo-create-service-with-data-proofs`.

.github/CODEOWNERS
```diff
@@ -7,12 +7,12 @@
 # Cryptography:
 crypto/* @darkeld3r @aleksuss @pavel-mukhanov @vldm @slowli
 
-# Cryptocurrency examples:
-examples/cryptocurrency/* @darkeld3r @aleksuss @vitvakatu @ilammy
-examples/cryptocurrency-advanced/* @darkeld3r @aleksuss @vitvakatu @ilammy
+# Demo service examples:
```
aleksuss: Just for curiosity, why not `demo-cryptocurrency` ? `demo-service` it's abstract name as for me.

vitvakatu: Why not `cryptocurrency-demo`? :D

MartaDubno: I understand that the PR might seems a bit strange, but its an attempt at making names of demos uniform and more user-friendly across docs and in the code.
In this case, the motivation was that cryptocurrency is a rather hyped up word, and it might make the wrong impression. We did not want the name to be instantly associated with ICO, etc., but rather with what it is - a demo showing how to implement a basic service using Exonum.
Hope I managed to explain the case a bit.

PR #1022, opened 2018-11-01
Replace VecMap dependency with HashMap<u16, _> [ECR-1803]

DarkEld3r:
Motivation: `vec_map` [crate](https://github.com/contain-rs/vec-map) is in "maintenance mode" and has not reached 1.0 version. Additionally it doesn't look as a crucial optimization, so it would be nice to get rid of additional dependency.

It is stated in the ECR-1803 issue that benchmarking should be done, but I don't see much sense in synthetic benchmarks comparing general `VecMap` and `HashMap ` performance.


exonum/src/blockchain/mod.rs
```diff
@@ -129,7 +128,7 @@ impl Blockchain {
     /// Returns the `VecMap` for all services. This is a map which
     /// contains service identifiers and service interfaces. The VecMap
     /// allows proceeding from the service identifier to the service itself.
-    pub fn service_map(&self) -> &Arc<VecMap<Box<dyn Service>>> {
+    pub fn service_map(&self) -> &Arc<HashMap<u16, Box<dyn Service>>> {
```
ilammy: Documentation comment should be updated accordingly as well. I guess we can avoid specifying concrete type of the map.

PR #1023, opened 2018-11-01
Implement DB Checkpoints and fast node synchronisation

eupn:
# Description

This is the second revision of #996.

This pull-request make nodes able to regularly produce storage checkpoints and to bootstrap storage state in `sync-mode` from checkpoints made by other nodes or create checkpoints from command line.

*`max_message_len` should be increased up to `10` MB to be able to transfer RocksDB's `.SST` files via `FileResponse` message!*

## Changes made

1. Added interface to the RocksDB's [checkpoint](https://github.com/facebook/rocksdb/wiki/Checkpoints) routines, modified `Database` trait to add `create_checkpoint` method which is implemented in `rocskdb.rs`

2. Added command to create RocksDB checkpoint from specified DB instance:

Example of usage:

```bash
exonum-node create-checkpoint <SOURCE> <CHECKPOINT>
```

3. Added module `node::checkpoints` with `CheckpointManager` with following responsibilities:
    - Produce checkpoints periodically
    - Keep track of created and currently receiving checkpoints

For now, creation of checkpoint is hooked to `Commit` phase of consensus (in `consensus.rs`).
In future it can easily be made asynchronously, but since creation of RocksDB checkpoint is done almost instantaneously, synchronous checkpoint creation shouldn't cause performance issues.

4. Added optional configuration file option `checkpoint_period_blocks`

Example of node's `.toml` configuration file that will instruct node to create checkpoint each 100 blocks:
```toml
# ... lines skipped ...

[database]
# ... lines skipped ...
checkpoint_period_blocks=100
```

5. Added optional limit of number of checkpoints. During checkpoint creation, oldest checkpoints are removed
```toml
# ... lines skipped ...

[database]
# ... lines skipped ...
max_checkpoints=3 # limit number of checkpoints up to 3
```

6. Added peer-to-peer messages:
    - `LastCheckpointRequest`
    - `LastCheckpointResponse`
    - `FileRequest`
    - `FileResponse`

# How Has This Been Tested?

## DB checkpoint creation from command line

1. Executed `./exonum-node run --db-path=db1 --node-config=node.toml`
2. Created some amount of blocks
3. Shut node down
4. Executed `./exonum-node create-checkpoint db1 checkpoint1`
5. Checkpoint created in `checkpoint1`
6. Ran node from checkpoint: `./exonum-node run --db-path=checkpoint1 --node-config=node.toml`
7. Observed same node's storage state as in `db1`

## DB sync mode

1. There were four (N1 ... N4) Exonum nodes set up with a timestamping service example
2. Nodes were allowed to create some amount of blocks and N1 was allowed to create checkpoints each 100 blocks and last checkpoint was produced on height `H`
3. One of the nodes (N4) was shut down
4. Database of the node N4 was deleted
5. N4 has been started with parameter `--sync-mode=true`
6. N4 has been requesting and downloading last created checkpoint (for height `H`) from node N1
7. After checkpoint was downloaded and applied, node was shut down
8. N4 started again as usual and continued to produce blocks and maintaining consensus starting from height `H`

# To do (open problems/questions)

- [ ] Gracefully shutdown node (instead of `panic!()`) when sync is finished, or: 
- [ ] What to do with node-specific data such as mempool (possibly delete from checkpoint after creation)
- [ ] Seamlessly replace current RocksDB instance with data from checkpoint without restart
- [ ] Make checkpoints asynchronously to consensus `Commit` stage
- [ ] Transfer files in small chunks to avoid increasing `max_message_len`

*Reviews and feedback are highly appreciated!*


slowli:
Just a couple of surface-level notes:

- I find it not the best design decision to embed synchronization directly into the P2P node protocol until we're reasonably sure about consequences. There should be a way to bootstrap nodes having less impact on the node behavior. As one of possibilities, imo, it would a reasonable workflow to bootstrap a node from a _local_ backup (i.e., not make an assumption that other nodes can send us files in the backup - they may have reasons not want to do that; for one, it seems like a DoS attack vector).
- Was it an intentional decision not to use [`BackupEngine`](https://docs.rs/exonum_rocksdb/0.7.4/exonum_rocksdb/backup/struct.BackupEngine.html)? As far as I understand, [it does use persistent snapshots / checkpoints internally](https://github.com/facebook/rocksdb/blob/21171615c10ee1a636ea28f2303a93a4bc39dbde/utilities/backupable/backupable_db.cc), but it may use internal optimizations that "user-space" code won't be able to reproduce.

aleksuss:
This feature will implement later and in another way.

PR #1024, opened 2018-11-02
Add ?Sized bound to the index_id attribute of new_in_family(..) 

blackbeam:
The motivation is to allow this:

```rust
let _ = ProofMapIndex::new_in_family("foo", "bar", &mut fork);
                                         // ^---^ Error is triggered here at the moment
                                         // because str is not Sized
```

The reason is that signature of `new_in_family` have only `StorageKey` bound on `index_id` attribute but `StorageKey` is also implemented for `str` and `[u8]` which is unsized.

Note that this PR only address the `new_in_family` and this issue still may exists for other functions with similar bounds.

### DIFF
--- a/exonum/src/storage/tests.rs
+++ b/exonum/src/storage/tests.rs
@@ -14,8 +14,8 @@
 
 use super::super::crypto::Hash;
 use super::{
-    Database, Fork, KeySetIndex, ListIndex, MapIndex, ProofListIndex, ProofMapIndex, Snapshot,
-    SparseListIndex, ValueSetIndex,
+    Database, Entry, Fork, KeySetIndex, ListIndex, MapIndex, ProofListIndex, ProofMapIndex,
+    Snapshot, SparseListIndex, ValueSetIndex,
 };
 
 const IDX_NAME: &'static str = "idx_name";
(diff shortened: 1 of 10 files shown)


aleksuss:
@blackbeam update CHANGELOG.md file please.

exonum/src/storage/tests.rs
```diff
@@ -259,3 +263,18 @@ mod rocksdb_tests {
         assert_eq!(index.len() as usize, iter.count());
     }
 }
+
+#[test]
+#[ignore]
+#[allow(unreachable_code, unused_variables)]
+fn should_compile() {
```
aleksuss: @blackbeam i'd prefer to see real tests for covering this feature.

exonum/src/storage/tests.rs
```diff
@@ -12,7 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::{Database, Fork, Snapshot};
+use super::super::crypto::Hash;
```
DarkEld3r: I think `use crypto::Hash` would look better.

exonum/src/storage/tests.rs
DarkEld3r: I suppose a compile-time test is OK, but I would add a comment because `#[ignore]` attribute is too generic.

blackbeam: Looks like it doesn't have to be a test.
Will do something similar to this: https://github.com/rust-lang/rust/pull/34951/files

vitvakatu: Actually, I think the current solution is better than `dead_code` function, because it will not affect compile time.

blackbeam: @vitvakatu, it's in separete test module annotated with `#[test]` so compile time won't be affected 🙃

PR #1030, opened 2018-11-07
Incremental service initialization

blackbeam:
### Motivation

The motivation is to conveniently load huge (more than available RAM) amount of data into the database during service initialization.

### Implementation

Implementation is quite controversial but I'm not sure that it is possible to come up with something much better than this.
Implementation inspired by:
1. [WriteUnprepared Transactions propose for RocksDb][1];
2. [futures::future::loop_fn][2].

**Logic is the following:**

`Service::initialize` now returns `InitResult` which is either `Continue` or `Done(Value)`. Similar to `futures::future::loop_fn`, `InitResult::Continue` is a way to reenter `Service::Initialize` with a new instance of `Fork`.
Set of "rollback actions" required to rollback changes done within subsequent `Service::Initialize` calls is stored in a separate index within the database. This set of actions will be applied to the database in case of panic in later call `Service::Initialize`.

### Alternatives

The only alternative i see (ideas are much appreciated) is something similar to this, but crafted in ad-hoc manner as a separate binary to run before first network startup.

[1]: https://github.com/facebook/rocksdb/wiki/WriteUnprepared-Transactions
[2]: https://docs.rs/futures/0.1.25/futures/future/fn.loop_fn.html


slowli:
We intend to improve storage API and internal organization (see [the roadmap](https://exonum.com/doc/roadmap/#storage-enhancements)). Ideally, a large-scope proposal regarding the storage like this one should be understood better in order to integrate it into these storage improvements. Thus, a couple of clarifying questions / comments:

- Do we still assume that _ordinary_ forks are still comfortably small in terms of RAM usage? That is, there is relatively few scenarios when the proposed functionality is necessary; the limitation in scope is intentional.
- Would it be fair to characterize these scenarios as data migration? We've been thinking about functionality related to migrating Exonum services; large-scale data migrations seems a substantial part of it. (I must warn that this research is still very much work in progress.) One idea how to solve this is to use a separate location in storage to build up necessary changes (ordinary `Fork`s can be used for this purpose) and then atomically move them to the target place in the storage. The process would be managed using _asynchronous tasks_, which would become 1st class notion in Exonum.
- Admittedly, we did not think that async tasks would be needed for service initialization, as it is unclear that initialization requires extensive data crunching to justify the async nature of the task. Do you have in mind loading data from external sources in service initialization?

blackbeam:
Thanks for your response.

> Do we still assume that ordinary forks are still comfortably small in terms of RAM usage? That is, there is relatively few scenarios when the proposed functionality is necessary; the limitation in scope is intentional.

The scope of this functionality is intentionally limited to the service initialization (before genesis block is created). Large (relative to RAM) forks are still possible for nodes on non-zero height, but i think fork size limitation in that case is somewhat healthy (in terms of sanity). However, situation during service initialization is quite different, because the amount of data required for node to start is unbound and it is perfectly sane to desire to start the blockchain with a lot of legacy data already in it.

> * Would it be fair to characterize these scenarios as data migration? We've been thinking about functionality related to migrating Exonum services; large-scale data migrations seems a substantial part of it. (I must warn that this research is still very much work in progress.) One idea how to solve this is to use a separate location in storage to build up necessary changes (ordinary Forks can be used for this purpose) and then atomically move them to the target place in the storage. The process would be managed using asynchronous tasks, which would become 1st class notion in Exonum.
> * Admittedly, we did not think that async tasks would be needed for service initialization, as it is unclear that initialization requires extensive data crunching to justify the async nature of the task. Do you have in mind loading data from external sources in service initialization?

Yes. Obviously all this async-tasks-solution looks over-engineered for the case of data loading during service initialization. The duty of a node in that case is simply to load data into the database, whatever time it takes, to create genesis block and to start consensus protocol.

There is nothing node can do until all services are initialized. Services may initialize in parallel until this "synchronization point" but i'm not sure about the benefits because storage back-end will start to interfere here (e.g. rocksdb internal synchronization). Also to me it looks a bit out of scope for this "service initialization ram wall" problem.

Anyway, this looks like a workaround to me because it's actually a workaround. I'm also somewhat confident that it's a necessary workaround.
The proper solution, i think, is to reimplement Fork for RocksDb in terms of WriteUnprepared Transactions and SavePoints (WriteUnprepared Transactions are not yet implemented) and use similar primitives for other backends where possible. However this primitives may not give all desired properties of a fork so this needs to be further investigated.

slowli:
> Obviously all this async-tasks-solution looks over-engineered for the case of data loading during service initialization.

I'm inclined to disagree here. It's not over-engineering as soon as we allow services to be deployed / initialized _after_ the genesis block, which is very much the direction we're going to pursue in our roadmap; [dynamically deployed Java services](https://exonum.com/doc/roadmap/#dynamic-java-services) are planned for Q2 '19, but we may allow to initialize (not deploy) Rust services dynamically earlier than that. In case if dynamic services, async tasks are reasonably required for service deployment: if we assume that services can be deployed from a software artifact, then a node needs time to download this artifact, check its integrity, possibly build it from sources, etc.

Thus, the question of motivation behind the change is very important: Do we assume that we may need a complex setup tied to the genesis block creation? Or is it tied to the service initialization after all? Also, it would be very helpful if you specified the intended scenario in more detail (e.g., the source of blockchain data; intended guarantees of availability of this source; guarantees of data agreement among blockchain nodes) in order to pinpoint the best design possible.

PR #1035, opened 2018-11-09
Added a temporary dependency ring for being compatible with rust 1.27.2

aleksuss:


ilammy:
An alternative would be to add and commit Cargo.lock for Exonum, pinning _all_ dependencies for a while. This will make sure that we have something that builds with 1.27. But I'm not sure if this is a right approach for a library.

aleksuss:
@ilammy I don't think that it's a good idea to commit `Cargo.lock`.

PR #1042, opened 2018-11-14
Fix `clear` in indexes belonging to a family

slowli:
This PR fixes incorrect behavior of the `clear()` method in indexes belonging to the same family. Before the fix, in certain situations the method cleared all indexes with the prefixes lexicographically greater than the one the method was invoked for. The explanatory test for `ListIndex` is supplied with the PR.

### DIFF
--- a/exonum/src/storage/db.rs
+++ b/exonum/src/storage/db.rs
@@ -449,7 +449,7 @@ impl Fork {
 
     /// Removes all keys starting with the specified prefix from the column family
     /// with the given `name`.
-    pub fn remove_by_prefix(&mut self, name: &str, prefix: Option<&Vec<u8>>) {
+    pub fn remove_by_prefix(&mut self, name: &str, prefix: Option<&[u8]>) {
         let changes = self
             .patch
             .changes_entry(name.to_string())
@@ -458,7 +458,7 @@ impl Fork {
         if let Some(prefix) = prefix {
             let keys = changes
                 .data
-                .range::<Vec<u8>, _>((Included(prefix), Unbounded))
+                .range::<[u8], _>((Included(prefix), Unbounded))
                 .map(|(k, _)| k.to_vec())
                 .take_while(|k| k.starts_with(prefix))
                 .collect::<Vec<_>>();
@@ -470,10 +470,10 @@ impl Fork {
         }
 
         // Remove keys from storage.
-        let prefix_as_slice = prefix.map_or(&[] as &[u8], |k| k.as_slice());
-        let mut iter = self.snapshot.iter(name, prefix_as_slice);
+        let prefix_or_empty_slice = prefix.unwrap_or_default();
+        let mut iter = self.snapshot.iter(name, prefix_or_empty_slice);
         while let Some((k, ..)) = iter.next() {
-            if !k.starts_with(prefix_as_slice) {
+            if !k.starts_with(prefix_or_empty_slice) {
                 break;
             }
 
--- a/exonum/src/storage/db.rs
+++ b/exonum/src/storage/db.rs
@@ -468,11 +468,15 @@ impl Fork {
         } else {
             changes.data.clear();
         }
-        // Remove from storage
-        let mut iter = self
-            .snapshot
-            .iter(name, prefix.map_or(&[], |k| k.as_slice()));
+
+        // Remove keys from storage.
+        let prefix_as_slice = prefix.map_or(&[] as &[u8], |k| k.as_slice());
+        let mut iter = self.snapshot.iter(name, prefix_as_slice);
         while let Some((k, ..)) = iter.next() {
+            if !k.starts_with(prefix_as_slice) {
+                break;
+            }
+
             let change = changes.data.insert(k.to_vec(), Change::Delete);
             if self.logged {
                 self.changelog.push((name.to_string(), k.to_vec(), change));
--- a/exonum/src/storage/list_index.rs
+++ b/exonum/src/storage/list_index.rs
@@ -589,6 +589,10 @@ mod tests {
         );
     }
 
+    // Parameters for the `list_index_clear_in_family` test.
+    const FAMILY_CLEAR_PARAMS: &[(u32, u32, bool)] =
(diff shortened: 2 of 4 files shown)


exonum/src/storage/db.rs
```diff
@@ -468,11 +468,15 @@ impl Fork {
         } else {
             changes.data.clear();
         }
-        // Remove from storage
-        let mut iter = self
-            .snapshot
-            .iter(name, prefix.map_or(&[], |k| k.as_slice()));
+
+        // Remove keys from storage.
+        let prefix_as_slice = prefix.map_or(&[] as &[u8], |k| k.as_slice());
```
vitvakatu: What about this syntax?
```suggestion
        let prefix_as_slice: &[u8] = prefix.map_or(&[], |k| k.as_slice());
```

By the way, why we use `&Vec` in the arguments? If we replace it with `&[u8]` (which is better for API) - we can use even:

```suggestion
        let prefix_as_slice = prefix.unwrap_or_default();
```

exonum/src/storage/list_index.rs
```diff
@@ -581,6 +639,23 @@ mod tests {
             let mut list_index = ListIndex::new_in_family(IDX_NAME, &vec![01], &mut fork);
             super::list_index_iter(&mut list_index);
         }
+
+        #[test]
+        fn list_index_clear_in_family() {
+            const PARAMS: &[(u32, u32, bool)] = &[
+                (0_u32, 5_u32, false),
```
vitvakatu: Why do you need `_u32` suffixes here? Looks a bit verbose

exonum/src/storage/list_index.rs
```diff
@@ -581,6 +639,23 @@ mod tests {
...
+                (5_u32, 0_u32, false),
+                (1_u32, 7_u32, true),
+                (7_u32, 1_u32, true),
+            ] as &[(u32, u32, bool)];
```
vitvakatu: 3 different ways of type indication is used. Is it really needed?

aleksuss: Agree with @vitvakatu. Parameters types in the tuple is enough I suppose.

ilammy: IIRC, `const` in Rust requires explicit type specification so we can leave only that:

```rust
const PARAMS: &[(u32, u32, bool)] = &[
    (0, 5, false),
    // ...
];
```

exonum/src/storage/list_index.rs
```diff
@@ -581,6 +639,23 @@ mod tests {
...
+
+            for &(x, y, merge_before_clear) in PARAMS {
+                let dir = TempDir::new(super::gen_tempdir_name().as_str()).unwrap();
```
ilammy: I guess we don't need to call `gen_tempdir_name()` here. We can simply use a string `"list_index_clear_in_family"` or something. `TempDir::new()` uses the prefix argument to generate a unique random name for the temporary directory so generating a random prefix for it is a bit excessive.

Ditto below in `rocksdb_tests`.

exonum/src/storage/list_index.rs
slowli: I *think* both explicit `_u32` suffixes and explicit cast to slice were required for the code to compile in Rust 1.27 (I used it to compile the code initially before seeing that we require 1.30 now), but apparently, 1.30 doesn't require them. I'm not going to install 1.27 back to check tho, so maybe it's just a brainfart on my side 🙃

exonum/src/storage/list_index.rs
slowli: Is this specific to the tests I've added, or are you proposing to refactor all tests in `storage::list_index` module?

aleksuss: @ilammy there was some problems with generated unique random name by `TempDir::new()`.

exonum/src/storage/list_index.rs
ilammy: Travis build has a job to test 1.27 compatibility.

I have 1.27.2 installed and this compiles fine with it:

```rust
fn main() {
    const THINGS: &[(u32, u32, bool)] = &[
        (0, 0, false),
        (9, 20, true),
    ];

    for stuff in THINGS {
        println!("{:?}", stuff);
    }
}
```

exonum/src/storage/list_index.rs
```diff
@@ -642,14 +646,7 @@ mod tests {
...
         fn list_index_clear_in_family() {
-            const PARAMS: &[(u32, u32, bool)] = &[
-                (0_u32, 5_u32, false),
-                (5_u32, 0_u32, false),
-                (1_u32, 7_u32, true),
-                (7_u32, 1_u32, true),
-            ] as &[(u32, u32, bool)];
-
-            for &(x, y, merge_before_clear) in PARAMS {
+            for &(x, y, merge_before_clear) in super::FAMILY_CLEAR_PARAMS {
```
ilammy: <span class="code-style-nit"&gt;
Though I definitely find these attack calls amusing (super family clear params! hiya!), I guess some `use super::*;` won't hurt.
&lt;/span>

exonum/src/storage/list_index.rs
ilammy: @aleksuss, hm... I'm curious, any references to that? TempDir [seems to do exactly the same thing](https://docs.rs/tempdir/0.3.7/src/tempdir/lib.rs.html#200-217) with generating random names, giving up only after 2^31 retries. So I find it really strange that it fails to work.

exonum/src/storage/db.rs
```diff
@@ -449,7 +449,7 @@ impl Fork {
 
     /// Removes all keys starting with the specified prefix from the column family
     /// with the given `name`.
-    pub fn remove_by_prefix(&mut self, name: &str, prefix: Option<&Vec<u8>>) {
+    pub fn remove_by_prefix(&mut self, name: &str, prefix: Option<&[u8]>) {
```
vitvakatu: Now it's breaking change and needs changelog update 🙃

aleksuss:
@slowli could you please add some entry related to the change ?

PR #1045, opened 2018-11-15
Use protobuf for transactions [ECR-*]

dvush:
WIP


exonum_derive/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "exonum_derive"
+version = "0.1.0"
+authors = ["Vitaly Drogan <vitaliy.drogan@gmail.com>"]
```
aleksuss:
```suggestion
authors = ["The Exonum Team <exonum@bitfury.com>"]
```

dvush: yes of course

examples/cryptocurrency-advanced/backend/build.rs
```diff
@@ -0,0 +1,14 @@
+extern crate protoc_rust;
+
+use protoc_rust::Customize;
+
+fn main() {
+    protoc_rust::run(protoc_rust::Args {
+        out_dir: "src/proto",
+        input: &["src/proto/cryptocurrency.proto"],
+        includes: &["src/proto", "../../../exonum/src/encoding/protobuf/proto"],
```
slowli: Somehow, this doesn't seem like a sustainable solution. How is it supposed to work for published crates?

examples/cryptocurrency-advanced/backend/src/proto/cryptocurrency.proto
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package exonum.examples.cryptocurrency_advanced;
+
+import "helpers.proto";
+
+message Transfer {
+ exonum.PublicKey to = 1;
```
slowli: Weird indentation (1). Is there canonical indentation for `proto` files, btw? The [examples](https://developers.google.com/protocol-buffers/docs/proto3) seem to use 2-space indentation.

examples/cryptocurrency-advanced/backend/src/proto/cryptocurrency.proto
```diff
@@ -0,0 +1,42 @@
...
+ uint64 amount = 2;
+ uint64 seed = 3;
+}
+
+message Issue {
+  uint64 amount = 1;
+  uint64 seed = 2;
```
slowli: Shouldn't `seed` be `fixed64` if it's chosen randomly?

examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -65,44 +66,62 @@ impl From<Error> for ExecutionError {
...
-            /// Amount of currency to transfer.
-            amount:  u64,
-            /// Auxiliary number to guarantee [non-idempotence][idempotence] of transactions.
-            ///
-            /// [idempotence]: https://en.wikipedia.org/wiki/Idempotence
-            seed:    u64,
-        }
+/// Transfer `amount` of the currency from one wallet to another.
+#[derive(Serialize, Deserialize, Clone, Debug, ProtobufConvert)]
+#[protobuf_convert("proto::Transfer")]
```
slowli: I vaguely remember that attributes support namespaces now (like `exonum::protobuf_convert`), judging by namespaced `clippy` attributes in the code of the main crate. Please check and use the namespaced version if it's possible (you can shorten the attribute name to something like `exonum::pb()` as well).

examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -65,44 +66,62 @@ impl From<Error> for ExecutionError {
...
+pub struct Transfer {
+    /// `PublicKey` of receiver's wallet.
```
slowli: Why aren't these comments present in the `proto` file?

examples/cryptocurrency-advanced/backend/src/wallet.rs
```diff
@@ -16,30 +16,48 @@
...
+    pub balance: u64,
+    /// Length of the transactions history.
+    pub history_len: u64,
+    /// `Hash` of the transactions history.
+    pub history_hash: Hash,
 }
 
 impl Wallet {
+    /// Create new Wallet.
+    pub fn new(
```
slowli: Is `new` used frequently? I'd think it could be safely removed.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+extern crate quote;
+
+use proc_macro::TokenStream;
+use proc_macro2::{Ident, Span};
+use syn::{Data, DeriveInput, Lit, Meta, NestedMeta, Path};
+
+use std::env;
+
+#[proc_macro_derive(ProtobufConvert, attributes(protobuf_convert))]
+pub fn protobuf_convert_derive(input: TokenStream) -> TokenStream {
```
slowli: Please split into several functions; this one is _far_ too long.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+    let to_pb_fn = {
+        let names1 = field_names.clone();
+        let setters = names1
+            .iter()
+            .map(|i| Ident::new(&format!("set_{}", i), Span::call_site()));
+        let names2 = field_names.clone();
+        quote! {
+            fn to_pb(&self) -> Self::ProtoStruct {
+                let mut msg = Self::ProtoStruct::new();
+                #( msg.#setters(self.#names2.to_pb().into()); )*
```
slowli: If `to_pb()` is not an inherent method (seems it's not), it may make more sense to use its full invocation form specifying the trait.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+        mod #mod_name {
+            use super::*;
+
+            use std::borrow::Cow;
+
+            use #cr::messages::BinaryForm;
+            use #cr::encoding::protobuf::ProtobufConvert;
+            use #cr::storage::StorageValue;
+            use #cr::encoding;
+            use protobuf::Message;
```
slowli: This is awkward; what if there is no `protobuf` crate in scope or it's shadowed by another declaration? IMO, we should reexport all stuff we need.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+            use crypto::{self, CryptoHash, Hash};
```
slowli: What's `crypto` here?

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+    };
+
+    let from_pb_fn = {
+        let names1 = field_names.clone();
+        let getters = names1
+            .iter()
+            .map(|i| Ident::new(&format!("get_{}", i), Span::call_site()));
+        let names2 = field_names.clone();
+        quote! {
+            fn from_pb(pb: Self::ProtoStruct) -> Result<Self, ()> {
```
slowli: Please use FQN for `Result`; it can reasonably be redefined by user code.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+    let cr = if env::var("CARGO_PKG_NAME").unwrap() == "exonum" {
+        quote!(crate)
+    } else {
+        quote!(exonum)
+    };
+
+    let expanded = quote! {
+        mod #mod_name{
+            use super::*;
+            use #cr::blockchain::{Transaction, TransactionSet};
```
slowli: I'm semi-against using *any* imports (except for `super::*`), as they may lead to shadowing and/or weird compile errors.

examples/cryptocurrency-advanced/backend/build.rs
```diff
@@ -0,0 +1,14 @@
+extern crate protoc_rust;
+
+use protoc_rust::Customize;
+
+fn main() {
+    protoc_rust::run(protoc_rust::Args {
+        out_dir: "src/proto",
```
slowli: (Semi) off-topic: there may be some problems pushing generated code to `src`; please see [Cargo docs for the workaround](https://doc.rust-lang.org/cargo/reference/build-scripts.html#case-study-code-generation).

examples/cryptocurrency-advanced/backend/Cargo.toml
```diff
@@ -16,14 +16,20 @@ circle-ci = { repository = "exonum/exonum" }
 
 [dependencies]
 exonum = { version = "0.9.0", path = "../../../exonum" }
+exonum_derive = { path = "../../../exonum_derive" }
 exonum-configuration = { version = "0.9.0", path = "../../../services/configuration" }
+exonum-crypto = { path = "../../../crypto" }
```
slowli: Isn't this crate re-exported via `exonum`?

examples/timestamping/backend/src/transactions.rs
```diff
@@ -46,18 +47,28 @@ impl From<Error> for ExecutionError {
     }
 }
 
-transactions! {
-    /// Transaction group.
-    pub TimeTransactions {
+/// Timestamping transaction.
+#[derive(Serialize, Deserialize, Clone, Debug, ProtobufConvert)]
+#[protobuf_convert("proto::TxTimestamp")]
```
slowli: Nit: I wonder if it's possible to come up with a reasonable default convention to skip specifying this attribute in most cases. Say, place Protobuf files in the `proto` module of the crate, so we can include them via `crate::proto::#name`.

exonum/src/encoding/protobuf/proto/tests.proto
```diff
@@ -0,0 +1,55 @@
+// Copyright 2018 The Exonum Team
```
slowli: If this file is used to test `exonum_derive`, it should be placed in `exonum_derive`, IMO. (Perhaps, as a separate crate, as it seems that cargo doesn't support build scripts for tests / examples.)

dvush:
@slowli 
Thanks for comments, I will address them.
exonum_derive now is not finished and it is written so that it can at least compile.

slowli:
@dvush As a possible source of inspiration for `exonum_derive`, you can refer to [this branch](https://github.com/slowli/exonum/tree/exonum-derive/exonum_derive) (it's abandoned, but IMO has some good code style choices; then again, I may be biased in favor of _my_ code).

examples/cryptocurrency-advanced/backend/build.rs
dvush: They have to get .proto files separately
* maybe there is a way to distribute them with cargo package that I did not find with brief search

examples/cryptocurrency-advanced/backend/src/proto/cryptocurrency.proto
dvush: clang-format can be used for this.

examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -65,44 +66,62 @@ impl From<Error> for ExecutionError {
...
+#[derive(Serialize, Deserialize, Clone, Debug, ProtobufConvert)]
+#[protobuf_convert("proto::CreateWallet")]
+pub struct CreateWallet {
+    /// Name of the new wallet.
+    pub name: String,
+}
+
+/// Transaction group.
+#[derive(Serialize, Deserialize, Clone, Debug, TransactionSet)]
+pub enum WalletTransactions {
```
ilammy: I believe later we could hide all this complexity in Protobuf codegen so that it generates correct Rust structures and enumerations from a *.proto file describing the service.

examples/cryptocurrency/Cargo.toml
```diff
@@ -18,12 +18,19 @@ circle-ci = { repository = "exonum/exonum" }
 
 [dependencies]
 exonum = { version = "0.9.0", path = "../../exonum" }
+exonum-crypto = { path = "../../crypto" }
+exonum_derive = { path = "../../exonum_derive" }
```
ilammy: Please add `version` to path dependencies. It is required in order to use a crate from crates.io.

Here and in other Cargo.toml files.

examples/timestamping/backend/src/proto/timestamping.proto
```diff
@@ -0,0 +1,35 @@
...
+
+message Timestamp {
+  exonum.Hash content_hash = 1;
+  string metadata = 2;
+}
+
+message TimestampEntry {
+  Timestamp timestamp = 1;
+  exonum.Hash tx_hash = 2;
+  google.protobuf.Timestamp time = 3;
```
ilammy: I wonder, is it okay to use `google.protobuf.Timestamp` directly in Protobufs? Maybe it would be better to wrap it into `exonum.DateTime` or something so that it's more clear that it will be mapped onto some Exonum-specific timestamp representation.

exonum_derive/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "exonum_derive"
+version = "0.1.0"
```
ilammy: I guess we can keep the version in sync with the main `exonum` crate.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+              })
+            }
+        }
+    };
+
+    let cr = if env::var("CARGO_PKG_NAME").unwrap() == "exonum" {
+        quote!(crate)
+    } else {
+        quote!(exonum)
+    };
```
ilammy: Isn't there a better way than special-casing `exonum`? It looks like a code smell to me.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+        .map(|f| f.ident.clone().unwrap())
+        .collect::<Vec<_>>();
+
+    let to_pb_fn = {
+        let names1 = field_names.clone();
+        let setters = names1
+            .iter()
+            .map(|i| Ident::new(&format!("set_{}", i), Span::call_site()));
+        let names2 = field_names.clone();
+        quote! {
```
ilammy: It is quite amusing for a person with some Lisp familiarity to see this old friend here.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+                type ProtoStruct = #proto_struct_name;
+
+                #to_pb_fn
+
+                #from_pb_fn
+
+            }
+
+            impl BinaryForm for #name
+            {
```
ilammy: This lone opening brace (together the empty line above) trigger and offend my perfectionism. I guess they are not covered by rustfmt so please don't let them stand out of the code style.

exonum_derive/src/lib.rs
ilammy: The same can be said about `std::borrow::Cow` and prelude imports (like `Into` below).

I'm not an expert in Rust's procedural macros, but it seems that `quote!` does not support hygiene in a good way (so that I can `use protobuf::Message` _outside_ of `quote!` and then just call it `Message` inside). I guess we'll have to explicitly say `use ::protobuf::Message` to import from the top-level `protobuf` crate.

.gitignore
```diff
@@ -26,3 +26,9 @@ npm-debug.log
 
 # protobuf generated
 exonum/src/encoding/protobuf/*.rs
+examples/timestamping/backend/src/proto/*.rs
```
alekseysidorov: Why not just `**/proto/*.rs`?

examples/cryptocurrency-advanced/backend/build.rs
alekseysidorov: I think that we just should create exonum proto files registry in the repository root?

examples/cryptocurrency-advanced/backend/src/transactions.rs
alekseysidorov: Can the rustdoc handle comments in the derive macroses?

examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -65,44 +66,62 @@ impl From<Error> for ExecutionError {
...
+    Transfer(Transfer),
+    /// Issue tx.
+    Issue(Issue),
+    /// CreateWallet tx.
+    CreateWallet(CreateWallet),
 }
 
 impl CreateWallet {
     #[doc(hidden)]
     pub fn sign(name: &str, pk: &PublicKey, sk: &SecretKey) -> Signed<RawTransaction> {
```
alekseysidorov: It seems that method like such can be rewritten like following code.
```rust
impl CreateWallet {
    pub fn sign(self, pk: &PublicKey, sk: &SecretKey) -> Signed<RawTransaction>;
}
```
And also this code can be generated automatically.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,336 @@
...
+                Some(name)
+            }).expect("protobuf_convert attribute not set")
+    };
+    let inside_crate = {
+        let derive_outer = input.attrs.iter().any(|attr| {
+            let meta = match attr.parse_meta() {
+                Ok(m) => m,
+                Err(_) => return false,
+            };
+            meta.name() == "exonum_derive_outer"
```
alekseysidorov: Can you write comment to explain what `derive_outer` do?

exonum_derive/src/lib.rs
dvush: Its one of the best for current rust, but it will be changes in future when you would have ability to write something like `export exonum as self` in exonum crate so you could write code as if exonum is external crate.
Before you could use `pub mod exonum{ pub use super::* } in crate root to do this, but it does not work now.

PR #1046, opened 2018-11-15
Change ProofPath encoding in hash calculations [ECR-2637]

alekseysidorov:
Encode `ProofPath` in `BranchNode::hash` as `proof_path.end() as leb128 || proof_path.bytes()[0..ceil(proof_path.end() / 8)..]`.


exonum/src/storage/proof_map_index/key.rs
```diff
@@ -462,6 +468,53 @@ impl PartialOrd for ProofPath {
     }
 }
 
+impl ProofPath {
+    /// Creates a compressed binary representation of the given proof path.
+    ///
+    /// # Binary format
```
vitvakatu: I believe this requires some expanding and more in-depth explanation

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -462,6 +468,53 @@ impl PartialOrd for ProofPath {
...
+    ///
+    /// - **bits_len** - total length in bits compressed by the leb128 algorithm.
+    /// - **bytes** - non-null bytes of the given ProofPath, i.e. the first `(bits_len + 7) / 8` bytes.
+    pub fn compress(&self) -> SmallVec<[u8; 64]> {
```
vitvakatu: Should it be the part of the public (user-available) API? Are there any scenarios when the user would need it?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -462,6 +468,53 @@ impl PartialOrd for ProofPath {
...
+        let key = &self.raw_key()[0..bytes_len];
+
+        let mut buf = smallvec![0u8; 64];
+        let bytes_written = {
+            let mut writer = Cursor::new(buf.as_mut());
+            let mut bytes_written = leb128::write::unsigned(&mut writer, bits_len).unwrap();
+            bytes_written + writer.write(&key).unwrap()
+        };
+        buf.resize(bytes_written, 0);
+        // Cuts of the bits that lie to the right of the end.
```
vitvakatu:
```suggestion
        // Cut the bits that lie to the right of the end.
```

To the end of what?

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -462,6 +468,53 @@ impl PartialOrd for ProofPath {
...
+        let last_bits_len = bits_len % 8;
```
vitvakatu: I suggest smth like this:
```suggestion
        let integer_bytes = (bits_len % 8) == 0;
```

although I don't like `integer_bytes` naming, but `whole_number_of_bytes` is much longer

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -462,6 +468,53 @@ impl PartialOrd for ProofPath {
...
+        if bytes_len > 0 && last_bits_len != 0 {
+            *buf.as_mut().last_mut().unwrap() &= !(255_u8 << (self.end() % 8));
```
vitvakatu: I definetely don't like that kind of one-liners. Can you explain what's going on here? (better in code comment)

ilammy: Initially I thought this masks off the trailing bits which are not zero for some reason.

Supposed the bit stream contains 13 bits (`bits_len == 13`) which is represented with the following two bytes:

```
0110001011101: stream

01000110 XXX10111: buf
```

then we would like to ensure that trailing bits (marked with X here) are actually zero. However, what we _seem_ to do here is this:

```
01000110 XXX10111: buf
         11111000: 255u8 << (bits_len % 8)
         00000111: !(255_u8 << (bits_len % 8))
01000110 00000111: buf.last_mut() &= !(255_u8 << (bits_len % 8))
```

First, I do not understand why we need to zero them out explicitly. Isn't `self.raw_key()` already property constructed?

Second, I think the correct expression would be 

```suggestion
            *buf.as_mut().last_mut().unwrap() &= 255_u8 >> (self.end() % 8);
```

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -511,6 +564,17 @@ mod tests {
...
+    #[test]
+    fn test_proof_path_compress_fuzz() {
+        let mut rng = rand::thread_rng();
+        for _ in 0..1000 {
+            let key = random_path(&mut rng);
+            let buf = key.compress();
+            let key2 = ProofPath::decompress(buf.as_ref());
+            assert_eq!(key2, key);
+        }
+    }
```
ilammy: In addition to fuzzing tests I would also recommend to add some test for binary representation. Take some simple proof, manually (on paper) construct the expected binary representation, and then compare the algorithm's output to the result.

It would be nice to cover the edge cases there as well:

* shortest possible proof
* longest possible proof
* proof with bit length that's not a multiple of 8

exonum/src/storage/proof_map_index/node.rs
```diff
@@ -84,7 +82,12 @@ impl BranchNode {
 
 impl CryptoHash for BranchNode {
     fn hash(&self) -> Hash {
-        hash(&self.raw)
+        HashStream::new()
+            .update(self.child_hash(ChildKind::Left).as_ref())
+            .update(self.child_hash(ChildKind::Right).as_ref())
+            .update(self.child_path(ChildKind::Left).compress().as_ref())
+            .update(self.child_path(ChildKind::Right).compress().as_ref())
+            .hash()
```
ilammy: Uh... Why do we need to change this?

exonum/src/storage/proof_map_index/key.rs
slowli: > Isn't self.raw_key() already property constructed?

Depends on what you mean by "properly" if a key is constructed as a prefix of another key, the bits are copied from the parent key and not zeroed out after the end of the prefix (for efficiency reasons, seemingly, although I'm not sure we ever  assessed performance impact).

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -462,6 +468,53 @@ impl PartialOrd for ProofPath {
...
+    /// # Binary format
+    ///
+    /// - **bits_len** - total length in bits compressed by the leb128 algorithm.
+    /// - **bytes** - non-null bytes of the given ProofPath, i.e. the first `(bits_len + 7) / 8` bytes.
+    pub fn compress(&self) -> SmallVec<[u8; 64]> {
+        let bits_len = self.end() as u64;
+        let bytes_len = ((bits_len + 7) / 8) as usize;
+        let key = &self.raw_key()[0..bytes_len];
+
+        let mut buf = smallvec![0u8; 64];
```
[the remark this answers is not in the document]
alekseysidorov: Small vec specialisation is absent for the array with given len. Nearest len is 64

exonum/src/storage/proof_map_index/node.rs
```diff
@@ -30,7 +30,7 @@ pub enum Node<T: StorageValue> {
     Branch(BranchNode),
 }
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 pub struct BranchNode {
     raw: Vec<u8>,
```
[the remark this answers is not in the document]
alekseysidorov: `Vec<u8>` buffer returns by the `rocksdb`.

exonum/src/storage/proof_map_index/key.rs
alekseysidorov: This needs only for exonum clients.

exonum/src/storage/proof_map_index/key.rs
alekseysidorov: Of raw buffer.

exonum/src/storage/proof_map_index/key.rs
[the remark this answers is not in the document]
DarkEld3r: Isn't it an implementation detail? Will `smallvec![0u8; 34]` use stack (64 bytes)? If yes, then I propose to request such size.

exonum/src/storage/proof_map_index/node.rs
```diff
@@ -82,12 +84,25 @@ impl BranchNode {
...
+        // a `[u8; 132]` newtype could improve performance.
+        let mut bytes: SmallVec<[u8; 256]> = SmallVec::new();
+        let left_path = self.child_path(ChildKind::Left);
+        let right_path = self.child_path(ChildKind::Right);
+
+        let hashes_end = HASH_SIZE * 2;
+        let right_path_start = hashes_end + left_path.compressed_len();
+        let total_len = right_path_start + right_path.compressed_len();
+        // You can do without `unsafe` by initializing `bytes` to `smallvec![0; len]`;
+        // as far as I can tell, the decrease in performance is negligible.
```
ilammy: So why not go for it then? I'd rather not try to outsmart the compiler by preventively inserting `unsafe`s without profiling results to back them up.

exonum/src/lib.rs
```diff
@@ -65,6 +65,7 @@ extern crate exonum_sodiumoxide as sodiumoxide;
 extern crate failure;
 extern crate futures;
 extern crate hex;
+extern crate leb128;
```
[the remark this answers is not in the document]
alekseysidorov: leb128 is supported crate with large amount of tests without external dependencies.
https://github.com/gimli-rs/leb128
Why not?

exonum/src/storage/proof_map_index/key.rs
alekseysidorov: * absent
* `test_proof_path_compress_leaf`
* `test_proof_path_compress_branch`

exonum/src/lib.rs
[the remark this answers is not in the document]
vitvakatu: I don't think we should implement some algorithms by hand (unless they are our own ones 🙃 ). So let's stick with `leb128`

exonum/Cargo.toml
```diff
@@ -80,4 +82,4 @@ sodiumoxide-crypto = ["exonum_sodiumoxide"]
 
 [build-dependencies]
 protoc-rust = "=2.2.0"
-walkdir = "2.2.7"
+walkdir = "2.2.7"
```
vitvakatu: New line 🙃

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -462,13 +474,72 @@ impl PartialOrd for ProofPath {
...
+    /// # Binary format
+    ///
+    /// - **bits_len** - total length in bits compressed by the leb128 algorithm.
+    /// - **bytes** - non-null bytes of the given ProofPath, i.e. the first `(bits_len + 7) / 8` bytes.
+    pub fn write_compressed(&self, buffer: &mut [u8]) -> usize {
+        let bits_len = u64::from(self.end());
+        let whole_bytes_len = div_ceil!(bits_len, 8) as usize;
+        let key = &self.raw_key()[0..whole_bytes_len];
+
+        let (buffer, bytes_written) = {
```
slowli: Minor nit: you can simplify this code with the help of reborrowing:

```suggestion
        let bytes_written = {
            let mut writer = Cursor::new(&mut *buffer);
            let bytes_written = leb128::write::unsigned(&mut writer, bits_len).unwrap();
            bytes_written + writer.write(&key).unwrap()
        };
```

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -462,13 +474,72 @@ impl PartialOrd for ProofPath {
...
+            let mut writer = Cursor::new(buffer);
+            let mut bytes_written = leb128::write::unsigned(&mut writer, bits_len).unwrap();
+            bytes_written += writer.write(&key).unwrap();
+            (writer.into_inner(), bytes_written)
+        };
+        // Cut the bits that lie to the right of the end.
+        let has_bits_in_last_byte = (bits_len % 8) != 0;
```
slowli: Minor nit: IMO, the following code is slightly more readable:

```suggestion
        let bits_in_last_byte = bits_len % 8;
        if whole_bytes_len > 0 && bits_in_last_byte > 0 {
            let zero_bits_mask = !(255_u8 << bits_in_last_byte);
            buffer[bytes_written - 1] &= zero_bits_mask;
        }
```

exonum/src/storage/proof_map_index/key.rs
```diff
@@ -29,9 +32,18 @@ pub const KEY_SIZE: usize = HASH_SIZE;
...
+pub const PROOF_PATH_KEY_POS: usize = 1;
 /// Position of the byte with total length of the branch.
 pub const PROOF_PATH_LEN_POS: usize = KEY_SIZE + 1;
 
+/// Performs division, rounding the result up.
+macro_rules! div_ceil {
+    ($a:expr, $b:expr) => {
+        ($a + $b - 1) / $b
+    };
+}
```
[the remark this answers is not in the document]
alekseysidorov: This macro is private and I can be sure that nobody will not use complex types as arguments.

dmitry-timofeev:
> Sorry, PR is blocked after JB team communication. We'll review this functionality soon.

https://jira.bf.local/browse/ECR-2637?focusedCommentId=29975&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-29975

exonum-dictionary.txt
```diff
@@ -21,18 +21,17 @@ clonned
 codecov
 compat
 concat
+CORS
```
dvush: Since you have already touched it.
AFAIK cspell is case insensitive, so it can be removed

PR #1052, opened 2018-11-18
Removed a build stage in CI by 1.27.2 version of Rust

aleksuss:
The stage became not actual because too many dependencies became not compatible with 1.27.2 and therefore this stage always is failed.


ozkriff:
Why remove the whole job? Shouldn't it be just upgraded to 1.28 or 1.29?

PR #1055, opened 2018-11-20
Use protobuf for transactions in exonum (p.1)  [ECR-2558] 

dvush:
1) Introduce exonum_derive crate for easy protobuf conversions, transaction enum generation
2) Use it for types inside exonum crate.

TODO:
exonum_derive tests and more refactoring.


exonum/src/encoding/protobuf/mod.rs
```diff
@@ -53,17 +55,17 @@ pub trait ProtobufConvert: Sized {
 
 impl<T> BinaryForm for T
 where
-    T: ProtobufConvert,
-    <T as ProtobufConvert>::ProtoStruct: Message,
+    T: Message,
```
slowli: I'm too lazy to fetch the PR locally right now, so could you please briefly explain the reason for this change?

dvush: The idea is that user can use rust-protobuf generated structs or convertible structs(that implement `ProtobufConvert` trait) when serialization is needed.
The main trait that is required to serialize struct in exonum is`Binary Form` but we cannot implement it for T where T is `ProtobufConvert` and for T that is `Message` because this implementations will conflict. So I chose to genearate explicit implementation of this trait for `ProtobufConvert` in `derive(ProtobufConvert)` and implement generic one for `protobuf::Message`

exonum/tests/explorer/blockchain.rs
```diff
@@ -23,32 +23,60 @@ use exonum::{
         TransactionContext, TransactionSet,
     },
     crypto::{self, Hash, PublicKey, SecretKey},
-    encoding::Error as EncodingError,
+    encoding::{protobuf::tests, Error as EncodingError},
```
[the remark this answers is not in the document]
dvush: Maybe this module can be included with "#[cfg(test)]"

exonum/benches/criterion/block.rs
```diff
@@ -142,29 +142,43 @@ mod timestamping {
...
     }
 
-    transactions! {
-        TimestampingTransactions {
-            struct Tx {
-                data: &Hash,
-            }
+    #[derive(Serialize, Deserialize, Clone, Debug, ProtobufConvert)]
+    #[protobuf_convert("tests::BenchTsTx")]
+    #[exonum_derive_outer]
```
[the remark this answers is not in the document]
dvush: I agree this is bad name, but there is only one place where this solution is needed(benchmarks and external tests in exonum crate).
Yes attribute map will look better.

dvush:
@slowli 
`ProtobufConvert` is just a simpler interface for protobuf structures, mainly to make validations of this structs simpler(like checking hash size, etc.).

`TransactionSet` ==`transactions!` macro that is there now.
I thought that `TransactionSet` will be left in one form or in the other if services will subscribe to generic receive transaction endpoint. But I'll check PoC of this approach to see how multiple txs will be handled.

exonum/src/encoding/protobuf/mod.rs
```diff
@@ -53,17 +55,17 @@ pub trait ProtobufConvert: Sized {
...
 {
     fn encode(&self) -> Result<Vec<u8>, Error> {
-        Ok(self.to_pb().write_to_bytes().unwrap())
+        Ok(self.write_to_bytes().unwrap())
```
alekseysidorov: Is `self.write_to_bytes()` never returns error?

dvush: It can but possible reasons are something like "unable to allocate memory" or smth. I thought that error handling will be revisited when someone will decide to remove old serialization, then all this method will be changed since encoding::Error probably will be gone.

exonum/Cargo.toml
```diff
@@ -51,11 +51,13 @@ chrono = { version = "=0.4.6", features = ["serde"] }
 uuid = { version = "=0.7.1", features = ["serde"] }
 snow = "=0.4.0"
 rust_decimal = "=0.10.2"
-protobuf = "=2.2.0"
+protobuf = { version = "=2.2.0", features = ["with-serde"] }
+ring = "=0.13.2"
```
aleksuss: @dvush As we deleted travis job with build 1.27.2 version of Rust so this dependency not needed anymore.

exonum/src/encoding/protobuf/mod.rs
```diff
@@ -53,17 +56,17 @@ pub trait ProtobufConvert: Sized {
...
```
aleksuss: I guess It will be better, won't it ? 
```suggestion
        self.write_to_bytes().map_err(|e| Error::Other(Box::new(e)))
```

exonum/benches/criterion/block.rs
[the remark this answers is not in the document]
dvush: I fixed it, and named it exonum_root_path = "crate/self". It will be removed after in future rust you could reexport exonum crate from itself under exonum namespace.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+pub fn transaction_set_derive(input: TokenStream) -> TokenStream {
+    tx_set::generate_transaction_set(input)
+}
+
+/// Exonum types should be imported with `crate::` prefix if inside crate
+/// or with `exonum::` when outside.
+fn get_exonum_types_prefix(attrs: &[Attribute]) -> impl quote::ToTokens {
+    let map_attrs = get_exonum_attributes(attrs);
+    let crate_path = map_attrs.into_iter().find_map(|nv| {
+        if nv.ident == "exonum_root_path" {
```
slowli: Nit: it could make sense to extract this value into a constant.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+            match nv.lit {
+                Lit::Str(path) => Some(path.parse::<Path>().unwrap()),
```
slowli: Please include a diagnostic message here via `expect` or a similar construct.

exonum_derive/src/lib.rs
slowli: Nit: could it make sense to match `nv` immediately?

```rust
match nv {
    Attribute { lit: Lit::Str(path), ident, .. } if ident == "exonum_root_path" => { /* ... */ }
    _ => None,
}
```

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+    });
+
+    // If exonum_root_path attribute is defined we use its value
+    if let Some(path) = crate_path {
+        return quote!(#path);
+    }
+
+    // Check cargo env variable to see if we are building inside exonum crate.
+    let pkg_name = env::var("CARGO_PKG_NAME").expect(
```
slowli: I think we should just use `exonum` by default. If we use the macro incorrectly within our crate, it's *our* problem.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+        quote!(crate)
+    } else {
+        quote!(exonum)
+    }
+}
+
+/// Extract attributes in the form of `#[exonum(name = "value")]`
+fn get_exonum_attributes(attrs: &[Attribute]) -> Vec<MetaNameValue> {
+    let exonum_meta = attrs.iter().find_map(|attr| {
+        let meta = match attr.parse_meta() {
```
slowli: Nit: you can seemingly use `Result::ok` and `Option::filter` here:

```rust
attr.parse_meta().ok().filter(|attr| attr.name() == "exonum")
```

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+            Err(_) => return None,
+        };
+        if meta.name() == "exonum" {
+            Some(meta)
+        } else {
+            None
+        }
+    });
+
+    let exonum_meta = if let Some(m) = exonum_meta {
```
slowli: Style nit: IMO, it could be more idiomatic to write

```rust
match exonum_meta {
    Some(Meta::List(meta_list)) => meta_list.nested./*...*/,
    // or
    Some(Meta::List(MetaList { nested: list, .. })) => list./*...*/,

    Some(_) => panic!("`exonum` attribute should not be empty."),
    None => vec![],
}
```

exonum_derive/src/pb_convert.rs
```diff
@@ -0,0 +1,165 @@
...
+        impl #cr::crypto::CryptoHash for #name {
+            fn hash(&self) -> #cr::crypto::Hash {
+                let v = self.to_pb().write_to_bytes().unwrap();
+                #cr::crypto::hash(&v)
+            }
+        }
+
+        impl #cr::storage::StorageValue for #name {
+            fn into_bytes(self) -> Vec<u8> {
+                self.to_pb().write_to_bytes().unwrap()
```
slowli: Nit: I'd switch `unwrap()` to `expect()` with more informative messages here and in `from_bytes`. Also, it could make sense to add a comment explaining why unwrapping is fine here (because we deal with trusted data source).

exonum_derive/src/pb_convert.rs
```diff
@@ -0,0 +1,165 @@
...
+        Data::Struct(x) => x,
+        _ => panic!("Protobuf convert can be derived for structs only."),
+    };
+    data.fields
+        .iter()
+        .map(|f| f.ident.clone().unwrap())
+        .collect()
+}
+
+pub fn generate_protobuf_convert(input: TokenStream) -> TokenStream {
```
slowli: Nit: Inconsistency in method naming - `generate_*` here vs `gen_*` earlier. I'm in favor of a third way, `implement`: `implement_binary_form`, `implement_storage_traits`, etc.

exonum_derive/src/tx_set.rs
```diff
@@ -0,0 +1,171 @@
...
+) -> impl quote::ToTokens {
+    let tx_set_impl = variants.iter().map(|TxSetVariant(n, id, ty)| {
+        quote! {
+            #n => Ok(#name::#id(#ty::decode(&vec)?)),
+        }
+    });
+
+    quote! {
+        impl #cr::blockchain::TransactionSet for #name {
+            fn tx_from_raw(raw: #cr::messages::RawTransaction) -> std::result::Result<Self, _EncodingError> {
```
slowli: Nit: Overly long line.

exonum_derive/src/tx_set.rs
```diff
@@ -0,0 +1,171 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use proc_macro::TokenStream;
+use proc_macro2::{Ident, Span};
+use syn::{Data, DataEnum, DeriveInput, Type};
+
+struct TxSetVariant(u16, Ident, Type);
```
slowli: Nit: I'm against using tuples where their contents aren't immediately obvious.

exonum_derive/src/tx_set.rs
```diff
@@ -0,0 +1,171 @@
...
+fn get_tx_variants(data: &DataEnum) -> Vec<TxSetVariant> {
+    if data.variants.is_empty() {
+        panic!("TransactionSet enum should not be empty");
+    }
+
+    data.variants
+        .iter()
+        .enumerate()
+        .map(|(n, v)| {
+            if v.fields.iter().len() > 1 {
```
slowli: - IMO, it would make sense to extract `v.fields.iter()` into a variable.
- It seems that we don't support named fields in variants, but this code seemingly doesn't check this.

exonum_derive/src/tx_set.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+fn gen_conversions_for_transactions(
+    name: &Ident,
+    variants: &[TxSetVariant],
+    cr: &quote::ToTokens,
+) -> impl quote::ToTokens {
+    let conversions = variants.iter().map(|TxSetVariant(_, id, ty)| {
+        quote! {
+          impl Into<#name> for #ty {
```
slowli: Why not `From<#ty> for #name`? Using `From` is preferable to `Into` if it's possible, as there is corresponding blanket `Into` implementation.

exonum_derive/src/tx_set.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+fn gen_into_service_tx(
+    name: &Ident,
+    variants: &[TxSetVariant],
+    cr: &quote::ToTokens,
+) -> impl quote::ToTokens {
+    let tx_set_impl = variants.iter().map(|TxSetVariant(n, id, _)| {
+        quote! {
+            #name::#id( ref tx) => ( #n, tx.encode().unwrap()),
```
slowli: Is this `encode()` from `BinaryForm`? I'm slightly concerned about the complexity of all these traits...

exonum/benches/criterion/block.rs
```diff
@@ -142,29 +143,47 @@ mod timestamping {
...
     }
 
-    transactions! {
-        TimestampingTransactions {
-            struct Tx {
-                data: &Hash,
-            }
+    #[derive(Serialize, Deserialize, Clone, Debug, ProtobufConvert)]
+    #[exonum(
+        protobuf_convert = "proto::TimestampTx",
```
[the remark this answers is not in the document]
dvush: Yes, I understood your point but I thought that current version which is longer is more easier to understand.
Changed to `pb` and `root`.

exonum/src/encoding/protobuf/mod.rs
dvush: I agree that it can create some confusion but I think that we should left it for now.
Suppose we will leave only one way inside and this way will be - use only ProtobufConvertible structs. Then it may happen that some of the original protobuf features are not available through this ProtobufConvert approach(unless some workaround is used), or this ProtobufConvertible structs can be redundant if they do not use types that need to be validated(like transactions with only one numeric field).
* even though I used ProtobufConvert for this trivial transactions to make code change minimal.

exonum_derive/src/tx_set.rs
dvush: Yes it is from `BinaryForm`.
What complexity are you talking about?
If you talk about serialization than there is only one important trait: `BinaryForm`(there is also `StorageValue` which is `BinaryForm` without fail).
There is also `ProtobufConvert` to make validation of protobuf structs simpler.
Or you talking about Transaction related traits?

exonum_derive/src/tx_set.rs
dvush: I agree, this was copy-pasted  from `transactions!`. Maybe author used it for consistency with other conversions in `transactions!` which you can write in `Into` form only.

exonum/benches/criterion/proto/transactions.proto
```diff
@@ -0,0 +1,38 @@
...
+
+syntax = "proto3";
+
+import "helpers.proto";
+
+package exonum.benchmarks;
+
+message TimestampTx { exonum.Hash data = 1; }
+
+message TimestampPanickingTx { exonum.Hash data = 1; }
```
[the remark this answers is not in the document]
dvush: Yes it is possible, do you suggest to simplify all tests with this feature?

exonum/build.rs
```diff
@@ -35,17 +50,28 @@ fn main() {
...
-        },
-    }).expect("protoc");
+    protoc_generate(
+        "src/encoding/protobuf",
+        "src/encoding/protobuf/proto/",
+        &["src/encoding/protobuf/proto"],
+    );
+
+    // Exonum external tests.
+    protoc_generate(
```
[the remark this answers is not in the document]
dvush: It will not work without creating separate crate for tests or something like that, and I think it is not an issue.
https://github.com/rust-lang/cargo/issues/4789

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+/// or with `exonum::` when outside.
+fn get_exonum_types_prefix(attrs: &[Attribute]) -> impl quote::ToTokens {
+    let map_attrs = get_exonum_attributes(attrs);
+    let crate_path = map_attrs.into_iter().find_map(|nv| match &nv {
+        MetaNameValue {
+            lit: Lit::Str(path),
+            ident,
+            ..
+        }
+            if ident == ROOT_PATH_ATTRIBUTE =>
```
slowli: Nit: Is this formatting produced by `cargo fmt`? Looks pretty much unintelligible, tbh 😞

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        _ => None,
+    });
+
+    // If exonum_root_path attribute is defined we use its value
+    if let Some(path) = crate_path {
+        return quote!(#path);
+    }
+
+    // Check cargo env variable to see if we are building inside exonum crate.
+    let inside_exonum = env::var("CARGO_PKG_NAME")
```
slowli: I think I've already commented on this: IMO, this code is redundant; it increases complexity without a clear purpose (it *should* be relatively easy to remember to use the option inside the Exonum crate).

exonum_derive/src/tx_set.rs
```diff
@@ -0,0 +1,204 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use proc_macro::TokenStream;
+use proc_macro2::{Ident, Span};
+use syn::{Data, DataEnum, DeriveInput, Fields, Type};
+
+struct TxSetVariant {
+    tx_id: u16,
```
slowli: Nit: `tx_` prefix seems redundant.

exonum_derive/src/tx_set.rs
```diff
@@ -0,0 +1,204 @@
...
+        }).collect()
+}
+
+fn implement_conversions_for_transactions(
+    name: &Ident,
+    variants: &[TxSetVariant],
+    cr: &quote::ToTokens,
+) -> impl quote::ToTokens {
+    let conversions = variants.iter().map(
+        |TxSetVariant {
```
slowli: Nit: again, weird formatting. I'd prefer something like

```rust
let conversions = variants.iter().map(|Variant { ident, typ, .. }| {
    // ...
})
```

exonum/benches/criterion/proto/transactions.proto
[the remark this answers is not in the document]
slowli: Yes, I do indeed.

exonum_derive/src/lib.rs
dvush: Ok, I'll change it, but it saves us from using this attribute in exonum/src

exonum_derive/src/lib.rs
dvush: Yes, looks wierd.

exonum_derive/src/tx_set.rs
```diff
@@ -27,55 +31,71 @@ fn get_tx_variants(data: &DataEnum) -> Vec<TxSetVariant> {
...
-          impl Into<#cr::messages::ServiceTransaction> for #ty {
-              fn into(self) -> #cr::messages::ServiceTransaction {
-                  let set: #name = self.into();
-                  set.into()
+    let conversions = variants.iter().map(
+        |TxSetVariant {
+             tx_variant: var,
+             tx_type: ty,
+             ..
+         }| {
```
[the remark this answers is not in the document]
dvush: I made it shorter, as @slowli  suggested and now it can fit in line.

dvush:
@ilammy 
IMO:
1) ProtobufConvert - simple solution to exonum types problem, which is quite generic and does not require to patch protobuf implementation.
2) TransactionSet - I found current derive code more readable than previous `transactions!` macro (content is pretty much the same).

exonum_derive/src/lib.rs
dvush: But I think it is still much better than previous version.

exonum_derive/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+/// or with `exonum::` when outside.
+fn get_exonum_types_prefix(attrs: &[Attribute]) -> impl quote::ToTokens {
+    let map_attrs = get_exonum_attributes(attrs);
+    let crate_path = map_attrs.into_iter().find_map(|nv| match &nv {
+        MetaNameValue {
+            lit: Lit::Str(path),
+            ident,
+            ..
+        }
+            if ident == CRATE_PATH_ATTRIBUTE =>
```
pavel-mukhanov: `cargo fmt` is weird here.

aleksuss: Maybe because it's a `proc_macro` ?

dvush: It is not in macro, this is just code.
Cargo fmt usually ignores arguments of the macros like `quote!`
Here it produced this formatting on purpose.

PR #1057, opened 2018-11-20
ECR-2636 change hashing in ProofListIndex

pavel-mukhanov:
Use separate hashing for leaf nodes and branch nodes according to [certificate transparency ](https://tools.ietf.org/html/rfc6962#section-2.1) specification. Leaf nodes hashes with 0x0 prefix, branch nodes with 0x1.

- [ ] Change hash tree structure(?)
- [ ] Add tests.
- [ ] Fix sanbox tests.

### DIFF
--- a/exonum/src/sandbox/sandbox_tests_helper.rs
+++ b/exonum/src/sandbox/sandbox_tests_helper.rs
@@ -25,6 +25,7 @@ use helpers::{Height, Milliseconds, Round, ValidatorId};
 use messages::{
     Precommit, Prevote, PrevotesRequest, Propose, ProposeRequest, RawTransaction, Signed,
 };
+use storage::proof_list_index::root_hash;
 use storage::Database;
 
 pub type TimestampingSandbox = Sandbox;
@@ -207,7 +208,9 @@ pub fn compute_txs_merkle_root(txs: &[Hash]) -> Hash {
     let mut fork = MemoryDB::new().fork();
     let mut hashes = ProofListIndex::new("name", &mut fork);
     hashes.extend(txs.iter().cloned());
-    hashes.merkle_root()
+    hashes.merkle_root();
+
+    root_hash(txs)
 }
 
 pub fn add_round_with_transactions(
--- a/exonum/src/storage/proof_list_index/mod.rs
+++ b/exonum/src/storage/proof_list_index/mod.rs
@@ -61,12 +61,15 @@ pub struct ProofListIndexIter<'a, V> {
 
 const LEAF_TAG: u8 = 0x0;
 const NODE_TAG: u8 = 0x1;
+const LIST_TAG: u8 = 0x2;
+
+fn hash_value<V: StorageValue>(value: V) -> Hash {
+    let value_bytes = value.into_bytes();
+    hash_with_prefix(LEAF_TAG, &value_bytes)
+}
 
 fn hash_one(h: &Hash) -> Hash {
-    HashStream::new()
-        .update(&[NODE_TAG])
-        .update(h.as_ref())
-        .hash()
+    hash_with_prefix(NODE_TAG, h.as_ref())
 }
 
 fn hash_pair(h1: &Hash, h2: &Hash) -> Hash {
@@ -75,6 +78,18 @@ fn hash_pair(h1: &Hash, h2: &Hash) -> Hash {
         .update(h1.as_ref())
         .update(h2.as_ref())
         .hash()
+}
+
+fn hash_pair_with_prefix(prefix: u8, v1: &[u8], v2: &[u8]) -> Hash {
+    HashStream::new()
+        .update(&[prefix])
+        .update(&v1)
+        .update(&v2)
+        .hash()
+}
+
+fn hash_with_prefix(prefix: u8, value: &[u8]) -> Hash {
+    HashStream::new().update(&[prefix]).update(&value).hash()
 }
 
 impl<T, V> ProofListIndex<T, V>
@@ -337,6 +352,15 @@ where
         self.get_branch(self.root_key()).unwrap_or_default()
     }
 
+    /// TBD
+    pub fn list_hash(&self) -> Hash {
+        HashStream::new()
+            .update(&[LIST_TAG])
+            .update(&[self.len() as u8])
+            .update(self.merkle_root().as_ref())
+            .hash()
+    }
+
     /// Returns the proof of existence for the list element at the specified position.
     ///
     /// # Panics
@@ -454,7 +478,7 @@ where
 
 impl<'a, V> ProofListIndex<&'a mut Fork, V>
 where
-    V: StorageValue,
+    V: StorageValue + Clone,
 {
     fn set_len(&mut self, len: u64) {
         self.base.put(&(), len);
@@ -486,7 +510,12 @@ where
         let len = self.len();
         self.set_len(len + 1);
         let mut key = ProofListKey::new(1, len);
-        self.base.put(&key, value.hash());
+
+        let hash = hash_value(value.clone());
+
+        println!("prooflist value hash {:?}", hash);
+
+        self.base.put(&key, hash);
         self.base.put(&ProofListKey::leaf(len), value);
         while key.height() < self.height() {
             let hash = if key.is_left() {
@@ -498,28 +527,8 @@ where
                 )
             };
             key = key.parent();
-            self.set_branch(key, hash);
-        }
-    }
 
-    ///
-    pub fn push_ct(&mut self, value: V) {
-        let len = self.len();
-        self.set_len(len + 1);
-        let mut key = ProofListKey::new(1, len);
-        self.base.put(&key, value.hash());
-        self.base.put(&ProofListKey::leaf(len), value);
-
-        while key.height() < self.height() {
-            let hash = if key.is_left() {
-                hash_one(&self.get_branch_unchecked(key))
-            } else {
-                hash_pair(
-                    &self.get_branch_unchecked(key.as_left()),
-                    &self.get_branch_unchecked(key),
-                )
-            };
-            key = key.parent();
+            println!("prooflist mode hash {:?}", hash);
             self.set_branch(key, hash);
         }
     }
@@ -579,7 +588,7 @@ where
             );
         }
         let mut key = ProofListKey::new(1, index);
-        self.base.put(&key, value.hash());
+        self.base.put(&key, hash_value(value.clone()));
         self.base.put(&ProofListKey::leaf(index), value);
         while key.height() < self.height() {
             let (left, right) = (key.as_left(), key.as_right());
@@ -650,24 +659,34 @@ where
     }
 }
 
-fn hash_value(value_bytes: &[u8]) -> Hash {
-    let mut value_to_hash = vec![LEAF_TAG];
-    value_to_hash.extend_from_slice(&value_bytes);
-    hash(&value_to_hash)
-}
-
 /// Computes Merkle root hash for a given list of hashes.
 ///
 /// If `hashes` are empty then `Hash::zero()` value is returned.
 pub fn root_hash(hashes: &[Hash]) -> Hash {
     match hashes.len() {
         0 => Hash::zero(),
-        1 => hashes[0],
+        1 => {
+            let hash = hash_value(hashes[0]);
+            println!("hash value {:?}", hash);
+            hash
+        }
         _ => {
-            let mut current_hashes = combine_hash_list(hashes);
+            let hashes: Vec<Hash> = hashes
+                .into_iter()
+                .map(|h| hash_with_prefix(LEAF_TAG, h.as_ref()))
+                .collect();
+
+            let mut current_hashes = combine_hash_list(&hashes);
+
+            println!("current_hashes {:#?}", current_hashes);
             while current_hashes.len() > 1 {
                 current_hashes = combine_hash_list(&current_hashes);
+
+                println!("current_hashes {:#?}", current_hashes);
             }
+            let node_hash = current_hashes[0];
+
+            println!("node_hash {:?}", node_hash);
             current_hashes[0]
         }
     }
@@ -690,11 +709,12 @@ pub fn root_hash_ct(hashes: &[Hash]) -> Hash {
     match len {
         0 => Hash::zero(),
         1 => {
-            let hash = hash_value(hashes[0].as_ref());
+            //            let hash = hash_value(hashes[0].as_ref());
 
-            println!("hash value {:?}", hash);
-            hash
-        },
+            //            println!("hash value {:?}", hash);
+            //            hash
+            unimplemented!()
+        }
         _ => {
             let k = len.next_power_of_two() / 2;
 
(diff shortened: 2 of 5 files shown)


exonum/src/sandbox/sandbox_tests_helper.rs
```diff
@@ -207,7 +208,9 @@ pub fn compute_txs_merkle_root(txs: &[Hash]) -> Hash {
     let mut fork = MemoryDB::new().fork();
     let mut hashes = ProofListIndex::new("name", &mut fork);
     hashes.extend(txs.iter().cloned());
-    hashes.merkle_root()
+    hashes.merkle_root();
+
+    root_hash(txs)
```
slowli: Why tho? I'd thought `ProofListIndex::merkle_root()` should be redefined as a part of this PR; leaving it as is and building a full proof on top just creates confusion, IMO. Now, `merkle_root` can be a confusing name (`list_hash` could be better indeed), but having both `merkle_root` and `list_hash` as public methods is IMO quite confusing.

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -24,13 +24,19 @@ use super::{
...
+use crypto::{Hash, HashStream};
+use byteorder::LittleEndian;
+use bytes::ByteOrder;
 
 mod key;
 mod proof;
 #[cfg(test)]
 mod tests;
 
+pub const LEAF_TAG: u8 = 0x0;
```
slowli: Tags are intended to be global for all `Proof*Index`es, so it could make sense to move them upper in the module hierarchy.

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -59,17 +65,28 @@ pub struct ProofListIndexIter<'a, V> {
...
 fn hash_pair(h1: &Hash, h2: &Hash) -> Hash {
     HashStream::new()
+        .update(&[NODE_TAG])
         .update(h1.as_ref())
         .update(h2.as_ref())
         .hash()
 }
 
+pub fn hash_with_prefix(prefix: u8, value: &[u8]) -> Hash {
+    HashStream::new().update(&[prefix]).update(&value).hash()
```
slowli: If I didn't know that `HashStream` is slow (see discussion in #1046 for details), I'd suggest to store `HashStream::new().update(&[prefix])` for allowed `prefix`es e.g. as `lazy_static` in order to speed up calculations. But I'm not sure they would be sped up at all 🙃 Also, as far as I can see, `HashStream` doesn't implement `Clone` and is passed by value for each `update`, which makes use of `lazy_static` non-trivial. (Don't know whether this is a problem in Rust binding that can be fixed.)

exonum/src/sandbox/sandbox_tests_helper.rs
```diff
@@ -207,7 +208,9 @@ pub fn compute_txs_merkle_root(txs: &[Hash]) -> Hash {
     let mut fork = MemoryDB::new().fork();
     let mut hashes = ProofListIndex::new("name", &mut fork);
     hashes.extend(txs.iter().cloned());
-    hashes.merkle_root()
+    hashes.merkle_root();
```
slowli: What's the point of this code?

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -330,6 +347,18 @@ where
         self.get_branch(self.root_key()).unwrap_or_default()
     }
 
+    /// TBD
+    pub fn list_hash(&self) -> Hash {
+        let mut length_bytes = vec![0; 8];
```
slowli: Why use `vec` here? Just use an array.

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -330,6 +347,18 @@ where
...
+        LittleEndian::write_u64(&mut length_bytes, self.len());
+
+        HashStream::new()
```
slowli: Did you test performance? After #1046, I'm inclined to believe that an implementation without `HashStream` (i.e., allocating an array with length `1 + 8 + HASH_LEN`, writing the parts of the hashed value to it and then hashing the array) could be faster.

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -447,7 +476,7 @@ where
 
 impl<'a, V> ProofListIndex<&'a mut Fork, V>
 where
-    V: StorageValue,
+    V: StorageValue + Clone,
```
slowli: That's a significant change; is it really justified? If `HashStream` worked ideally, I'd suggest to extend `CryptoHash` to

```rust
trait CryptoHash {
    fn hash_to(&self, state: HashStream) -> HashStream;
    // or even
    fn hash_to(&self, state: &mut HashStream);
}
```

But given that `HashStream` doesn't work as efficiently as expected, your solution may be appropriate.

exonum/src/storage/proof_list_index/tests.rs
```diff
@@ -822,3 +830,49 @@ mod root_hash_tests {
         assert_root_hash_correct(&hash_list(&[]));
     }
 }
+
+mod list_hash_tests {
```
slowli: Why is this a separate module?

exonum/src/storage/proof_list_index/tests.rs
```diff
@@ -822,3 +830,49 @@ mod root_hash_tests {
...
+
+        let hash = list_hash(existed_index, &root_hash.as_ref());
+        assert_eq!(hash, actual_list_hash);
+
+        let non_existed_index = 6u64;
+        let hash = list_hash(non_existed_index, &index.merkle_root().as_ref());
+        assert_ne!(hash, actual_list_hash);
+    }
+
+    fn list_hash(index: u64, root_hash: &[u8]) -> Hash {
```
slowli: Why does this duplicate the method in `ProofListIndex`?

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -59,17 +65,28 @@ pub struct ProofListIndexIter<'a, V> {
     base_iter: BaseIndexIter<'a, ProofListKey, V>,
 }
 
+/// TBD
+pub fn hash_leaf<V: StorageValue>(value: V) -> Hash {
```
alekseysidorov: I wonder, can we modify `StorageValue` signature to avoid needless clone here?

exonum/src/storage/proof_list_index/mod.rs
ilammy: If we care for it that much then I'd rather look pinpoint that performance issue in `HashStream` implementation and fix it. Otherwise leave this code as is.

Even if we do care to avoid using `HashStream`, I think that doing

```rust
let mut value = value.to_vec();
value.insert(0, prefix);
```

will be much more concise and readable than any trickery with `lazy_static` and partially-computed hashes.

It's _a single-byte prefix_ and we're computing a hash sum for multi-byte value. Allocating a new hash state and populating it with a single byte should not be significantly more expensive than cloning an existing hash state.

exonum/src/storage/proof_list_index/mod.rs
ilammy: Like this?

```rust
pub trait StorageValue: UniqueHash + Sized {
    fn into_bytes(&self) -> Vec<u8>;

    fn from_bytes(value: &[u8]) -> Self;
}
```

As far as I understand, currently `into_bytes()` consumes `self` and `from_bytes()` takes `Cow<[u8]>` in order to use `std::String::into_bytes()` and `std::Vec::into_owned()` in `StorageValue` impls for `String` and `Vec`. Thus we avoid doing copies of possibly large strings and vectors. (Other types implementing `StorageValue` have fixed size so copying is at least predictable for them.)

However, with proposed changes it seems we will be doing clones of _all_ values so this optimization may be moot.

If this optimization is important then maybe we could _add_ a method to `StorageValue` to avoid breaking everything. For example, `fn as_bytes(&self) -> Cow<[u8]>` which will try avoiding expensive copies.

exonum/src/storage/proof_list_index/mod.rs
ilammy: `StorageValue` requires `UniqueHash`, not `CryptoHash`.

exonum/src/sandbox/sandbox_tests_helper.rs
pavel-mukhanov: Debugging artifact, I forgot to remove it.

exonum/src/storage/proof_list_index/mod.rs
slowli: Sorry for my brainfart; caching hash state most likely won't do anything good, as SHA-256 is processed in chunks of 64 bytes. Still, I'd run benchmarks and would not use `HashStream` if it turns out to be slow. I have a suspicion that `Vec::insert` may be slow, too; creating a vector with sufficient length from the start and then filling it could be faster (in some cases, we can use an array, as we know length of `value`).

pavel-mukhanov: Bench is [here](https://github.com/pavel-mukhanov/sandbox/blob/master/src/proof.rs)

exonum/src/storage/proof_list_index/mod.rs
pavel-mukhanov: I tried to avoid reimplementing `StorageValue` for all types. I thought about extending `UniqueHash` 
```
impl<T: CryptoHash + AsRef<[u8]> UniqueHash for T {
    fn hash(&self) -> Hash {
        CryptoHash::hash(self)
    }

    fn hash_to(&self, stream: HashStream) -> HashStream {
        stream.update(&self.as_ref())
    }
}
```
But in this case `T` is need to be bound with something like `AsRef`. And we also decided to get rid of `HashStream`.

exonum/src/storage/hash.rs
```diff
@@ -31,3 +43,48 @@ impl<T: CryptoHash> UniqueHash for T {
...
+}
+
+/// Convenient method to obtain prefixed value of `Hash`.
+pub fn hash_one(h: &Hash) -> Hash {
+    hash_with_prefix(NODE_TAG, h.as_ref())
+}
+
+/// Convenient method to obtain prefixed value of concatenation of two hashes.
+pub fn hash_pair(h1: &Hash, h2: &Hash) -> Hash {
+    let mut hash_bytes = [0u8; HASH_SIZE * 2 + PREFIX_SIZE];
```
slowli: Do we have a final decision on whether to use `HashStream`? (I'm fine with either of choices.)

Going with the previous comment about an enum with tags, it seems convenient to have the interface for its hashing:

```rust
#[repr(u8)]
pub enum Tag {
    Node = 2,
    // ...
}

impl Tag {
    // if we use streams
    pub fn hash_stream(&self) -> HashStream;
    // or if we don't
    pub fn hash(&self, values: impl Iterator<Item = &[u8]>) -> Hash;
}
```

To go even further, this API could be private, and users may use more high-level abstractions:

```rust
impl Tag {
    fn hash_stream(&self) -> HashStream;

    pub fn hash_node(left_hash: &Hash, right_hash: &Hash) -> Hash {
        Tag::Node
            .hash_stream()
            .update(left_hash.as_ref())
            .update(right_hash.as_ref())
            .hash()
    }
}
```

exonum/src/storage/hash.rs
```diff
@@ -31,3 +43,48 @@ impl<T: CryptoHash> UniqueHash for T {
...
+    hash_bytes[PREFIX_SIZE..HASH_SIZE + PREFIX_SIZE].copy_from_slice(h1.as_ref());
+    hash_bytes[HASH_SIZE + PREFIX_SIZE..HASH_SIZE * 2 + PREFIX_SIZE].copy_from_slice(h2.as_ref());
+    hash(&hash_bytes)
+}
+
+/// Calculate hash value with specified prefix.
+///
+/// Different hashes for leaf and branch nodes are used to secure merkle tree from pre-image attack.
+/// More information here: https://tools.ietf.org/html/rfc6962#section-2.1
+pub fn hash_with_prefix(prefix: u8, value: &[u8]) -> Hash {
```
slowli: See comments about the `Tag` enum above.

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -347,12 +346,9 @@ where
     /// ```
     pub fn get_proof(&self, index: u64) -> ListProof<V> {
         if index >= self.len() {
-            panic!(
-                "Index out of bounds: the len is {} but the index is {}",
-                self.len(),
-                index
-            );
+            return ListProof::Absent(index, self.list_hash());
```
ilammy: I believe we should document this behavior in rustdocs and maybe even mention in changelog. This is a new feature and users should be aware then they can use `get_proof()` with any index to get a proof of absence as well.

exonum/src/storage/proof_list_index/mod.rs
```diff
@@ -377,21 +373,18 @@ where
...
         if to <= from {
             panic!(
                 "Illegal range boundaries: the range start is {:?}, but the range end is {:?}",
                 from, to
             )
         }
 
-        self.construct_proof(self.root_key(), from, to)
+        if to > self.len() {
+            ListProof::Absent(to, self.list_hash())
```
ilammy: This feature should be documented as well. It's not self-evident that we should return a proof of absence of the last element in range if it does not overlap with valid indices.

exonum/src/storage/proof_list_index/proof.rs
```diff
@@ -52,24 +54,30 @@ impl<V: StorageValue + Clone> ListProof<V> {
...
                 if key.height() > 1 {
                     return Err(ListProofError::UnexpectedLeaf);
                 }
                 vec.push((key.index(), value));
-                hash_leaf(value.clone())
+                HashTag::hash_leaf(value.clone())
             }
+            ListProof::Absent(_, _) => {
+                //TODO: modify to use in validate method
+                unreachable!()
```
ilammy: Hm... Is it really unreachable or just `unimplemented!()` at the moment?

exonum/src/storage/proof_list_index/tests.rs
```diff
@@ -838,32 +831,43 @@ mod list_hash_tests {
...
+            }
+            _ => {
+                panic!("Unexpected proof {:?}", proof);
+            }
+        }
+    }
+
+    #[test]
+    fn proof_of_absence_range() {
+        //TBD
```
ilammy: I'd suggest using something like this for marking tests that need to be implemented.

```rust
#[test]
#[ignore]
fn proof_of_absence_range() {
    unimplemented!()
}
```

This way they are more conspicuous in test runner output.

exonum/src/storage/hash.rs
pavel-mukhanov: We decided to use `HashStream`.

exonum/src/storage/proof_list_index/proof.rs
pavel-mukhanov: my mistake, is's `unimplemented()`

exonum/src/node/requests.rs
```diff
@@ -128,7 +128,7 @@ impl NodeHandler {
 
     /// Handles `BlockRequest` message. For details see the message documentation.
     pub fn handle_request_block(&mut self, msg: &Signed<BlockRequest>) {
-        trace!(
+        info!(
```
ilammy: Won't this result in log spam?

exonum/src/storage/proof_list_index/tests.rs
```diff
@@ -827,47 +830,34 @@ mod root_hash_tests {
...
+            assert!(index >= list.len());
+        }
+        _ => {
+            panic!("Unexpected proof {:?}", proof);
         }
-    }
-
-    #[test]
-    fn proof_of_absence_range() {
-        //TBD
```
ilammy: Did we abadon the idea of writing tests for this?

exonum/src/node/requests.rs
pavel-mukhanov: Forgot to remove after debugging.

exonum/src/storage/proof_list_index/tests.rs
pavel-mukhanov: I plan to write it after fixing sandbox tests.

exonum/src/storage/proof_list_index/proof.rs
```diff
@@ -79,12 +84,28 @@ impl<V: StorageValue> ListProof<V> {
     ///
     /// If the proof is valid, a vector with indices and references to elements is returned.
     /// Otherwise, `Err` is returned.
-    pub fn validate(&self, merkle_root: Hash, len: u64) -> Result<Vec<(u64, &V)>, ListProofError> {
+    pub fn validate(
+        &self,
+        expected_list_hash: Hash,
+        len: u64,
```
slowli: Forgot to mention in the review: I see on compelling reasons to provide the list length from the external source. Where are you supposed to get it from in most cases? It could make more sense to supply the length with the proof.

PR #1067, opened 2018-11-23
Improve storage benchmarks

alekseysidorov:
This changes were extracted from the https://github.com/exonum/exonum/pull/1046 PR

- Replace old nightly only benchmarks by the criterion based.
- Use elements/sec metric instead of total time.
- Sort dependencies in the toml files.


exonum/benches/criterion/storage.rs
```diff
@@ -45,58 +52,119 @@ fn generate_random_kv(len: usize) -> Vec<([u8; KEY_SIZE], Vec<u8>)> {
...
+        || {
+            (0..*len)
+                .map(|_| {
+                    let mut chunk = vec![0; 64];
+                    rng.fill_bytes(&mut chunk);
+                    chunk
+                }).collect::<Vec<_>>()
+        },
+        |data| {
+            for item in data {
```
slowli: Just to be clear: the table *is not* cleared on each iteration, is it? Is this intended? I *could* let it slide in case of `ProofMapIndex`, but in case of lists, not clearing the list makes the performance of each successive iteration increasingly worse.

exonum/benches/criterion/storage.rs
```diff
@@ -45,58 +52,119 @@ fn generate_random_kv(len: usize) -> Vec<([u8; KEY_SIZE], Vec<u8>)> {
...
+                table.push(item);
+            }
+        },
+    );
+}
+
+fn proof_map_insert_without_merge<T: Database>(b: &mut Bencher, db: &T, len: &usize) {
+    let data = generate_random_kv(*len);
```
slowli: Nit: Is there a particular reason why `proof_list_append` generates table contents on the setup stage, but here it's generated once? It looks like both approaches are viable, but I'm for consistency.

exonum/benches/criterion/storage.rs
```diff
@@ -45,58 +52,119 @@ fn generate_random_kv(len: usize) -> Vec<([u8; KEY_SIZE], Vec<u8>)> {
...
+fn proof_map_insert_with_merge<T: Database>(b: &mut Bencher, db: &T, len: &usize) {
+    let data = generate_random_kv(*len);
+    b.iter_with_large_setup(
+        || {
+            let patch = {
+                let mut fork = db.fork();
+                {
+                    let mut table: ProofMapIndex<_, Hash, Vec<u8>> =
+                        ProofMapIndex::new(NAME, &mut fork);
+                    table.clear();
```
slowli: Why don't we clear the table on each iteration? Is it expensive enough to distort the benchmark?

exonum/benches/criterion/storage.rs
```diff
@@ -105,34 +173,86 @@ fn create_rocksdb(tempdir: &TempDir) -> RocksDB {
...
 }
 
 pub fn bench_storage(c: &mut Criterion) {
     ::exonum::crypto::init();
 
-    c.bench_function("append merkle table", bench_merkle_table_append_rocksdb);
-    c.bench_function(
-        "insert without merge merkle table",
-        bench_merkle_patricia_table_insertion_fork_rocksdb,
+    let item_counts = vec![1_000, 10_000, 100_000];
```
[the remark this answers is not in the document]
alekseysidorov: For const arrays I would have to write code like following.

```rust
ITEM_COUNTS.as_ref().iter().map(|x| *x)
```

For each `ParameterizedBenchmark`

exonum/benches/criterion/storage.rs
alekseysidorov: No, in each setup I create a new empty fork and wouldn't merge it after bench iteration.

exonum/benches/criterion/storage.rs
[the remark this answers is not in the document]
slowli: ...or something like

```rust
const ITEM_COUNTS: &[usize] = &[1_000, 10_000, 100_000];

fn bench(c: &mut Criterion, id: &str, f: impl Fn(&mut Bencher, usize) + 'static) {
    let benchmark = ParameterizedBenchmark::new(
        "items",
        move |b, &&count| f(b, count),
        ITEM_COUNTS,
    );
    c.bench(
        id,
        benchmark
            .throughput(|&&count| Throughput::Elements(count as u32))
            .plot_config(PlotConfiguration::default().summary_scale(AxisScale::Logarithmic))
            .sample_size(SAMPLE_SIZE),
    );
}

// later...
bench(c, "storage/proof_list/append", bench_proof_list_append);
```

IMO, looks pretty good, although I understand that it's subjective.

exonum/benches/criterion/storage.rs
slowli: The setup closure is just above in the code, and the fork isn't created there (but rather earlier) as of 275ef0c526ae90670ee7a8c57e64b8c9a9570be1 😕 If you add `assert!(table.is_empty())` before the `for` cycle, it will fail.

exonum/benches/criterion/storage.rs
```diff
@@ -45,58 +52,119 @@ fn generate_random_kv(len: usize) -> Vec<([u8; KEY_SIZE], Vec<u8>)> {
...
         exists_keys.insert(k);
-        (k, v)
+        (Hash::new(k), v)
     };
 
     (0..len).map(kv_generator).collect::<Vec<_>>()
 }
 
-fn merkle_table_insertion<T: Database>(b: &mut Bencher, db: &T) {
+fn proof_list_append<T: Database>(b: &mut Bencher, db: &T, len: &usize) {
```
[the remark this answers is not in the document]
ilammy: Oh hey... I did not know about this 'auto-deferencing' in Rust function arguments. Thanks :heart:

Why not just `usize` though? Should it be immutable and un`&mut`able without an explicit `mut` here?

exonum/benches/criterion/storage.rs
slowli: Update: On the second look, I don't really understand what `iter_with_large_setup` tries to accomplish. Have you looked into its docs or implementation? The setup step is supposed to generate some data which is then funneled into iterations. Here, we don't actually generate any data, but rather effectively do a no-op a bunch of times, and then fill the table with the same values (i.e., do something that could be theoretically optimized into no-op on every iteration barring the first one).

It could make sense to generate random values for each iteration and then write them under the same keys; in this case, we reasonably don't need table clearing (although, for surety, we can initially fill the keys before any iterations). *Or* we can generate both keys and values and clear the table on each iteration.

alekseysidorov: Yes `clear` just iterate over the whole keys.

exonum/benches/criterion/storage.rs
alekseysidorov: `assert!(table.keys().next().is_none());` doesn't fail.

PR #1069, opened 2018-11-23
Improve TransactionContext ergonomics

alekseysidorov:
Add some useful methods to the TransactionContext

### DIFF
--- a/services/time/src/transactions.rs
+++ b/services/time/src/transactions.rs
@@ -21,7 +21,7 @@ use exonum::{
     blockchain::{ExecutionError, ExecutionResult, Schema, Transaction, TransactionContext},
     crypto::{PublicKey, SecretKey},
     messages::{Message, RawTransaction, Signed},
-    storage::{Fork, Snapshot},
+    storage::Fork,
 };
 
 use super::SERVICE_ID;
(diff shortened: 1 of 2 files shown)


services/time/src/transactions.rs
```diff
@@ -136,11 +122,13 @@ impl TxTime {
 
 impl Transaction for TxTime {
     fn execute(&self, mut context: TransactionContext) -> ExecutionResult {
+        // Checks that transaction was signed by the validator.
+        context.validator_id().ok_or_else(|| Error::UnknownSender)?;
```
ilammy: This line reads very strange. The comment says we verify that _the transaction_ has been signed by a validator, but the code _does not touch_ `self` which contains the transaction. It only uses the execution context. Yeah, after a while I realize that the execution context actually keeps the authorship information about the transaction, but the line still reads strange.

Anyhow, why do we perform this check only in timestamping service? Shouldn't other services have the same behavior? If so then why not make this check before calling `Transation::execute()` in the core?

alekseysidorov: This check needs only for oracles like `exonum-time` or `exonum-btc-anchoring`, where transactions generates only by the validators.

PR #1070, opened 2018-11-27
Fix Signed<T> deserialization [ECR-2689]

alekseysidorov:
`Signed<T>` serialization contains a bug, that omit payload verification.

### DIFF
--- a/exonum/src/messages/authorization.rs
+++ b/exonum/src/messages/authorization.rs
@@ -144,7 +144,7 @@ impl SignedMessage {
         public_key: &PublicKey,
     ) -> Result<(), Error> {
         if !crypto::verify(signature, &full_buffer, &public_key) {
-            bail!("Can not verify message.");
+            bail!("Cannot verify message.");
         }
         Ok(())
     }
--- a/exonum/src/messages/authorization.rs
+++ b/exonum/src/messages/authorization.rs
@@ -144,7 +144,7 @@ impl SignedMessage {
         public_key: &PublicKey,
     ) -> Result<(), Error> {
         if !crypto::verify(signature, &full_buffer, &public_key) {
-            bail!("Can't verify message.");
+            bail!("Can not verify message.");
         }
         Ok(())
     }
--- a/exonum/src/messages/mod.rs
+++ b/exonum/src/messages/mod.rs
@@ -35,6 +35,7 @@ use byteorder::{ByteOrder, LittleEndian};
 use failure::Error;
 use hex::{FromHex, ToHex};
 use serde::de::{self, Deserialize, Deserializer};
+use serde::ser::{Serialize, Serializer};
 
 use std::{borrow::Cow, cmp::PartialEq, fmt, mem, ops::Deref};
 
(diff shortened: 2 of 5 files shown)


ilammy:
I've always had a feeling that our addiction to those automagically derived `Serialize` and `Deserialize` is no good...

exonum/src/messages/mod.rs
```diff
@@ -284,3 +285,36 @@ impl PartialEq<Signed<RawTransaction>> for SignedMessage {
...
+            payload: P,
+            #[serde(with = "HexStringRepresentation")]
+            message: SignedMessage,
+        };
+
+        let inner: SignedLayout<T> = SignedLayout::deserialize(deserializer)?;
+        // Checks that signed message correspond to the payload.
+        let encoded_payload = inner
+            .payload
+            .encode()
```
ilammy: I think it would be better to check the signature on the payload instead of reencoding it. Is is possible?

I remember that one of the pros of the new message layout was that we don't have to do this weird back-and-forth where deserialization requires serialization and vice versa.

exonum/src/messages/tests.rs
```diff
@@ -151,11 +152,61 @@ fn test_block() {
...
+            &hash(&[1, 2, 3]),
+            &hash(&[3, 2, 1]),
+            ts,
+        ),
+        pub_key,
+        &secret_key,
+    );
+    // Break signature.
+    let raw_len = precommit.message.raw.len();
+    precommit.message.raw[raw_len - 2] /= 2 + 1;
```
dvush: What if `precommit.message.raw[raw_len - 2] == 0`
It will not change. Or it could not be 0?

alekseysidorov: Yes I think that more good solution is to fill signature by zeros.

exonum/src/messages/authorization.rs
```diff
@@ -144,7 +144,7 @@ impl SignedMessage {
         public_key: &PublicKey,
     ) -> Result<(), Error> {
         if !crypto::verify(signature, &full_buffer, &public_key) {
-            bail!("Can't verify message.");
+            bail!("Can not verify message.");
```
ilammy: One-word _cannot_ is much more prevalent when it expresses inability.

Why change it anyway, though? Does that single quote break some tests or something? :)

alekseysidorov: Because `'` is a part of a quotes and sometimes breaks parsers.

ilammy: (IMO, low-quality parsers deserve it.)

exonum/src/messages/tests.rs
```diff
@@ -151,11 +152,65 @@ fn test_block() {
...
+        &secret_key,
+    );
+
+    let precommit_json = serde_json::to_string(&precommit).unwrap();
+    let precommit2: Signed<Precommit> = serde_json::from_str(&precommit_json).unwrap();
+    assert_eq!(precommit2, precommit);
+}
+
+#[test]
+#[should_panic(expected = "Can not verify message.")]
```
vitvakatu: Tests failed, here should be `Cannot`

PR #1072, opened 2018-11-28
Fix documentation for TransactionResult

vitvakatu:

### DIFF
--- a/exonum/src/blockchain/transaction.rs
+++ b/exonum/src/blockchain/transaction.rs
@@ -35,7 +35,7 @@ const TRANSACTION_STATUS_PANIC: u16 = TRANSACTION_STATUS_OK + 1;
 /// failed. Errors consist of an error code and an optional description.
 pub type ExecutionResult = Result<(), ExecutionError>;
 /// Extended version of `ExecutionResult` (with additional values set exclusively by Exonum
-/// framework) that can be obtained through `Schema` `transaction_statuses` method.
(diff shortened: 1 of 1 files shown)


exonum/src/blockchain/transaction.rs
```diff
@@ -35,7 +35,7 @@ const TRANSACTION_STATUS_PANIC: u16 = TRANSACTION_STATUS_OK + 1;
 /// failed. Errors consist of an error code and an optional description.
 pub type ExecutionResult = Result<(), ExecutionError>;
 /// Extended version of `ExecutionResult` (with additional values set exclusively by Exonum
-/// framework) that can be obtained through `Schema` `transaction_statuses` method.
+/// framework) that can be obtained through `Schema::transaction_results` method.
```
ilammy: Maybe make it an actual link so that `cargo deadlinks` will notice and complain when the name changes (again)?

I'd probably do that with _all_ references to structures and methods in documentation. It's quite handy to be able to click if right away instead of bothering with copying and pasting names into the search bar.

vitvakatu: Totally agree with you, but I think it's better to make this change in a separate PR, together with other updates

ozkriff: Maybe it's better to wait for https://github.com/rust-lang/rust/issues/43466 ?

PR #1073, opened 2018-11-28
Update clippy and polish ci files.

alekseysidorov:


exonum/src/lib.rs
```diff
@@ -23,7 +23,6 @@
     bare_trait_objects
 )]
 #![cfg_attr(feature = "long_benchmarks", feature(test))]
-#![cfg_attr(feature = "cargo-clippy", deny(clippy::clippy_pedantic))]
```
ilammy: Why? Was it that annoying? Or is this switch not supported now?

.travis.yml
```diff
@@ -171,16 +158,16 @@ jobs:
  
  # Clippy linting.
   - name: clippy
-    rust: nightly-2018-10-23
+    rust: nightly-2018-11-27
     install:
-    - cargo clippy --version | rustup component add clippy-preview --toolchain=nightly-2018-10-23
+    - cargo clippy --version | rustup component add clippy-preview --toolchain=2018-11-27
```
ilammy: Is _pipe_ (`|`) here intentional? Shouldn't it be

```suggestion
    - cargo clippy --version || rustup component add clippy-preview --toolchain=2018-11-27
```

which installs clippy if it is not already installed?

.travis.yml
```diff
@@ -50,29 +52,12 @@ env:
   global:
   - DEADLINKS_VERS=0.3.0
   - SODIUM_VERS=1.0.13
-  - CARGO_INCREMENTAL=1
```
ilammy: What are Cargo defaults now?

.travis.yml
```diff
@@ -136,24 +121,26 @@ jobs:
...
-        - zlib1g-dev
-        - build-essential
+        - libsnappy-dev
+        - libsodium-dev        
+        - libssl-dev
         - pkg-config
+        - powershell
         - protobuf-compiler
-        - libprotobuf-dev
+        - zlib1g-dev
```
ilammy: Please remind me, why do tests have their own package dependency list? It seems to be more or less the same as the global one.

.travis.yml
```diff
@@ -50,29 +52,12 @@ env:
   global:
   - DEADLINKS_VERS=0.3.0
   - SODIUM_VERS=1.0.13
```
ilammy: I guess we don't need this variable anymore since we install whatever version is in the PPA.

exonum/src/lib.rs
alekseysidorov: Yes.

ilammy: Which one? ;)

alekseysidorov: Oh, sorry, lint has been renamed to `clippy::pedantic`

dvush:
@alekseysidorov Nice
what do you think about removing benchmark test job and adding something like
`cargo +nighly-compiler-version check --benches` which is much faster?

aleksuss:
> @alekseysidorov Nice
> what do you think about removing benchmark test job and adding something like
> `cargo +nighly-compiler-version check --benches` which is much faster?

don't forget to add `--features long_benchmarks`

.travis.yml
alekseysidorov: Test stage contains additional `powershell` package.

.appveyor.yml
```diff
@@ -22,4 +22,4 @@ test_script:
 
 cache:
   - '%USERPROFILE%\.cargo -> Cargo.toml'
-  - target -> Cargo.toml
+  - target -> Cargo.toml
```
aleksuss: It would be good to add new line here 😄

.travis.yml
```diff
@@ -171,18 +132,18 @@ jobs:
...
     script:
     - cargo clippy --all -- -D warnings
 
   # Benchmarks (compilation only).
   - name: benchmarks
-    rust: nightly-2018-10-23
+    rust: nightly-2018-11-28
     script:
-    - RUST_LOG=off cargo bench --verbose --manifest-path exonum/Cargo.toml --features long_benchmarks --no-run
+    - RUST_LOG=off cargo check --benches --features "long_benchmarks"
```
aleksuss: Just out of curiosity, why have you added quotes around `long_benchmards`. As I know it works without it as well.

PR #1076, opened 2018-11-30
Introduce exonum_build

dvush:
Tiny crate to simplify protobuf code generation.

## Overview

---

### Definition of Done


exonum_build/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
```
vitvakatu: license, keywords, description, categories fields are missing.

exonum_build/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+
+extern crate protoc_rust;
+extern crate walkdir;
+
+use protoc_rust::Customize;
+use walkdir::WalkDir;
+
+use std::{env, fs::File, io::Write, path::Path};
+
+/// Find all .proto files in `path` and subfolders.
```
vitvakatu:
```suggestion
/// Finds all .proto files in `path` and subfolders and returns a vector of their paths.
```

exonum_build/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+fn generate_mod_rs(out_dir: &str, proto_files: &[String], mod_file: &str) {
+    let mod_file_content = {
+        proto_files
+            .iter()
+            .map(|f| {
+                let mod_name = Path::new(f)
+                    .file_stem()
+                    .unwrap()
+                    .to_str()
+                    .expect("proto file name is not &str");
```
vitvakatu:
```suggestion
                    .expect(".proto file name is not convertible to &str");
```

exonum_build/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+/// It is impossible to `include!` .rs files generated by rust-protobuf
+/// so we generate piece of `mod.rs` which includes generated files as public submodules.
+///
+/// tests.proto .rs file will be included with `#[cfg(test)]`.
+fn generate_mod_rs(out_dir: &str, proto_files: &[String], mod_file: &str) {
+    let mod_file_content = {
+        proto_files
+            .iter()
+            .map(|f| {
+                let mod_name = Path::new(f)
```
vitvakatu: 🤔 You're transforming Path to String and then again to Path. Seems useless.

exonum_build/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+                }
+            }).collect::<String>()
+    };
+    let dest_path = Path::new(&out_dir).join(mod_file);
+    let mut file = File::create(dest_path).expect("Unable to create output file");
+    file.write_all(mod_file_content.as_bytes())
+        .expect("Unable to write data to file");
+}
+
+/// Generate .rs files from .proto files.
```
vitvakatu:
```suggestion
/// Generates .rs files from .proto files.
```

exonum_build/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+/// In `src/proto/mod.rs`
+/// ```ignore
+/// extern crate exonum;
+///
+/// include!(concat!(env!("OUT_DIR"), "/example_mod.rs"));
+///
+/// // If you use types from exonum .proto files.
+/// use exonum::encoding::protobuf::*;
+/// ```
+pub fn protobuf_generate(input_dir: &str, includes: &[&str], mod_file: &str) {
```
vitvakatu:
```suggestion
pub fn protobuf_generate<P: AsRef<Path>>(input_dir: P, includes: &[&Path], mod_file_name: &str) {
```

Or even `impl Iterator` instead of slice

exonum/build.rs
```diff
@@ -77,16 +20,14 @@ fn main() {
     file.write_all(user_agent.as_bytes())
         .expect("Unable to write data to file");
 
-    protoc_generate(
-        &out_dir,
+    protobuf_generate(
         "src/encoding/protobuf/proto/",
```
alekseysidorov: You can share this path to the other crates via `cargo:stuff=/your/path` and import them by the env variable `EXONUM_STUFF`.
Thus dependent crates will be able to use exonum proto files.

exonum_build/src/lib.rs
dvush: Ok, but the reason it is like this is that protoc crate uses &str everywhere.
Interface may look a little bit better but it will contain same conversions to String.

exonum/build.rs
dvush: I'll do this in the future PR for services.

exonum_build/src/lib.rs
dvush: @aleksuss Thanks for suggesting correct path.
You solution is close but does not work perfectly, it requires you to write
`protobuf_generate(&"src/proto", &["src/proto"], &"example_mod.rs")`
I found correct one to be `impl IntoIterator<Item=impl AsRef<Path>` this way it does not tie together types of first argument of the function(input_dir) and second one(includes).

Also this solution has another flaw - input_dir and mod_file_name types now tied together and it will be type error to use different types which are `AsRef<Path>` there, so if you want to be really generic you have to put another template parameter for mod_file_name.

exonum_build/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }).collect()
+}
+
+/// Workaround for https://github.com/stepancheg/rust-protobuf/issues/324
+/// It is impossible to `include!` .rs files generated by rust-protobuf
+/// so we generate piece of `mod.rs` which includes generated files as public submodules.
+///
+/// tests.proto .rs file will be included with `#[cfg(test)]`.
+fn generate_mod_rs<P: AsRef<Path>, Q: AsRef<Path>>(
```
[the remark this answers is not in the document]
dvush: Bounds are the same but types are not. If first argument will be '&str' second one must me the same in your approach. With this one we can have first argument to be '&str' and second one 'PathBuf'

exonum_build/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+/// // If you use types from exonum .proto files.
+/// use exonum::encoding::protobuf::*;
+/// ```
+pub fn protobuf_generate<P, Q, R, I>(input_dir: P, includes: I, mod_file_name: Q)
+where
+    P: AsRef<Path>,
+    Q: AsRef<Path>,
+    R: AsRef<Path>,
+    I: IntoIterator<Item = R>,
+{
```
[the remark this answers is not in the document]
dvush: I think there are two choices:
1) As I did before use &str everywhere. This one is simple and IMHO sufficient.
2) Do as you say and make everything generic, but it has to be generic all the way down otherwise there is no reason to do it.(IMHO this change make code worse, but I want it to be merged ASAP)

Technical reason: even though trait bounds are the same types for this parameters should not be the same. It can be `&str`, `PathBuf`, `Path`, etc. e.g. mod_file_name can be `&str`, but input_dir can be `PathBuf`.

CHANGELOG.md
```diff
@@ -80,6 +80,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   of the result is now the equivalent of there being no description
   of the result. (#1075)
 
+- `exonum_build` crate has been added to simplify writing `build.rs` files
+  for services that use protobuf code generation. (#1076)
```
aleksuss: This description should be isolate into separate paragraph. See description for other crates (i.e. exonum-time, exonum-configuration)

dvush: I've added it in a new feature paragraph.

exonum_build/src/lib.rs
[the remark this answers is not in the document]
aleksuss: @dvush I propose to finalize such signature like this: 
```rust
pub fn protobuf_generate<P, T, I>(input_dir: P, includes: I, mod_file_name: T)
 where
     P: AsRef<Path>,
     T: AsRef<str>,
     I: IntoIterator<Item = P>;
```
1. `input_dir` and `includes` are always paths and as for me they should bound to Path trait. Because path.  
2. `mode_file_name` - name of module. Should be `&str` or `String`.
@alekseysidorov @ilammy @pavel-mukhanov what do you think ?

exonum_build/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+    let out_dir: PathBuf = env::var("OUT_DIR")
+        .map(|i| i.into())
```
aleksuss:
```suggestion
        .map(PathBuf::from)
```

exonum_build/src/lib.rs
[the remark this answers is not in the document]
aleksuss: Yep. I've made a mistake.

PR #1078, opened 2018-12-04
Testkit to protobuf

dvush:
## Overview

---

### Definition of Done

### DIFF
--- a/examples/cryptocurrency-advanced/backend/build.rs
+++ b/examples/cryptocurrency-advanced/backend/build.rs
@@ -1,9 +1,9 @@
 extern crate exonum_build;
 
-use exonum_build::protobuf_generate;
+use exonum_build::{get_exonum_protobuf_files_path, protobuf_generate};
 
 fn main() {
-    let exonum_protos = exonum_build::get_exonum_protobuf_files_path();
+    let exonum_protos = get_exonum_protobuf_files_path();
     protobuf_generate(
         "src/proto",
         &["src/proto", &exonum_protos],
--- a/examples/cryptocurrency-advanced/backend/build.rs
+++ b/examples/cryptocurrency-advanced/backend/build.rs
@@ -1,10 +1,9 @@
 extern crate exonum_build;
 
 use exonum_build::protobuf_generate;
-use std::env;
 
 fn main() {
-    let exonum_protos = env::var("DEP_EXONUM_PROTOBUF_PROTOS").unwrap();
+    let exonum_protos = exonum_build::get_exonum_protobuf_files_path();
     protobuf_generate(
         "src/proto",
         &["src/proto", &exonum_protos],
--- a/testkit/build.rs
+++ b/testkit/build.rs
@@ -1,0 +1,15 @@
+extern crate protoc_rust;
+
+use protoc_rust::Customize;
+
+fn main() {
+    protoc_rust::run(protoc_rust::Args {
+        out_dir: "src/proto",
+        input: &["src/proto/testkit.proto"],
+        includes: &["src/proto", "../exonum/src/encoding/protobuf/proto"],
+        customize: Customize {
+            serde_derive: Some(true),
+            ..Default::default()
+        },
+    }).expect("protoc");
+}
--- a/testkit/build.rs
+++ b/testkit/build.rs
@@ -2,13 +2,31 @@ extern crate exonum_build;
 
 use exonum_build::protobuf_generate;
 
-use std::env;
-
 fn main() {
-    let exonum_protos = env::var("DEP_EXONUM_PROTOBUF_PROTOS").unwrap();
+    protobuf_generate("src/proto", &["src/proto"], "testkit_protobuf_mod.rs");
+
+    let exonum_protos = exonum_build::get_exonum_protobuf_files_path();
     protobuf_generate(
-        "src/proto",
-        &["src/proto", &exonum_protos],
-        "protobuf_mod.rs",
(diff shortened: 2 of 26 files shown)


testkit/build.rs
```diff
@@ -0,0 +1,32 @@
...
+        &["tests/service_hooks/proto"],
+        "hooks_example_protobuf_mod.rs",
+    );
+
+    protobuf_generate(
+        "examples/timestamping/proto",
+        &["examples/timestamping/proto"],
+        "timestamping_example_protobuf_mod.rs",
+    );
+}
```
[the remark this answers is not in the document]
aleksuss: Or even to create some structure in `exonum_build` crate for holding this data:
```rust
struct ProtoData {
    proto_path: String,
    includes: Vec<String>,
   mod_file_name: String
}
```

examples/cryptocurrency-advanced/backend/build.rs
```diff
@@ -1,10 +1,9 @@
 extern crate exonum_build;
 
 use exonum_build::protobuf_generate;
-use std::env;
 
 fn main() {
-    let exonum_protos = env::var("DEP_EXONUM_PROTOBUF_PROTOS").unwrap();
+    let exonum_protos = exonum_build::get_exonum_protobuf_files_path();
```
ilammy: It would be nicer to import `get_exonum_protobuf_files_files` along with `protobuf_generate`, just as you write in the rustdoc example.

testkit/build.rs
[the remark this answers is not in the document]
dvush: I think they are equivalent because there are no reduction in the code size.
To me current version looks more simple.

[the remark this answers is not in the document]
aleksuss: Really ? What about DRY ?

testkit/examples/timestamping/proto/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Module of the rust-protobuf generated files.
+
+// For protobuf generated files.
+#![allow(bare_trait_objects)]
+#![allow(renamed_and_removed_lints)]
```
ilammy: Is it possible to teach rust-protobuf print this on its own? I'm not a fan of boilerplate that we _have_ to keep forever in out files because someone else does something stupid. rust-protobuf seems to output [quite a few](https://github.com/stepancheg/rust-protobuf/blob/43641931ad427c0d2874f298d6d86c9c128e47a2/protobuf/src/descriptor.rs#L4-L18) clippy annotations, why don't we ask them nicely to add some more?

dvush: Unfortunately no, it is known issue. Maybe it is even fixed, but not released yet.

testkit/build.rs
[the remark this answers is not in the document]
ilammy: I'd support @dvush in that repeated function calls and repeated tuples/structures with a loop (or `map`) have the same readability. We have to repeat either function calls or struct/tuple constructors so it does not make any difference here, so I'm personally fine with function calls. Verbosity here comes from essential complexity of having 4 sets of 3 distinct paths that are not going anywhere.

There may be other considerations for choosing structs (and _not_ tuples) because with something like this

```rust
Proto {
    proto_dir: "tests/service_hooks/proto",
    include_dirs: vec!["tests/service_hooks/proto"],
    rust_module: "hooks_example_protobuf_mod.rs",
}
```

it's easier to see what path stands for what, when compared to

```rust
(
    "tests/service_hooks/proto",
    vec!["tests/service_hooks/proto"],
    "hooks_example_protobuf_mod.rs",
)
```

Such 'named arguments' may actually improve readability.

testkit/examples/timestamping/proto/mod.rs
dvush: Why not just use current workaround without patching generated files? It seems simpler.

Unfortunately we cannot add this annotations to file which we `import!` for the same reason this workaround with `import!`ing piece of the mod file exists.

vitvakatu: I don't think we should waste our limited resources, so current solution is preferable.

ilammy: [This](https://github.com/stepancheg/rust-protobuf/issues/331) [(PR)](https://github.com/stepancheg/rust-protobuf/pull/332) seems to be the tracking issue for this in rust-protobuf. It seems to be blocked by clippy's interface insanity w.r.t. stable vs. nighty and scoped vs. legacy attributes.

testkit/examples/timestamping/proto/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Module of the rust-protobuf generated files.
+
+// For protobuf generated files.
```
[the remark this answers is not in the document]
dvush: This is comment attributed to next two lines - which are attributes that allow some bad style in code generated files.

I think since it is inner doc it is informative enough.

PR #1079, opened 2018-12-05
Add Blockchain::broadcast_raw_transaction method [ECR-2401]

vitvakatu:
## Overview

This method will be used for broadcasting transactions by Java Bindings.
Any ideas about nicer interface are welcome

Additionally, `TransactionContext::new` method made public.

---

See: https://jira.bf.local/browse/ECR-2401

### Definition of Done

- [X] There are no TODOs left in the merged code
- [X] Change is covered by automated tests
- [X] Benchmark results are attached (if applicable)
- [X] The [coding guidelines] are followed
- [X] Public API has proper documentation
- [X] Changelog is updated if needed (in case of notable or breaking changes)

### DIFF
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -289,6 +289,12 @@ impl Blockchain {
     #[doc(hidden)]
     pub fn broadcast_raw_transaction(&self, tx: RawTransaction) -> Result<(), failure::Error> {
         let service_id = tx.service_id();
+        if !self.service_map.contains_key(&service_id) {
+            return format_err!(
+                "Unable to broadcast transaction: no service with ID={} found",
+                service_id
+            );
+        }
         let msg = Message::sign_transaction(
             tx.service_transaction(),
             service_id,
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -290,10 +290,10 @@ impl Blockchain {
     pub fn broadcast_raw_transaction(&self, tx: RawTransaction) -> Result<(), failure::Error> {
         let service_id = tx.service_id();
         if !self.service_map.contains_key(&service_id) {
-            return format_err!(
+            return Err(format_err!(
                 "Unable to broadcast transaction: no service with ID={} found",
                 service_id
-            );
+            ));
         }
         let msg = Message::sign_transaction(
             tx.service_transaction(),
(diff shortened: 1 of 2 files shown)


exonum/src/blockchain/mod.rs
```diff
@@ -286,6 +286,19 @@ impl Blockchain {
         crypto::hash(&vec)
     }
 
+    #[doc(hidden)]
+    pub fn broadcast_raw_transaction(&self, tx: RawTransaction) -> Result<(), failure::Error> {
+        let service_id = tx.service_id();
```
slowli: Should we check that `service_id` is valid (i.e., corresponds to an existing service) here?

exonum/src/blockchain/mod.rs
```diff
@@ -286,6 +286,19 @@ impl Blockchain {
         crypto::hash(&vec)
     }
 
+    #[doc(hidden)]
+    pub fn broadcast_raw_transaction(&self, tx: RawTransaction) -> Result<(), failure::Error> {
```
slowli: Where do you plan to call this method from? Asking to better understand possible invariants we could enforce. E.g., the method is only called from service code, it could make sense to check that the service can only send transactions to itself (not saying that it will necessarily cover all use cases, but it sounds like an understandable restriction).

exonum/src/blockchain/mod.rs
vitvakatu: Probably good idea.

exonum/src/blockchain/mod.rs
vitvakatu: This method is called by internal code of Java Bindings (see NodeContext::submit method [here](https://github.com/exonum/exonum-java-binding/blob/new_message_format/exonum-java-binding-core/rust/src/proxy/node.rs#L67). User will be able to submit some transaction from inside `Service#afterCommit` method, for example.

I don't think we need any restrictions in 0.10, as they probably make some of use cases impossible to implement.

aleksuss: @vitvakatu does it make sense to add the check like this ?
```rust
if self.service_map.contains_key(&service_id) {
```

vitvakatu: @aleksuss i.e. with positive-if instead of negative-if? I've added this check in the latest commit

aleksuss: @vitvakatu yeah i've seen it. It's ok.

PR #1081, opened 2018-12-05
Other services to protobuf

dvush:
## Overview

---

### Definition of Done

### DIFF
--- a/examples/cryptocurrency/Cargo.toml
+++ b/examples/cryptocurrency/Cargo.toml
@@ -18,7 +18,6 @@ circle-ci = { repository = "exonum/exonum" }
 
 [dependencies]
 exonum = { version = "0.9.0", path = "../../exonum" }
-exonum-crypto = { path = "../../crypto" }
 exonum_derive = { version = "0.9.0", path = "../../exonum_derive" }
 failure = "0.1.2"
 serde = "1.0.0"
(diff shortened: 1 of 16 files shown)


examples/cryptocurrency/Cargo.toml
```diff
@@ -18,12 +18,18 @@ circle-ci = { repository = "exonum/exonum" }
 
 [dependencies]
 exonum = { version = "0.9.0", path = "../../exonum" }
+exonum_derive = { version = "0.9.0", path = "../../exonum_derive" }
 failure = "0.1.2"
 serde = "1.0.0"
 serde_derive = "1.0.0"
 serde_json = "1.0.0"
+protobuf = "=2.2.0"
```
vitvakatu: Is strict version number necessary?

aleksuss: I suppose not

vitvakatu: So I think we should remove strict versions if it's possible

aleksuss:
@dvush don't forget about CHANGELOG please

examples/cryptocurrency/Cargo.toml
aleksuss: I guess it's actual for crates with version less then 1.0

dvush:
@vitvakatu 
If you substract .proto files than changes would be small.
And it is not temporary, at least I do not think so, since you have to have .proto files to have interaction between service and client.

examples/cryptocurrency/Cargo.toml
dvush: Ok, I'll remove it in the batch, not in this PR.

PR #1088, opened 2018-12-06
Remove old serialization [ECR-2733]

dvush:
## Overview

---

### Definition of Done


exonum/src/encoding/protobuf/mod.rs
```diff
@@ -247,7 +218,29 @@ impl ProtobufConvert for Vec<u8> {
...
+            }
+        }
+    };
+}
+
+impl_protobuf_convert_scalar!(bool);
+impl_protobuf_convert_scalar!(u32);
+impl_protobuf_convert_scalar!(u64);
+impl_protobuf_convert_scalar!(i32);
+impl_protobuf_convert_scalar!(i64);
```
ilammy: What about `u8`, `u16`, `i8`, `i16`?

examples/cryptocurrency-advanced/backend/src/lib.rs
```diff
@@ -39,11 +39,12 @@ pub mod schema;
 pub mod transactions;
 pub mod wallet;
 
+use self::failure::Error as EncodingError;
```
alekseysidorov: What the following line do? Why not to use `failure::Error` directly?

examples/timestamping/backend/src/lib.rs
```diff
@@ -43,11 +43,12 @@ pub mod proto;
 pub mod schema;
 pub mod transactions;
 
+use failure::Error as StreamStructError;
```
alekseysidorov: Also.

exonum/src/blockchain/tests.rs
```diff
@@ -622,3 +527,23 @@ mod rocksdb_tests {
...
+    fixed_u32: u32,
+    fixed_u64: u64,
+    fixed_i32: i32,
+    fixed_i64: i64,
+    float_32: f32,
+    float_64: f64,
+    boolean: bool,
+    s_i32: i32,
+    s_i64: i64,
+}
```
alekseysidorov: Can you write encode/decode tests? And I think that blockchain mod is not good place for this test.
Maybe we need to revamp modules structure.

exonum/src/encoding/protobuf/mod.rs
```diff
@@ -46,21 +46,20 @@ pub trait ProtobufConvert: Sized {
...
+    fn from_pb(pb: Self::ProtoStruct) -> Result<Self, Error>;
 }
 
 impl<T> BinaryForm for T
 where
     T: Message,
 {
     fn encode(&self) -> Result<Vec<u8>, Error> {
-        self.write_to_bytes().map_err(|e| Error::Other(Box::new(e)))
+        self.write_to_bytes().map_err(|e| e.into())
```
alekseysidorov: You can just write `map_err(Error::from)`

exonum/src/encoding/protobuf/mod.rs
```diff
@@ -74,12 +73,13 @@ impl ProtobufConvert for crypto::Hash {
         hash
     }
 
-    fn from_pb(pb: Hash) -> Result<Self, ()> {
+    fn from_pb(pb: Hash) -> Result<Self, Error> {
         let data = pb.get_data();
         if data.len() == crypto::HASH_SIZE {
```
alekseysidorov: You can use a `ensure!` macro which similar to the `assert!` macro but throws error instead of panic.

exonum/src/messages/mod.rs
```diff
@@ -124,9 +121,9 @@ impl BinaryForm for RawTransaction {
     }
 
     /// Converts a serialized byte array into a transaction.
-    fn decode(buffer: &[u8]) -> Result<Self, encoding::Error> {
+    fn decode(buffer: &[u8]) -> Result<Self, Error> {
         if buffer.len() < mem::size_of::<u16>() {
-            Err("Buffer too short in RawTransaction deserialization.")?
+            bail!("Buffer too short in RawTransaction deserialization.");
```
alekseysidorov: Use `ensure!` https://docs.rs/failure/0.1.3/failure/macro.ensure.html

exonum/src/encoding/protobuf/mod.rs
dvush: u8 will not be supported since Vec<u8> is specific case for protobuf bytes.(conflict with Vec<T: ProtobufConvert>)
other one should be converted with bound checks.

examples/cryptocurrency-advanced/backend/src/lib.rs
dvush: Yeah, this is WIP implementation that at least compiles, It is still not refactored. 
Thanks for comments, btw.

exonum/src/blockchain/tests.rs
dvush: Exactly I am tried changing mod struct - since protobuf is the only encoding now + tests are planned to.

exonum/src/encoding/protobuf/mod.rs
dvush: Ok, I already received similar comments but I preferred this form because you do not name type. But I change it.

exonum/src/encoding/protobuf/mod.rs
```diff
@@ -46,21 +46,20 @@ pub trait ProtobufConvert: Sized {
     fn to_pb(&self) -> Self::ProtoStruct;
 
     /// ProtoStruct -> Struct
-    fn from_pb(pb: Self::ProtoStruct) -> Result<Self, ()>;
+    fn from_pb(pb: Self::ProtoStruct) -> Result<Self, Error>;
```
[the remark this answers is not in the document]
dvush: I used failure here since it is common error type for everyone.

Lets say that we have `E1` type for errors for type `A` and `E2` error type for type `B`. If we include struct `B` into `A` than `E2` should be convertible to `E1`(with ? operator), if we include `A` into `B` than `E1` should be convertible to `E2`(with ? operator).

Obvious solution would be to have common error type for everyone to have freedom in composition.

exonum/src/proto/mod.rs
```diff
@@ -46,21 +92,20 @@ pub trait ProtobufConvert: Sized {
...
+    fn from_pb(pb: Self::ProtoStruct) -> Result<Self, Error>;
 }
 
 impl<T> BinaryForm for T
 where
     T: Message,
 {
     fn encode(&self) -> Result<Vec<u8>, Error> {
-        self.write_to_bytes().map_err(|e| Error::Other(Box::new(e)))
+        self.write_to_bytes().map_err(Error::from)
```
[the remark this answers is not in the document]
dvush: It does support conversion, but this function returns result.

exonum/src/blockchain/service.rs
```diff
@@ -167,9 +172,9 @@ pub trait Service: Send + Sync + 'static {
     /// these operations should be performed in the `Transaction::verify` and
     /// `Transaction::execute` methods.
     ///
-    /// `transactions!` macro generates code that allows simple implementation, see
+    /// `#[derive(TransactionSet)]` macro generates code that allows simple implementation, see
```
aleksuss: @dvush `#[derive...` is an attribute. I guess this section of doc should be changed.

exonum/src/storage/indexes_metadata.rs
```diff
@@ -65,79 +76,37 @@ impl From<u8> for IndexType {
...
-        Ok(())
-    }
-
-    fn serialize_field(&self) -> Result<json::reexport::Value, Box<dyn Error + Send + Sync>> {
-        Ok(json::reexport::Value::from(*self as u8))
+impl From<IndexType> for u32 {
+    fn from(idx: IndexType) -> Self {
+        use self::IndexType::*;
+        match idx {
+            Entry => 0,
```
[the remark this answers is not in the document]
dvush: I agree, this code is no longer needed.

exonum/src/blockchain/tests.rs
```diff
@@ -384,7 +288,7 @@ impl Service for ServiceGood {
         vec![]
     }
 
-    fn tx_from_raw(&self, _raw: RawTransaction) -> Result<Box<dyn Transaction>, MessageError> {
+    fn tx_from_raw(&self, _raw: RawTransaction) -> Result<Box<dyn Transaction>, failure::Error> {
         unimplemented!()
```
alekseysidorov: In my opinion `unreachable` is more appropriate here.

PR #1093, opened 2018-12-11
[WIP] Pb json frontend structs

dvush:
## Overview

---

### Definition of Done


exonum_derive/src/pb_convert.rs
```diff
@@ -140,12 +143,37 @@ fn implement_storage_traits(name: &Ident, cr: &quote::ToTokens) -> impl quote::T
...
+
+    impl<'de> _serde::Deserialize<'de> for #name {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: _serde::Deserializer<'de>,
+        {
+            let pb = <#name as ProtobufConvert>::ProtoStruct::deserialize(deserializer)?;
+            ProtobufConvert::from_pb(pb).map_err(_serde::de::Error::custom)
+        }
+    }
```
ilammy: That's some weird indentation here...

exonum_derive/src/lib.rs
```diff
@@ -27,32 +27,58 @@ mod tx_set;
...
-/// `#[exonum( crate = "path" )]`
-/// Optional. `path` is prefix of the exonum crate(usually "crate" or "exonum")
+///
+/// * `#[exonum( pb = "path" )]`
+/// Required. `path` is the name of the corresponding protobuf generated struct.
+///
+/// * `#[exonum( crate = "path" )]`
+/// Optional. `path` is prefix of the `exonum` crate(usually "crate" or "exonum").
+///
+/// * `#[exonum( serde_pb_convert )]`
```
ilammy: Are we sure that we need this as an optional attribute? I mean, why make it opt-in so that some structs implement Serialize and Deserialize while others do not? Why not automatically implement these traits for all transaction structs without marking them with an attribute?

ilammy:
Some tests that ensure `Serialize`/`Deserialize` impls would also be nice.

exonum_derive/src/lib.rs
dvush: Yes, I agree it can be done by default, but I would like to leave it as an option, because conventional serde can be required for some application.

It can be done by default(and option would be to disable it) but it would require to add feature `with-serde` to .toml file of every crate since only this way rust-protobuf structs derive serde for it's structs.

exonum_derive/src/pb_convert.rs
dvush: Thanks! Thats weird.

components/derive/src/lib.rs
```diff
@@ -27,32 +27,58 @@ mod tx_set;
...
+/// Required. `path` is the name of the corresponding protobuf generated struct.
+///
+/// * `#[exonum( crate = "path" )]`
+/// Optional. `path` is prefix of the `exonum` crate(usually "crate" or "exonum").
+///
+/// * `#[exonum( serde_pb_convert )]`
+/// Optional. Implements `serde::{Serialize, Deserialize}` using structs that were generated with
+/// protobuf.
+/// ```json
+/// // For example struct with `exonum::crypto::Hash` with this
```
slowli: Nit: there are no comments in JSON, so this may render weirdly in the docs.

examples/cryptocurrency-advanced/backend/Cargo.toml
```diff
@@ -32,3 +32,7 @@ hex = "=0.3.2"
 
 [build-dependencies]
 exonum-build = { version = "0.9.0", path = "../../../components/build" }
+
+[features]
+default = ["with-serde"]
```
slowli: Sorry for being stupid: what does this feature influence?

components/derive/src/lib.rs
```diff
@@ -27,32 +27,58 @@ mod tx_set;
...
+/// * Conversion from Transaction types into this enum.
+/// * Conversion from Transaction types and this enum into `ServiceTransaction`.
+/// * Conversion from this enum into `Box<dyn Transaction>`.
 ///
 /// Attributes:
-/// `#[exonum( crate = "path" )]`
-/// Optional. `path` is prefix of the exonum crate(usually "crate" or "exonum")
+///
+/// * `#[exonum( crate = "path" )]`
+/// Optional. `path` is prefix of the `exonum` crate(usually "crate" or "exonum").
```
vitvakatu:
```suggestion
/// Optional. `path` is a prefix of the `exonum` crate (usually "crate" or "exonum").
```
This comment is rather unclear though. I suggest the following:
```
Optional. `path` is a current `exonum` crate alias. Usually it is "exonum", if renaming such as `extern crate exonum as my_exonum_name;` is not used. For internal usage this should be "crate".
```

components/derive/src/lib.rs
```diff
@@ -27,32 +27,58 @@ mod tx_set;
...
-/// Optional. `path` is prefix of the exonum crate(usually "crate" or "exonum")
+///
+/// * `#[exonum( pb = "path" )]`
+/// Required. `path` is the name of the corresponding protobuf generated struct.
+///
+/// * `#[exonum( crate = "path" )]`
+/// Optional. `path` is prefix of the `exonum` crate(usually "crate" or "exonum").
+///
+/// * `#[exonum( serde_pb_convert )]`
+/// Optional. Implements `serde::{Serialize, Deserialize}` using structs that were generated with
```
vitvakatu: This doc comment doesn't mention situations when I should use this attribute

components/derive/src/lib.rs
```diff
@@ -27,32 +27,58 @@ mod tx_set;
 use proc_macro::TokenStream;
 use syn::{Attribute, Lit, Meta, MetaList, MetaNameValue, NestedMeta, Path};
 
-/// Exonum derive attribute names, used as `#[exonum( ATTRIBUTE_NAME = .. )]`
+/// Exonum derive attribute names, used as `#[exonum( ATTRIBUTE_NAME = .. or ATTRIBUTE_NAME )]`
```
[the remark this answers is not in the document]
vitvakatu: As multiple attributes are supported, I suggest the following change:
```suggestion
/// Exonum derive attribute names, used as `#[exonum( [ ATTRIBUTE_NAME = ATTRIBUTE_VALUE or ATTRIBUTE_NAME ],* )]`
```

examples/cryptocurrency-advanced/backend/Cargo.toml
dvush: rust-protobuf generates structs with serde support with the following annotations:
`#[cfg_attr(feature = "with-serde", derive(Serialize, Deserialize))]`

components/derive/src/lib.rs
dvush: I think it could be supported in the future, but I found issue about this closed.
I'll replace it with `text`

dvush:
@slowli 
Sorry for confusing naming. It is not proper protobuf JSON(it is not implemented by rust-protobuf yet).

This is `serde_json` for protobuf generated structs and this representation is more convenient for light client since it can be transformed into protobuf(for ex. to check hash) without any additional complex transformations.

There is other solution of this problem -- send only protobuf bytes to client.

exonum_derive/src/lib.rs
ilammy: But we seem to already litter the crates with `with-serde` features, just look at all the examples which you had to change. I wanted to comment on that first, but then I noticed that this feature is already present in many other places and let it past.

Is there some way to avoid that? I think I read somewhere that this is exactly why usage of features is discouraged unless absolutely necessary and that separate crates usually work better. Would having some _exonum-rust-protobuf_ crate with pre-enabled `with-serde` feature help?

examples/cryptocurrency-advanced/backend/Cargo.toml
ilammy: Maybe we could file an issue to _rust-protobuf_ to stop throwing feature-gates like this into _generated_ code. I'd find it more natural to have some boolean flag in protoc invocation which will tell protoc whether to add `#[derive]` or not. Or have a feature on rust-protobuf itself which enables this everywhere. I'm definitely not a fan of having to add default features to our crates for this.

components/derive/src/lib.rs
dvush: I find second variant too elaborate.

vitvakatu: What do you mean? It's unclear what is `prefix of exonum crate` and what is its purpose.

PR #1096, opened 2018-12-11
Store secret keys in separate files [ECR-2518]

mvlabat:
## Overview

This PR changes the way we store secret keys for our nodes. CLI has been updated for these changes as well.

---

See:
- https://jira.bf.local/browse/ECR-2518
- https://jira.bf.local/browse/ECR-2519

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum
 
+- Node secret keys are now stored in separate files in a secure way.
+  CLI for generating node configs and starting nodes has been extended
+  in order to reflect these changes.
+
 - Trait `TransactionSend` was removed.
   `ApiSender` now contains `broadcast_transaction` method. (#943)
 
@@ -109,7 +113,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum-crypto
 
-- Added `utils` module with functions `create_keys_file` for creating
+- Added `utils` module with functions `generate_keys_file` for creating
   and `read_keys_from_file` for reading files that contain a
   public key and encrypted secret key. (#1056)
 
--- a/exonum/Cargo.toml
+++ b/exonum/Cargo.toml
@@ -53,6 +53,7 @@ uuid = { version = "=0.7.1", features = ["serde"] }
 snow = "=0.4.0"
 rust_decimal = "=0.10.2"
 protobuf = { version = "2.2.0", features = ["with-serde"] }
+rpassword = "2.0.0"
 
 exonum-crypto = { version = "0.9.0", path = "../crypto" }
 exonum_rocksdb = "0.7.4"
--- a/exonum/Cargo.toml
+++ b/exonum/Cargo.toml
@@ -53,7 +53,7 @@ uuid = { version = "=0.7.1", features = ["serde"] }
 snow = "=0.4.0"
 rust_decimal = "=0.10.2"
 protobuf = { version = "2.2.0", features = ["with-serde"] }
-rpassword = "2.0.0"
+rpassword = "=2.1.0"
 
 exonum-crypto = { version = "0.9.0", path = "../crypto" }
 exonum_rocksdb = "0.7.4"
--- a/exonum/src/helpers/fabric/builder.rs
+++ b/exonum/src/helpers/fabric/builder.rs
@@ -14,6 +14,7 @@
 
 use std::{
     collections::HashMap,
+    env,
     ffi::OsString,
     fmt,
     panic::{self, PanicInfo},
@@ -26,10 +27,13 @@ use super::{
     internal::{CollectedCommand, Command, Feedback},
     keys,
     maintenance::Maintenance,
-    CommandName, ServiceFactory,
+    CommandName, Context, ServiceFactory,
 };
 use blockchain::Service;
 use node::Node;
+
+const EXONUM_CONSENSUS_PASS: &str = "EXONUM_CONSENSUS_PASS";
+const EXONUM_SERVICE_PASS: &str = "EXONUM_SERVICE_PASS";
 
 /// `NodeBuilder` is a high level object,
 /// usable for fast prototyping and creating app from services list.
@@ -66,6 +70,9 @@ impl NodeBuilder {
         T: Into<OsString> + Clone,
     {
         let feedback = ClapBackend::execute_cmd_string(&self.commands, cmd_line);
+        if let Feedback::RunNode(ref ctx) = feedback {
+            self.read_run_context(ctx);
+        }
         feedback != Feedback::None
     }
 
@@ -73,17 +80,7 @@ impl NodeBuilder {
     pub fn parse_cmd(self) -> Option<Node> {
         match ClapBackend::execute(&self.commands) {
             Feedback::RunNode(ref ctx) => {
-                let config_file_path = ctx.get(keys::NODE_CONFIG_PATH).ok();
-                let config = ctx
-                    .get(keys::NODE_CONFIG)
-                    .expect("could not find node_config");
-                let db = Run::db_helper(ctx, &config.database);
-                let services: Vec<Box<dyn Service>> = self
-                    .service_factories
-                    .into_iter()
-                    .map(|mut factory| factory.make_service(ctx))
-                    .collect();
-                let node = Node::new(db, services, config, config_file_path);
+                let node = self.read_run_context(ctx);
                 Some(node)
             }
             _ => None,
@@ -138,6 +135,31 @@ impl NodeBuilder {
         ].into_iter()
         .map(|c| (c.name(), CollectedCommand::new(c)))
         .collect()
+    }
+
+    fn read_run_context(self, ctx: &Context) -> Node {
+        let config_file_path = ctx
+            .get(keys::NODE_CONFIG_PATH)
+            .expect("Could not find node_config_path");
+        let config = ctx
+            .get(keys::NODE_CONFIG)
+            .expect("could not find node_config");
+        let db = Run::db_helper(ctx, &config.database);
+        let services: Vec<Box<dyn Service>> = self
+            .service_factories
+            .into_iter()
+            .map(|mut factory| factory.make_service(ctx))
+            .collect();
+
+        let consensus_passphrase = env::var(EXONUM_CONSENSUS_PASS).unwrap_or_default();
+        let service_passphrase = env::var(EXONUM_SERVICE_PASS).unwrap_or_default();
+
+        let config = config.read_secret_keys(
+            &config_file_path,
+            consensus_passphrase.as_bytes(),
+            service_passphrase.as_bytes(),
+        );
+        Node::new(db, services, config, Some(config_file_path))
     }
 }
 
--- a/exonum/src/helpers/fabric/details.rs
+++ b/exonum/src/helpers/fabric/details.rs
@@ -17,17 +17,20 @@
 //! This module implement all core commands.
 // spell-checker:ignore exts, rsplitn
 
-use crypto;
+use crypto::generate_keys_file;
+use rpassword;
 use toml;
 
 use std::{
     collections::{BTreeMap, HashMap},
-    fs,
+    env, fs,
+    io::{self, BufRead, Write},
     net::{IpAddr, SocketAddr},
     path::{Path, PathBuf},
 };
 
 use super::{
+    super::path_relative_from,
     internal::{CollectedCommand, Command, Feedback},
     keys,
     shared::{
@@ -37,10 +40,13 @@ use super::{
 };
 use api::backends::actix::AllowOrigin;
 use blockchain::{config::ValidatorKeys, GenesisConfig};
+use crypto::PublicKey;
 use helpers::{config::ConfigFile, generate_testnet_config};
 use node::{ConnectListConfig, NodeApiConfig, NodeConfig};
 use storage::{Database, DbOptions, RocksDB};
 
+const EXONUM_CONSENSUS_PASS: &str = "EXONUM_CONSENSUS_PASS";
+const EXONUM_SERVICE_PASS: &str = "EXONUM_SERVICE_PASS";
 const DATABASE_PATH: &str = "DATABASE_PATH";
 const OUTPUT_DIR: &str = "OUTPUT_DIR";
 const PEER_ADDRESS: &str = "PEER_ADDRESS";
@@ -50,6 +56,9 @@ const PUBLIC_API_ADDRESS: &str = "PUBLIC_API_ADDRESS";
 const PRIVATE_API_ADDRESS: &str = "PRIVATE_API_ADDRESS";
 const PUBLIC_ALLOW_ORIGIN: &str = "PUBLIC_ALLOW_ORIGIN";
 const PRIVATE_ALLOW_ORIGIN: &str = "PRIVATE_ALLOW_ORIGIN";
+const CONSENSUS_KEY_PATH: &str = "CONSENSUS_KEY_PATH";
+const SERVICE_KEY_PATH: &str = "SERVICE_KEY_PATH";
+const NO_PASSWORD: &str = "NO_PASSWORD";
 
 /// Run command.
 pub struct Run;
@@ -68,7 +77,7 @@ impl Run {
             .unwrap_or_else(|_| panic!("{} not found.", NODE_CONFIG_PATH))
     }
 
-    fn node_config(path: String) -> NodeConfig {
+    fn node_config(path: String) -> NodeConfig<PathBuf> {
         ConfigFile::load(path).expect("Can't load node config file")
     }
 
@@ -184,6 +193,8 @@ impl RunDev {
         let pub_config_path = Self::artifacts_path("public.toml", &ctx);
         let sec_config_path = Self::artifacts_path("secret.toml", &ctx);
         let output_config_path = Self::artifacts_path("output.toml", &ctx);
+        let consensus_key_path = Self::artifacts_path("consensus.toml", &ctx);
+        let service_key_path = Self::artifacts_path("service.toml", &ctx);
 
         // Arguments for common config command.
         ctx.set_arg("COMMON_CONFIG", common_config_path.clone());
@@ -194,6 +205,9 @@ impl RunDev {
         ctx.set_arg("PUB_CONFIG", pub_config_path.clone());
         ctx.set_arg("SEC_CONFIG", sec_config_path.clone());
         ctx.set_arg(PEER_ADDRESS, peer_addr.into());
+        ctx.set_arg(CONSENSUS_KEY_PATH, consensus_key_path);
+        ctx.set_arg(SERVICE_KEY_PATH, service_key_path);
+        ctx.set_flag_occurrences(NO_PASSWORD, 1);
 
         // Arguments for finalize config command.
         ctx.set_arg_multiple("PUBLIC_CONFIGS", vec![pub_config_path.clone()]);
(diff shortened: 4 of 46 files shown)


CHANGELOG.md
```diff
@@ -109,7 +113,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum-crypto
 
-- Added `utils` module with functions `create_keys_file` for creating
+- Added `utils` module with functions `generate_keys_file` for creating
```
aleksuss: 👍

exonum/src/helpers/fabric/details.rs
```diff
@@ -392,6 +406,47 @@ impl GenerateNodeConfig {
...
+                let stdin_locked = stdin.lock();
+                Self::prompt_passphrase(stdin_locked, io::stdout()).ok()
+            }).unwrap()
+            .into_bytes()
+    }
+
+    fn prompt_passphrase<R: BufRead, W: Write>(mut reader: R, mut writer: W) -> io::Result<String> {
+        loop {
+            write!(&mut writer, "Enter passphrase (empty for no passphrase): ").unwrap();
+            let password = rpassword::read_password_with_reader(Some(&mut reader))?;
```
ilammy: I'd suggest using `rpassword::read_password_from_tty` instead of explicitly manipulating stdin and stdout.
It is more correct to read passwords directly from terminal in UNIX world.

I guess you have used `BufRead` exactly to be able to write unit-tests for this function. However, I'm not sure that we need to test this behavior, retrying not that complex and we can trust `rpassword` to do its thing correctly. This test seems to provide little value. It would be more useful to have an integration test which verifies environment variable handling together with that `NO_PASSWORD` thing.

Also, I'm curious: how does this behave if there is no interactive terminal? E.g., when Exonum node is launched as a service with stdin and stdout bound to `/dev/null`?

exonum/src/helpers/fabric/details.rs
```diff
@@ -392,6 +406,47 @@ impl GenerateNodeConfig {
...
+            if password == rpassword::read_password_with_reader(Some(&mut reader))? {
+                return Ok(password);
+            }
+            writeln!(&mut writer, "Passphrases do not match.  Try again.").unwrap();
+        }
+    }
+
+    fn create_secret_key_file(
+        secret_key_path: impl AsRef<Path>,
+        passphrase: impl AsRef<[u8]>,
```
ilammy: I'd suggest calling `get_passphrase` from `create_secret_key_file` directly instead of passing the passphrase in. These is not need to use a passphrase outside of this function, and the function itself is an internal utility rather than general-purpose public function. This should be a nicer API for where it's actually used.

I'm not sure whether passing EXONUM_SERVICE_PASS here will read better though...

exonum/src/helpers/fabric/details.rs
```diff
@@ -392,6 +406,47 @@ impl GenerateNodeConfig {
 
         (external_address, listen_address)
     }
+
+    /// * `key` - expects EXONUM_SERVICE_PASS or EXONUM_CONSENSUS_PASS.
+    fn get_passphrase(context: &Context, key: &str) -> Vec<u8> {
```
ilammy: I'm a bit concerned that someone may reuse `get_passphrase` somewhere later to read an existing passphase, thinking that it's a general passphrase-reading function. Currently it is geared towards this only use: reading a fresh new consensus or service password for key files. Maybe we should consider giving this function more unwieldy name like `get_new_exonum_key_file_passphrase` or something alike so that it stays an internal utility.

exonum/src/helpers/fabric/details.rs
```diff
@@ -392,6 +406,47 @@ impl GenerateNodeConfig {
...
+        }
+    }
+
+    fn create_secret_key_file(
+        secret_key_path: impl AsRef<Path>,
+        passphrase: impl AsRef<[u8]>,
+    ) -> PublicKey {
+        let secret_key_path = secret_key_path.as_ref();
+        if secret_key_path.exists() {
+            panic!("{}: File exists", secret_key_path.to_string_lossy(),);
```
ilammy: These panics and unwraps make me cringe a bit. We can be more nice and return a Result from this function and _then_ panic in one place if we wish to. A function like this should not make decisions on error handling.

I'm not against leaving it as is (because of gazillion of other unwraps that are already present in this module), but following the boy scout rule is good for one's karma.

exonum/src/helpers/fabric/details.rs
```diff
@@ -459,8 +545,24 @@ impl Command for GenerateNodeConfig {
...
+        let service_public_key =
+            Self::create_secret_key_file(&service_secret_key_path, &service_passphrase);
+
+        let pub_config_folder = Path::new(&pub_config_path).parent().unwrap();
+        let consensus_secret_key = if consensus_secret_key_path.is_absolute() {
+            consensus_secret_key_path
+        } else {
+            path_relative_from(&consensus_secret_key_path, &pub_config_folder).unwrap()
+        };
+        let service_secret_key = if service_secret_key_path.is_absolute() {
```
ilammy: It feels that it would be better to have this `is_absolute` branch as a part of `path_relative_from`.

I also believe that it will be more nice to read if the arguments switch places:

```rust
path_relative_from(&pub_config_folder, service_secret_key_path)
```

exonum/src/helpers/fabric/mod.rs
```diff
@@ -79,9 +79,35 @@ pub struct Argument {
     pub required: bool,
     /// Help message.
     pub help: &'static str,
+    /// Used for named arguments: defines whether the argument is a flag or not (false for flags).
+    pub takes_value: bool,
 }
 
 impl Argument {
+    /// Creates a new falg with `long` and optionally `short` names.
```
ilammy:
```suggestion
    /// Creates a new flag with `long` and optionally `short` names.
```

exonum/src/helpers/mod.rs
```diff
@@ -77,3 +79,62 @@ pub fn generate_testnet_config(count: u16, start_port: u16) -> Vec<NodeConfig> {
...
+#[test]
+fn test_path_relative_from() {
+    let baz: PathBuf = "/foo/bar/baz".into();
+    let bar: PathBuf = "/foo/bar".into();
+    let quux: PathBuf = "/foo/bar/quux".into();
+    assert_eq!(path_relative_from(&bar, &baz), Some("../".into()));
+    assert_eq!(path_relative_from(&baz, &bar), Some("baz".into()));
+    assert_eq!(path_relative_from(&quux, &baz), Some("../quux".into()));
+    assert_eq!(path_relative_from(&baz, &quux), Some("../baz".into()));
+    assert_eq!(path_relative_from(&bar, &quux), Some("../".into()));
```
ilammy: I'd like to see more edge cases in tests for a utility function which may be reused in surprising places later.

- relative paths (as both arguments)
- special path components (`.` and `..`)
- trying to escape the root directory (`/foo/../../../../../../../../../../../bar`)
- any other stupid things that you may think of (e.g., `("/", "/")`, referring the directory itself, etc.)

exonum/src/node/mod.rs
```diff
@@ -268,6 +269,51 @@ pub struct NodeConfig {
...
+impl NodeConfig<PathBuf> {
+    /// Converts `NodeConfig<PathBuf>` to `NodeConfig<SecretKey>` reading the key files.
+    pub fn read_secret_keys(
+        self,
+        config_file_path: impl AsRef<Path>,
+        consensus_passphrase: &[u8],
+        service_passphrase: &[u8],
+    ) -> NodeConfig {
+        let config_folder = config_file_path.as_ref().parent().unwrap();
+        let consensus_key_path = if self.consensus_secret_key.is_absolute() {
```
ilammy: I'd suggest to extract this branch into a utility function.

exonum/tests/config.rs
```diff
@@ -32,14 +32,17 @@ use exonum::{
...
-    fs::{File, OpenOptions},
+    fs::{self, File, OpenOptions},
     io::{Read, Write},
     panic,
-    path::Path,
+    path::{Path, PathBuf},
 };
 
+const EXONUM_CONSENSUS_PASS: &str = "EXONUM_CONSENSUS_PASS";
+const EXONUM_SERVICE_PASS: &str = "EXONUM_SERVICE_PASS";
```
ilammy: Can't we somehow export these constants for tests instead of duplicating them? Or it looks too ugly?

exonum/src/helpers/fabric/mod.rs
vitvakatu:
```suggestion
    /// Creates a new flag with `long` and optionally `short` names.
```

exonum/src/helpers/fabric/details.rs
```diff
@@ -416,6 +471,29 @@ impl Command for GenerateNodeConfig {
                 "listen-address",
                 false,
             ),
+            Argument::new_named(
+                CONSENSUS_KEY_PATH,
+                false,
+                "Path to the pem file storing consensus private key (default: ./consensus.toml)",
```
aleksuss: `pem` is redundant here.

exonum/tests/config.rs
mvlabat: I looked through the Exonum code and found out that we do the same for some other constants (I don't remember which ones exactly). If I make them public, I'll have to reexport them in several modules and write documentation... Meh )

exonum/Cargo.toml
```diff
@@ -53,6 +53,7 @@ uuid = { version = "=0.7.1", features = ["serde"] }
 snow = "=0.4.0"
 rust_decimal = "=0.10.2"
 protobuf = { version = "2.2.0", features = ["with-serde"] }
+rpassword = "2.0.0"
```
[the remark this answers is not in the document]
vitvakatu: @ilammy why?

exonum/src/helpers/fabric/details.rs
mvlabat: I didn't want to edit the function which once was in the `stdlib`. I'd suggest to leave it as-is. It was somehow tested and reliable, I fell like it's better to keep that expected original behavior.

exonum/src/helpers/fabric/details.rs
mvlabat: It's a private method for this command.. Why would someone reuse it?
I'd rather not use such a verbose name, though I can add some comments in order to make it clear what this function does.

exonum/src/helpers/fabric/details.rs
mvlabat: Yeah.. I didn't want to pass variable name to this function, that why I kept it this way. Really dunno which way is really better

exonum/Cargo.toml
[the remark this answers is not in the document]
aleksuss: @ilammy we pin to a latest version if version less than 1.0.

exonum/src/helpers/fabric/details.rs
mvlabat: > This test seems to provide little value.

I think it's important to test our prompt behavior: what exactly is displayed to used and under what conditions. I can't see a way to do it in our integration tests.

> Also, I'm curious: how does this behave if there is no interactive terminal? E.g., when Exonum node is launched as a service with stdin and stdout bound to `/dev/null`?

You can always pass the password with environment variables, so you won't be prompted to enter the password manually.

exonum/src/helpers/fabric/details.rs
aleksuss: @mvlabat To hold a passphrase in `Vec` it's an unsecured approach. For that propose you can use https://docs.rs/smallvec/0.6.6/smallvec/struct.SmallVec.html.

exonum/src/node/mod.rs
```diff
@@ -268,6 +269,51 @@ pub struct NodeConfig {
...
+            config_folder.join(&self.consensus_secret_key)
+        };
+        let service_key_path = if self.service_secret_key.is_absolute() {
+            self.service_secret_key
+        } else {
+            config_folder.join(&self.service_secret_key)
+        };
+
+        let consensus_secret_key = read_keys_from_file(&consensus_key_path, consensus_passphrase)
+            .expect("Could not read consensus_secret_key from file")
```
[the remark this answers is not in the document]
mvlabat: It's always appended to the `expect` message. See
https://github.com/rust-lang/rust/blob/master/src/libcore/result.rs#L835
https://github.com/rust-lang/rust/blob/master/src/libcore/result.rs#L1008

exonum/src/helpers/fabric/details.rs
mvlabat: Saw the same comment for your PR and forgot about it.. Shame! )

exonum/src/helpers/fabric/details.rs
ilammy: If it were still around in stdlib as a deprecated function then I'd suggest using it instead. However, it was removed then I don't really buy the sentimental argument of keeping all the old quirks in naming, behavior, and argument order when we can do better now. It's not like new developers looking at this code will remember the old behavior. Plus, we have our own tests to verify that this function behaves as we expect it to.

Though, argument naming and order is a subjective matter so I'm most certainly not blocking this PR if you believe that we should keep the original naming and argument order.

exonum/tests/config.rs
ilammy: I'm not talking about making a really 'public' public export, with docs and stuff. Something like this in the Exonum module

```rust
#[cfg(test)]
pub use module::of::the::EXONUM_SERVICE_PASS;
```

Not sure if it will work though. If it's too hard or too ugly to read then forget it.

exonum/Cargo.toml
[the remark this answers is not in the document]
ilammy: Hm... that's news to me. I thought we pin _everything_. I guess that's a false impression made by 99% of Rust stuff still having 0.x.x versions :rofl: 

However, I also see some 1.0.0+ crates being pinned, such as `os_info`, `protobuf`. As well as some pre-1.0.0 stuff not pinned, such as `failure`, `erased_serde`.

I guess we need to spell this rule out in CONTRIBUTING.md or somewhere if that's our strategy on dependency management.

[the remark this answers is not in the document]
aleksuss: Be sure we will do it

exonum/src/helpers/fabric/details.rs
mvlabat: Yeah, I haven't brought up the main argument and I'm talking not against renaming the function right now.. I don't want to put the branch there. It really has nothing to do with the purpose of that function. The logic is to keep absolute paths as-is, because that's what user, I think, expects to see in the config if they pass an absolute path to the command line.

exonum/src/helpers/fabric/details.rs
ilammy: I'm not sure how `SmallVec` helps there. Isn't it (slightly) insecure because we should securely zero out memory before dropping it?

exonum/src/helpers/fabric/details.rs
ilammy: > You can always pass the password with environment variables, so you won't be prompted to enter the password manually.

But what if I forget to set the environment variable? Will Exonum node hang indefinitely, assume an empty password, or prompt the user on some random terminal for the password?

exonum/src/helpers/fabric/builder.rs
```diff
@@ -139,6 +136,31 @@ impl NodeBuilder {
...
+            .get(keys::NODE_CONFIG)
+            .expect("could not find node_config");
+        let db = Run::db_helper(ctx, &config.database);
+        let services: Vec<Box<dyn Service>> = self
+            .service_factories
+            .into_iter()
+            .map(|mut factory| factory.make_service(ctx))
+            .collect();
+
+        let consensus_passphrase = env::var(EXONUM_CONSENSUS_PASS).unwrap_or_default();
```
slowli: Am I reading this correctly that the password can only be input from env when starting a node? This doesn't sound right.

exonum/src/helpers/fabric/details.rs
```diff
@@ -416,6 +471,29 @@ impl Command for GenerateNodeConfig {
...
+            Argument::new_named(
+                SERVICE_KEY_PATH,
+                false,
+                "Path to the file storing service private key (default: ./service.toml)",
+                "s",
+                "service-path",
+                false,
+            ),
+            Argument::new_flag(
+                NO_PASSWORD,
```
slowli: What is the utility of this flag? I'm not an expert in designing secure UX, but I think that the *only* case where we might not ask for a password is creating a test network. We would want to provide domain separation (i.e., make it impossible to run a test network node in the "real" network, or vice versa; this might include different address format if/when address appear in Exonum) and otherwise make clear that a test network is not secure and should not be used in production.

exonum/src/helpers/fabric/details.rs
```diff
@@ -392,6 +406,47 @@ impl GenerateNodeConfig {
 
         (external_address, listen_address)
     }
+
+    /// * `key` - expects EXONUM_SERVICE_PASS or EXONUM_CONSENSUS_PASS.
+    fn get_passphrase(context: &Context, key: &str) -> Vec<u8> {
+        context
+            .get_flag_occurrences(NO_PASSWORD)
+            .map(|_| "".to_string())
+            .or_else(|| env::var(key).ok())
```
slowli: Using a fixed env var can potentially leak a passphrase from another Exonum instance here, right? Doesn't look like a good decision to me.

exonum/src/helpers/fabric/details.rs
```diff
@@ -392,6 +406,47 @@ impl GenerateNodeConfig {
...
+            }).unwrap()
+            .into_bytes()
+    }
+
+    fn prompt_passphrase<R: BufRead, W: Write>(mut reader: R, mut writer: W) -> io::Result<String> {
+        loop {
+            write!(&mut writer, "Enter passphrase (empty for no passphrase): ")?;
+            let password = rpassword::read_password_with_reader(Some(&mut reader))?;
+            write!(&mut writer, "Enter same passphrase again: ")?;
+            if password == rpassword::read_password_with_reader(Some(&mut reader))? {
```
slowli: I'd check that the password is non-empty, tbh. Empty passwords are acceptable for test networks, but otherwise an obvious security issue.

exonum/src/helpers/fabric/details.rs
```diff
@@ -392,6 +406,47 @@ impl GenerateNodeConfig {
...
+        passphrase: impl AsRef<[u8]>,
+    ) -> PublicKey {
+        let secret_key_path = secret_key_path.as_ref();
+        if secret_key_path.exists() {
+            panic!("{}: File exists", secret_key_path.to_string_lossy(),);
+        } else {
+            if let Some(dir) = secret_key_path.parent() {
+                fs::create_dir_all(dir).unwrap();
+            }
+            generate_keys_file(&secret_key_path, &passphrase).unwrap()
```
slowli: Slightly off-topic: I'd be wary of using `unwrap()`s: they can leak diagnostic information useful to an adversary.

exonum/src/node/mod.rs
```diff
@@ -268,6 +269,51 @@ pub struct NodeConfig {
...
+            .expect("Could not read service_secret_key from file")
+            .1;
+        NodeConfig {
+            consensus_secret_key,
+            service_secret_key,
+            genesis: self.genesis,
+            listen_address: self.listen_address,
+            external_address: self.external_address,
+            network: self.network,
+            consensus_public_key: self.consensus_public_key,
```
slowli: What happens if the keypair in the key file doesn't correspond to this public key?

exonum/src/helpers/fabric/builder.rs
mvlabat: As far as I know, that's what on we agreed for the first iteration.

exonum/src/helpers/fabric/details.rs
mvlabat: Sounds reasonable... But it's not quite clear to me what kind of interface we should provide instead

exonum/src/node/mod.rs
mvlabat: Nothing right now. Should we throw an error in this case?

I really discovered that we store public key in the key files when I published the PR. I think that another option is to remove public keys from our final config and read them from key files, so we can get rid of redundancy and duplication of data. What do you think about it?

exonum/src/node/mod.rs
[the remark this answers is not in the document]
vitvakatu: Okay, sorry

exonum/src/helpers/fabric/details.rs
```diff
@@ -416,6 +471,29 @@ impl Command for GenerateNodeConfig {
...
+                false,
+                "Path to the pem file storing service private key (default: ./service.toml)",
+                "s",
+                "service-path",
+                false,
+            ),
+            Argument::new_flag(
+                NO_PASSWORD,
+                "Don't prompt for passwords when generating private keys (leave empty)",
+                "n",
```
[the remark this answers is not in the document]
vitvakatu: Well, I think this functionality is potentially dangerous and every user should understand, that he's using "--no-password" flag, not some invisible "-n" :)

If you think that it's okay - I will not mind against.

exonum/src/node/mod.rs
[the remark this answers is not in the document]
mvlabat: No problem :)

dvush:
Moved to #1222

exonum/src/helpers/fabric/details.rs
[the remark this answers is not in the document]
dvush: I've left it as is.

exonum/src/helpers/fabric/details.rs
dvush: I think since this function is used in "testnet" where there is no passphrase it can be left as is.
I changed scoping to address your comment in new PR.

exonum/src/helpers/fabric/details.rs
dvush: I used https://crates.io/crates/zeroize here.

exonum/src/helpers/fabric/details.rs
dvush: I thing that current arguments named better since function is called "path relative from" - first argument is "path" second argument "from".
Regarding moving is_absolute branch into some function I would consider premature since there are only two use cases, also this thing with paths may change(see comments later)

exonum/src/helpers/fabric/details.rs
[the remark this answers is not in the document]
dvush: It makes sence but for now I left it as is. Maybe UX will be slightly changed keeping this in mind.

exonum/src/helpers/fabric/details.rs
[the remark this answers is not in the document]
dvush: I've added zeroizing Vec<u8> and string produced by read_password.

exonum/src/helpers/mod.rs
dvush: This function does not work some of this cases properly.

>relative paths

Sort of works.

>special path components
>trying to escape the root directory

it does not "collapse" arguments so it does not work in all cases.

>("/", "/"), referring the directory itself

It will return ""...

PR #1102, opened 2018-12-13
Fix windows path issue

dvush:
## Overview

### Definition of Done

### DIFF
--- a/components/build/src/lib.rs
+++ b/components/build/src/lib.rs
@@ -36,7 +36,7 @@ fn get_proto_files<P: AsRef<Path>>(path: &P) -> Vec<PathBuf> {
         .filter_map(|e| {
             let e = e.ok()?;
             if e.path().extension()?.to_str() == Some("proto") {
-                Some(e.path().into())
+                Some(clean_path(e.path()))
             } else {
                 None
             }
@@ -112,6 +112,9 @@ where
     I: IntoIterator<Item = R>,
     T: AsRef<str>,
 {
+    let input_dir = clean_path(input_dir);
+    let includes = includes.into_iter().map(clean_path).collect::<Vec<_>>();
+
     let out_dir = env::var("OUT_DIR")
         .map(PathBuf::from)
         .expect("Unable to get OUT_DIR");
@@ -132,8 +135,7 @@ where
         includes: &includes
             .iter()
             .map(|s| {
-                s.as_ref()
-                    .to_str()
+                s.to_str()
                     .expect("Include dir name is not convertible to &str")
             }).collect::<Vec<_>>(),
         customize: Customize {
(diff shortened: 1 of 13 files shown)


components/build/src/lib.rs
```diff
@@ -173,3 +174,13 @@ where
...
     env::var("DEP_EXONUM_PROTOBUF_PROTOS").expect("Failed to get exonum protobuf path")
 }
+
+// The problem is that
+// 1) On Windows path can contain both / and \.
+// 2) `protoc_rust` compares paths as strings but same path can have different representation as a string.
+fn clean_path<P: AsRef<Path>>(path: P) -> PathBuf {
+    path.as_ref()
+        .to_str()
+        .map(|s| PathBuf::from(s.replace("\\", "/")))
```
aleksuss: I against such hacks in the code. I'd prefer to create PR into `protobuf-rust` which will fix that issue.

components/build/src/lib.rs
```diff
@@ -173,3 +174,13 @@ where
 pub fn get_exonum_protobuf_files_path() -> String {
     env::var("DEP_EXONUM_PROTOBUF_PROTOS").expect("Failed to get exonum protobuf path")
 }
+
+// The problem is that
+// 1) On Windows path can contain both / and \.
+// 2) `protoc_rust` compares paths as strings but same path can have different representation as a string.
+fn clean_path<P: AsRef<Path>>(path: P) -> PathBuf {
```
ilammy: I think that this may be used as a <i>temporary</i>™ workaround, but it sounds like a bug that has to be reported to **rust-protobuf**. I would strongly consider taking time to pinpoint and report it (there doesn't seem to be a similar issue yet). The whole reason behind `std::fs::Path` existence is to prevent exactly this kind of bugs.

I'd also rename this function to `canonicalize_separators` because 'clean' is not obvious enough.

It would also be great to see some basic tests for this function so that its effect is more clear.

I'm also slightly curious whether this implementation correctly handles all intricacies of Windows paths, like disk labels (`C:\Some\Directory` not turning into `/C:/Some/Directory` with a leading slash) and UNC paths (`\\some\path` not turning into `/some/path`, with a single slash). This `to_str` also looks fishy because strictly speaking paths are not required to be valid Unicode text (on Windows and UNIX alike). Another thing that I would like to note is that backslash `\` can be used in file names on UNIX platforms so we probably should not convert it into a directory separator.

However, I'm willing to sacrifice a bit of my karma and can ignore all these minutiae considering that this is a workaround which hopefully will be removed in the future. But I would strongly prefer this to be accompanied by at least a bug report to rust-protobuf (or a fix in rust-protobuf, but that may be to costly to do at the moment), as well as a giant comment `// THIS IS A HACK DON'T DO LIKE THIS IN PRODUCTION CODE KIDS`.

PR #1165, opened 2019-01-04
CHANGELOG actualization

aleksuss:
Added info about patch releases.

### DIFF
--- a/components/derive/Cargo.toml
+++ b/components/derive/Cargo.toml
@@ -8,7 +8,7 @@ documentation = "https://docs.rs/exonum-derive"
 license = "Apache-2.0"
 keywords = ["protobuf", "exonum"]
 categories = ["development-tools"]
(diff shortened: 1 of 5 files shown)


components/derive/Cargo.toml
```diff
@@ -5,10 +5,10 @@ authors = ["The Exonum Team <exonum@bitfury.com>"]
 homepage = "https://exonum.com/"
 repository = "https://github.com/exonum/exonum"
 documentation = "https://docs.rs/exonum-derive"
-readme = "README.md"
 license = "Apache-2.0"
 keywords = ["protobuf", "exonum"]
 categories = ["development-tools"]
+description = "Helper macroses for serialization structs in protobuf."
```
popzxc: AFAIK "macros" is already plural and shouldn't be pluralized even more :)

PR #1166, opened 2019-01-04
WIP: Update rust version

popzxc:


aleksuss:
@popzxc what do you think about to update nightly rust too ?

PR #1230, opened 2019-01-17
Update to Rust 2018 edition [ECR-2816]

ilammy:
## Overview

We'll have to modernize the code at some point. Let's do it now.

Usually it is enough to do `cargo fix --edition && cargo fix --edition-idioms && cargo fmt`, but there are some things which require manual editing. Sometimes `cargo fix` will encounter a bug and overwrite the whole line with garbage. Take that into account when you have to migrate a project a newer edition.

## Rust 2018 guidelines

Using Rust 2018 edition means that we require rustc 1.31 or later. IIRC, Exonum's dependencies require at least this version already, but if you're slow then please upgrade.

Most of the changes here are related to crate and module imports, but there are some other minor fixes interspersed between them. Using old style usually results in warnings from compiler, rustfmt, or clippy so you're pretty safe. However, here's a bunch of things that I had to fix in this PR and that we have to maintain from now on:

- `extern crate foo;` **is not required now** (in most cases)

    The compiler is finally able to figure out crate imports (with Cargo's help) so these lines are not necessary now.

    Using them explicitly does not result in warnings. However, I've heard rumors of `extern crate` being deprecated in some later edition, so let's go ahead and remove what we can.

  There are a bunch of uses that we leave. Mostly these are for `#[macro_use]` with crates that provide a whole bunch of macros (e.g., `log`) or custom derives (`serde_derive`). Importing this stuff directly is either too tedious now, or did not work for me, so I've left them as is. (There's also a special case of `proc_macro`, because reasons.)

- **You can import macros like any other item**

    If you just need one macro then just `use quote::quote` to import `quote!`, for example. There's no need to `#[macro_use]`

- **It is not necessary to import top-level crate module**

    Due to uniform paths ([read more here](https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html)] you don't have to `use exonum;` in order to write `exonum::init_log()` or something. Just use the crate name directly. There are many other simplifications in paths and imports due to this change.

- **Use** `crate`/`self`/`super` **to import from the current context**

    Uniform paths have a price. Previously if was possible to refer to currently visible modules implicitly. For example, to import module `foo` from current crate you wrote `use foo` and that's it. However, sometimes this conflicted with another `foo` imported from other places.

    Now you have to explicitly name the source of the import. All paths start either with a top-level crate name, or one of the keywords:

    * `crate::thing` to address the current crate starting from its root

    * `dependency::item` to address the some other extrnal crate starting from its root

    * `self::stuff` to import from modules you have imported earlier

    * `super::*` to import from your parent module

    [Read the docs](https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html) to learn more.

- `mod.rs` **is now not required**

    Previously, if you had a module named `foo` living in `foo.rs` and wanted to add a submodule `foo::bar` to it then you _had_ to rename this:

  ```
  src
  |- foo.rs
  \- zog.rs
  ```

    Into this:

  ```
  src
  |- foo
  |  |- mod.rs  (previously foo.rs)
  |  \- bar.rs
  \- zog.rs
  ```

  With Rust 2018 edition you can add submodules more directly: 

  ```
  src
  |- foo.rs
  |- foo
  |  \- bar.rs
  \- zog.rs
  ```

    Currently our code has a lot of `mod.rs` files. I did not rename them to avoid diff noise. However, please avoid them for new modules. Rename the old ones if that makes sense to do during a refactoring.

- **It is not necessary to prefix top-level paths with** `::`

    For example, now you can simply write `std::ptr::null()`. In Rust 2015 you had to either spell it `::std::ptr::null()`, or first `use std::ptr;` then call `ptr::null()`.

    Currently our code has _tons_ of paths like this. I did not change them to avoid diff noise. However, please don't add new ones. And please do fix old ones if you refactor the code around them.

- **Trait objects now require** `dyn` **prefix**

    It is now `dyn Snapshot`, `&mut dyn Fork`, `Box<dyn Transaction>`, etc. Previously this was optional, but in Rust 2018 this is required.

- **Trait methods now require argument names**

    In Rust 2015 trait method were allowed to have unnamed parameters. Now you have to give them names.

```diff
 pub trait ProofMapKey
 where
     Self::Output: ProofMapKey,
 {
     type Output;
 
-    fn write_key(&self, &mut [u8]);
+    fn write_key(&self, key: &mut [u8]);
 
-    fn read_key(&[u8]) -> Self::Output;
+    fn read_key(key: &[u8]) -> Self::Output;
 }
```

- **Prefer new syntax for inclusive (closed) ranges**

    Write `0..=9` instead of `0...9` if you mean [`0`, ..., `9`].

---

See: https://jira.bf.local/browse/ECR-2816

### Definition of Done

- [x] ~~Change is covered by automated tests~~
- [x] ~~Benchmark results are attached (if applicable)~~
- [x] The [coding guidelines] are updated
- [x] ~~Public API has proper documentation~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 ### Internal improvements
 
 - All Exonum crates have been updated to Rust 2018 edition. This means that
-  we require at least Rust 1.31 for compilation. (#1230)
+  it is required to use Rust 1.31 or newer for compilation. (#1230)
 
 #### exonum
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Internal improvements
 
+- All Exonum crates have been updated to Rust 2018 edition. This means that
+  we require at least Rust 1.31 for compilation. (#1230)
+
 #### exonum
 
 - Added `allow-origin` for `localhost` for public and private api in
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,9 +62,9 @@ but we have several additional conventions:
 
 - Use [Rust 2018 edition].
 
-  * Avoid `extern crate` unless it is required.
+  - Avoid `extern crate` unless it is required.
 
-  * Avoid `mod.rs` files, simply use directories for nested modules.
+  - Avoid `mod.rs` files, simply use directories for nested modules.
 
 - Prefer [nested imports] whenever possible.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ Notice that the repository uses a set of linters for static analysis:
 - [clippy]
 - [cargo audit]
 - [cargo-deadlinks]
-- [rustfmt 0.9.0]
+- [rustfmt]
 - [cspell]
 - [markdownlint-cli]
 
@@ -105,7 +105,7 @@ but we have several additional conventions:
 [clippy]: https://github.com/rust-lang-nursery/rust-clippy
 [cargo audit]: https://github.com/RustSec/cargo-audit
 [cargo-deadlinks]: https://github.com/deadlinks/cargo-deadlinks
-[rustfmt 0.9.0]: https://github.com/rust-lang-nursery/rustfmt
+[rustfmt]: https://github.com/rust-lang-nursery/rustfmt
 [cspell]: https://github.com/Jason3S/cspell
 [markdownlint-cli]: https://github.com/igorshubovych/markdownlint-cli
 [Travis script]: .travis.yml
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,13 +60,20 @@ but we have several additional conventions:
   Public method should be documented, but meaningful parameter names are also
   helpful for better understanding.
 
+- Use [Rust 2018 edition].
+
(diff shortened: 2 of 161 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Internal improvements
 
+- All Exonum crates have been updated to Rust 2018 edition. This means that
+  we require at least Rust 1.31 for compilation. (#1230)
```
vitvakatu: I have an information that even before this change build is impossible because some of our dependencies have already transfered to 2018 :wink:

vitvakatu: As additional note, I think we should not use pronouns in the changelog. So:
```suggestion
  it is required to use Rust 1.31 or newer for compilation. (#1230)
```

ilammy:
@vitvakatu I've added a couple of lines about preferring Rust 2018 in new code. I think we can leave it at that now. We can extend the guidelines later as necessary, _if_ we encounter some recurring flaming topics during code review. Otherwise let's expect devs to follow a sane code style.

As for leading `::`, it's not that it is _impossible_ in Rust 2018. This syntax still works and it has the same meaning as in Rust 2015 (importing from a top-level crate). It's just not required currently because the first term in paths always refers to a top-level crate (or the current module, on nightly). It may still be useful for resolving conflcts (as an alternative to renaming). And I assume that it might have some weird uses inside macros.

Anyway, I've updated the guideline to keep its spirit: we prefer to use imports to use short names instead of fully-qualified names.

PR #1235, opened 2019-01-21
Improve key files password entry [ECR-2519]

dvush:
## Overview

This PR allows following passphrase entry methods while generating configs and running node.

```--{service,consensus}-key-pass [stdin,env:VAR,pass:PASS]```
Default password entry method is stdin.
If variable in env is not set default one is used. (EXONUM_SERVICE_PASS, EXONUM_CONSENSUS_PASS)

---

See: https://jira.bf.local/browse/ECR-2519

### Definition of Done

### DIFF
--- a/exonum/src/helpers/fabric/builder.rs
+++ b/exonum/src/helpers/fabric/builder.rs
@@ -16,7 +16,6 @@ use ctrlc;
 
 use std::{
     collections::HashMap,
-    env,
     ffi::OsString,
     fmt,
     panic::{self, PanicInfo},
@@ -29,14 +28,11 @@ use super::{
     internal::{CollectedCommand, Command, Feedback},
     keys,
     maintenance::Maintenance,
+    password::{PassInputMethod, SecretKeyType},
     CommandName, Context, ServiceFactory,
 };
 use blockchain::Service;
-use helpers::ZeroizeOnDrop;
 use node::{ExternalMessage, Node};
-
-const EXONUM_CONSENSUS_PASS: &str = "EXONUM_CONSENSUS_PASS";
-const EXONUM_SERVICE_PASS: &str = "EXONUM_SERVICE_PASS";
 
 /// `NodeBuilder` is a high level object,
 /// usable for fast prototyping and creating app from services list.
@@ -162,15 +158,25 @@ impl NodeBuilder {
             .map(|mut factory| factory.make_service(ctx))
             .collect();
 
-        let consensus_passphrase =
-            ZeroizeOnDrop(env::var(EXONUM_CONSENSUS_PASS).unwrap_or_default());
-        let service_passphrase = ZeroizeOnDrop(env::var(EXONUM_SERVICE_PASS).unwrap_or_default());
+        let config = {
+            let run_config = ctx.get(keys::RUN_CONFIG).unwrap();
+            let consensus_passphrase = run_config
+                .consensus_pass_method
+                .parse::<PassInputMethod>()
+                .expect("Incorrect passphrase input method for consensus key.")
+                .get_passphrase(SecretKeyType::Consensus, true);
+            let service_passphrase = run_config
+                .service_pass_method
+                .parse::<PassInputMethod>()
+                .expect("Incorrect passphrase input method for service key.")
+                .get_passphrase(SecretKeyType::Service, true);
 
-        let config = config.read_secret_keys(
-            &config_file_path,
-            consensus_passphrase.0.as_bytes(),
-            service_passphrase.0.as_bytes(),
-        );
+            config.read_secret_keys(
+                &config_file_path,
+                consensus_passphrase.0.as_bytes(),
+                service_passphrase.0.as_bytes(),
+            )
+        };
         Node::new(db, services, config, Some(config_file_path))
     }
 }
--- a/exonum/src/helpers/fabric/builder.rs
+++ b/exonum/src/helpers/fabric/builder.rs
@@ -19,6 +19,7 @@ use std::{
     ffi::OsString,
     fmt,
     panic::{self, PanicInfo},
+    str::FromStr,
 };
 
 use super::{
@@ -160,21 +161,17 @@ impl NodeBuilder {
 
         let config = {
             let run_config = ctx.get(keys::RUN_CONFIG).unwrap();
-            let consensus_passphrase = run_config
-                .consensus_pass_method
-                .parse::<PassInputMethod>()
+            let consensus_passphrase = PassInputMethod::from_str(&run_config.consensus_pass_method)
                 .expect("Incorrect passphrase input method for consensus key.")
                 .get_passphrase(SecretKeyType::Consensus, true);
-            let service_passphrase = run_config
-                .service_pass_method
-                .parse::<PassInputMethod>()
+            let service_passphrase = PassInputMethod::from_str(&run_config.service_pass_method)
                 .expect("Incorrect passphrase input method for service key.")
                 .get_passphrase(SecretKeyType::Service, true);
 
             config.read_secret_keys(
                 &config_file_path,
-                consensus_passphrase.0.as_bytes(),
-                service_passphrase.0.as_bytes(),
+                consensus_passphrase.as_bytes(),
+                service_passphrase.as_bytes(),
             )
         };
         Node::new(db, services, config, Some(config_file_path))
--- a/exonum/src/helpers/fabric/details.rs
+++ b/exonum/src/helpers/fabric/details.rs
@@ -593,7 +593,7 @@ impl Command for GenerateNodeConfig {
                 consensus_key_pass_method,
                 SecretKeyType::Consensus,
             );
-            create_secret_key_file(&consensus_secret_key_path, &passphrase.0)
+            create_secret_key_file(&consensus_secret_key_path, passphrase.as_bytes())
         };
         let service_public_key = {
             let passphrase = Self::get_passphrase(
@@ -601,7 +601,7 @@ impl Command for GenerateNodeConfig {
                 service_key_pass_method,
                 SecretKeyType::Service,
             );
-            create_secret_key_file(&service_secret_key_path, &passphrase.0)
+            create_secret_key_file(&service_secret_key_path, passphrase.as_bytes())
         };
 
         let pub_config_dir = Path::new(&pub_config_path)
--- a/exonum/src/helpers/fabric/details.rs
+++ b/exonum/src/helpers/fabric/details.rs
@@ -94,7 +94,7 @@ impl Run {
             SecretKeyType::Consensus => CONSENSUS_KEY_PASS_METHOD,
             SecretKeyType::Service => SERVICE_KEY_PASS_METHOD,
         };
-        ctx.arg(arg_key).unwrap_or_else(|_| "".into())
+        ctx.arg(arg_key).unwrap_or_default()
     }
 }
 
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: exonum/src/helpers/fabric/builder.rs [hunk 1]
```diff
@@ -162,15 +158,25 @@ impl NodeBuilder {
...
+        let config = {
+            let run_config = ctx.get(keys::RUN_CONFIG).unwrap();
+            let consensus_passphrase = run_config
+                .consensus_pass_method
+                .parse::<PassInputMethod>()
+                .expect("Incorrect passphrase input method for consensus key.")
+                .get_passphrase(SecretKeyType::Consensus, true);
+            let service_passphrase = run_config
+                .service_pass_method
+                .parse::<PassInputMethod>()
```
ilammy: Using `FromStr` directly may look more nice and tight. (YMMV of course.)

```rust
let service_passphrase = PassInputMethod::from_str(& service_pass_method)
    .expect("Incorrect passphrase input method for service key.")
    .get_passphrase(SecretKeyType::Consensus, true);
```

exonum/src/helpers/fabric/password.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    loop {
+        let password = ZeroizeOnDrop(read_password_from_tty(Some(prompt))?);
+        if password.0.is_empty() {
+            eprintln!("Passphrase must not be empty. Try again.");
+            continue;
+        }
+
+        let second_password = ZeroizeOnDrop(read_password_from_tty(Some(
+            "Enter same passphrase again: ",
```
ilammy: I'd call this variable `confirmation` instead of `second_password` which reads as if this is some new, unrelated password requested from the user.

Also, correct grammar would be

```suggestion
            "Enter the same passphrase again: ",
```

exonum/src/helpers/fabric/password.rs
```diff
@@ -0,0 +1,168 @@
...
+        ))?);
+
+        if password.0 == second_password.0 {
```
ilammy: It may be nice to implement `Deref` for `ZeroizeOnDrop` so that we don't have to put `.0` everywhere.

exonum/src/helpers/fabric/password.rs
```diff
@@ -0,0 +1,168 @@
...
+        if s == "" {
+            return Ok(Default::default());
+        }
+
+        if s == "stdin" {
+            return Ok(PassInputMethod::Terminal);
+        }
+
+        if let Some(prefix) = s.get(0..3) {
+            if prefix == "env" {
```
ilammy: This comparison can be done in a more simple and robust way: `s.starts_with("env")`

dvush: Thanks. For some reason I was trying to use `is_prefix_of` which is unstable)

exonum/src/helpers/fabric/password.rs
ilammy: Touché.

exonum/src/helpers/fabric/password.rs
```diff
@@ -0,0 +1,168 @@
+// Copyright 2018 The Exonum Team
```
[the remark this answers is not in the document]
ilammy: Please no. There's no need in additional diff clutter. We'll have enough of it when the Rust 2018 update lands which touches a whole lot of files due to import updates. But that has technical reasons and merit at least.

IANAL, but from what I've read on the internets, the year in copyright statements may be interpreted as the year of the first publication of the original work. Therefore it's okay to put the current year when you create a file and then just leave it at that.

exonum/src/helpers/fabric/builder.rs
```diff
@@ -162,15 +158,25 @@ impl NodeBuilder {
...
-            consensus_passphrase.0.as_bytes(),
-            service_passphrase.0.as_bytes(),
-        );
+        let config = {
+            let run_config = ctx.get(keys::RUN_CONFIG).unwrap();
+            let consensus_passphrase = run_config
+                .consensus_pass_method
+                .parse::<PassInputMethod>()
+                .expect("Incorrect passphrase input method for consensus key.")
+                .get_passphrase(SecretKeyType::Consensus, true);
```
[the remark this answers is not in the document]
dvush: Overall I agree but not in this case since I do not think that this enum is going to be extended and this makes function call bigger. 
I renamed this argument to `node_run`, also this function has comment about this parameter.

exonum/src/helpers/fabric/password.rs
```diff
@@ -0,0 +1,164 @@
...
+            )));
+        }
+
+        bail!("Failed to parse passphrase input method");
+    }
+}
+
+fn prompt_passphrase(prompt: &str, node_run: bool) -> io::Result<ZeroizeOnDrop<String>> {
+    if node_run {
+        return Ok(ZeroizeOnDrop(read_password_from_tty(Some(prompt))?));
```
pavel-mukhanov: Can we use an empty passphrase when running node?

dvush: Yes. That is why this code is here, if we generated key files without password we should be able to use them using all methods of entering passphrase including terminal.

pavel-mukhanov: How can we generate files without password? Using `run-dev`? According to this method, we can't enter an empty password from the terminal of maybe I missing something.

dvush: Using run dev.
Using "no password option".
Using empty password in environment variables.

exonum/src/helpers/fabric/details.rs
```diff
@@ -420,18 +456,16 @@ impl GenerateNodeConfig {
...
-            .or_else(|| prompt_passphrase(secret_key_type).ok())
-            .map(|s| ZeroizeOnDrop(s.0.as_bytes().to_vec()))
-            .unwrap()
+    fn get_passphrase(
+        context: &Context,
+        method: PassInputMethod,
+        secret_key_type: SecretKeyType,
+    ) -> ZeroizeOnDrop<String> {
+        if context.get_flag_occurrences(NO_PASSWORD).is_some() {
+            ZeroizeOnDrop("".to_owned())
```
alekseysidorov: Can you explain in comment what the following line do?

exonum/src/helpers/fabric/details.rs
```diff
@@ -514,6 +568,16 @@ impl Command for GenerateNodeConfig {
             .arg::<String>(SERVICE_KEY_PATH)
             .unwrap_or_else(|_| "service.toml".into())
             .into();
+        let consensus_key_pass_method: PassInputMethod = context
+            .arg::<String>(CONSENSUS_KEY_PASS_METHOD)
+            .unwrap_or_else(|_| "".to_owned())
```
alekseysidorov: Why not just `String::new`?

exonum/src/helpers/fabric/password.rs
```diff
@@ -0,0 +1,164 @@
...
+            return Ok(PassInputMethod::Terminal);
+        }
+
+        if s.starts_with("env") {
+            let env_var = s.split(':').nth(1).map(|s| s.to_owned());
+            return Ok(PassInputMethod::EnvVariable(env_var));
+        }
+
+        if s.starts_with("pass") {
+            let pass = s.split(':').nth(1).unwrap_or("");
```
alekseysidorov: `unwrap_or_default`

exonum/src/helpers/fabric/password.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn default() -> Self {
+        PassInputMethod::Terminal
+    }
+}
+
+impl FromStr for PassInputMethod {
+    type Err = failure::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s == "" {
```
alekseysidorov: Why not `s.is_empty()`?

exonum/src/helpers/fabric/password.rs
```diff
@@ -0,0 +1,164 @@
...
+            return Ok(Default::default());
+        }
+
+        if s == "stdin" {
+            return Ok(PassInputMethod::Terminal);
+        }
+
+        if s.starts_with("env") {
+            let env_var = s.split(':').nth(1).map(|s| s.to_owned());
```
alekseysidorov: `map(String::to_owned)` should works too.

exonum/src/helpers/fabric/details.rs
```diff
@@ -93,6 +88,14 @@ impl Run {
     fn private_api_address(ctx: &Context) -> Option<SocketAddr> {
         ctx.arg(PRIVATE_API_ADDRESS).ok()
     }
+
+    fn pass_input_method(ctx: &Context, key_type: SecretKeyType) -> String {
+        let arg_key = match key_type {
+            SecretKeyType::Consensus => CONSENSUS_KEY_PASS_METHOD,
+            SecretKeyType::Service => SERVICE_KEY_PASS_METHOD,
+        };
+        ctx.arg(arg_key).unwrap_or_else(|_| "".into())
```
aleksuss:
```suggestion
        ctx.arg(arg_key).unwrap_or_default()
```

exonum/src/helpers/fabric/details.rs
aleksuss: Or `.unwrap_or_default()` could be ok as well here.

exonum/src/helpers/fabric/details.rs
```diff
@@ -514,6 +568,16 @@ impl Command for GenerateNodeConfig {
...
             .unwrap_or_else(|_| "service.toml".into())
             .into();
+        let consensus_key_pass_method: PassInputMethod = context
+            .arg::<String>(CONSENSUS_KEY_PASS_METHOD)
+            .unwrap_or_else(|_| "".to_owned())
+            .parse()
+            .expect("expected correct passphrase input method for consensus key");
+        let service_key_pass_method: PassInputMethod = context
+            .arg::<String>(SERVICE_KEY_PASS_METHOD)
+            .unwrap_or_else(|_| "".to_owned())
```
aleksuss: Same here

exonum/src/helpers/types.rs
```diff
@@ -362,11 +362,25 @@ impl Iterator for RoundRangeIter {
 }
 
 /// Struct used to call zeroize on inner type on drop.
-#[derive(Debug)]
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
```
aleksuss:
```suggestion
#[derive(Debug, Default, Serialize, Deserialize, PartialEq)]
```

exonum/src/helpers/fabric/details.rs
dvush: Returns empty password.

exonum/src/helpers/fabric/password.rs
dvush: I preferred this variant.
Fine, changed.

exonum/src/helpers/fabric/password.rs
dvush: I always get this kind of comment and I do not see any benefit.
If you all really prefer it you should add it to CONTRIBUTIND.md :)

dvush: Also it does not work. It should be`str::to_owned` or something like that.

vitvakatu: FYI, https://rust-lang.github.io/rust-clippy/master/#redundant_closure

vitvakatu: However, I'm not insisting on this specific case. If clippy is satisfied - leave it as it is

PR #1242, opened 2019-01-30
Support disabling services [ECR-1849]

ilammy:
## Overview

Teach Exonum a notion of _disabled_ service. Such services exist in the network, but cannot execute any transactions and change the blockchain state in any way. Their data is retained intact and is still available via read requests though. The service may be born disabled since the genesis, it may also be disabled and reenabled back later by network administrators.

Service status is a part of _global_ (consensus) configuration. It is configurable on network startup via the node configuration file. It may be updated later via the configuration service. (Currently it is quite possible to disable the configuration service itself, thus locking down the network configuration.)

The configuration starts in the common configuration template generated by `generate-template`:

```toml
[services.cryptocurrency]
enabled = true

[services.configuration]
enabled = true
```

and completes its journey in the genesis configuration section of the final configuration:

```toml
[genesis.services.cryptocurrency]
enabled = true

[genesis.services.configuration]
enabled = true
```

After the initial network startup the service status is a part of the node configuration. It has to be updated via the configuration service.

All services known to the network are enabled by default initially. The administrators have to manually disable them if they wish so.

---

See: https://jira.bf.local/browse/ECR-1849

### Definition of Done

### DIFF
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -64,7 +64,7 @@ pub struct StoredConfiguration {
     pub validator_keys: Vec<ValidatorKeys>,
     /// Consensus algorithm parameters.
     pub consensus: ConsensusConfig,
-    /// Servic-specific configuration.
+    /// Service-specific configuration.
     /// Keys are `service_name` from the `Service` trait.
     #[serde(default)]
     pub services: BTreeMap<String, ServiceConfig>,
--- a/exonum/src/blockchain/config.rs
+++ b/exonum/src/blockchain/config.rs
@@ -64,10 +64,10 @@ pub struct StoredConfiguration {
     pub validator_keys: Vec<ValidatorKeys>,
     /// Consensus algorithm parameters.
     pub consensus: ConsensusConfig,
-    /// Services specific variables.
-    /// Keys are `service_name` from the `Service` trait and values are the serialized JSON.
+    /// Servic-specific configuration.
+    /// Keys are `service_name` from the `Service` trait.
     #[serde(default)]
-    pub services: BTreeMap<String, serde_json::Value>,
+    pub services: BTreeMap<String, ServiceConfig>,
 }
 
 /// Consensus algorithm parameters.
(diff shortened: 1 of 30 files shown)


ilammy:
This definitely warrants a changelog entry. It will be added later when the discussion finalizes.

I'd like some second opinion on the design and naming. I do not really like how it's tacked onto existing behavior, and the naming surely could be better (we have an overload of _config_ words in the code).

You may have also noticed that these changes are not automatically tested, but I think that they can be. I would like to get some pointers for that.

This also affects ongoing work by @popzxc on dynamic service management. just keep it mind.

exonum/src/blockchain/config.rs
```diff
@@ -123,6 +123,15 @@ pub struct ConsensusConfig {
     pub propose_timeout_threshold: u32,
 }
 
+/// Service parameters.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct ServiceConfig {
+    /// This service is considered enabled.
+    pub enabled: bool,
+    /// Opaque service-specific configuration serialized as JSON.
+    pub private: serde_json::Value,
```
pavel-mukhanov: Nit: `private` variable name is confusing for me, it reminds about reserved keyword from other languages.

slowli: I also think `private` is not a very good name. Maybe, something like `details` can be better?

exonum/src/blockchain/mod.rs
```diff
@@ -433,8 +445,15 @@ impl Blockchain {
         fork.checkpoint();
 
         let catch_result = panic::catch_unwind(panic::AssertUnwindSafe(|| {
-            let context = TransactionContext::new(&mut *fork, &raw);
-            tx.execute(context)
+            if service_enabled(service_name, fork) {
+                let context = TransactionContext::new(&mut *fork, &raw);
+                tx.execute(context)
+            } else {
+                Err(ExecutionError::with_description(
```
slowli: What if the service uses an error with code 255? Should we document somewhere that 255 is a reserved code? Maybe, it could make sense to handle this error by analogy with panics (i.e., by introducing a new variant in `TransactionErrorType`)?

PR #1251, opened 2019-02-19
Update configuration service docker image

ivan-ochc:


services/configuration/doc/testnet-api-tutorial.md
```diff
@@ -9,7 +9,7 @@
 Simply run the following command to configure and launch testnet with 4 nodes:
 
 ```bash
-docker run -p 8000-8007:8000-8007 vitvakatu/exonum-configuration-example 4
+docker run -p 8000-8007:8000-8007 exonumhub/exonum-configuration-service:example 4
```
aleksuss: Checker doesn't like a length of the string

PR #1252, opened 2019-02-20
Changed a response for healthcheck end-point [ECR-2925]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-2925

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   CLI for generating node configs and starting nodes has been extended
   in order to reflect these changes. (#1222, #1096, #1235)
 
-- Changed a response for `/healthcheck` end-point. (#1252)
+- Changed a response for `/healthcheck` endpoint. (#1252)
 
 #### exonum-crypto
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   CLI for generating node configs and starting nodes has been extended
   in order to reflect these changes. (#1222, #1096, #1235)
 
+- Changed a response for `/healthcheck` end-point. (#1252)
+
 #### exonum-crypto
 
 - Renamed `create_keys_file` function to `generate_keys_file`
--- a/exonum/src/api/node/public/system.rs
+++ b/exonum/src/api/node/public/system.rs
@@ -25,20 +25,31 @@ pub struct MemPoolInfo {
     pub size: u64,
 }
 
-/// Information about the amount of peers connected to the node.
-#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
-pub struct PeersAmount {
-    /// Amount of connected peers.
-    pub amount: usize,
-}
-
 /// Information about whether the node is connected to other peers.
 #[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
-pub enum ConnectivityStatus {
+pub struct ConnectivityStatus {
     /// The node has no connected peers.
-    NotConnected,
+    connected: bool,
     /// The node has connected peers. Amount of connected peers is stored within this variant.
-    Connected(PeersAmount),
+    connected_peers: usize,
+}
+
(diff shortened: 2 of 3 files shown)


exonum/src/api/node/public/system.rs
```diff
@@ -25,20 +25,31 @@ pub struct MemPoolInfo {
...
-}
-
 /// Information about whether the node is connected to other peers.
 #[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
-pub enum ConnectivityStatus {
+pub struct ConnectivityStatus {
     /// The node has no connected peers.
-    NotConnected,
+    connected: bool,
     /// The node has connected peers. Amount of connected peers is stored within this variant.
```
dmitry-timofeev: The number of ...

dmitry-timofeev: Apparently, not a variant.

exonum/src/api/node/public/system.rs
```diff
@@ -25,20 +25,31 @@ pub struct MemPoolInfo {
...
-    pub amount: usize,
-}
-
 /// Information about whether the node is connected to other peers.
 #[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
-pub enum ConnectivityStatus {
+pub struct ConnectivityStatus {
     /// The node has no connected peers.
-    NotConnected,
+    connected: bool,
```
dmitry-timofeev: Is the flag needed if we include the number of connections (peers=0 -> not connected)?

exonum/src/api/node/public/system.rs
```diff
@@ -109,11 +120,9 @@ impl SystemApi {
         let out_conn = self.shared_api_state.outgoing_connections().len();
 
         if in_conn == 0 && out_conn == 0 {
-            ConnectivityStatus::NotConnected
+            ConnectivityStatus::not_connected()
         } else {
-            ConnectivityStatus::Connected(PeersAmount {
-                amount: in_conn + out_conn,
-            })
+            ConnectivityStatus::connected(in_conn + out_conn)
```
dmitry-timofeev: Don't know what are these connections, but does the sum of incoming and outgoing connections correspond to the _number of peers_ or of connections? Can't we have both an ingoing and outcoming connection to the same node?

aleksuss: @dmitry-timofeev No, we can't.

exonum/src/api/node/public/system.rs
aleksuss: @dmitry-timofeev I asked you and @bullet-tooth in the ticket. If you see in the ticket so the answer was "Format is ok"

exonum/src/api/node/public/system.rs
dmitry-timofeev: I'd add a comment in this case, because it is far from immediately apparent.

exonum/src/api/node/public/system.rs
dmitry-timofeev: But that doesn't answer if including a boolean makes sense. Is it possible to have peers != 0 but connected = false? If connected always implies > 0 connections and 0 connections imply not connected, then it's just redundant data.

aleksuss: Basically, you're right. If connections number is 0 then flag is always `false`. Provide a format of response in the ticket please.

dmitry-timofeev: I see, thanks! I'll get back to that shortly (maybe @bullet-tooth would be able to clarify that first?)

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   CLI for generating node configs and starting nodes has been extended
   in order to reflect these changes. (#1222, #1096, #1235)
 
+- Changed a response for `/healthcheck` end-point. (#1252)
```
bullet-tooth:
```suggestion
- Changed a response for `/healthcheck` endpoint. (#1252)
```

exonum/src/api/node/public/system.rs
bullet-tooth: Agree with @dmitry-timofeev for removing this flag. Ticket updated with new format

PR #1253, opened 2019-02-20
Interface mocks for dynamic services

popzxc:
## Overview

Interface mocks for dynamic services

### Definition of Done


exonum/src/runtime/rust.rs
```diff
@@ -40,14 +39,14 @@ struct RustRuntimeInner {
     initialized: HashMap<ServiceInstanceId, RustArtifactData>,
 }
 
-#[derive(Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct RustArtifactSpec {
     name: String,
-    version: (usize, usize, usize),
+    version: (u32, u32, u32),
```
aleksuss: Is it temporary solution, isn't it ? There is a `semver` create exists for such goal.

popzxc: Yep, it is. I think it will be rewritten (almost) completely when we'll work on `RustRuntime`. For now it's important only as an interface part.

exonum/src/runtime/dispatcher.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+}
+
+#[derive(Default)]
+struct Dispatcher {
+    runtimes: HashMap<String, Box<dyn RuntimeEnvironment>>,
+}
+
+impl Dispatcher {}
```
alekseysidorov: What does it mean?

exonum/src/runtime/rust.rs
```diff
@@ -98,44 +113,40 @@ impl RuntimeEnvironment for RustRuntime {
...
     fn from_env_ctx(env_context: &'a EnvContext<'a>) -> Self {
-        Self { env_context }
+        Self {
+            _env_context: env_context,
+        }
     }
 }
 
 struct Handler {
     pub fun_untyped: Box<dyn Fn(&mut TransactionContext, &[u8]) -> Result<(), ExecutionError>>,
```
alekseysidorov: Method name sounds a little bit weird

exonum/src/runtime/dispatcher.rs
popzxc: It means that most probably `Dispatcher` entity will have some methods in the future, but currently it doesn't :)

exonum/src/runtime/rust.rs
popzxc: Yea, agree. I guess in the future there will be `fun_typed`, but not sure.
What will you suggest?

exonum/src/runtime/dispatcher.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl RuntimeEnvironment for Dispatcher {
+    fn start_deploy(&self, _artifact: ArtifactSpec) -> Result<(), DeployError> {
```
vitvakatu: Why this method name is `start_deploy`, when it probably does _loading_ of the artifact? Will it also instantiate the artifact?

popzxc: Hmm. As far as I understand it, artifact != service. Artifact is a template, based on which we can instantiate a service. Thus this method is called `start_deploy`, since it initiates the process of receiving & initializing of the artifact (as a template). Later, when artifact will be deployed, we can use it to initialize a service instance. Am I wrong in this one?

vitvakatu: Okay, thank you. You're right, I just thought that the process of service instantiation will consist of two steps: load -> instantiate

exonum/src/runtime/mod.rs
```diff
@@ -57,24 +59,30 @@ pub enum ArtifactSpec {
 /// Service runtime environment.
 /// It does not assign id to services/interfaces, ids are given to runtime from outside.
 pub trait RuntimeEnvironment {
-    /// Deploy artifact.
-    fn deploy(&self, artifact: ArtifactSpec) -> Result<(), DeployError>;
+    /// Start artifact deploy.
+    fn start_deploy(&self, artifact: ArtifactSpec) -> Result<(), DeployError>;
+
+    /// Check deployment status.
+    fn check_deploy_status(&self, artifact: ArtifactSpec) -> Result<DeployStatus, DeployError>;
```
dvush: I find this result to be a little bit redundant since `DeployStatus` contain `DeployError` variant.

exonum/src/runtime/mod.rs
```diff
@@ -14,27 +14,23 @@
...
-    ServiceIdExists,
+pub enum DeployStatus {
+    DeployInProgress,
+    Deployed,
+    DeployErrored(DeployError),
 }
 
 type ServiceInstanceId = u32;
-type InterfaceId = u32;
+type MethodId = String;
```
dvush: Is this change permanent? Because I think integer method can be useful to minimize space storage + it is quite easy to assign them.

exonum/src/runtime/rust.rs
```diff
@@ -98,44 +113,40 @@ impl RuntimeEnvironment for RustRuntime {
...
-    }
-
-    fn execute(&self, context: &mut EnvContext, dispatch: DispatchInfo, payload: &[u8]) {
+    fn execute(
+        &self,
+        context: &mut EnvContext,
+        dispatch: CallInfo,
+        payload: &[u8],
+    ) -> Result<(), ExecutionError> {
         let inner = self.inner.read().unwrap();
```
dvush: It is out of scope of this PR and I wrote this line of code, but I just wanted to mention that maybe it makes more sense to not have locks inside runtime environment so we don't acquire lock every time we execute transaction.

exonum/src/runtime/mod.rs
popzxc: Nope, it's temporary. Until we create an entity which will provide handy interface for (name => number) mapping, I think identifiers as strings are a bit more usable.

exonum/src/runtime/mod.rs
popzxc: Ah, I forgot to remove this variant. I decided that using a `Result` is a bit more convenient than implementing it inside of enum as variant. I'll remove `DeployError` status.

exonum/src/runtime/rust.rs
popzxc: Sure. I guess we'll clarify how to do it when we'll start work on `RustRuntime` task :)

exonum/src/blockchain/mod.rs
```diff
@@ -108,40 +108,43 @@ impl StaticServiceDispatcher {
     }
 
     // TODO clarify if we need this method
-    pub fn services(&self) -> Arc<HashMap<u16, Box<dyn Service>>>
-    {
+    pub fn services(&self) -> Arc<HashMap<u16, Box<dyn Service>>> {
```
aleksuss: I don't know how it is actual for now, but why don't you wrap `HashMap` with `RwLock` ? Is it redundant ?

popzxc: I just copy-pasted it from the [old version](https://github.com/exonum/exonum/blob/master/exonum/src/blockchain/mod.rs#L80) :) 
Anyway, this dispatcher will be removed soon. It was created mostly by copy-pasting just to make new dispatched integration easier in the future.

exonum/src/runtime/rust.rs
aleksuss: What about `fn_untyped` or `func_untyped` ? But `fun_...` it's really weird 🙃

popzxc: Sure!

PR #1261, opened 2019-02-26
Basic dispatcher functionality

popzxc:
## Overview

Dispatcher entity implementation.

### Definition of Done

### DIFF
--- a/exonum/src/runtime/dispatcher.rs
+++ b/exonum/src/runtime/dispatcher.rs
@@ -20,7 +20,7 @@ use super::{
     ServiceInstanceId,
 };
 
-#[derive(Debug, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum RuntimeIdentifier {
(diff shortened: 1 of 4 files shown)


exonum/src/runtime/dispatcher.rs
```diff
@@ -16,143 +16,387 @@ use std::collections::HashMap;
...
-        _payload: &[u8],
+        context: &mut EnvContext,
+        call_info: CallInfo,
+        payload: &[u8],
     ) -> Result<(), ExecutionError> {
-        Ok(())
+        let runtime_id = self.runtime_lookup.get(&call_info.instance_id);
+
+        if runtime_id.is_none() {
+            return Err(ExecutionError::with_description(0x00, "Wrong runtime"));
```
dvush: Services use errors with id 0 and it can cause problems. I think we should reserve higher error numbers for runtime errors. 
EDIT: Ok, I see todo there, so yeah we should just reserve some error numbers for system errors like (service not found, method not found, etc.)

PR #1263, opened 2019-02-27
Add new rust services interface

dvush:
## Overview

exonum/src/runtime/rust/tests.rs - sample service definition and implementation

### Definition of Done


exonum/src/runtime/rust/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        let service = inner.services.remove(&artifact).unwrap();
+        inner.initialized.insert(init.instance_id, service);
+        Ok(())
+    }
+
+    fn execute(
+        &self,
+        context: &mut EnvContext,
```
alekseysidorov: EnvContext sounds a little bit unclear, imo `Environment` or `RuntimeContext` is clearly.

exonum/src/runtime/rust/service.rs
```diff
@@ -0,0 +1,88 @@
...
+pub trait ServiceDispatcher {
+    fn call(
+        &self,
+        method: MethodId,
+        ctx: TransactionContext,
+        payload: &[u8],
+    ) -> Result<Result<(), ExecutionError>, Error>;
+}
+
+pub trait SystemService: ServiceDispatcher + std::fmt::Debug {
```
alekseysidorov: SystemService also sounds a little bit strange.

exonum/src/runtime/rust/tests.rs
```diff
@@ -0,0 +1,136 @@
...
+    service::SystemService, ArtifactSpec, RustArtifactSpec, RustRuntime, TransactionContext,
+};
+use crate::crypto::{Hash, PublicKey};
+use crate::runtime::{
+    error::ExecutionError, CallInfo, DeployStatus, EnvContext, InstanceInitData, RuntimeEnvironment,
+};
+use crate::storage::{Database, Entry, MemoryDB};
+use protobuf::Message;
+
+service_interface! {
```
alekseysidorov: I would prefer if possible a regular `trait` or `impl` blocks with the derive attribute than the IDE-unfriendly macroses.

exonum/src/runtime/rust/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+        dispatch: CallInfo,
+        payload: &[u8],
+    ) -> Result<(), ExecutionError> {
+        let inner = self.inner.borrow();
+        let instance = inner.initialized.get(&dispatch.instance_id).unwrap();
+
+        let ctx = TransactionContext::new(context, self);
```
aleksuss: Why wouldn't include payload in `TransactionContext`? Does the payload relate to transaction here, doesn't it ?

exonum/src/runtime/rust/service.rs
```diff
@@ -0,0 +1,88 @@
...
+    // constructor method
+    // other hooks such as "on node startup"
+}
+
+#[macro_export]
+macro_rules! service_interface {
```
popzxc: A pure magic. I guess it's OK for now, but later we need to document this thing very carefully.

exonum/src/runtime/rust/mod.rs
dvush: It is passed as a separate argument and converted to typed structure. Why would anyone need payload bytes?

exonum/src/runtime/rust/tests.rs
dvush: Totally agree.

exonum/src/runtime/rust/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+};
+
+use crate::crypto::{Hash, PublicKey};
+use crate::storage::Fork;
+
+use self::service::SystemService;
+
+#[derive(Debug, Default)]
+struct RustRuntime {
+    inner: RefCell<RustRuntimeInner>,
```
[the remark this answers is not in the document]
dvush: Yes, but here start_deploy does not call check_deploy_status so it would not happen and this code would not compile if we will try to run in multiple threads. Do you propose to have lock here? But then, I think, it may be an issue if we will have to take a lock every time we execute tx.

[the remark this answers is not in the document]
popzxc: I propose to keep this thing in mind :) Later we'll test this carefully and find a proper solution. I guess we don't have to solve the problem that doesn't yet exists :)
Maybe just mark it with `TODO`?

exonum/src/runtime/rust/mod.rs
aleksuss: I mean to add `payload` as a field of `TransactionContext` struct. And then:
```rust
let ctx = TransactionContext::new(context, self, payload);
instance
      .call(dispatch.method_id, ctx)
      .expect("Dispatch error")
```

Logically a payload is more related to `TransactionContext` as for me. Because it's a part of the transaction, isn't it? But I could be wrong.

dvush: Here is an example of usage of transaction context https://github.com/exonum/exonum/pull/1263/files#diff-b86eb31f9bf79bec1ba1ec769dc47638R37
`TransactionContext` structure passed as is, but payload is converted to `TimestampTx`. `TransactionContext` here is something other that transaction itself(payload). If we would add payload to Context we would have duplication of payload.

aleksuss: Ok. I got it 👍🏻

PR #1267, opened 2019-03-05
Move configuration service to the core

popzxc:
## Overview

Configuration service will be a part of the core.

**TODO:**

- [x] Make it compile
- [x] Fix `cryptocurrency-advanced` and `timestamping` examples
- [x] Make configuration service run by default
- [x] Check if everything works fine


exonum/src/helpers/fabric/builder.rs
```diff
@@ -47,7 +48,7 @@ impl NodeBuilder {
     pub fn new() -> Self {
         Self {
             commands: Self::commands(),
-            service_factories: Vec::new(),
+            service_factories: vec![Box::new(configuration::ServiceFactory)],
```
popzxc: I'm not really sure about this thing. Is there any way to make this more explicit?

PR #1270, opened 2019-03-12
Dynamic configuration service

popzxc:
## Overview

Attempt to adapt static configuration service to the new architecture.


exonum/src/proto/schema/exonum/configuration.proto
```diff
@@ -47,3 +47,11 @@ message ProposeData {
...
   uint64 num_validators = 3;
 }
+
+// Init data for Configuration service.
+message ConfigurationServiceInit {
+  // Check whether we want to use a custom majority count.
+  bool is_custom_majority_count = 1;
+  // Custom majority count.
+  uint32 majority_count = 2;
+}
```
aleksuss: new line ? 🙃

exonum/src/runtime/configuration_new/config.rs
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Config for Configuration service.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ConfigurationServiceConfig {
+    /// Number of votes required to commit the new configuration.
+    /// This value should be greater than 2/3 and less or equal to the
+    /// validators count.
+    pub majority_count: Option<u16>,
```
alekseysidorov: What do you think about `quorum` name?

exonum/src/runtime/configuration_new/schema.rs
```diff
@@ -0,0 +1,275 @@
...
+    /// Returns a `Propose` transaction with a particular configuration hash.
+    pub fn propose(&self, cfg_hash: &Hash) -> Option<Propose> {
+        self.propose_data_by_config_hash()
+            .get(cfg_hash)?
+            .tx_propose
+            .into()
+    }
+
+    /// Returns a list of votes for the proposal corresponding to the given configuration hash.
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::let_and_return))]
```
alekseysidorov: You can just write `#[allow(clippy::let_and_return)]`

exonum/src/runtime/configuration_new/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+use schema::VotingDecision;
+use transactions::{enough_votes_to_commit, VotingContext};
+use config::ConfigurationServiceConfig;
+
+/// Service identifier for the configuration service.
+pub const SERVICE_ID: u16 = 1;
+/// Configuration service name.
+pub const SERVICE_NAME: &str = "configuration";
+
+service_interface! {
```
alekseysidorov: Is it possible to avoid macros?

exonum/src/runtime/configuration_new/config.rs
popzxc: I think that it makes not much difference from the used one.

exonum/src/runtime/configuration_new/mod.rs
popzxc: For now not really.

popzxc:
> As I understand it is just part 1 and it is just old config service.

Exactly. This PR only gives old config service a new interface.

PR #1275, opened 2019-03-17
Dynamic services integration

popzxc:
## Overview

This PR will contain functionality to deploy services at the node start through config file with deploy/init transactions.

---

See: https://jira.bf.local/browse/ECR-2917

### Definition of Done

- [x] Deploy and Init transactions.
- [x] Service deploy interface.
- [ ] Deserializable config file with deploy/init transactions.
- [ ] Passing transactions from config file to Configuration service.
- [ ] Working prototype.

### DIFF
--- a/exonum/src/runtime/configuration_new/mod.rs
+++ b/exonum/src/runtime/configuration_new/mod.rs
@@ -165,10 +165,12 @@ impl ConfigurationService for ConfigurationServiceImpl {
         let artifact_spec = arg.get_artifact_spec();
 
         let dispatcher = self.dispatcher.borrow();
-        let _result = dispatcher.start_deploy(artifact_spec);
+        dispatcher.start_deploy(artifact_spec).map_err(|err| {
+            error!("Service instance deploy failed: {:?}", err);
+            ServiceError::DeployError(err)
+        })?;
 
         // TODO add result into deployable (to check deploy status in before_commit).
-        // TODO return result.
 
         Ok(())
     }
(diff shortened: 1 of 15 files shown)


exonum/src/runtime/configuration_new/mod.rs
```diff
@@ -124,6 +156,104 @@ impl ConfigurationService for ConfigurationServiceImpl {
...
+                .init_service(ctx.env_context(), artifact_spec, &init_data)
+                .map_err(|err| {
+                    error!("Service instance initialization failed: {:?}", err);
+                    ServiceError::InitError(err)
+                })?;
+        }
+
+        Ok(())
+    }
 }
```
aleksuss: It seems to me that a code duplication here, doesn't it ? 🙃

popzxc: Yep, it is. And there is a comment for that slightly above: `// TODO reduce copy-paste.`. However, it can't be fixed easily, since `TransactionContext` is consumed in tx handling and doesn't implement `Clone`. 

Since it's a draft I decided to left this thing as-is for now and move to the more important parts of this task.

popzxc:
Hm.. About artifact spec:

```rust
pub struct ArtifactSpec {
    runtime_id: u32,
    raw_spec: Vec<u8>,
}
```

It's already stored in it, innit?

Other question is that services should have a method like `into_artifact` which will give us a generic `ArtifactSpec` (not `RustArtifactSpec` or something).

PR #1285, opened 2019-04-02
Fix concurrent patch merge in testkit server

slowli:
Currently, it's possible to break a testkit server by concurrently sending transactions and creating blocks. The problem seems to be specific to the testkit server (i.e., does not affect real Exonum nodes), since in real nodes all merges are performed on a single thread.

I didn't have time to write targeted tests for this bug, but the scenario to reproduce is described above.

### DIFF
--- a/testkit/src/lib.rs
+++ b/testkit/src/lib.rs
@@ -170,7 +170,7 @@ pub mod proto;
 use futures::{sync::mpsc, Future, Stream};
 use tokio_core::reactor::Core;
 
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
 use std::{fmt, net::SocketAddr};
 
 use exonum::{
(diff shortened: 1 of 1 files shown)


testkit/src/lib.rs
```diff
@@ -433,27 +434,29 @@ impl TestKit {
...
-                        | ExternalMessage::Rebroadcast
-                        | ExternalMessage::Shutdown => { /* Ignored */ }
                     }
+                    ExternalMessage::PeerAdd(_)
+                    | ExternalMessage::Enable(_)
+                    | ExternalMessage::Rebroadcast
+                    | ExternalMessage::Shutdown => { /* Ignored */ }
                 }
                 blockchain.merge(fork.into_patch()).unwrap();
+                drop(guard);
```
aleksuss: @slowli Is it really necessary to call `drop` explicitly on the `guard` here ?

PR #1286, opened 2019-04-02
Use rocksdb crate instead of exonum_rocksdb.

pavel-mukhanov:
Preparing to move to merkledb.


aleksuss:
@pavel-mukhanov I suppose it's reasonable to add a few lines in CHANGELOG about this change?

PR #1287, opened 2019-04-03
Remove macro from service interface trait definition

dvush:
## Overview

---

### Definition of Done

### DIFF
--- a/components/derive/src/service_interface.rs
+++ b/components/derive/src/service_interface.rs
@@ -14,7 +14,10 @@
 
 use proc_macro::TokenStream;
 use quote::{quote, ToTokens};
-use syn::{FnArg, Ident, ItemTrait, TraitItem, TraitItemMethod, Type};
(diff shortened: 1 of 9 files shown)


components/derive/src/service_interface.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub fn impl_service_interface(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut trait_item: ItemTrait =
+        syn::parse(item.clone()).expect("service_dispatcher can be used on traits only");
+
+    let vec_items = trait_item
+        .items
+        .iter()
+        .filter(|item| {
+            if let TraitItem::Method(_) = item {
```
popzxc: Maybe use `match`?

dvush: I prefer match too, but I thought that clippy is against this (its not).

PR #1291, opened 2019-04-09
Change service builder and web api.

dvush:
TODO
- [x] Store link to runtime in config service and rust runtime.
- [x] Call config service init tx in genesis block creation.
- [x] Add web api restart.
- [x] Fix tests compilation


exonum/src/helpers/fabric/mod.rs
```diff
@@ -349,14 +349,12 @@ pub trait CommandExtension {
 ///
 /// Services should provide implementation of this trait.
 pub trait ServiceFactory: 'static {
```
alekseysidorov: It seems that naming became completely confused.

exonum/src/runtime/rust/mod.rs
```diff
@@ -33,16 +32,28 @@ use super::{
...
-use crate::messages::{BinaryForm, CallInfo};
+use crate::messages::{AnyTx, BinaryForm, CallInfo};
 use crate::proto::schema;
 use crate::storage::{Error as StorageError, Fork, Snapshot};
 
-use self::service::Service;
+use self::service::{Service, ServiceFactory};
+use crate::runtime::configuration_new;
+use crate::runtime::dispatcher::Dispatcher;
+use protobuf::well_known_types::Any as PbAny;
```
popzxc: IMHO `well_known_types::Any` looks better than `PbAny`.

exonum/src/runtime/rust/mod.rs
```diff
@@ -37,12 +36,25 @@ use crate::messages::{BinaryForm, CallInfo};
...
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct RustRuntime {
-    // TODO: think about ways to share runtime.
-    inner: RefCell<RustRuntimeInner>,
+    inner: RustRuntimeInner,
+    dispatcher: *mut Dispatcher,
```
popzxc: IMHO in the future we should rework all that pointers and unsafe code... It just smells bad.

examples/cryptocurrency/src/lib.rs
```diff
@@ -314,28 +327,60 @@ pub mod contracts {
...
+    impl ServiceFactory for ServiceFactoryImpl {
+        fn artifact(&self) -> RustArtifactSpec {
+            artifact_spec()
+        }
+
+        fn new_instance(&self) -> Box<dyn Service> {
+            Box::new(CryptocurrencyServiceImpl)
+        }
+
+        fn genesis_init_info(&self) -> Vec<GenesisInitInfo> {
```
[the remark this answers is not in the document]
dvush: We could use rust types for services that require constructor data here and this way we don't care about how to store protobuf structs in config files.

exonum/src/helpers/fabric/mod.rs
dvush: Yes. Maybe some suggestions about how this could be named?
There is struct that creates service instances when new service is initialized and it is called (runtime::ServiceFactrory).
And there is struct that creates that ServiceFactory using arguments from config/command line (helpers::ServiceFactory).

exonum/src/runtime/rust/mod.rs
dvush: Yes, but how this should be implemented?
I tried using Weak-Rc. (Blockchain has Rc on Dispatcher), Config service has Weak on Dispatcher. 
But then when we execute Init transaction we have borrow on Dispatcher (since all transactions are calls to Dispatcher method) and we have to init service which will require mutable borrow on dispatcher. That is why I think all safe primitives such as RefCell and Rc can't be used here.
Maybe I am missing something.

popzxc: Currently I have no answer, that's a topic for further investigation.
Maybe one day we'll start using actor approach and then it'll be separate actors that interact asynchonously. Or we'll find other solution.

exonum/src/helpers/fabric/mod.rs
popzxc: I'd suggest `helpers::ServiceFactory` => `helpers::ArtifactProvider` and `runtime::ServiceFactory` => `runtime::Deployable`.
@alekseysidorov  what do you think about such a naming?

dvush:
Some details about this implementation.

1) To make basic service deployment and initialization work I used *mut reference to Dispatcher.
Here is why, lets say that we agree that Dispatcher will own other runtimes and all transactions should go through the dispatcher. Than we had to have some reference to dispatcher in rust runtime(for example to dispatch transactions from services to other runtimes, or to make initialization in genesis block like in this PR). and we should have some reference to dispatcher in configuration service since deploys are transactions executed by this service and we call deploy method of dispatcher when we execute transaction. 

All safe primitives such as RefCell and rc::Rc-rc::Weak won't help in this scenario since they all enforce rule: only one mutable reference is allowed at once. Since our transaction execution borrows Dispatcher (all txs go through Dispatcher) we can't have mutable reference to the same dispatcher. 

One way to go around it is to pass &mut Dispatcher to all methods of runtimes and then to services, this will require small change in API.

2) Service restart. Actix-web starts service in the same thread as the actor that spawns it, that is why service restart is made in this strange way: from ApiManager we create future that will send to the service "ServiceStop" message and after this future will be completed we send to ApiManager message that "ServiceStop" is finished and we can start new service on the same port. This future will be executed on the same thread as ApiManager and main service actor are executing.

PR #1298, opened 2019-04-19
Use `BinaryValue` trait instead of `BinaryForm`

pavel-mukhanov:
These traits are very similar, so we decided to combine them.

### Definition of Done

### DIFF
--- a/components/merkledb/Cargo.toml
+++ b/components/merkledb/Cargo.toml
@@ -32,6 +32,7 @@ smallvec = "0.6"
 tempfile = "3.0"
 uuid = { version="0.7", features = ["v4"] }
 rand = "0.6.5"
+protobuf = { version = "2.4.0", features = ["with-serde"] }
 
 [dev-dependencies]
 criterion = "0.2"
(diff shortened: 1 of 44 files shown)


components/merkledb/Cargo.toml
```diff
@@ -32,6 +32,7 @@ smallvec = "0.6"
 tempfile = "3.0"
 uuid = { version="0.7", features = ["v4"] }
 rand = "0.6.5"
+protobuf = { version = "2.4.0", features = ["with-serde"] }
```
aleksuss: AFAIK `2.5.0` is actual version.

components/merkledb/src/macros.rs
```diff
@@ -57,3 +57,22 @@ macro_rules! impl_object_hash_for_binary_value {
         )*
     };
 }
+
+#[macro_export]
+macro_rules! impl_binary_value_for_message {
+    ($( $type:ty ),*) => {
+        $(
+            impl BinaryValue for $type {
+                fn to_bytes(&self) -> Vec<u8> {
```
aleksuss: Does exist any reason why we don't return `Result<Vec<u8>, failure::Error>` here ?

PR #1327, opened 2019-05-30
Remove unsafe code from runtimes 

alekseysidorov:


exonum/src/runtime/dispatcher.rs
```diff
@@ -75,16 +71,17 @@ impl Dispatcher {
     }
 }
 
-impl RuntimeEnvironment for Dispatcher {
-    fn start_deploy(&mut self, artifact: ArtifactSpec) -> Result<(), DeployError> {
+// TODO think about runtime environment traits. [ECR-3222]
+impl Dispatcher {
```
popzxc: We can merge 2 `impl Dispatcher` blocks.

exonum/src/runtime/dispatcher.rs
```diff
@@ -137,7 +134,12 @@ impl RuntimeEnvironment for Dispatcher {
         }
 
         if let Some(runtime) = self.runtimes.get(&runtime_id.unwrap()) {
-            runtime.execute(context, call_info, payload)
+            runtime.execute(context, call_info, payload)?;
+            // Executes pending dispatcher actions.
+            context
```
popzxc: Maybe add a check if there is any action so we won't create an empty `Vec` on each transaction execute?

alekseysidorov: Empty vec doesn't allocate memory and thus `Vec::new` is very cheap.

PR #1328, opened 2019-05-30
WIP: Proof list lazy hash

pavel-mukhanov:
## Overview

Introduced `LazyListIndex` with lazy hash calculations. Hash is now calculated on demand by invoking `update_hashes` method. Currently `LazyListIndex` is slower than `ProofListIndex`.

---

### Definition of Done

### DIFF
--- a/components/merkledb/src/proof_list_index/lazy.rs
+++ b/components/merkledb/src/proof_list_index/lazy.rs
@@ -16,20 +16,28 @@
 
 pub use super::proof::{ListProof, ListProofError};
 
+use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
+
 use std::{
+    io::{Read, Write},
     marker::PhantomData,
+    mem::size_of,
     ops::{Bound, RangeBounds},
 };
 
-use exonum_crypto::Hash;
+use exonum_crypto::{Hash, HASH_SIZE};
 
 use super::{key::ProofListKey, proof::ProofOfAbsence};
-use crate::views::IndexAddress;
 use crate::{
     hash::HashTag,
-    views::{AnyObject, IndexAccess, IndexBuilder, IndexState, IndexType, Iter as ViewIter, View},
+    views::{
+        AnyObject, BinaryAttribute, IndexAccess, IndexAddress, IndexBuilder, IndexState, IndexType,
+        Iter as ViewIter, View,
+    },
     BinaryKey, BinaryValue, ObjectHash,
 };
+
+const LEN_SIZE: usize = size_of::<u64>();
 
 // TODO: Implement pop and truncate methods for Merkle tree. (ECR-173)
 
@@ -42,7 +50,7 @@ use crate::{
 #[derive(Debug)]
 pub struct LazyListIndex<T: IndexAccess, V> {
     base: View<T>,
-    state: IndexState<T, u64>,
+    state: IndexState<T, ProofListState>,
     _v: PhantomData<V>,
 }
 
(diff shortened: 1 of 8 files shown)


components/merkledb/src/proof_list_index/lazy.rs
```diff
@@ -0,0 +1,876 @@
...
+        IndexType::ProofList
+    }
+
+    fn metadata(&self) -> Vec<u8> {
+        self.state.metadata().to_bytes()
+    }
+}
+
+#[derive(Debug, Default, Copy, Clone)]
+struct ProofListState {
```
popzxc: It's a bit confusing that `LazyListIndex` has `ProofListState`.

components/merkledb/src/proof_list_index/lazy.rs
```diff
@@ -0,0 +1,876 @@
...
+    ///
+    /// let default_hash = index.object_hash();
+    /// assert_eq!(HashTag::empty_list_hash(), default_hash);
+    ///
+    /// index.push(1);
+    /// let hash = index.object_hash();
+    /// assert_ne!(hash, default_hash);
+    /// ```
+    fn object_hash(&self) -> Hash {
+        self.state.get().hash.expect("Hash is not calculated yet")
```
popzxc: That looks like a possibility to shoot yourself in the foot. Maybe if hash is required, it should be calculated right here right now? Or at least documentation should tell you that this method may panic.

pavel-mukhanov: Good point, but to calculate hash we need to obtain a mutable reference to self. It's not possible in some cases, for example when we creating index from the snapshot.

components/merkledb/src/proof_list_index/lazy.rs
pavel-mukhanov: My bad, I should say about it in PR description. I plan to replace ProofListIndex with LazyListIndex if the latter shows the required performance. Currently, two indices are needed for benchmarking.

components/merkledb/src/proof_list_index/lazy.rs
popzxc: Hmm... You have a method `calculate_and_save_hashes`. You can add method `calculate_hashes` without saving, it would calculate hash once and won't store it anywhere. It wouldn't require mutable reference, would it?

However, there should be a note that if this method is calles without `update_hashes` it will be time-consuming. But at least it won't panic.

alekseysidorov: You can rewrite index a little bit to make `calculate_and_save_hashes` immutable, thus for fork `object_hash` invocation will be update index but for snapshot you just only calculate object hash without saving.

slowli:
@pavel-mukhanov @aleksuss Here's a couple of thoughts on performance:

According to the initial idea for lazy Merkle root updates, we know all items in the list that were added or updated since the last hash calculation. If I understand correctly, this information is already stored as a part of `Fork` (in RAM, i.e., with faster access times than RocksDB). This means a couple of things:

- We don't need to overwrite intermediate hashes with the "null" value (which impacts performance).
- We don't need to traverse *all* items in `calculate_and_save_hashes`, but only necessary ones (which also impacts performance).

The tricky part is that the initial design assumed we don't need to recalculate object hashes when working with a `Fork`, mostly due to implementation of object hierarchies. I assume that this implementation is built with the opposite requirement (we do want to be able to compute hashes of objects, otherwise we'll break the existing code, e.g., the advanced cryptocurrency demo). It seems that the optimizations I've mentioned are still possible with this in mind, but they require complicating `Fork` internal structure.

PR #1335, opened 2019-06-03
Extended websocket interface

Flowneee:
## Overview

This PR introduce additional functionality to websocket interface. Old version allowed to subscribe only for new blocks events. New version add support of subscribing to new transactions (all or only from specific services with specific ID) and sending transactions via websocket connection.

### Details

Added new explorer endpoints:
 - `v1/transactions/subscribe`: this endpoint automatically subscribe to new transaction events. This endpoint accept optional parameters: `service_id` and `message_id` (`message_id` as in derive macro `TransactionSet`);
 - `v1/ws`: this endpoint open websocket connection and allow to set multiple subscription (for new blocks and transaction, filtered by service and transaction id) and send transactions (in hex, like explorer) to blockchain.

Messages, sent to websocket, must have format: `{"type": "...", "payload": ...}` (currently possible types: `set-subsctiptions`, `transaction`), where payload differs for different types.

Response for each sent message can be:
 - success: `{"result": "success"}` or `{"result": "success", "response": ...}`, if action assume some response;
 - error: `{"result": "error", "description": "..."}`.

**Note**: you can also set multiple subscription and set transactions through `v1/blocks/subscribe` and `v1/transactions/subscribe`, so they are like `v1/ws`, but with default filter.

### `v1/blocks/subscribe`

*  for subscribing to new blocks: `ws://localhost:8080/api/explorer/v1/blocks/subscribe`.

### `v1/transactions/subscribe`

*  for subscribing to all new transactions: `ws://localhost:8080/api/explorer/v1/transactions/subscribe`;
*  for subscribing to new transactions of service with id 1: `ws://localhost:8080/api/explorer/v1/transactions/subscribe?service_id=1`;
*  for subscribing to new transactions of service with id 1 and message id 2: `ws://localhost:8080/api/explorer/v1/transactions/subscribe?service_id=1&message_id=2`.

### `v1/ws` (multiple filters)

1.  Initiate connection: `ws://localhost:8080/api/explorer/v1/ws`;
2.  Send text message (via websocket) with list of filter (for example, blocks and transactions of service1 with message id 2): `{"type":"set-subscriptions","payload":[{"type":"blocks"},{"type":"transactions","filter":{"service_id":1,"message_id":2}}]}`;
3.  Update filter (leave only subscription for blocks), send text message: `{"type":"set-subscriptions","payload":[{"type":"blocks"}]}`.

### `v1/ws` (send transaction)

1.  Initiate connection: `ws://localhost:8080/api/explorer/v1/ws`;
2. Send text message (via websocket) with transaction hex (like in explorer): `{'type":"transaction","payload":{"tx_body":"53ef10e1b06e60d69ad72b21c628d11ef5beb1a1cfe828d6dcf1e4055093e1470000000000000a220a2053ef10e1b06e60d69ad72b21c628d11ef5beb1a1cfe828d6dcf1e4055093e1471205416c696365c50061fe2442b846179420c51cd19e6efafb4c20dc5272247dbb1d2549bcead9aee0805161fac631ed8483a41fdff3957a44cbacdbd8b13fcb827a0f913a120e"}}`.


exonum/src/api/websocket.rs
```diff
@@ -33,20 +137,33 @@ pub(crate) struct Message(pub String);
...
 
+#[derive(Message)]
+#[rtype("Result<TransactionResponse, failure::Error>")]
+pub(crate) struct Transaction {
+    tx: TransactionHex,
+}
+
 pub(crate) struct Server {
-    pub subscribers: HashMap<usize, Recipient<Message>>,
+    pub subscribers: HashMap<SubscriptionType, HashMap<usize, Recipient<Message>>>,
```
alekseysidorov: Subscription types are not many, and `HashMap` is less effective for them than the simple `BTreeMap`.

exonum/src/api/websocket.rs
```diff
@@ -33,20 +137,33 @@ pub(crate) struct Message(pub String);
 #[rtype(usize)]
 pub(crate) struct Subscribe {
     pub address: Recipient<Message>,
+    pub sub_types: Vec<SubscriptionType>,
```
alekseysidorov: Such name isn't so clear because it can be confused with subtypes.

exonum/src/api/websocket.rs
```diff
@@ -33,20 +137,33 @@ pub(crate) struct Message(pub String);
...
 }
 
 #[derive(Message)]
 pub(crate) struct Unsubscribe {
     pub id: usize,
 }
 
+#[derive(Message)]
+pub(crate) struct UpdateSubscriptions {
+    pub id: usize,
```
alekseysidorov: `usize` is platform dependent type and it shouldn't be used in network related code, use `u64` or `u32`.

Flowneee: Agreed, but original code (like `Unsubscribe` structure) also uses `usize`, so I should either update all places or leave it "as is". And since this id is just an internal id of session (and never leave server), I don't see how it could cause an error.

alekseysidorov: So it means that there was a flaw in original `Unsubscribe` code :) I think you may hardily fix it.

exonum/src/api/websocket.rs
Flowneee: Done

exonum/src/api/websocket.rs
Flowneee: Done (renamed to `subscriptions`)

exonum/src/api/websocket.rs
Flowneee: Done (now `u64` used)

aleksuss:
@Flowneee good job. Thanks a lot. But what do you think about to rename a query parameter `tranisaction_id` to `message_id` ? We use a `message_id` definition to differ messages (transactions) in our codebase. Take a look [here](https://exonum.com/doc/version/latest/architecture/transactions/#messages) for example.

Flowneee:
@aleksuss  Yes, this make sense. Done. Also added in changelog how to determine `message_id` (just mentioned `TransactionSet` macro).

Flowneee:
@aleksuss [Error](https://travis-ci.com/exonum/exonum/jobs/205271381) occurred during tests, on command `cd $TRAVIS_BUILD_DIR/testkit/server && npm run proto && npm run test:unix`. I don't think that it is because of my change. Can you look into it?

aleksuss:
@Flowneee I've restarted the job. I suppose it will fix that issue. It happens some times.

Flowneee:
> At the first sight, this PR looking good for me except of a several moments. Also may you write entry to the `CHANGELOG.md`?

@alekseysidorov I added entry to CHANGELOG.md and fixed all your comments. Can you check it?

exonum/src/api/websocket.rs
```diff
@@ -19,58 +19,206 @@ use actix_web::ws;
...
+    location: TxLocation,
+    proof: ListProof<Hash>,
+}
+
+impl TransactionInfo {
+    fn new<T>(schema: &Schema<T>, tx_hash: &Hash) -> Self
+    where
+        T: AsRef<dyn Snapshot> + IndexAccess,
+    {
+        let tx = schema.transactions().get(tx_hash).unwrap();
```
aleksuss: @Flowneee What if we don't have a transaction with hash `tx_hash` in the blockchain ? What do you think about to return a `Result` or `Option` here ?

Flowneee: @aleksuss *Currently* there is always will be a transaction, because the only place where `TransactionInfo` called is [here](https://github.com/exonum/exonum/pull/1335/files/5b833cf0865ec25bf7e54a50dc563f48eea4521f#diff-581fbced55c30b0ca3a20d76b965c630R277) (`tx_hash` in that case is always already committed transaction). Generally yes, probably good idea. And I even think that method should be not `new`, but something like `new_committed` to indicate that we want info about committed transaction, and this method should return `None` instead of each `.unwrap()`. Another option is to rename type to something like `CommittedTransactionInfo`. What do you say?

aleksuss: @Flowneee I've got your point and it's true. But `unwrap()` hurts my eyes in this place. I'd prefer a case with `None` instead of each `.unwrap()` but also you can use `.expect()` with a message like ("BUG. We don't have committed transaction in `transactions` index with hash: {:?}, tx_hash). Because the transaction is really MUST BE in the index. So. It's up to you.

Flowneee: Okay, done. `TransactionInfo` renamed to `CommittedTransactionSummary`, added logging of situation, when summary cannot be built.

Flowneee: @alekseysidorov @aleksuss @popzxc Can you check it again?

PR #1351, opened 2019-06-12
Create fork from patch.

pavel-mukhanov:
Create fork from patch


aleksuss:
@pavel-mukhanov Do you think some tests would be redundant for this functionality ?

components/merkledb/src/db.rs
```diff
@@ -527,9 +471,19 @@ pub trait Iterator {
     fn peek(&mut self) -> Option<(&[u8], &[u8])>;
 }
 
-impl Snapshot for FlushedFork {
+impl Patch {
+    pub fn changes(&self) -> HashMap<String, Changes> {
+        self.changes.clone()
+    }
+
+    pub fn iter(&self) -> HmIter<String, Changes> {
```
aleksuss: @pavel-mukhanov what is a `HmIter` ???

aleksuss: Ok. I got it.

exonum/src/node/state.rs
```diff
@@ -352,8 +352,8 @@ impl BlockState {
     }
 
     /// Returns the changes that should be made for block committing.
-    pub fn patch(&self) -> &Patch {
-        &self.patch
+    pub fn patch(&mut self) -> Patch {
+        self.patch.take().unwrap()
```
aleksuss: Is the `unwrap` is really safe here ?

components/merkledb/src/db.rs
```diff
@@ -567,12 +521,24 @@ impl Snapshot for FlushedFork {
 }
 
 impl Fork {
+    /// Creates a fork based on the provided `patch`.
+    ///
+    /// Note: using created fork to modify data already present in `patch` may lead
+    /// to an inconsistent database state. Hence, this method is useful only if you
+    /// are sure that the fork and `patch` interacted with different indices.
+    pub fn from_patch(patch: Patch) -> Self {
```
alekseysidorov: What do you think about implementing `From<Patch> for Fork`?

exonum/src/node/state.rs
pavel-mukhanov: Yes, because this patch must be used only once. I change `unwrap` to `expect` for clarity.

components/merkledb/src/db.rs
pavel-mukhanov: I've changed this method to accept `snapshot` for consistency, therefore currently I can't implement `From<Patch`.

aleksuss:
Wait for merging https://github.com/rust-rocksdb/rust-rocksdb/pull/313

components/merkledb/Cargo.toml
```diff
@@ -23,7 +23,9 @@ failure = "0.1"
 hex = "0.3.2"
 leb128 = "0.2"
 num-traits = "0.2"
-rocksdb = "0.12.0"
+
+# TODO: revert to original rocksdb
+rocksdb = { git = "https://github.com/pavel-mukhanov/rust-rocksdb", branch = "sync_snapshot" }
```
aleksuss:
```suggestion
rocksdb = "0.12.3"
```
0.12.3 has been released !

components/merkledb/Cargo.toml
```diff
@@ -23,7 +23,8 @@ failure = "0.1"
 hex = "0.3.2"
 leb128 = "0.2"
 num-traits = "0.2"
-rocksdb = "0.12.0"
+
+rocksdb = "0.12.3"
```
aleksuss: @pavel-mukhanov Why is an empty line here ?

components/merkledb/src/db.rs
```diff
@@ -567,12 +521,27 @@ impl Snapshot for FlushedFork {
 }
 
 impl Fork {
+    /// Creates a fork based on the provided `patch` and `snapshot`.
+    ///
+    /// Note: using created fork to modify data already present in `patch` may lead
+    /// to an inconsistent database state. Hence, this method is useful only if you
+    /// are sure that the fork and `patch` interacted with different indices.
+    pub fn from_patch(patch: Patch, snapshot: Box<dyn Snapshot>) -> Self {
```
aleksuss: Do we really need `snapshot` parameter here ? We have it in Patch ? 
```rust
Self {
             patch: Patch {
                 changes: patch.changes,
                 snapshot: patch.snapshot,
             },
             working_patch: WorkingPatch::new(),
         }
}
```

pavel-mukhanov: We can't use snapshot from patch, because it's doesn't have changes from mempool made by `handle_tx`. We can use it only after we will rewrite mempool.

aleksuss: Yeah. I got it. It's a really pain. It makes public api ugly.

alekseysidorov: This method is full of dangerous dark corners and looks very ugly. It should not be public method.
And moreover, we have to find solution that does not need such methods.
For example, we can keep fork in transactions pool and at the moment of commit block use this fork as current blockchain state. But during the proposal execution we can rollback changes from fork after that propose will be created.

components/merkledb/src/db.rs
aleksuss: @pavel-mukhanov Do we really need this method in case we implement `From<Patch` trait for `Fork` ? I suppose we can get rid of it at all, can't we ?

aleksuss:
@pavel-mukhanov, could be closed ?

pavel-mukhanov:
@aleksuss yes

PR #1361, opened 2019-06-19
FIrst step of persistent dynamic services implementation [ECR-3276]

alekseysidorov:


exonum/src/node/mod.rs
```diff
@@ -275,6 +275,13 @@ pub struct NodeConfig<T = SecretKey> {
     pub thread_pool_size: Option<u8>,
 }
 
+impl<T: Clone> NodeConfig<T> {
+    /// Returns service keypair.
```
aleksuss: Maybe specify why we used generic here ?

exonum/src/blockchain/builder.rs
```diff
@@ -0,0 +1,213 @@
...
+            })?;
+            // Commits genesis block.
+            blockchain.create_genesis_block(self.genesis_config)?;
+            blockchain
+        })
+    }
+}
+
+#[derive(Debug)]
+pub struct ServiceInstances {
```
aleksuss: `ServiceInstances` is a little bit weird as for me. What about `ServiceInstancesContainer` or `ServiceInstancesHolder` or something like this but in the singular ?

exonum/src/runtime/rust/tests.rs
```diff
@@ -174,10 +174,10 @@ fn test_basic_rust_runtime() {
             },
         };
 
-        let fork = db.fork();
+        let mut fork = db.fork();
```
aleksuss: `mut` is redundant here, isn't it ?

PR #1365, opened 2019-06-21
Add some basic node config validation [ECR-3125]

ozkriff:
## Overview

This PR adds some basic validation for `mempool.events_pool_capacity` parameters.

I was going to use the [Keats/validator](https://github.com/Keats/validator) crate initially, but it doesn't save much code and it's `#[validate(range(min = "18", max = "20"))]` attribute requires both min and max for some reason. So I decided to just add a plain validation function that can be modified easily in case of any refactorings of the configuration logic.

---

See: https://jira.bf.local/browse/ECR-3125

### DIFF
--- a/exonum/src/node/mod.rs
+++ b/exonum/src/node/mod.rs
@@ -314,6 +314,38 @@ impl NodeConfig<PathBuf> {
     }
 }
 
+impl<T> NodeConfig<T> {
+    fn validate_or_panic(&self) {
+        let capacity = &self.mempool.events_pool_capacity;
+        if capacity.internal_events_capacity < 3 {
+            panic!(
+                "internal_events_capacity({}) must be strictly larger than 2",
+                capacity.internal_events_capacity
+            );
+        }
+        if capacity.network_requests_capacity == 0 {
+            panic!(
+                "network_requests_capacity({}) must be strictly larger than 0",
+                capacity.network_requests_capacity
+            );
+        }
+        // Sanity checks for cases of accidental negative overflows.
+        let sanity_max = 2_usize.pow(16);
+        if capacity.internal_events_capacity > sanity_max {
+            panic!(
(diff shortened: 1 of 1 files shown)


aleksuss:
@ozkriff Why do not use custom validator ? https://github.com/Keats/validator#custom

ozkriff:
@aleksuss A custom validator (something like `fn min_value`) can be added, but the top-level `.validate()` method must be called manually anyway, so it doesn't actually save much code relatively to a plain and simple validation function.

What we really want is serde to support validation attributes. :(

exonum/src/node/mod.rs
```diff
@@ -314,6 +314,24 @@ impl NodeConfig<PathBuf> {
     }
 }
 
+impl<T> NodeConfig<T> {
+    fn validate_or_panic(&self) {
+        let capacity = &self.mempool.events_pool_capacity;
+        if capacity.internal_events_capacity == 0 {
```
aleksuss: I'd say that `internal_events_capacity` should be more than zero. But there is a little problem here. `internal_events_capacity` has a `usize` type. So if you specify negative number the overflow occur and we'll get huge size of the capacity.

aleksuss: And take into account this case in tests please.

ivan-ochc:
@ozkriff, @aleksuss What is the minimum value available to set for internal_events_capacity? Seems, that node doesn't work in case of internal_events_capacity<=2.

exonum/src/node/mod.rs
ozkriff: Added more sanity checks and a few tests.

ozkriff:
Added more sanity checks and tests + rebased on top of master.

ozkriff:
ugh, i'm not sure what kcov wants from me

PR #1367, opened 2019-06-25
Rename ArtifactSpec to ArtifactId [ECR-3291]

alekseysidorov:
Implemented first point of task.


exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -26,21 +26,13 @@ message AnyTx {
...
+  uint32 runtime = 1;
+  string raw_id = 2;
 }
 
 message InstanceSpec {
   uint32 id = 1;
   string name = 2;
-  ArtifactSpec artifact = 3;
+  ArtifactId artifact = 3;
 }
```
aleksuss: New line

PR #1382, opened 2019-07-12
Replace impl_service_dispatcher by the attribute in service_interface [ECR-3222]

alekseysidorov:


examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -95,7 +95,7 @@ pub struct CreateWallet {
 }
 
 /// Cryptocurrency service transactions.
-#[service_interface]
+#[service_interface(exonum(dispatcher = "CryptocurrencyService"))]
```
slowli: Is it necessary to wrap macro arguments into `exonum(..)`? This seems redundant.

alekseysidorov: This is only for uniformity with the `#[derive(ProtobufConvert)]`. 
I suppose that if we don't need to support it, I can rename `service_interface` into `exonum_service` and remove redundant nesting level.

PR #1398, opened 2019-07-30
Cache transactions before adding them to pool

pavel-mukhanov:
## Overview
Added tx_cache to store incoming transactions. 

---

### Definition of Done


exonum/benches/criterion/block.rs
```diff
@@ -106,7 +106,12 @@ fn create_blockchain(db: impl Database, services: Vec<Box<dyn Service>>) -> Bloc
 }
 
 fn execute_block(blockchain: &Blockchain, height: u64, txs: &[Hash]) -> (Hash, Patch) {
-    blockchain.create_patch(ValidatorId::zero(), Height(height), txs)
+    blockchain.create_patch(
+        ValidatorId::zero(),
+        Height(height),
+        txs,
+        &mut BTreeMap::new(),
```
aleksuss:
```suggestion
        None,
```
I suppose `BTreeMap::new` not free operation.

alekseysidorov: https://github.com/rust-lang/rust/blob/master/src/liballoc/collections/btree/node.rs#L127
It seems that `BtreeMap::new` is cheap operation.

exonum/src/blockchain/schema.rs
```diff
@@ -463,8 +464,16 @@ where
     }
 
     /// Changes the transaction status from `in_pool`, to `committed`.
-    pub(crate) fn commit_transaction(&mut self, hash: &Hash) {
-        self.transactions_pool().remove(hash);
+    pub(crate) fn commit_transaction(&mut self, hash: &Hash, tx: Signed<RawTransaction>) {
+        if !self.transactions().contains(hash) {
```
alekseysidorov: Not a very clear case, what means if we are trying to commit a transaction that is not in `self.transactions()`?

exonum/src/blockchain/schema.rs
```diff
@@ -510,3 +519,13 @@ where
         Height(self.block_hashes_by_height().len())
     }
 }
+
+/// Return transaction from persistent pool. If transaction is not present in pool, try
+/// to return it from transactions cache.
+pub fn get_tx<T: IndexAccess>(
```
alekseysidorov: Why this helper function is public?

exonum/src/blockchain/tests.rs
```diff
@@ -468,7 +473,7 @@ mod rocksdb_tests {
     fn create_blockchain_with_service(path: &Path, service: Box<dyn Service>) -> Blockchain {
         let db = create_database(path);
         let service_keypair = gen_keypair();
-        let api_channel = mpsc::unbounded();
+        let api_channel = mpsc::channel(1);
```
alekseysidorov: `channel(0)` is more useful, in my opinion.

exonum/benches/criterion/block.rs
```diff
@@ -106,7 +106,12 @@ fn create_blockchain(db: impl Database, services: Vec<Box<dyn Service>>) -> Bloc
 }
 
 fn execute_block(blockchain: &Blockchain, height: u64, txs: &[Hash]) -> (Hash, Patch) {
-    blockchain.create_patch(ValidatorId::zero(), Height(height), txs)
+    blockchain.create_patch(
+        ValidatorId::zero(),
+        Height(height),
+        txs,
```
alekseysidorov: So I think that we have to unify this to arguments by the single trait `TransactionsSource` with the good interface in future.

exonum/benches/criterion/block.rs
aleksuss: But benchmarks say otherwise :)

pavel-mukhanov: This code with `BtreeMap::new` is used only in benchmarks or tests. It doesn't affect the production code. But adding `Option` will result in unnecessary unwraps.

exonum/src/blockchain/schema.rs
pavel-mukhanov: Because now we have transactions that stored only in cache during the round and added to storage after commit(or before the node is stopped).

exonum/src/blockchain/schema.rs
pavel-mukhanov: No reason, I will change this.

exonum/src/api/node/public/system.rs
```diff
@@ -74,15 +76,17 @@ impl SystemApi {
     }
 
     fn handle_stats_info(self, name: &'static str, api_scope: &mut ServiceApiScope) -> Self {
+        let self_ = self.clone();
```
aleksuss: I guess we could get rid of the clone. We just can get a `tx_cache_size` here and this value will move into the closure scope. Nope ?

aleksuss:
```rust
fn handle_stats_info(self, name: &'static str, api_scope: &mut ServiceApiScope) -> Self {
    let tx_cache_size = self.shared_api_state.tx_cache_size();
    api_scope.endpoint(name, move |state: &ServiceApiState, _query: ()| {
        let snapshot = state.snapshot();
        let schema = Schema::new(&snapshot);
        Ok(StatsInfo {
            tx_pool_size: schema.transactions_pool_len(),
            tx_count: schema.transactions_len(),
            tx_cache_size,
        })
    });
    self
}
```

pavel-mukhanov: In your variant, we will obtain tx_cache_size only once and then use this captured value for all requests.

aleksuss: Sure. My fault.

PR #1403, opened 2019-08-05
Create fork from patch.

pavel-mukhanov:
## Overview
Implement `From<Patch>` trait for `Fork`. 
---

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -182,7 +182,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Added `len` method to `KeySetIndex` and `ValueSetIndex`. (#1319)
 
-- Added `From<Patch>` trait implementation to `Fork`. (#1345)
+- Added `From<Patch>` trait implementation to `Fork`. (#1403)
 
 #### exonum-testkit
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -182,6 +182,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Added `len` method to `KeySetIndex` and `ValueSetIndex`. (#1319)
 
+- Added `From<Patch>` trait implementation to `Fork`. (#1345)
```
```suggestion
- Added `From<Patch>` trait implementation to `Fork`. (#1403)
```

components/merkledb/src/db.rs
```diff
@@ -561,7 +515,7 @@ impl Snapshot for FlushedFork {
     }
 
     fn contains(&self, name: &str, key: &[u8]) -> bool {
```
aleksuss: Could it be replace for `self.get(key).is_some()` ?

aleksuss: But if use such approach we need a third private method which should return a `Option<&[u8]>` and then in `get` method use clone of the result. It'd slightly complicate the code. So it's better to leave as is, I suppose.

pavel-mukhanov: We can remove `contains` at all because we have a default implementation that works as your said(`self.get(key).is_some()`). But current implementation of `contains` doesn't require cloning the value.

PR #1411, opened 2019-08-09
Refine Runtime trait [ECR-3412]

alekseysidorov:


dmitry-timofeev:
@alekseysidorov is the base branch correct/shan't it be rebased atop of anything (documentation branch)?

exonum/src/runtime/mod.rs
```diff
@@ -255,9 +258,20 @@ pub trait Runtime: Send + Debug + 'static {
...
+    #[doc(hidden)]
+    fn api_endpoints(&self, _context: &ApiContext) -> Vec<(String, ServiceApiBuilder)> {
         Vec::new()
     }
+
+    /// Notify runtime about changes in list of available in API service instances.
+    ///
+    /// The purpose of this method is provide building blocks to create your own API processing
+    /// mechanisms.
+    fn notify_api_changes(&self, _context: &ApiContext, _changes: &[ApiChange]) {}
```
dmitry-timofeev: 1. What is the relative order of `api_endpoints` and `notify_api_changes`? 
Can the runtime implement both?
2. (will be more important when we get proper stop_service) What is the relative order between `stop_service` and a request to make service "A" API inaccessible?
3. (will be more important when we get proper stop_service)  Can this method be blocking?

Also, I wonder if it would make sense to combine `api_endpoints` and `notify_api_changes` into a single method that allows to either create API handlers for Exonum API server, or manage runtime-specific API connections 🤔

exonum/src/runtime/mod.rs
```diff
@@ -359,10 +373,34 @@ impl<'a> ExecutionContext<'a> {
...
 }
+
+impl From<InstanceDescriptor<'_>> for (InstanceId, String) {
+    fn from(descriptor: InstanceDescriptor) -> Self {
+        (descriptor.id, descriptor.name.to_owned())
+    }
+}
+
+/// Change in list of service instances.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
```
vitvakatu: What is the purpose of deriving `Ord` and `PartialOrd`, as this struct can't be compared to others?

exonum/src/runtime/mod.rs
```diff
@@ -359,10 +373,34 @@ impl<'a> ExecutionContext<'a> {
...
+}
 
-    pub(crate) fn take_actions(&mut self) -> Vec<dispatcher::Action> {
-        let mut other = Vec::new();
-        std::mem::swap(&mut self.actions, &mut other);
-        other
+/// Instance descriptor contains information to access running service instance.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct InstanceDescriptor<'a> {
+    /// The current service instance identifier.
```
dmitry-timofeev: ~current~

exonum/src/runtime/mod.rs
```diff
@@ -359,10 +373,34 @@ impl<'a> ExecutionContext<'a> {
...
     }
+}
 
-    pub(crate) fn take_actions(&mut self) -> Vec<dispatcher::Action> {
-        let mut other = Vec::new();
-        std::mem::swap(&mut self.actions, &mut other);
-        other
+/// Instance descriptor contains information to access running service instance.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct InstanceDescriptor<'a> {
```
dmitry-timofeev: I'd suggest expanding the docs on id/name slightly: 

https://github.com/dmitry-timofeev/exonum-java-binding-1/blob/ddadb22fd66a52b51d0541fcb99aee67dce65b62/exonum-java-binding/core/src/main/java/com/exonum/binding/core/runtime/ServiceInstanceSpec.java#L12-L25

exonum/src/runtime/mod.rs
```diff
@@ -255,9 +258,20 @@ pub trait Runtime: Send + Debug + 'static {
...
+    /// Collect the full list of API handlers from the runtime for the built-in Exonum API server.
+    ///
+    /// This method is called during an API server restart. Use this method if you do not plan to
+    /// use your own API processing mechanism.
+    #[doc(hidden)]
+    fn api_endpoints(&self, _context: &ApiContext) -> Vec<(String, ServiceApiBuilder)> {
         Vec::new()
     }
+
+    /// Notify runtime about changes in list of available in API service instances.
```
```suggestion
    /// Notify runtime about changes in list of service instances.
```

exonum/src/runtime/mod.rs
alekseysidorov: `api_endpoints` is a transition method, not to break the existing `RustRuntime` code, in the future it will be replaced entirely by the `notify_api_change`.

exonum/src/runtime/mod.rs
```diff
@@ -262,6 +262,9 @@ pub trait Runtime: Send + Debug + 'static {
     ///
     /// This method is called during an API server restart. Use this method if you do not plan to
     /// use your own API processing mechanism.
+    /// 
+    /// Warning! It is a temporary method which not to break the existing `RustRuntime` code,
```
dmitry-timofeev: "It is a temporary method [that exists/existing/present] to not break the existing `RustRuntime` code,"? Maria might suggest a better option. Though it is internal thing, and we as intended audience understand it ^_^

exonum/src/runtime/types.rs
```diff
@@ -214,9 +214,15 @@ impl FromStr for ArtifactId {
...
+    ///  Exonum assigns it to the service on instantiation. It is mainly used to route the
+    /// transaction messages belonging to this instance.
     pub id: InstanceId,
-    /// Unique service instance name.
+    /// The unique name of the service instance. 
+    /// 
+    /// It serves as the primary identifier of this service in most operations.
+    /// It is assigned by the network administrators.
     ///
     /// The name must correspond to the following regular expression: `[a-zA-Z0-9/\.:-_]+`
```
dmitry-timofeev: Isn't 'match' a proper term: "The name must match the following regular expression:"

exonum/src/runtime/mod.rs
```diff
@@ -255,9 +258,23 @@ pub trait Runtime: Send + Debug + 'static {
...
+    #[doc(hidden)]
+    fn api_endpoints(&self, _context: &ApiContext) -> Vec<(String, ServiceApiBuilder)> {
         Vec::new()
     }
+
+    /// Notify runtime about changes in list of service instances.
+    ///
+    /// The purpose of this method is provide building blocks to create your own API processing
+    /// mechanisms.
+    fn notify_api_changes(&self, _context: &ApiContext, _changes: &[ApiChange]) {}
```
dmitry-timofeev: Is it called before or after the service has been configured? I.e. will its API become externally available to requests before configuration?

exonum/src/api/mod.rs
```diff
@@ -316,89 +205,142 @@ pub trait ExtendApiBackend {
...
+        let sender = context.sender().clone();
+        self::node::private::SystemApi::new(sender, NodeInfo::new(), shared_api_state.clone())
             .wire(builder.private_scope());
-        self::node::public::SystemApi::new(shared_api_state).wire(builder.public_scope());
+        self::node::public::SystemApi::new(context, shared_api_state).wire(builder.public_scope());
         builder
     }
 }
+
+/// Provide the current state of the blockchain to the API handlers.
```
popzxc: NItpick: since this is a struct and not a function, I'd suggest "This entity provides".

exonum/src/api/mod.rs
```diff
@@ -316,89 +205,142 @@ pub trait ExtendApiBackend {
...
+///
+/// This context contains a necessary parts to interaction with blockchain
```
popzxc: Nitpick: "for the blockchain interaction"/"to interact with the blockchain".

exonum/src/api/mod.rs
```diff
@@ -316,89 +205,142 @@ pub trait ExtendApiBackend {
...
+            service_keypair: blockchain.service_keypair.clone(),
+            api_sender: blockchain.api_sender.clone(),
+            database: blockchain.db.clone(),
+        }
+    }
+
+    /// Create a new blockchain database state snapshot.
+    ///
+    /// Be careful with this method! It is recommended to get a snapshot only once per request
+    /// to ensure all database accesses see the same state. When a snapshot is requested
```
popzxc: 1. "all the database accesses".
2. "When the shapshot is requested".

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -235,16 +247,17 @@ impl Dispatcher {
...
-                    e
-                })
+                Self::configure_service(runtime.as_ref(), fork, spec.as_descriptor(), constructor)
+                    .map_err(|e| {
+                        error!(
+                            "An error occurred while configuring the service {}: {}",
+                            spec.name, e
+                        );
+                        if let Err(e) = runtime.stop_service(spec.as_descriptor()) {
+                            panic!(FatalError::new(e.to_string()))
```
popzxc: I don't really like this panic. If the service didn't stop correctly, I guess it's better to write an error to the log, and just mark somehow this service as not available.
Service code is client-side and I think it shouldn't be able to crash the core.

exonum/src/runtime/mod.rs
```diff
@@ -255,9 +258,23 @@ pub trait Runtime: Send + Debug + 'static {
...
+    /// use your own API processing mechanism.
+    ///
+    /// Warning! It is a temporary method which not to break the existing `RustRuntime` code,
+    /// in future it will be removed.
+    #[doc(hidden)]
+    fn api_endpoints(&self, _context: &ApiContext) -> Vec<(String, ServiceApiBuilder)> {
         Vec::new()
     }
+
+    /// Notify runtime about changes in list of service instances.
```
popzxc: "Notify runtime about changes in the list"

exonum/src/runtime/mod.rs
```diff
@@ -255,9 +258,23 @@ pub trait Runtime: Send + Debug + 'static {
...
+    ///
+    /// The purpose of this method is provide building blocks to create your own API processing
```
popzxc: "The purpose of this method is to provide the building blocks"

exonum/src/runtime/mod.rs
alekseysidorov: This method is called last even after `after_commit`.

exonum/src/runtime/dispatcher/mod.rs
alekseysidorov: Ok I add `TODO` to handle this case.

exonum/src/explorer/mod.rs
```diff
@@ -512,12 +512,12 @@ pub struct BlockchainExplorer<'a> {
 }
 
 impl<'a> BlockchainExplorer<'a> {
-    /// Creates a new `BlockchainExplorer` instance.
+    /// Create a new `BlockchainExplorer` instance.
```
dmitry-timofeev: Isn't it common to use 3rd person verb phrase as a summary fragment: https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment 

The standard Rust docs also follow this principle: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.with_capacity

popzxc: I think that should be a limit for nitpicking PRs. Otherwise every PR may become an endless sequense of "little fixes".

According to the pointed problem, I think it's okay. We have a lot of little inconsistencies in the docs. When we'll have a tech writers team for this, we'll solve it. But now I think it's more important to write new code, than support completely brilliant documentation.

Maria-Nosyk: I am already confused which variant you prefer more. Both have explanations behind them. Just use the one that is more convenient for you. I will follow your preferences in my corrections

dmitry-timofeev: There is a difference between _API_ documentation (which follows the linked format), and _implementation_ documentation (which is usually written as commands). We use both consistently in appropriate places, with no 'preferences'.

exonum/src/runtime/mod.rs
```diff
@@ -255,9 +258,23 @@ pub trait Runtime: Send + Debug + 'static {
...
+    #[doc(hidden)]
+    fn api_endpoints(&self, _context: &ApiContext) -> Vec<(String, ServiceApiBuilder)> {
         Vec::new()
     }
+
+    /// Notify the runtime about the changes in the list of service instances.
+    ///
+    /// The purpose of this method is to provide building blocks to create your own
+    /// API processing mechanisms.
+    fn notify_api_changes(&self, _context: &ApiContext, _changes: &[ApiChange]) {}
```
dmitry-timofeev: What shall the runtime implementation do if some services fail to mount the API (e.g., an Exception occurs)? How does it communicate that to the core? Shall this error affect other services-to-be-connected?

Shan't we instead have a method notifying that _a single_ service is to be connected or disconnected from the web server? Otherwise the communication of potential errors in API configuration becomes complicated.

exonum/src/explorer/mod.rs
Maria-Nosyk: We had a similar situation with EJB documentation where my corrections from infinitive to 3rd person forms were rejected for the reason that infinitive is allowable in such cases...

I absolutely agree with what you say about _implementation_ and _API_ docs. However, there **is** a problem in your following these rules. Anyway, we need a style guide and I will develop one for common use to avoid such situations in the future.

dmitry-timofeev: The EJB docs used and continue to use the infinitive in implementation docs (which is a different species from method API docs). Would you expand on **the** problem — we don't seem to have any with it?

Regarding this particular case, nowhere I said it is anything more than a suggestion — keep whatever you are happy with, as it does not affect the meaning.

PR #1412, opened 2019-08-09
Remove len from MapIndex and ProofMapIndex

pavel-mukhanov:


vitvakatu:
Do we have a Jira link explaining why this change is reverted? We've already implemented the support in Java Bindings

aleksuss:
@vitvakatu No, we don't. The reason we reverted is wrong implementation. But if fix this we will get regression in performance. We need to check snapshot when we insert or remove element for its existence if don't have it in changes. As you know It's an expensive operation. So we've decided to rollback this functionality.

PR #1414, opened 2019-08-12
Add length to ListProof.

pavel-mukhanov:
## Overview
Add length to ListProof.

```json
{
   "length":10,
   "proof":{
      "left":"8b5a8d3579661f029c8fe29d951006142ccd409560b8e27f543ae0e7e83ddc38",
      "right":{
         "left":{
            "left":{
               "left":"f5d670eedd61a826276abeb5484f734d6bf30b03931258c3f0e17182ddf171bc",
               "right":{
                  "val":9
               }
            }
         }
      }
   }
}
```


components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -58,17 +58,75 @@ impl ProofOfAbsence {
...
+
+impl<V> ListProof<V>
+where
+    V: BinaryValue,
+{
+    pub fn new(length: u64, proof: ProofVariant<V>) -> Self {
+        Self { length, proof }
+    }
+
+    // Length of the corresponding ProofList.
```
aleksuss: Obviously it should be comment doc ?

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -58,17 +58,75 @@ impl ProofOfAbsence {
...
+pub struct ListProof<V> {
+    length: u64,
+    proof: ProofVariant<V>,
+}
+
+impl<V> ListProof<V>
+where
+    V: BinaryValue,
+{
+    pub fn new(length: u64, proof: ProofVariant<V>) -> Self {
```
aleksuss: Is `pub(crate)` visibility not enough in this case ?

alekseysidorov: I wonder if one want to construct synthetic proofs to check API endpoints.

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -58,17 +58,75 @@ impl ProofOfAbsence {
...
+        Self { length, proof }
+    }
+
+    // Length of the corresponding ProofList.
+    pub fn length(&self) -> u64 {
+        self.length
+    }
+
+    // Proof.
```
alekseysidorov: There is a lack of documentation.

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -309,3 +330,69 @@ where
...
+        D: Deserializer<'a>,
+    {
+        fn format_err_string(type_str: &str, value: &Value, err: &SerdeJsonError) -> String {
+            format!(
+                "Couldn't deserialize {} from serde_json::Value: {}, error: {}",
+                type_str, value, err
+            )
+        }
+
+        let json: Value = <Value as Deserialize>::deserialize(deserializer)?;
```
alekseysidorov: This code looks very strange for me. We use generic serialization and concrete json for deserialization. 
It seems that we should write a special serde wrapper structure for this one.

Can you add `TODO` with the `[ECR-3222]`  marker about this problem or fix it if fix is trivial?

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -58,17 +58,77 @@ impl ProofOfAbsence {
...
+    pub fn length(&self) -> u64 {
+        self.length
+    }
+
+    /// Underlying proof variant.
+    pub fn proof_variant(&self) -> &ProofVariant<V> {
+        &self.proof
+    }
+
+    /// Verifies the correctness of the proof by the trusted list hash and the number of
```
dmitry-timofeev: 'the number of elements in the tree' is no longer true, as this method accepts the expected list hash only.

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -58,17 +58,77 @@ impl ProofOfAbsence {
...
+    /// elements in the tree.
+    ///
+    /// To validate the proof one needs to provide `expected_list_hash` calculated as follows:
```
dmitry-timofeev: Is it relevant here how the list hash is calculated if the client does not have access to `merkle_root`, but usually gets the `expected_list_hash` from the block header and/or aggregating collection of state hashes?

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -58,17 +58,77 @@ impl ProofOfAbsence {
...
+    /// and `length` of the `ProofListIndex`.
+    ///
+    /// If the proof is valid, a vector with indices and references to elements is returned.
+    /// Otherwise, `Err` is returned.
+    ///
+    /// If the proof is the proof of absence, then empty vector will be returned.
+    pub fn validate(&self, expected_list_hash: Hash) -> Result<Vec<(u64, &V)>, ListProofError> {
+        let mut vec = Vec::new();
+        let height = self.length.next_power_of_two().trailing_zeros() as u8 + 1;
+        let root_hash =
```
dmitry-timofeev: is it _root_ hash (= the Merkle root of the tree) or the _list_ hash (the Merkle root combined with the length and the tag)? Below `list_hash` is compared with `root_hash` which does not seem correct.

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -58,17 +58,77 @@ impl ProofOfAbsence {
...
+            self.proof_variant()
+                .collect(ProofListKey::new(height, 0), self.length, &mut vec)?;
+
+        match self.proof_variant() {
+            ProofVariant::Absent(_) => {
+                if expected_list_hash != root_hash {
```
dmitry-timofeev: ⚠️  🔼

components/merkledb/src/proof_list_index/proof.rs
pavel-mukhanov: Clients must provide `expected_list_hash`, currently `merkle_root` method is private.

components/merkledb/src/proof_list_index/proof.rs
pavel-mukhanov: Collect method returns either root_hash and merkle_root. I'll fix that.

components/merkledb/src/proof_list_index/proof.rs
dmitry-timofeev: That is what I am saying: they must provide it, but they don't calculate it themselves from the merkle root but get it from somewhere.

components/merkledb/src/proof_list_index/mod.rs
```diff
@@ -431,11 +433,13 @@ where
             )
         }
 
-        if to > self.len() {
-            ListProof::Absent(ProofOfAbsence::new(self.len(), self.merkle_root()))
+        let proof = if to > self.len() {
```
slowli: Semi-offtopic: I find it slightly weird / unexpected that in case the range proof is *partially* valid (e.g., we request elements `5..10` in the three with 8 elements), we get no elements in return.

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -19,58 +19,83 @@ use exonum_crypto::Hash;
 
 use super::{super::BinaryValue, key::ProofListKey, HashTag};
 
-/// Encapsulates a proof of absence for `ProofListIndex`.
+/// An enum that represents a proof of existence/absence for a proof list elements.
```
slowli: Nit: "for a proof list elements" -> "for elements in a proof list".

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -19,58 +19,83 @@ use exonum_crypto::Hash;
...
+    /// A branch of proof in which both children contain requested elements.
+    Full(Box<ProofVariant<V>>, Box<ProofVariant<V>>),
+    /// A branch of proof in which only the left child contains requested elements.
+    Left(Box<ProofVariant<V>>, Option<Hash>),
+    /// A branch of proof in which only the right child contains requested elements.
+    Right(Hash, Box<ProofVariant<V>>),
+    /// A leaf of the proof with the requested element.
+    Leaf(V),
+    /// Proof of absence of an element with the specified index.
+    Absent(Hash),
```
slowli: Am I correct that `Absent` variant can occur only as the direct value of `ListProof.proof`? In this case, the variant doc comment seems misleading: the variant does not proof an absence of a specific element, but rather a proof of absence of a range. (Arguably, even this would be incorrect; see the comment about partially valid ranges above.)

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -87,90 +112,52 @@ pub enum ListProofError {
...
                 if key.height() > 1 {
                     return Err(ListProofError::UnexpectedLeaf);
                 }
                 vec.push((key.index(), value));
                 HashTag::hash_leaf(&value.to_bytes())
             }
-            ListProof::Absent(ref proof) => {
-                HashTag::hash_list_node(proof.length, proof.merkle_root)
-            }
+            ProofVariant::Absent(merkle_root) => merkle_root,
```
slowli: This looks incorrect / dangerous at worst, or misleading at best. We should either explicitly check that `Absent` vairant cannot occur anywhere but in the tree root, or rename it to something like `Hash` / `HashedNode` and treat it correspondingly.

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -311,3 +298,58 @@ where
         Ok(res)
     }
 }
+
+impl<'a, V> Deserialize<'a> for ListProof<V>
```
slowli: Having to write such crutches is one of the reasons I'd prefer non-recursive `ListProof` definition. I've written [an implementation](https://github.com/slowli/exonum/tree/list-proof-this-time-its-personal) (although it's severely outdated), and I'll draft the corresponding internal RFC soon.

components/merkledb/src/proof_list_index/mod.rs
dmitry-timofeev: If I understand correctly, it won't require anything atop of a proof for the present elements (as it already includes the length, which affects the list state hash, and allows to infer that the other elements are absent).

pavel-mukhanov: I thought about creating a separate task to modify range proofs of absence to accommodate existing elements. But is it relevant in the case of new flat proofs proposal?

slowli: IMO, it's an orthogonal to the internal proof organization, so you can still create the corresponding task 😉

aleksuss:
Closed due to #1450

PR #1417, opened 2019-08-13
Fix cryptocurrency test for Windows

ivan-ochc:


ivan-ochc:
> LGTM as of [726dc60](https://github.com/exonum/exonum/commit/726dc6025bd4e01cbed1ac375f9cfd3d8bc344b4). It seems that in order for the script to return non-zero status on error, you should add explicit `exit 1;` at the end:
> 
> ```powershell
> try {
>   Main;
> } finally {
>   kill $server.id;
>   exit 1;
> }
> ```

As far as I understand **Exit status 1** is already returned in case of an error. I've checked it with `echo $?`

aleksuss:
@slowli If I understand correct the `Finally` block runs every time, regardless of whether there was an error or not. Probably you have meant this: 
```powershell
Try {
    Main;
} Catch {
    exit 1;
} Finally {
   kill $server.id;
}
```

slowli:
> As far as I understand Exit status 1 is already returned by test in case of an error. I've checked it with echo $?

Sorry for misleading you. I've run the script in Ubuntu 16.04 (i.e., the environment used in Travis CI) and it *seemed* like the script returns 0 code even on error, but I've re-checked now, and this is not the case.

PR #1427, opened 2019-08-16
Try to re deploy artifact before registration.

alekseysidorov:
In some cases runtime forgets about deployed but not registered artifacts, for example after restart between artifact confirmation and registration.

To fix this issue we have to add `is_artifact_deployed` to runtime trait and mark artifact deploy error during the registration as fatal.

### DIFF
--- a/exonum/src/blockchain/tests.rs
+++ b/exonum/src/blockchain/tests.rs
@@ -17,10 +17,12 @@ use exonum_merkledb::{
 };
 use futures::{sync::mpsc, Future};
 
-use std::{collections::BTreeMap, sync::Mutex};
+use std::{collections::BTreeMap, panic, sync::Mutex};
 
 use crate::{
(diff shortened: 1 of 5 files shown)


exonum/src/blockchain/tests.rs
```diff
@@ -653,19 +655,19 @@ fn test_dispatcher_register_rollback() {
...
-    );
-    let snapshot = blockchain.snapshot();
-    assert!(!dispatcher::Schema::new(snapshot.as_ref())
-        .artifacts()
-        .contains(&artifact_id.name));
-    assert_eq!(
-        Entry::<_, u64>::new(IDX_NAME, snapshot.as_ref()).get(),
-        None
-    );
+    let error_string = panic::catch_unwind(panic::AssertUnwindSafe(|| {
```
ozkriff: Maybe this statement be broken into a few lines with the help of intermediate variable for better readability?

PR #1429, opened 2019-08-19
Review unwraps in dynamic services [ECR-3419]

alekseysidorov:


exonum/src/runtime/rust/service.rs
```diff
@@ -116,10 +116,12 @@ impl<'a, 'b> TransactionContext<'a, 'b> {
         self.runtime_context.fork
     }
 
+    // Method for backward compatibility with the existing code.
```
aleksuss: Have you missed one slash ? Comment-doc should be here, right ?

PR #1430, opened 2019-08-19
Expand documentation on configuration parameters usage [ECR-3463]

alekseysidorov:


exonum/src/runtime/mod.rs
```diff
@@ -186,6 +186,10 @@ pub trait Runtime: Send + Debug + 'static {
     fn start_service(&mut self, spec: &InstanceSpec) -> Result<(), ExecutionError>;
 
     /// Configure a service instance with the given parameters.
+    /// 
+    /// The configuration parameters passed to the method are not saved anywhere,
```
```suggestion
    /// The configuration parameters which passed to the method are not saved anywhere, 
```

vitvakatu: Not sure that it is valid, it must be `parameters which are passed`. The current variant `parameters passed` is also valid though, iirc. @Maria-Nosyk ?

exonum/src/runtime/mod.rs
```diff
@@ -187,6 +187,10 @@ pub trait Runtime: Send + Debug + 'static {
 
     /// Configure a service instance with the given parameters.
     ///
+    /// The configuration parameters passed to the method are not saved anywhere,
+    /// so the service instance should save them by itself if it is important for
```
dmitry-timofeev: It seems important that this method is only called once (or once per configuration change, when that is implemented) to understand the need to make some changes to persistent storage in the implementation of this method (not necessarily save them as is, but update the state based on them), therefore, I'd consider moving this paragraph _after_ the subsequent.

dmitry-timofeev: On the other hand — these are internal docs, so any order will do here, but in `Service` docs it must be very clear.

exonum/src/runtime/mod.rs
Maria-Nosyk: From the grammar point of view @vitvakatu  is tight. However, it is recommended to avoid negative verb forms, epecially such a large ones. In the present case I would suggest something like 
```suggestion
    /// The configuration parameters passed to the method are deleted immediately. 
```

alekseysidorov: "deleted" is not proper verb in this context, I suggest to use "forgot", "discarded" or something like that.

Maria-Nosyk: "discarded" is a nice one!

PR #1448, opened 2019-09-03
Use keys derived from master key

pavel-mukhanov:
Attempt to use specific identity-keys(x25519) for Noise-handshake, without converting them from ed25519 consensus keys.  All validator keys are derived from one master key. 

This a prototype and it is not intended for merge.  There are some workarounds and todo's in code that will be fixed in the future.


exonum/src/blockchain/config.rs
```diff
@@ -29,6 +29,7 @@ use std::collections::{BTreeMap, HashSet};
 use crate::crypto::{hash, CryptoHash, Hash, PublicKey};
 use crate::helpers::{Height, Milliseconds};
 use crate::messages::EMPTY_SIGNED_MESSAGE_SIZE;
+use exonum_crypto::PublicKeyKx;
```
alekseysidorov: Why not something like following?

```rust
use exonum::crypto::kx;

...

kx::PublicKey;
```

exonum/src/node/mod.rs
```diff
@@ -850,6 +837,8 @@ pub struct ConnectInfo {
     pub address: String,
     /// Peer public key.
     pub public_key: PublicKey,
+    /// Peer identity key.
+    pub identity_key: Option<PublicKeyKx>,
```
slowli: Why is `identity_key` optional? What are the scenarios in which `identity_key == None`?

components/crypto/src/utils.rs
```diff
@@ -138,6 +141,121 @@ impl EncryptedKeys {
...
+    pub service_pk: PublicKey,
+    pub service_sk: SecretKey,
+    pub identity_pk: PublicKeyKx,
+    pub identity_sk: SecretKeyKx,
+}
+
+pub fn save_master_key<P: AsRef<Path>, W: AsRef<[u8]>>(
+    path: P,
+    pass_phrase: W,
+    key: Salt,
```
slowli: Why does `key` have this particular type? Semantically, `key` is not a salt for `pwhash` operation; it is rather a seed for the `SecretTree` (i.e., essentially an input to Blake2b-based KDF [as used in libsodium](https://download.libsodium.org/doc/key_derivation#deriving-keys-from-a-single-high-entropy-key)). Maybe, it could make sense to make a type alias or a newtype for `[u8; secret_tree::SEED_LEN]` in the `secret_tree` crate?

pavel-mukhanov: Salt is a wrapper around libsodium's `randombytes_buf` which is also used to generate master key via `crypto_kdf_keygen`. But I agree that newtype will be better.

components/crypto/src/utils.rs
```diff
@@ -138,6 +122,124 @@ impl EncryptedKeys {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
+pub struct Keys {
```
popzxc: Well, that's more like preferences question, but I think it would be better to have something like:

```rust
struct KeyPair {
  pub pk: PublicKey,
  pub sk: SecretKey,
}

pub struct Keys {
  pub consensus: KeyPair,
  pub service: KeyPair,
  pub identity: KeyPair,
}
```

components/crypto/src/utils.rs
```diff
@@ -138,6 +122,124 @@ impl EncryptedKeys {
...
+    }
+}
+
+pub fn generate_keys<P: AsRef<Path>>(path: P, passphrase: &[u8]) -> Keys {
+    let tree = SecretTree::new(&mut thread_rng());
+    save_master_key(path, passphrase, tree.seed()).expect("Error generating master key.");
+    generate_keys_from_master_password(tree)
+}
+
+fn generate_keys_from_master_password(tree: SecretTree) -> Keys {
```
aleksuss: Why don't you return `Result<Keys, Error` here ? It would let you to get rid of unwraps.

components/crypto/Cargo.toml
```diff
@@ -16,15 +16,17 @@ byteorder = { version = "1.2.7", features = [ "i128" ] }
...
 rust_decimal = "1.0"
 serde = "1.0.10"
 serde_derive = "1.0.64"
 serde_json = "1.0.19"
+failure = "0.1.5"
 toml = "0.5.0"
 uuid = "0.7.1"
-exonum_sodiumoxide = { version = "0.0.22", optional = true }
+exonum_sodiumoxide = { git = "http://github.com/pavel-mukhanov/exonum_sodiumoxide/", branch = "key-exchange-mod", optional = true }
+secret-tree = { version = "0.2.0-pre.0", path = "../../../secret-tree/" }
```
aleksuss: Is it really a part of the repository ?

pavel-mukhanov: No, I after release of updated versions of `pwhash`, `secret-tree` and `exonum-sodiumoxide` I will change that.

components/crypto/src/utils.rs
slowli: To argue in favor of `KeyPair`. `KeyPair` may encode the invariant that the secret key and public key correspond to each other, which is beneficial at least from the safety perspective. (Also, sodium's secret keys actually contain corresponding public keys inside them with offset 32; while the offset should probably be regarded as an implementation detail, ability to derive a public key from the secret one [is not](https://download.libsodium.org/doc/public-key_cryptography/public-key_signatures#extracting-the-seed-and-the-public-key-from-the-secret-key).)

components/crypto/src/crypto_lib/sodiumoxide/kx.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+#[derive(Debug, Default, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct KeyPair {
+    pub(crate) public_key: PublicKey,
+    pub(crate) secret_key: SecretKey,
+}
+
+impl KeyPair {
+    pub fn from_keys(public_key: PublicKey, secret_key: SecretKey) -> Self {
```
alekseysidorov: I think that `From<(PublicKey, SecretKey>)` implementation can be very useful also.

components/crypto/src/utils.rs
```diff
@@ -73,128 +40,250 @@ fn validate_file_mode(mode: u32) -> Result<(), Error> {
...
+            public_key,
+            secret_key,
+        }
     }
+}
 
-    fn from_bytes(bytes: &[u8]) -> Result<PublicKey, String> {
-        PublicKey::from_slice(bytes)
-            .ok_or_else(|| "Couldn't create PublicKey from slice".to_owned())
+/// Struct containing all validator key pairs.
```
alekseysidorov: It seems that this structure should be placed in the exonum crate.

exonum/Cargo.toml
```diff
@@ -56,7 +56,7 @@ ctrlc = "3.1.1"
 rpassword = "4.0.1"
 zeroize = "0.10.0"
 
-exonum_sodiumoxide = { version = "0.0.23", optional = true }
+exonum_sodiumoxide = {version = "0.0.23", optional = true}
```
aleksuss: An old version is more preferable I guess.

exonum/src/events/noise/tests.rs
```diff
@@ -72,7 +72,7 @@ fn noise_converted_keys_handshake() {
     use crate::crypto::{gen_keypair, x25519::into_x25519_keypair};
 
     const MSG_SIZE: usize = 4096;
-    static PATTERN: &'static str = "Noise_XK_25519_ChaChaPoly_SHA256";
+    static PATTERN: &str = "Noise_XK_25519_ChaChaPoly_SHA256";
```
aleksuss:
```suggestion
    const PATTERN: &str = "Noise_XK_25519_ChaChaPoly_SHA256";
```

components/crypto/src/utils.rs
```diff
@@ -73,128 +40,250 @@ fn validate_file_mode(mode: u32) -> Result<(), Error> {
...
 }
 
+/// Creates a TOML file that contains encrypted master and returns `Keys` derived from it.
+pub fn generate_keys<P: AsRef<Path>>(path: P, passphrase: &[u8]) -> Keys {
+    let tree = SecretTree::new(&mut thread_rng());
+    save_master_key(path, passphrase, tree.seed()).expect("Error generating master key.");
+    generate_keys_from_master_password(tree).expect("Error deriving keys from master key.")
+}
+
+fn generate_keys_from_master_password(tree: SecretTree) -> Option<Keys> {
```
aleksuss: This method is related to `exonum` framework rather then `exonum-crypto`.

pavel-mukhanov:
Closed in favor of #1459

PR #1451, opened 2019-09-05
Add the service interface name option to the proto files

popzxc:
This PR adds an 'option' entry to the proto files which contains service interface name.

It is supposed that 'service.proto' contains transactions for exactly one interface, and if there are more than one interface is implemented for a service, transactions will be splitted into several files.

This is needed for client-side code (e.g. for the python light client) to make easier calls to the interfaces.


popzxc:
Due to [details](https://github.com/exonum/exonum/pull/1446#issuecomment-528318617) discovered in the Service interface MVP pull request, this PR is turned out to be useless.

PR #1453, opened 2019-09-05
Database create checkpoint

blackbeam:
## Overview

~~This PR:~~
* ~~exposes [RocksDb checkpoints][1] via the `Database` trait method;~~
* ~~reimplements internal `CheckpointDb` structure of the `exonum-testkit` crate using this new API.~~

New version of this PR switches paradigm a bit:
* [RocksDb checkpoints][1] functionality now exposed via `RocksDB::create_checkpoint` method.
* `dyn Database` trait object can now be downcasted to a concrete implementation.

---

See: https://wiki.bf.local/pages/viewpage.action?pageId=31306417

### Definition of Done

[1]: https://github.com/facebook/rocksdb/wiki/Checkpoints

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `create_checkpoint` method is implemented for the `RocksDB` struct. This method uses
   [RocksDb checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints) functionality under the hood.
+
 - `Database` trait object can now be downcasted to a concrete implementation.
 
 ## 0.12.0 - 2019-08-14
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ## Unreleased
 
+- `create_checkpoint` method is implemented for the `RocksDB` struct. This method uses
+  [RocksDb checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints) functionality under the hood.
+- `Database` trait object can now be downcasted to a concrete implementation.
+
 ## 0.12.0 - 2019-08-14
 
 ### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `create_checkpoint` method is implemented for the `RocksDB` struct. This method uses
   [RocksDb checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints) functionality under the hood.
 
-- `Database` trait object can now be downcasted to a concrete implementation.
+- `Database` trait object can now be casted to a concrete implementation.
 
 ## 0.12.0 - 2019-08-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,6 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   [RocksDB checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints)
   functionality under the hood.
 
-- `Database` trait object can now be casted to a concrete implementation.
-
 ## 0.12.0 - 2019-08-14
 
 ### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ## Unreleased
 
-- `create_checkpoint` method is implemented for the `RocksDB` struct. This method uses
-  [RocksDb checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints) functionality under the hood.
+- `create_checkpoint` method has been implemented for the `RocksDB` struct.
+  This method uses
+  [RocksDB checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints)
+  functionality under the hood.
 
 - `Database` trait object can now be casted to a concrete implementation.
 
--- a/components/merkledb/src/backends/rocksdb.rs
+++ b/components/merkledb/src/backends/rocksdb.rs
@@ -20,6 +20,7 @@ pub use rocksdb::{BlockBasedOptions as RocksBlockOptions, WriteOptions as RocksD
 
 use std::{fmt, iter::Peekable, mem, path::Path, sync::Arc};
 
+use rocksdb::checkpoint::Checkpoint;
 use rocksdb::{self, ColumnFamily, DBIterator, Options as RocksDbOptions, WriteBatch};
 
 use crate::{
@@ -85,6 +86,18 @@ impl RocksDB {
         };
         check_database(&mut db)?;
         Ok(db)
+    }
+
+    /// Creates checkpoint of this database in the given directory. (see [RocksDb docs][1] for
+    /// details).
+    ///
+    /// Successfully created checkpoint can be opened using `RocksDB::open`.
+    ///
+    /// [1]: https://github.com/facebook/rocksdb/wiki/Checkpoints
+    pub fn create_checkpoint<T: AsRef<Path>>(&self, path: T) -> crate::Result<()> {
+        let checkpoint = Checkpoint::new(&*self.db)?;
+        checkpoint.create_checkpoint(path)?;
+        Ok(())
     }
 
     fn do_merge(&self, patch: Patch, w_opts: &RocksDBWriteOptions) -> crate::Result<()> {
--- a/components/merkledb/src/db.rs
+++ b/components/merkledb/src/db.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use std::{
-    any::Any,
     cell::RefCell,
     cmp::Ordering::{Equal, Greater, Less},
     collections::{
@@ -383,24 +382,6 @@ enum NextIterValue {
     Finished,
 }
 
-/// Helper trait to get `&self` as `&dyn Any`.
-pub trait AsAny: Any {
-    fn as_any(&self) -> &dyn Any;
-    fn as_any_mut(&mut self) -> &mut dyn Any;
-}
-
-impl<T: Any> AsAny for T {
-    /// Returns `&self` as `&dyn Any`.
-    fn as_any(&self) -> &dyn Any {
-        self
-    }
-
-    /// Returns `&mut self` as `&mut dyn Any`.
-    fn as_any_mut(&mut self) -> &mut dyn Any {
-        self
-    }
-}
-
 /// Low-level storage backend implementing a collection of named key-value stores
 /// (aka column families).
 ///
@@ -434,7 +415,7 @@ impl<T: Any> AsAny for T {
 /// [`fork`]: #method.fork
 /// [`merge`]: #tymethod.merge
 /// [interior-mut]: https://doc.rust-lang.org/book/second-edition/ch15-05-interior-mutability.html
-pub trait Database: AsAny + Send + Sync + 'static {
+pub trait Database: Send + Sync + 'static {
     /// Creates a new snapshot of the database from its current state.
     fn snapshot(&self) -> Box<dyn Snapshot>;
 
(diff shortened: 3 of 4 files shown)


components/merkledb/src/backends/rocksdb.rs
```diff
@@ -152,6 +153,17 @@ impl Database for RocksDB {
         w_opts.set_sync(true);
         self.do_merge(patch, &w_opts)
     }
+
+    fn create_checkpoint(&self, path: &Path) -> crate::Result<Box<dyn Database>> {
```
aleksuss:
```suggestion
    fn create_checkpoint<P: AsRef<Path>>(&self, path: P) -> crate::Result<Box<dyn Database>> {
```

components/merkledb/src/backends/rocksdb.rs
```diff
@@ -152,6 +153,17 @@ impl Database for RocksDB {
...
+        let checkpoint = Checkpoint::new(&*self.db)?;
+        checkpoint.create_checkpoint(path)?;
+        let options = DbOptions {
```
vitvakatu: Why the options are not preserved? I can understand `create_if_missing: false`, but can't understand why other options are default.

components/merkledb/src/db.rs
```diff
@@ -453,6 +454,15 @@ pub trait Database: Send + Sync + 'static {
     /// will be returned. In case of an error, the method guarantees no changes are applied to
     /// the database.
     fn merge_sync(&self, patch: Patch) -> Result<()>;
+
+    /// Creates checkpoint of this database in the given directory.
```
vitvakatu: Please add a link to the corresponding RocksDB wiki page. 
Also a couple of questions which needs to be mentioned in docs:
1) What will happen to the checkpointed snapshot if the original database file will be deleted.
2) RocksDB docs mention two modes: Read and Read-Write. Which one is used here?

components/merkledb/src/tests.rs
```diff
@@ -32,3 +33,67 @@ fn should_compile() {
...
     let _: ValueSetIndex<_, ()> = ValueSetIndex::new_in_family("", "", &fork);
 }
+
+#[test]
+fn checkpoints() {
+    fn test_checkpoint(db: Box<dyn Database>) {
+        let temp_dir = TempDir::new().unwrap();
+        let root_hashes = {
+            let fork = db.fork();
+            let mut root_hashes = Vec::new();
```
vitvakatu: What the purpose of 10 root_hashes if they are all the same?

components/merkledb/src/tests.rs
```diff
@@ -32,3 +33,67 @@ fn should_compile() {
...
+                let mut index: ProofListIndex<_, Vec<u8>> = ProofListIndex::new(index_name, &fork);
+                assert_eq!(index.object_hash(), root_hashes[i]);
+                for j in 0..64 {
+                    index.push(vec![0_u8; j]);
+                }
+            }
+            db.merge_sync(fork.into_patch()).unwrap();
+        }
+
+        {
```
vitvakatu: Please add a short comments explaining what's going on here.

As I can see, the changes in parent database are visible in the checkpoint, and vice versa?

components/merkledb/src/tests.rs
```diff
@@ -32,3 +33,67 @@ fn should_compile() {
...
+            for i in 0..10 {
+                let index_name = format!("idx{}", i);
+                let index: ProofListIndex<_, Vec<u8>> = ProofListIndex::new(index_name, &fork);
+                assert_eq!(index.len(), 256);
+            }
+        }
+    }
+
+    let temp_dir = TempDir::new().unwrap();
+    let rocks_db = RocksDB::open(temp_dir.path().join("db"), &DbOptions::default()).unwrap();
```
vitvakatu: Just wondering why we need to `join("db")` here, as the temp_dir is different and empty anyway.

testkit/src/checkpoint_db.rs
```diff
@@ -128,66 +151,28 @@ impl<T: Database> CheckpointDbInner<T> {
...
-        Ok(())
+        self.db.merge(patch)
     }
 
     fn checkpoint(&mut self) {
-        self.backup_stack.push(Vec::new())
+        let name: String = Alphanumeric.sample_iter(thread_rng()).take(32).collect();
+        let checkpoint = self
+            .db
+            .create_checkpoint(&*self.temp_dir.path().join(name))
```
vitvakatu: Why not create fresh temp_dir for each checkpoint?

components/merkledb/src/tests.rs
vitvakatu: If so, please also mention this in the docs

components/merkledb/src/backends/rocksdb.rs
blackbeam: This will, unfortunately, break object safety for the `Database` trait.

components/merkledb/src/backends/rocksdb.rs
blackbeam: This will, unfortunately, break object safety for the `Database` trait.

components/merkledb/src/tests.rs
blackbeam: RocksDb will error in case of existing checkpoint directory. I believe to not to accidentally overwrite something if API is misused.

components/merkledb/src/backends/rocksdb.rs
alekseysidorov: And even for `Where: Self`?

components/merkledb/src/db.rs
```diff
@@ -453,6 +454,15 @@ pub trait Database: Send + Sync + 'static {
...
+    ///
+    /// Returns new database instance created from checkpoint.
```
alekseysidorov: I think you should specify that snapshot is persistent.

testkit/src/checkpoint_db.rs
blackbeam: To store one `TempDir` instead of one per checkpoint.

vitvakatu: TempDir is just a `Option<PathBuf>`, according to the [source code](https://docs.rs/tempfile/3.1.0/src/tempfile/dir.rs.html#194-196) of the tempfile. Here you're implementing part of the `tempfile` functionality yourself by involving additional dependency.

components/merkledb/src/tests.rs
```diff
@@ -32,3 +33,67 @@ fn should_compile() {
     let _: SparseListIndex<_, ()> = SparseListIndex::new_in_family("", "", &fork);
     let _: ValueSetIndex<_, ()> = ValueSetIndex::new_in_family("", "", &fork);
 }
+
+#[test]
+fn checkpoints() {
```
slowli: I wonder if it's possible to reliably generate enough data for the DB in order for it to dump some of it into SST files. As far as I understand, checkpoints hard-link SST files and copy the newer logs, so it could be interesting to check that these both mechanisms work as expected.

testkit/src/checkpoint_db.rs
```diff
@@ -259,65 +222,10 @@ mod tests {
         assert_eq!(stack_len(&db), 1);
     }
 
-    #[test]
-    fn test_backup() {
```
slowli: Why is this test removed?

components/merkledb/src/db.rs
blackbeam: > Please add a link to the corresponding RocksDB wiki page.

Should i? This will tie `Database` to `RocksDb` too much.

1, 2. I'll clarify that checkpoint is writable and independent.

vitvakatu: So then I'd explain what `checkpoint` is, cause it may be surprising for the users who do not know about RocksDB implementation details.

components/merkledb/src/tests.rs
vitvakatu: 🤔 But you can't overwrite anything because RocksDB will error in case of the same directory. The code is okay, it was just a little surprising to me before I spotted that there are actually two `temp_dir`s.

testkit/src/checkpoint_db.rs
```diff
@@ -12,38 +12,51 @@
...
+};
 
 /// Implementation of a `Database`, which allows to rollback its state
 /// to the last made checkpoint.
 ///
 /// **Note:** Intended for testing purposes only. Probably inefficient.
 #[derive(Debug)]
-pub struct CheckpointDb<T> {
-    inner: Arc<RwLock<CheckpointDbInner<T>>>,
+pub struct CheckpointDb {
```
slowli: Are the changes in `CheckpointDb` caused by the desire to test checkpoint functionality? Or is there demand to run `TestKit` with the persistent database? In the first case, the changes look like an overkill to me (the old implementation is still functional, isn't it?).

components/merkledb/src/db.rs
slowli: Sorry for not asking previously: do we _really_ need this method in the `Database` trait, or would implementing it for a specific implementation (`RocksDB`) suffice?

components/merkledb/src/backends/rocksdb.rs
blackbeam: > And even for where: Self?

This will, somewhat, break the purpose because `create_checkpoint` will be unusable on trait object when database [is only available via trait object][1] most of the time.
Option i see is to extend [`Database` bounds][2] with `Any` to allow downcast but i think it's too much for this.

[1]: https://docs.rs/exonum/0.12.0/src/exonum/blockchain/mod.rs.html#82
[2]: https://docs.rs/exonum-merkledb/0.12.0/src/exonum_merkledb/db.rs.html#418

components/merkledb/src/tests.rs
blackbeam: > I wonder if it's possible to reliably generate enough data for the DB in order for it to dump some of it into SST files.

I believe closing and reopening a database should flush memtable into sst.

aleksuss: Potentially, the method `merge_sync()` should flush memtables into sst files as well.

components/merkledb/src/db.rs
blackbeam: > do we really need this method in the Database trait

Actually no. However it was hard to find a clean way to downcast `dyn Database` to `&RocksDb`. I'll need to add additional dependency on [`downcast-rs`][1] or to reimplement the same functionality.

[1]: https://crates.io/crates/downcast-rs

components/merkledb/src/backends/rocksdb.rs
blackbeam: Addressed in new version.

components/merkledb/src/backends/rocksdb.rs
blackbeam: Not applicable to a new version.

components/merkledb/src/db.rs
blackbeam: Please look into the new version of this PR.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ## Unreleased
 
+- `create_checkpoint` method is implemented for the `RocksDB` struct. This method uses
+  [RocksDb checkpoints](https://github.com/facebook/rocksdb/wiki/Checkpoints) functionality under the hood.
+- `Database` trait object can now be downcasted to a concrete implementation.
```
aleksuss: Split entities with an empty line please

components/merkledb/src/db.rs
slowli: Can you describe in more detail how you plan to use `RocksDB` for backup? It seems strange to me that the current approach assumes we've just got a pointer to `dyn Database` and need to downcast it. Where in reality, the workflow may look something like this:

```rust
fn create_blockchain(db: Arc<dyn Database>) {
    // This emulates `Blockchain::new(..)`
}

fn backup(db: Arc<RocksDB>) {
    // ...
}

// We need to initialize a concrete DB *somewhere*.
let db: Arc<RocksDB> = Arc::new(RocksDB::new(..));
// Here, we upcast the `Arc` pointer to use dynamic dispatch.
create_blockchain(Arc::clone(&db) as Arc<dyn Database>);
// But we still know that the original pointer points to `RocksDB`!  
backup(db);
```

blackbeam: > Can you describe in more detail how you plan to use RocksDB for backup?

Functionality will be exposed as a Blockchain method and used in service hooks and API handlers.

aleksuss:
@blackbeam fix the lints pls 
`/home/travis/build/exonum/exonum/CHANGELOG.md:11:38 - Unknown word (downcasted)`

PR #1455, opened 2019-09-06
Batch operations for `ProofListIndex` (and other improvements)

slowli:
## Overview

This PR implements more efficient extension of `ProofListIndex` (~5-10x faster compared to the current implementation on the crate benchmarks), and also implements `truncate()` and `pop()` methods.

Also, there are some tangentially related changes:

- Implementations of `IndexAccess` were reworked. I've wanted to implement `IndexAccess` for `Rc<Fork>` (in order to be able to get a mutable index with `'static` lifetime), and it all snowballed from here.
- Using `Rc<Fork>`, the prop-tests for indexes were rewritten almost without macro magic.
- `HashTag::hash_list()` was extended to support values of any appropriate type, not only `Hash`.

Depends on #1450.

### Definition of Done


components/merkledb/src/proof_list_index/mod.rs
```diff
@@ -505,6 +488,49 @@ where
...
+    /// - Value hashes (i.e., tree branches on level 1) are assumed to be updated.
+    fn update_range(&mut self, mut first_index: u64, mut last_index: u64) {
+        let mut last_level_index = self.len() - 1;
+        for height in 1..self.height() {
+            // Check consistency of the index range.
+            debug_assert!(first_index <= last_index);
+            // Check consistency with the level length.
+            debug_assert!(last_index <= last_level_index);
+
+            let mut index = first_index & !1; // make the starting index even
```
popzxc: Is that bit magic really required? Maybe use something more readable like
```rust
let mut level_start_index = first_index;
let mut level_end_index = last_index;
...
// If the value with the odd index was updated, hash of the value before it should be used.
// So we decrement (if needed) start index to be odd.
level_start_index -= level_start_index % 2;
// Same for the end index: if an even value was updated, the next hash should also be used.
// So we increment (if needed) end index to be even.
level_end_index += if level_end_index % 2 == 0 { 1 } else { 0 };
// Check for boundaries
level_end_index = cmp::min(level_end_index, last_level_index);

File: components/merkledb/src/proof_list_index/mod.rs [hunk 2]
```diff
@@ -505,6 +488,49 @@ where
...
 
+    /// Updates levels of the tree with heights `2..` after the values in the range
+    /// `[first_index, last_index]` were updated.
+    ///
+    /// # Invariants
+    ///
+    /// - `self.len()` / `self.height()` is assumed to be correctly set.
+    /// - Value hashes (i.e., tree branches on level 1) are assumed to be updated.
+    fn update_range(&mut self, mut first_index: u64, mut last_index: u64) {
+        let mut last_level_index = self.len() - 1;
```
popzxc: I'd name it `last_index_of_level` to avoid ambiguity (now it can be considered to be an index of a last level).

components/merkledb/src/proof_list_index/mod.rs
```diff
@@ -560,9 +569,24 @@ where
     where
         I: IntoIterator<Item = V>,
     {
+        let first_index = self.len();
+        let mut last_index = first_index;
```
popzxc: I think name like `new_list_len` (and `old_list_len` in the line above) will be more clear.

components/merkledb/src/proof_list_index/mod.rs
```diff
@@ -595,21 +619,110 @@ where
...
+        let old_height = self.height();
+        self.set_len(new_length);
+        let mut last_index = Some(new_length - 1);
+
+        // Remove values.
+        for index in new_length..=old_last_index {
+            self.base.remove(&ProofListKey::leaf(index));
+        }
+
+        let mut started_updating_hashes = false;
```
popzxc: Maybe create a function `remove_range` similar to `update_range`?

components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -184,67 +170,58 @@ fn merge(
...
-fn hash_layer(layer: &mut Vec<HashedEntry>, last_index: u64) -> Result<(), ListProofError> {
-    let new_len = (layer.len() + 1) / 2;
-    for i in 0..new_len {
-        let x = &layer[2 * i];
-        layer[i] = if let Some(y) = layer.get(2 * i + 1) {
-            // To be able to zip two hashes on the layer, they need to be adjacent to each other,
-            // and the first of them needs to have an even index.
-            if !x.key.is_left() || y.key.index() != x.key.index() + 1 {
-                return Err(ListProofError::MissingHash);
+fn hash_layer(layer: &[HashedEntry], last_index: u64) -> Result<Vec<HashedEntry>, ListProofError> {
```
popzxc: Seems that merging of `dynamic_services` into this branch was not quite successful.

components/merkledb/src/views/tests.rs
```diff
@@ -681,6 +679,53 @@ fn mutable_and_immutable_prefixed_borrows() {
     _mutable_and_immutable_borrows(&TemporaryDB::new(), PREFIXED_IDX);
 }
 
+#[test]
+fn views_based_on_rc_fork() {
```
vitvakatu: What are we checking in this test? Does it make sense to split it in a smaller pieces?

components/merkledb/src/proof_list_index/mod.rs
vitvakatu: I'd suggest even something like `last_element_index`. It's the same thing, isn't it? (I'm literally blowing my brain now)

components/merkledb/src/proof_list_index/mod.rs
```diff
@@ -459,6 +454,49 @@ where
...
+                index += 2;
+            }
+
+            if stop_index % 2 == 0 {
+                let key = ProofListKey::new(height, stop_index);
+                let branch_hash = HashTag::hash_single_node(&self.get_branch_unchecked(key));
+                self.base.put(&key.parent(), branch_hash);
+            }
+
+            first_index >>= 1;
```
vitvakatu: Why not `first_index /= 2`?

components/merkledb/src/proof_list_index/mod.rs
popzxc: `last_element_index` may be considered to be the index of the last element in the list, which is only true for the first iteration of the loop below.

aleksuss:
LGTM, but update the `CHANGELOG` please.

components/merkledb/benches/storage.rs
```diff
@@ -232,6 +232,27 @@ fn proof_list_append(b: &mut Bencher, len: usize) {
     );
 }
 
+fn proof_list_extend(b: &mut Bencher, len: usize) {
```
alekseysidorov: May you attach benchmark results?

slowli: Where to? Here are some results from my laptop:

### New implementation

```text
storage/proof_list/extend/items/1000
        time:   [3.2060 ms 3.2233 ms 3.2460 ms]
        thrpt:  [308.07 Kelem/s 310.24 Kelem/s 311.92 Kelem/s]
storage/proof_list/extend/items/10000
        time:   [35.507 ms 35.960 ms 36.368 ms]
        thrpt:  [274.97 Kelem/s 278.09 Kelem/s 281.64 Kelem/s]
storage/proof_list/extend/items/100000
        time:   [379.63 ms 381.62 ms 384.91 ms]
        thrpt:  [259.80 Kelem/s 262.04 Kelem/s 263.42 Kelem/s]
```

### Old implementation

```text
storage/proof_list/extend/items/1000
        time:   [14.093 ms 14.502 ms 14.857 ms]
        thrpt:  [67.308 Kelem/s 68.955 Kelem/s 70.956 Kelem/s]
  change:
        time:   [+336.65% +354.39% +378.13%] (p = 0.00 < 0.05)
        thrpt:  [-79.085% -77.992% -77.099%]
storage/proof_list/extend/items/10000
        time:   [204.36 ms 206.86 ms 209.83 ms]
        thrpt:  [47.657 Kelem/s 48.341 Kelem/s 48.934 Kelem/s]
  change:
        time:   [+466.72% +481.71% +500.94%] (p = 0.00 < 0.05)
        thrpt:  [-83.359% -82.809% -82.354%]
storage/proof_list/extend/items/100000
        time:   [3.1680 s 3.3873 s 3.5454 s]
        thrpt:  [28.206 Kelem/s 29.522 Kelem/s 31.565 Kelem/s]
  change:
        time:   [+656.77% +716.51% +772.67%] (p = 0.00 < 0.05)
        thrpt:  [-88.541% -87.753% -86.786%]
```

components/merkledb/src/proof_list_index/mod.rs
```diff
@@ -548,21 +657,77 @@ where
...
+
+        // Remove values.
+        for index in new_length..=old_last_index {
+            self.base.remove(&ProofListKey::leaf(index));
+        }
+
+        self.remove_range(old_last_index, old_height);
+    }
+
+    /// Removes the last element from the list and returns it, or returns `None`
```
alekseysidorov: I suppose that data structures, such as the proof list index, are often used as append only structures.
So that do you think about creating append only wrapper for this structure with disallowed mutation operations.

popzxc: Why? If service author don't want to remove anything from the index, he just won't call those methods.

We should not protect users like "oh I accidently called `index.clear()`", we're not babysitters.

aleksuss: IMO I don't see a reason why to do so. It's a common method for DS with access to its elements by index. So it's ok as for me.

PR #1458, opened 2019-09-10
Implement Initialize interface for services [ECR-3508]

alekseysidorov:


exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,69 @@
+// Copyright 2019 The Exonum Team
```
vitvakatu: Is Initialize trait Rust-specific? If not, why is it placed inside `rust` module?

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,69 @@
...
+
+//! Important interservice communication interfaces.
+
+use crate::runtime::{dispatcher::Error as DispatcherError, MethodId};
+
+use super::{ExecutionError, Interface, TransactionContext};
+
+pub const INITIALIZE_METHOD_ID: MethodId = 0;
+
+/// This trait describes a service interface to pass initial configuration parameters into
```
vitvakatu: `This trait` is completely redundant. I suggest:
`Describes a service able to accept initial configuration parameters for the new service instance.` @Maria-Nosyk can help with better wording

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,69 @@
...
+/// the started service instance.
+pub trait Initialize {
+    /// Initialize a service instance with the given parameters.
+    ///
+    /// The configuration parameters passed to the method are discarded immediately.
```
vitvakatu: `... are not saved by the Dispatcher`. `discarded` means that they are not used in method.

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// So the service instance should save them by itself if it is important for
+    /// the service business logic.
+    ///
+    /// This method is called after creating a new service instance by the [`start_service`]
+    /// invocation. In this case if an error during this action occurs, the dispatcher will
+    /// invoke [`stop_service`].
+    ///
+    ///
```
vitvakatu: Too much of empty lines

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,69 @@
...
+    ///
+    /// ['start_service`]: ../../trait.Runtime.html#start_service
+    /// ['stop_service`]: ../../trait.Runtime.html#stop_service
+    fn initialize(&self, context: TransactionContext, params: &[u8]) -> Result<(), ExecutionError>;
```
vitvakatu: Why the method accepts `&[u8]`, given the fact that runtime probably wants to clone them and the Dispatcher won't use them after deserialization?

exonum/examples/sample_runtime.rs
```diff
@@ -156,15 +140,30 @@ impl Runtime for SampleRuntime {
...
-            "Executing method {} of service {}",
-            call_info.method_id, call_info.instance_id
+            "Executing method {}#{} of service {}",
+            context.interface_name, call_info.method_id, call_info.instance_id
         );
 
-        // Simple transaction executor.
-        match call_info.method_id {
+        const SERVICE_INTERFACE: &str = "";
+        match (context.interface_name, call_info.method_id) {
```
vitvakatu: Documentation of `execute` method is now not sufficient. Please expand on:
1. Requirements of handling Initialize interface by Runtimes.
2. Default interface name
3. Requirements on error handling in case of invalid runtime name passed.

exonum/src/runtime/rust/interfaces.rs
alekseysidorov: Yes this trait is Rust specific and you have to implement in Java same interface without binding to Rust.

exonum/src/runtime/rust/interfaces.rs
alekseysidorov: Because the `call` method passes arguments as` & [u8] `and then they are passed unchanged to the interface.

exonum/src/runtime/rust/interfaces.rs
alekseysidorov: @MakarovS Help please.

MakarovS: I believe you wanted to mention @Maria-Nosyk 
But I agree with @vitvakatu's suggestion btw

exonum/src/runtime/rust/interfaces.rs
vitvakatu: So than interface name and methods ids must be available in the `runtime` module, as well as the list of the interfaces which must be supported by the runtime implementation.

examples/timestamping/backend/src/lib.rs
```diff
@@ -52,28 +54,31 @@ use crate::{
 };
 
 #[derive(Debug, ServiceFactory)]
-#[exonum(proto_sources = "proto", implements("TimestampingInterface"))]
+#[exonum(
+    proto_sources = "proto",
+    implements("TimestampingInterface", "Initialize")
```
slowli: @alekseysidorov Interestingly, the following changes seem to work without any changes to proc macros:

```rust
pub trait Initialize {
    type Params: BinaryValue;
    fn initialize(&self, context: TransactionContext, params: Self::Params)
        -> Result<(), ExecutionError>;
}

impl<T: BinaryValue> Interface for dyn Initialize<Params = T> {
    const INTERFACE_NAME: &'static str = "Initialize";
    // probably, this constant needs to be exposed separately,
    // because call sites looks awkward otherwise.
    
    // snip
}

#[derive(Debug, ServiceFactory)]
#[exonum(
    // snip
    implements("TestService", "Initialize<Params = Init>")
)]
pub struct TestServiceImpl;

impl Initialize for TestDispatcherService {
    type Params = TestExecute;

    fn initialize(
        &self,
        _context: TransactionContext,
        params: TestExecute,
        // snip
    }
}
```

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -418,19 +423,54 @@ impl Dispatcher {
...
+        };
+        let args = constructor.into_bytes();
+
+        catch_panic(|| {
+            runtime
+                .execute(&context, &call_info, args.as_ref())
+                .or_else(|err| {
+                    // Default behavior for case if the service does not implement an `Initialize`
+                    // interface and constructor is empty.
+                    if constructor_is_empty && err.kind == Error::NoSuchInterface.into() {
```
vitvakatu: Why can't I initialize a service without any arguments?

exonum/src/runtime/mod.rs
```diff
@@ -221,11 +194,23 @@ pub trait Runtime: Send + Debug + 'static {
     /// * If panic occurs, the runtime must ensure that it is in a consistent state.
     fn stop_service(&mut self, descriptor: InstanceDescriptor) -> Result<(), ExecutionError>;
 
-    /// Execute service transaction.
+    /// Execute service instance method.
+    ///
+    /// # Important interfaces
+    ///
+    /// In addition to its own methods, a service can implement methods of additional interfaces.
+    /// You should provide in your own runtime following most important service interfaces:
```
```suggestion
    /// In your own runtime implementation, you must provide support of the following interfaces:
```

exonum/src/runtime/mod.rs
```diff
@@ -221,11 +194,23 @@ pub trait Runtime: Send + Debug + 'static {
     /// * If panic occurs, the runtime must ensure that it is in a consistent state.
     fn stop_service(&mut self, descriptor: InstanceDescriptor) -> Result<(), ExecutionError>;
 
-    /// Execute service transaction.
+    /// Execute service instance method.
```
vitvakatu: This line is redundant, isn't it?

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,94 @@
...
+/// [`Initialize::initialize`]: trait.Initialize.html#tymethod.initialize
+pub const INITIALIZE_METHOD_ID: MethodId = 0;
+/// Fully qualified name of the `Initialize` interface.
+pub const INITIALIZE_INTERFACE_NAME: &str = "Initialize";
+
+/// This trait describes a service interface to pass initial configuration parameters into
+/// the started service instance.
+///
+/// This interface is optional, therefore if initialize was called on a service
+/// that does not implement this interface, and then if the configuration parameters are empty,
```
vitvakatu: The initialization is successful if both statements are valid:
1. Service does not implement this interface
2. Configuration parameters are empty
? 
If so, why can't I initialize a service without any parameters?

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,94 @@
...
+    type Params: BinaryValue;
+
+    /// Initialize a service instance with the given parameters.
+    ///
+    /// The configuration parameters passed to the method are discarded immediately.
+    /// So the service instance should save them by itself if it is important for
+    /// the service business logic.
+    ///
+    /// This method is called after creating a new service instance by the [`start_service`]
+    /// invocation. In this case if an error during this action occurs, the dispatcher will
```
vitvakatu:
```suggestion
    /// invocation. In the case of an error occuring during this action, the dispatcher will
```

exonum/src/runtime/mod.rs
```diff
@@ -368,6 +345,15 @@ impl Caller {
             None
         }
     }
+
+    /// Try to reinterpret caller as blockchain.
+    pub fn as_blockchain(&self) -> Option<()> {
```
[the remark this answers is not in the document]
vitvakatu: Does it? The typical usage is `as_blockchain().is_none()` or `as_blockchain().is_some()`, which is longer and confusing. How such consistency helps with readability of the code?

exonum/src/runtime/dispatcher/mod.rs
alekseysidorov: You can do it if you implement `Initialize` interface

exonum/src/runtime/rust/interfaces.rs
alekseysidorov: But in fact it means that parameters exist only in RAM and after initialize dissapear from it.

exonum/src/runtime/mod.rs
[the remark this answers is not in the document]
alekseysidorov: `as_blockchain().ok_or(Error::MyFooError)?;`

exonum/src/runtime/mod.rs
```diff
@@ -194,11 +203,21 @@ pub trait Runtime: Send + Debug + 'static {
     /// * If panic occurs, the runtime must ensure that it is in a consistent state.
     fn stop_service(&mut self, descriptor: InstanceDescriptor) -> Result<(), ExecutionError>;
 
-    /// Execute service transaction.
+    /// Dispatch payload to the method of a specific service instance.
+    /// Service instance name and method ID are provided in the `call_info` argument and
+    /// interface name is provided as the corresponding field of the `context` argument.
```
dmitry-timofeev: Why isn't the interface name (required for dispatch) the part of the call info?

exonum/src/runtime/mod.rs
```diff
@@ -68,6 +68,15 @@
 //!
 //! 6. After execution the transaction [execution status] is written into the blockchain.
 //!
+//! # Service interfaces
+//!
+//! In addition to its own methods, a service can implement methods of additional interfaces.
+//! In your own runtime implementation, you must provide support of the following interfaces:
+//!
+//! ## Initialize
```
dmitry-timofeev: I remember from the conversation that the runtime was responsible to control the caller of some interfaces, requiring some special handling in the implementation, which negates any benefits from fewer number of methods. Is it still the case?

exonum/src/runtime/mod.rs
alekseysidorov: Because at the moment, "services interfaces" is private a feature and thus, service writers are limited in their ability to create their own interfaces.

Cast @popzxc to more explanations.

exonum/src/runtime/mod.rs
alekseysidorov: Each service instance should validate caller in each method regardless of interface. In this case there are no special cases for the `Initialize` interface.

exonum/src/runtime/mod.rs
vitvakatu: Okay, are there any plans on moving it into `CallInfo`? We need to know to plan changes in EJB

exonum/src/runtime/mod.rs
dmitry-timofeev: I don't see anywhere that any validations are required (it didn't use to be the case!)

exonum/src/runtime/mod.rs
dmitry-timofeev: I see, but the Runtime trait isn't for service developers, it is for Runtime developers.

alekseysidorov: > Okay, are there any plans on moving it into `CallInfo`? We need to know to plan changes in EJB

Yes, we have such plans.

exonum/src/runtime/mod.rs
alekseysidorov: For example you should verify that transaction author is authorized to transfer funds. In other words, anyway you have to write code like following.

```rust
let (author, tx_hash) = context.caller().as_transaction().ok_or(Error::WrongCaller)?;
if !is_authorized(context.fork(), &author) {
    return Err(Error::UnauthorizedOperation);
}
```

And even more safe way is write code like this:

```
let (author, fork) = context.verify_caller(|caller, snapshot| 
    caller.as_transaction().and_then(|(author, _)|
         if is_authorized(snapshot, author) { Some(author) }
         else { None }
    )
).ok_or(Error::SomeErrorType);
... logic
```

So i think you can write the same code in Java.

dmitry-timofeev: > you should verify that transaction author is authorized to transfer funds

That is a natural responsibility of the application code to authorize the operation. We discuss the built-in "interface", and I don't see why either the runtime or the appication code has to perform any authorization of an operation that is supposed to be performed by the framework. It does not make sense to authorize the operations performed by the framework — why can't the framework itself ensure that this operation is performed at the right time and by the authorized entity (= the framework!)?

popzxc: Why would framework suppose that some caller is authorized (or not) for some kind of runtime or user application?

If framework will perform `some` checks, it will result in checking the same thing twice (some checks by the framework, and some by end-side code) and create a field to shoot yourself in a leg (by expecting the framework to perform all checks for you).

IMHO letting the end-side code check everything is OK.

dmitry-timofeev: Because it is not an arbitrary operation, it is an operation that is suppossed to be performed by the **framework itself**. Isn't it ridiculous to have to check that your framework works as expected (or to enable misuse of the framework operations by implementing them in such a way)? Why don't we check then that 'deploy-artifact' is invoked by Exonum, and not some wonderful forces?

popzxc: I think that discussing in the abstract field won't lead to anything useful.

What's your concrete proposal?

dmitry-timofeev: If you look at the start, there is nothing abstract.
1. Use a normal method that does not require fancy documentation and implementation instead of half-baked interfaces, or
1. Make the framework responsible for controlling the caller.

In the very least, if it can't be fixed, then document what the implementation shall do with this interface!

popzxc: 1. That's not a proposal, just like "Do things the correct way and not how you're doing it now".
2. Ok, framework sees that `Caller::Transaction` attempts to call an user-defined interface in some runtime. How exactly should it check if that kind of caller is authorized?

dmitry-timofeev: Do you see anywhere the discussion on user-defined interfaces and if the core shall control the caller there? Me neither — [we discuss the built-in configuration interface that must be invoked by the framework only](https://github.com/exonum/exonum/pull/1458#discussion_r325171020), and the little sense it makes to place a burden of controlling the caller in this case on client code, be it the runtime or the service implementation.

aleksuss: Guys. Let's discuss this inside ?

exonum/src/runtime/rust/interfaces.rs
Maria-Nosyk: I agree that we can eliminate "This trait". As for the rest, I like the old variant more.

alekseysidorov:
This PR is closed in favor of https://github.com/exonum/exonum/pull/1467

PR #1459, opened 2019-09-10
Use keys derived from master key (ECR-3522)

pavel-mukhanov:
New PR pointing to `dynamic_services` instead of master.


cli/src/command/generate_config.rs
```diff
@@ -81,27 +81,15 @@ pub struct GenerateConfig {
...
-    pub consensus_key_pass: Option<PassInputMethod>,
-    /// Passphrase entry method for service key.
-    ///
-    /// Possible values are: `stdin`, `env{:ENV_VAR_NAME}`, `pass:PASSWORD`.
-    /// Default Value is `stdin`.
-    /// If `ENV_VAR_NAME` is not specified `$EXONUM_SERVICE_PASS` is used
-    /// by default.
-    #[structopt(long)]
-    pub service_key_pass: Option<PassInputMethod>,
+    pub master_key_pass: Option<PassInputMethod>,
```
vitvakatu: Documentation in cli/lib.rs needs updates.

CHANGELOG.md
```diff
@@ -71,6 +71,12 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   An additional field in the response of the endpoint was added. The field
   corresponds to the total number of transactions in the blockchain. (#1289)
 
+- Added new pair of keys to validator keys. This identity keys are now used
+  in noise handshake instead of converted consensus keys. (#1448)
+
+- All validator keys are now derived from single master key. Master key is
```
vitvakatu: Changelog must provide useful information to users, so please mention changes in node configuration process (i.e.  what parameters are chaged).

cli/src/command/generate_config.rs
```diff
@@ -139,31 +127,21 @@ impl ExonumCommand for GenerateConfig {
 
         let pub_config_path = self.output_dir.join(PUB_CONFIG_FILE_NAME);
         let private_config_path = self.output_dir.join(SEC_CONFIG_FILE_NAME);
-        let consensus_secret_key_path = self.output_dir.join(CONSENSUS_SECRET_KEY_NAME);
-        let service_secret_key_path = self.output_dir.join(SERVICE_SECRET_KEY_NAME);
+        let master_key_file_name = "master.key.toml";
```
aleksuss: It'd be good to use a constant instead the string literal here.

cli/src/command/generate_config.rs
```diff
@@ -194,20 +170,17 @@ impl ExonumCommand for GenerateConfig {
...
 
-fn create_secret_key_file(
-    secret_key_path: impl AsRef<Path>,
-    passphrase: impl AsRef<[u8]>,
-) -> Result<PublicKey, Error> {
+fn create_keys_and_files(secret_key_path: impl AsRef<Path>, passphrase: impl AsRef<[u8]>) -> Keys {
     let secret_key_path = secret_key_path.as_ref();
     if secret_key_path.exists() {
-        bail!(
+        panic!(
```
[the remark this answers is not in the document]
pavel-mukhanov: It's not intentional change, it's merge artifact.

cli/src/command/generate_config.rs
```diff
@@ -179,10 +157,8 @@ impl ExonumCommand for GenerateConfig {
         let private_config = NodePrivateConfig {
             listen_address,
             external_address: self.peer_address.to_string(),
-            consensus_public_key,
-            consensus_secret_key: CONSENSUS_SECRET_KEY_NAME.into(),
-            service_public_key,
-            service_secret_key: SERVICE_SECRET_KEY_NAME.into(),
+            master_key_path: master_key_file_name.into(),
```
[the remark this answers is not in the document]
pavel-mukhanov: Bad naming, it's actually file name, not path. I will rename it.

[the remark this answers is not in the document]
vitvakatu: Does it make sense to store name (which is always the same) in the config files? I suggest you to store full path to the file.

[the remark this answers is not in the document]
pavel-mukhanov: You are right, this is certainly path, but in case of `generate-config` command the path is relative, so we use `MASTER_KEY_FILE_NAME`. I think it's covers most of the cases, but you can later specify absolute path to master key manually if you need it.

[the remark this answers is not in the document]
vitvakatu: Sorry, I just don't see any use in such field if it's always equal to `MASTER_KEY_FILE_NAME`. I see two options here:
1. Remove the field completely, the master key will always be in `MASTER_KEY_FILE_NAME`, we clearly state this in documentation.
2. Store actual file location (preferably absolute path, but can be relative if the former one is not possible), possibly allow user to change it (with an optional configuration parameter), and return the path to the file in `StandardResult` so the user can expand the command later without any problems.

Right now the code shows a bit weird behaviour between these two options: we have a field, but it always contain constant value.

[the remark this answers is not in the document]
pavel-mukhanov: I choose second variant.

CHANGELOG.md
aleksuss: And one more thing. Why did you add the entities in the 0.12 release section ?

cli/src/command/generate_config.rs
```diff
@@ -33,17 +33,15 @@ use crate::{
...
 
-/// Name for a file containing consensus secret key.
-pub const CONSENSUS_SECRET_KEY_NAME: &str = "consensus.key.toml";
-/// Name for a file containing service secret key.
-pub const SERVICE_SECRET_KEY_NAME: &str = "service.key.toml";
 /// Name for a file containing the public part of the node configuration.
 pub const PUB_CONFIG_FILE_NAME: &str = "pub.toml";
 /// Name for a file containing the secret part of the node configuration.
 pub const SEC_CONFIG_FILE_NAME: &str = "sec.toml";
+/// Name for a file containing the secret part of the node configuration.
```
vitvakatu: Invalid doc string

cli/src/command/generate_config.rs
```diff
@@ -81,27 +79,18 @@ pub struct GenerateConfig {
...
     /// by default.
     #[structopt(long)]
-    pub consensus_key_pass: Option<PassInputMethod>,
-    /// Passphrase entry method for service key.
-    ///
-    /// Possible values are: `stdin`, `env{:ENV_VAR_NAME}`, `pass:PASSWORD`.
-    /// Default Value is `stdin`.
-    /// If `ENV_VAR_NAME` is not specified `$EXONUM_SERVICE_PASS` is used
-    /// by default.
+    pub master_key_pass: Option<PassInputMethod>,
```
vitvakatu: Invalid docs for field, it say "... for consensus key"

cli/src/command/generate_config.rs
```diff
@@ -137,33 +126,26 @@ impl ExonumCommand for GenerateConfig {
     fn execute(self) -> Result<StandardResult, Error> {
         let common_config: CommonConfigTemplate = load_config_file(self.common_config.clone())?;
 
-        let pub_config_path = self.output_dir.join(PUB_CONFIG_FILE_NAME);
-        let private_config_path = self.output_dir.join(SEC_CONFIG_FILE_NAME);
-        let consensus_secret_key_path = self.output_dir.join(CONSENSUS_SECRET_KEY_NAME);
-        let service_secret_key_path = self.output_dir.join(SERVICE_SECRET_KEY_NAME);
+        let public_config_path = self.output_dir.join(PUB_CONFIG_FILE_NAME);
+        let secret_config_path = self.output_dir.join(SEC_CONFIG_FILE_NAME);
```
vitvakatu: Thanks for proper renaming 👍

cli/src/command/generate_config.rs
```diff
@@ -137,33 +126,26 @@ impl ExonumCommand for GenerateConfig {
...
+        let master_key_path = self
+            .master_key_path
+            .unwrap_or_else(|| "".into())
```
vitvakatu: nit: PathBuf implements `Default`. Would `unwrap_or_default` work here?

cli/src/command/generate_config.rs
```diff
@@ -137,33 +126,26 @@ impl ExonumCommand for GenerateConfig {
...
-                self.no_password,
-                self.service_key_pass.unwrap_or_default(),
-                SecretKeyType::Service,
-            )?;
-            create_secret_key_file(&service_secret_key_path, passphrase.as_bytes())?
-        };
+        let keys = {
+            let passphrase =
+                Self::get_passphrase(self.no_password, self.master_key_pass.unwrap_or_default());
+            let master_key_path = self.output_dir.join(master_key_path.clone());
```
vitvakatu: What about moving this line up, to line 131? Then the private config and StandardResult will use correct path (relative to execute directory in case of relative `master_key_path`, absolute if `master_key_path` is absolute). Now you need to do `output_dir.join(private_config.master_key_path)` every time you want to get correct master key path.

Am I missing something? Some tests would be nice, just to make sure relative/absolute paths work as desired.

cli/src/command/generate_config.rs
```diff
@@ -137,33 +126,26 @@ impl ExonumCommand for GenerateConfig {
...
+            create_keys_and_files(&master_key_path, passphrase.unwrap().as_bytes())
```
vitvakatu: Please replace unwrap with `?`

cli/src/command/generate_config.rs
```diff
@@ -174,40 +156,39 @@ impl ExonumCommand for GenerateConfig {
...
     if secret_key_path.exists() {
         bail!(
             "Failed to create secret key file. File exists: {}",
             secret_key_path.to_string_lossy(),
-        );
+        )
     } else {
         if let Some(dir) = secret_key_path.parent() {
-            fs::create_dir_all(dir)?;
+            fs::create_dir_all(dir).unwrap();
```
vitvakatu: Please revert this line.

cli/src/password.rs
```diff
@@ -67,8 +65,7 @@ pub enum PassInputMethod {
     Terminal,
     /// Get passphrase from the environment variable with given name.
     ///
-    /// Default values are `EXONUM_CONSENSUS_PASS` and `EXONUM_SERVICE_PASS` for
-    /// consensus and service secret keys correspondingly.
+    /// Default value `EXONUM_MASTER_PASS` for master key.
```
vitvakatu:
```suggestion
    /// Default value is `EXONUM_MASTER_PASS`.
```

cli/src/command/generate_config.rs
```diff
@@ -128,18 +128,21 @@ impl ExonumCommand for GenerateConfig {
...
         let public_config_path = self.output_dir.join(PUB_CONFIG_FILE_NAME);
         let secret_config_path = self.output_dir.join(SEC_CONFIG_FILE_NAME);
-        let master_key_path = self
+        let mut master_key_path = self
             .master_key_path
-            .unwrap_or_else(|| "".into())
+            .unwrap_or_default()
             .join(MASTER_KEY_FILE_NAME);
 
+        if !master_key_path.is_absolute() {
```
vitvakatu: :+1: Just a nit: I didn't know, but it looks like `join` will properly work with absolute paths. [Docs](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.push) say that the path will be replaced by the argument if the latter one is absolute.

pavel-mukhanov:
@slowli This PR's introduces key derivation and some refactoring related to encrypted file creation. Maybe we should merge it anyway and implement key decoupling in the following task?

cli/src/command/generate_config.rs
pavel-mukhanov: Yes, from doc: `Pushing an absolute path replaces the existing path:`.

vitvakatu:
Does changelog need changes now, when identity keys have been removed?
Also, module-level docs in `cli/lib.rs` need updates about master key.

cli/src/lib.rs
```diff
@@ -26,8 +26,9 @@
 //!   Generated `.toml` file must be spread among all the nodes and must be used in the following
 //!   configuration step.
 //! 2. Generate public and secret (private) parts of the node configuration using `generate-config`
-//!   command. At this step, Exonum will generate two secret keys: consensus and service. Consensus
-//!   secret key is used for communications between the nodes, while service secret key is used
+//!   command. At this step, Exonum will generate master key from witch consensus and service
```
vitvakatu: typo: `witch`

components/crypto/src/lib.rs
```diff
@@ -483,10 +497,203 @@ implement_index_traits! {SecretKey}
...
+    }
+}
+
+impl CryptoHash for i8 {
+    fn hash(&self) -> Hash {
+        hash(&[*self as u8])
+    }
+}
+
+impl CryptoHash for i16 {
```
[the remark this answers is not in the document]
pavel-mukhanov: I created a macros, but then checked `dynamic services` branch and found that we have removed `CryptoHash` already. This is a merge artifact from master branch.

exonum/src/keys/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    save_master_key(path, passphrase, tree.seed())?;
+    generate_keys_from_master_password(tree)
+        .ok_or_else(|| format_err!("Error deriving keys from master key."))
+}
+
+fn generate_keys_from_master_password(tree: SecretTree) -> Option<Keys> {
+    let mut buffer = [0_u8; 32];
+
+    tree.child(Name::new("consensus")).fill(&mut buffer);
+    let seed = Seed::from_slice(&buffer)?;
```
slowli: This isn't directly related to this PR, but I'm not quite fond that unlike lower-level `sodiumoxide` types, `Seed`s in `exonum-crypto` cannot be initialized securely from the external source. I'm talking something like this:

```rust
let mut seed = Seed::new([0; SEED_LENGTH]);
tree.child(Name::new("consensus")).fill(&mut seed.0);
// We know that `Seed` is zeroed on drop, so at no point in time,
// seed bytes are contained in a non-secure container.
```

Alternatively, I could imagine `Seed` API like

```rust
impl Seed {
    pub fn generate<R: RngCore + CryptoRng>(rng: &mut R) -> Self { /* ... */ }
}
```

exonum/src/proto/schema/exonum/blockchain.proto
```diff
@@ -46,6 +46,8 @@ message ValidatorKeys {
   // Service key is used for services, for example, the configuration
   // updater service, the anchoring service, etc.
   exonum.PublicKey service_key = 2;
+  /// Identity key is used for secure communication between peers.
+  exonum.PublicKey identity_key = 3;
```
slowli: Is this a code artifact?

exonum/src/keys/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+            .seal(pass_phrase, key)
+            .map_err(|_| Error::new(ErrorKind::Other, "Couldn't create a pw box"))?;
+        let encrypted_key = eraser
+            .erase(&pwbox)
+            .map_err(|_| Error::new(ErrorKind::Other, "Couldn't convert a pw box"))?;
+
+        Ok(EncryptedMasterKey { key: encrypted_key })
+    }
+
+    fn decrypt(self, pass_phrase: impl AsRef<[u8]>) -> Result<secret_tree::Seed, Error> {
```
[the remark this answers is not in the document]
pavel-mukhanov: `SensitiveData` field and methods are private, i can't create instance of it. Is it possible to make them public, or it will break something?

popzxc:
I suggest to rewrite all the code in this PR several more times, just for fun.

b64

exonum/src/keys/mod.rs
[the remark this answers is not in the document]
slowli: You don't need access to `SensitiveData` internal structure, though, since it implements `Deref`:

```patch
@@ -14,7 +14,7 @@
 
 use crate::crypto::{gen_keypair_from_seed, PublicKey, SecretKey, Seed, SEED_LENGTH};
 use failure::format_err;
-use pwbox::{sodium::Sodium, ErasedPwBox, Eraser, Suite};
+use pwbox::{sodium::Sodium, ErasedPwBox, Eraser, SensitiveData, Suite};
 use rand::thread_rng;
 use secret_tree::{Name, SecretTree};
 
@@ -128,20 +128,16 @@
         Ok(EncryptedMasterKey { key: encrypted_key })
     }
 
-    fn decrypt(self, pass_phrase: impl AsRef<[u8]>) -> Result<secret_tree::Seed, Error> {
+    fn decrypt(self, pass_phrase: impl AsRef<[u8]>) -> Result<SensitiveData, Error> {
         let mut eraser = Eraser::new();
         eraser.add_suite::<Sodium>();
         let restored = eraser
             .restore(&self.key)
             .map_err(|_| Error::new(ErrorKind::Other, "Couldn't restore a secret key"))?;
         assert_eq!(restored.len(), SEED_LENGTH);
-        let seed_bytes = restored
+        restored
             .open(pass_phrase)
-            .map_err(|_| Error::new(ErrorKind::Other, "Couldn't open an encrypted key"))?;
-
-        let mut seed: [u8; 32] = [0; 32];
-        seed.copy_from_slice(&seed_bytes[..]);
-        Ok(seed)
+            .map_err(|_| Error::new(ErrorKind::Other, "Couldn't open an encrypted key"))
     }
 }
 
@@ -222,7 +218,7 @@
         let decrypted_seed = key
             .decrypt(pass_phrase)
             .expect("Couldn't decrypt master key ");
-        assert_eq!(seed, &decrypted_seed);
+        assert_eq!(&seed[..], &decrypted_seed[..]);
     }
 
     #[test]
@@ -249,7 +245,7 @@
         let seed = keys.decrypt(pass_phrase).expect("Couldn't decrypt key");
 
         assert_eq!(
-            hex::encode(&seed),
+            hex::encode(&*seed),
             "a05a82575d5f9d1f9469df31896f5b3c14ec4d18b3948cd7c8b09a7eed48b4e0"
         );
     }
```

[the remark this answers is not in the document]
pavel-mukhanov: Got it! I missed that RestoredPwBox::open already returns Result<SensitiveData, ..>

PR #1463, opened 2019-09-12
Fix check incoming message length

aleksuss:
## Overview

Set a payload limit equal to consensus configuration `max_message_len` value in the transaction handler.

---

### Definition of Done


exonum/src/api/node/public/explorer.rs
```diff
@@ -246,24 +250,51 @@ impl ExplorerApi {
...
+                .map_err(|_| format_err!("Couldn't deserialize transaction message."))?;
+            let _ = state
+                .sender()
+                .broadcast_transaction(signed)
+                .map_err(ApiError::from);
+            Ok(TransactionResponse { tx_hash })
+        };
+
+        let index = move |request: HttpRequest| {
+            let state = request.state().clone();
```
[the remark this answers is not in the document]
slowli: Well, it would allow to send transactions without 2x overhead; I wouldn't call that redundant 🙃 We probably should experiment a little, but it seems that sending binary data is quite feasible at least in the case of the JS client [via `fetch`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch):

```javascript
const body = new Uint8Array(32); // serialize message to binary...
fetch('/api/explorer/v1/transactions', {
    method: 'POST',
    headers: { 'Content-Type': 'application/octet-stream' },
    body,
})
```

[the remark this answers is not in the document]
aleksuss: What about to leave this PR as a hot fix. In a PR with fresh `actix` we will add a possibility to send transactions as a binary data. Ok ?

exonum/tests/websocket/main.rs
```diff
@@ -427,3 +427,53 @@ fn test_node_shutdown_with_active_ws_client_should_not_wait_for_timeout() {
...
+    let mut client =
+        create_ws_client("ws://localhost:8087/api/explorer/v1/ws").expect("Cannot connect to node");
+    client
+        .stream_ref()
+        .set_read_timeout(Some(Duration::from_secs(10)))
+        .unwrap();
+
+    // Send transaction.
+    let (pk, sk) = gen_keypair();
+    let name = String::from_utf8(vec![64; max_message_len / 2]).unwrap();
```
slowli: Please use the length here (or as a separate successful case) so that

```rust
assert_eq!(tx.signed_message().raw().len(), max_message_len);
```

I suppose that the best way to find the correct value is by using binary search, but it's possible to hard-code it (the value is 371).

Also, it's possible to use `"A".repeat(x)` instead of `String::from_utf8`.

exonum/tests/websocket/main.rs
```diff
@@ -427,3 +427,53 @@ fn test_node_shutdown_with_active_ws_client_should_not_wait_for_timeout() {
...
+    let (pk, sk) = gen_keypair();
+    let name = String::from_utf8(vec![64; max_message_len]).unwrap();
+    let tx = Message::sign_transaction(CreateWallet::new(&pk, name.as_str()), SERVICE_ID, pk, &sk);
+    let tx_json =
+        serde_json::to_string(&json!({ "type": "transaction", "payload": { "tx_body": tx }}))
+            .unwrap();
+    client.send_message(&OwnedMessage::Text(tx_json)).unwrap();
+
+    // Check response on set message when the message is bigger than 512 bytes.
+    assert!(recv_text_msg(&mut client).is_none());
```
slowli: Curiously, WS code does not return an error if we set the maximum accepted frame size too high; it still returns the transaction hash despite transaction not being broadcast. As far as I can tell, the underlying problem is that `impl Handler<Transaction> for Server` does not propagate an error from `ApiSender::broadcast_transaction`, and the latter does not check the message size.

exonum/tests/websocket/main.rs
```diff
@@ -427,3 +427,53 @@ fn test_node_shutdown_with_active_ws_client_should_not_wait_for_timeout() {
...
+    assert_eq!(
+        serde_json::from_str::<serde_json::Value>(&resp_text).unwrap(),
+        json!({
+            "result": "success",
+            "response": { "tx_hash": tx_hash }
+        })
+    );
+
+    let (pk, sk) = gen_keypair();
+    let name = String::from_utf8(vec![64; max_message_len]).unwrap();
```
slowli: Please modify or add a case when the `tx` length is exactly `max_message_len + 1`.

exonum/src/api/node/public/explorer.rs
```diff
@@ -307,6 +306,7 @@ impl ExplorerApi {
     ) where
         Q: Fn(&HttpRequest) -> Result<SubscriptionType, ActixError> + Send + Sync + 'static,
     {
+        let max_message_len = calculate_message_len(&service_api_state);
```
slowli: This is too strict; wrapping the transaction into `{ "type": "transaction", "payload": { "tx_body": tx } }` incurs additional overhead of >=45 bytes compared to the `calculated_message_len(..)`. We cannot specify an overhead with some leeway though (which I would otherwise recommend - a client implementation may insert some whitespace into JSON) because of the previously mentioned problem with WS processing; it swallows errors *and* does not check message size.

exonum/src/api/node/public/explorer.rs
```diff
@@ -297,6 +280,22 @@ impl ExplorerApi {
         });
     }
 
+    fn tx_handler(
+        state: &ServiceApiState,
+        query: TransactionHex,
+    ) -> Result<TransactionResponse, ApiError> {
+        let buf: Vec<u8> = ::hex::decode(query.tx_body).map_err(into_failure)?;
```
slowli: This code is very similar to `impl Handler<Transaction> for Server` in the `websocket` module. Is it possible to deduplicate it?

exonum/src/api/node/public/explorer.rs
```diff
@@ -297,6 +280,22 @@ impl ExplorerApi {
...
+        let signed = SignedMessage::from_raw_buffer(buf)?;
+        let tx_hash = signed.hash();
+        let signed = RawTransaction::try_from(Message::deserialize(signed)?)
+            .map_err(|_| format_err!("Couldn't deserialize transaction message."))?;
+        let _ = state
+            .sender()
+            .broadcast_transaction(signed)
+            .map_err(ApiError::from);
```
slowli: What's the point of mapping an error if it's swallowed by `let _ = `? Also, even if the error wasn't swallowed, `broadcast_transaction` doesn't seem to return an error on oversized transaction (which I'd expect it should), which makes it difficult to have any leeway in specifying maximum acceptable JSON size.

exonum/src/api/node/public/explorer.rs
```diff
@@ -419,3 +431,10 @@ impl<'a> From<explorer::BlockInfo<'a>> for BlockInfo {
         }
     }
 }
+
+fn calculate_message_len(api_state: &ServiceApiState) -> usize {
+    let snapshot = api_state.blockchain().snapshot();
+    let schema = Schema::new(&snapshot);
+    let max_message_len_in_bytes = schema.actual_configuration().consensus.max_message_len;
+    max_message_len_in_bytes as usize * 2 + 2
```
slowli: Nit: a comment explaining this formula should be appreciated.

exonum/src/api/node/public/explorer.rs
```diff
@@ -246,55 +252,112 @@ impl ExplorerApi {
...
+    pub(crate) fn tx_handler(
         state: &ServiceApiState,
         query: TransactionHex,
+        max_message_len: usize,
     ) -> Result<TransactionResponse, ApiError> {
-        use crate::events::error::into_failure;
-        use crate::messages::ProtocolMessage;
-
         let buf: Vec<u8> = ::hex::decode(query.tx_body).map_err(into_failure)?;
+        if buf.len() > max_message_len {
```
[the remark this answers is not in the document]
aleksuss: Here we check exact size of message in bytes rather than payload. Because payload can contain additional 64 bytes potentially. But yes, we can just compare `query.tx_body.len() / 2` and `max_message_len` here.

exonum/src/api/node/public/explorer.rs
```diff
@@ -246,55 +252,112 @@ impl ExplorerApi {
...
+        backend: &mut actix_backend::ApiBuilder,
+        service_api_state: ServiceApiState,
+    ) {
+        let (max_message_len, max_payload_len) = get_message_limits(&service_api_state);
+        let index = move |request: HttpRequest| {
+            let state = request.state().clone();
+            let content_length = request
+                .headers()
+                .get(CONTENT_LENGTH)
+                .unwrap()
```
popzxc: It's [possible](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4) to send a valid post request without Content-Length header. So it this `unwrap` safe (and correct) here?

exonum/tests/websocket/main.rs
```diff
@@ -427,3 +427,63 @@ fn test_node_shutdown_with_active_ws_client_should_not_wait_for_timeout() {
...
+    assert_eq!(
+        serde_json::from_str::<serde_json::Value>(&resp_text).unwrap(),
+        json!({
+            "result": "success",
+            "response": { "tx_hash": tx_hash }
+        })
+    );
+
+    let (pk, sk) = gen_keypair();
+    let name = "a".repeat(372);
```
popzxc: Maybe move magic numbers into variables with description (same for 371 above)?
Ideally I would expect those values to be calculated using `max_message_len`.

exonum/src/api/error.rs
```diff
@@ -63,3 +69,23 @@ impl From<failure::Error> for Error {
...
 }
+
+pub(crate) fn convert_error(
+    error: JsonPayloadError,
+    max_payload_len: usize,
+    content_len: String,
+) -> Error {
+    match error {
+        JsonPayloadError::Overflow => Error::PayloadTooLarge(format!(
+            "Allowed payload length is: {}, but try to send: {}",
```
popzxc: I would suggest `but it was attepted to send` or `but got` (same in one other place).

exonum/src/api/error.rs
```diff
@@ -63,3 +69,23 @@ impl From<failure::Error> for Error {
         Error::InternalError(e)
     }
 }
+
+pub(crate) fn convert_error(
```
popzxc: I'd name it `into_api_error`.

exonum/tests/websocket/main.rs
aleksuss: As for me it's obvious but I'll add a comment here.

exonum/src/api/backends/actix.rs
```diff
@@ -133,7 +133,13 @@ impl ResponseError for ApiError {
             ApiError::Storage(err) => HttpResponse::InternalServerError().body(err.to_string()),
             ApiError::NotFound(err) => HttpResponse::NotFound().body(err.to_string()),
             ApiError::Unauthorized => HttpResponse::Unauthorized().finish(),
-            ApiError::PayloadTooLarge(err) => HttpResponse::PayloadTooLarge().body(err.to_string()),
+            ApiError::PayloadTooLarge {
+                length_limit,
+                content_length,
+            } => HttpResponse::PayloadTooLarge().body(format!(
+                "Payload too large: Allowed {}, but got to send: {}",
```
slowli: Micro-nit: I'm not quite sure this is good English... I'd say something like "Payload too large: received $x-byte message, while the allowed message limit is $x bytes" (if necessary, replace "message" with "JSON").

PR #1467, opened 2019-09-18
Implement Configure interface [ECR-3306]

alekseysidorov:
At the moment this PS is based on the `Initialize` interface PR.


vitvakatu:
It looks like this PR contains changes from https://github.com/exonum/exonum/pull/1465

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,161 @@
+// Copyright 2019 The Exonum Team
```
popzxc: Creating one file for common interfaces is OK for now, but later it'll be better to have a separate folder with `mod.rs` just exporing each interface one by one.

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,161 @@
...
+use super::{ExecutionError, Interface, TransactionContext};
+
+/// Fully qualified name of the ['Configure`] interface.
+///
+/// ['Configure`]: trait.Configure.html
+pub const CONFIGURE_INTERFACE_NAME: &str = "Configure";
+/// Identifier of the [`Configure::verify_config`] method.
+///
+/// [`Configure::verify_config`]: trait.Configure.html#tymethod.verify_config
+pub const VERIFY_CONFIG_METHOD_ID: MethodId = 0;
```
popzxc: Can't method IDs be deduced automatically based on their order in the trait (like for service methods)?

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// Invoke the corresponding [method](trait.Configure.html#tymethod.apply_config)
+    /// of the interface.
+    pub fn apply_config(&self, params: impl BinaryValue) -> Result<(), ExecutionError> {
+        self.0
+            .call(CONFIGURE_INTERFACE_NAME, APPLY_CONFIG_METHOD_ID, params)
+    }
+}
+
+/// Verify that the caller of this method is supervisor service.
+pub fn caller_is_supervisor(caller: &Caller, _: &dyn Snapshot) -> Result<(), ExecutionError> {
```
popzxc: `caller_is_supervisor` is expected to return `bool` because of its name. Maybe `verify_caller_is_supervisor`?

exonum/src/runtime/rust/service.rs
```diff
@@ -170,7 +169,20 @@ impl<'a, 'b> TransactionContext<'a, 'b> {
...
-        CallContext::new(self.inner, self.instance.id, called)
+        CallContext::from_execution_context(self.inner, self.instance.id, called)
+    }
+
+    /// Check the caller of this method with the specified closure.
+    ///
+    /// If the closure returns `Ok(value)`, then the method returns `Ok((value, fork))` thus you
+    /// get a write access to the blockchain state. Otherwise this method returns
+    /// an occurred error.
+    pub fn check_caller<F, T, E>(&self, check: F) -> Result<(T, &Fork), E>
```
popzxc: Does it really have to be that complicated?..

exonum/src/runtime/rust/service.rs
```diff
@@ -183,22 +195,52 @@ pub struct BeforeCommitContext<'a> {
...
+    pub fn call_context(&self, called: InstanceId) -> CallContext<'a> {
+        CallContext::new(self.fork, self.dispatcher, self.instance.id, called)
+    }
+
+    /// Add a configuration update to pending actions. These changes will be applied immediately
+    /// before the block commit.
+    ///
+    /// Only the supervisor service is allowed to perform this action.
+    #[doc(hidden)]
+    pub fn update_config(&self, changes: Vec<ConfigChange>) {
```
popzxc: If only supervisor is allowed to perform this method, why make it a part of the interface for all services?

exonum/src/runtime/supervisor/mod.rs
```diff
@@ -124,6 +124,16 @@ impl Service for Supervisor {
 }
 
 impl Supervisor {
-    pub const BUILTIN_ID: InstanceId = 0;
+    pub const BUILTIN_ID: InstanceId = SUPERVISOR_SERVICE_ID;
```
popzxc: Supervisor asking for its service id from the runtime seems counter-intuitive for me, to be honest.

exonum/src/runtime/mod.rs
```diff
@@ -185,18 +205,15 @@ pub trait Runtime: Send + Debug + 'static {
...
     /// the service business logic.
     ///
-    /// This method is called in two cases:
-    ///
-    /// * After creating a new service instance by the [`start_service`] invocation. In this case,
-    /// if an error during this action occurs, the dispatcher will invoke [`stop_service`].
-    /// Make sure that this invocation will not fail.
-    /// * During the configuration change procedure. [ECR-3306]
+    /// This method is called after creating a new service instance by the [`start_service`].
+    /// In this case, if an error during this action occurs, the dispatcher will invoke
```
```suggestion
    /// If an error occurs during invocation of this method, the dispatcher invokes
```

exonum/src/runtime/mod.rs
```diff
@@ -185,18 +205,15 @@ pub trait Runtime: Send + Debug + 'static {
...
+    /// [`stop_service`]. Make sure that this invocation will not fail.
```
vitvakatu: What invocation? Of `initialize_service`?

exonum/src/runtime/rust/interfaces.rs
vitvakatu: Better to have them explicitly defined, as other runtimes do not have information about order of methods.

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,161 @@
...
+/// Identifier of the [`Configure::apply_config`] method.
+///
+/// [`Configure::apply_config`]: trait.Configure.html#tymethod.apply_config
+pub const APPLY_CONFIG_METHOD_ID: MethodId = 1;
+
+/// Describes a procedure for updating the configuration of a service instance.
+pub trait Configure {
+    /// The specific type of parameters passed during the service instance configuration.    
+    type Params: BinaryValue;
+    /// Verify a new configuration parameters before before their actual application.
```
vitvakatu:
```suggestion
    /// Verifies a new configuration parameters before their actual application.
```

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,161 @@
...
+    ///
+    /// This method is called by the new configuration change proposal. If the proposed
+    /// parameters do not fit for this service instance, it should return a corresponding
+    /// error to discard this proposal. Thus only a configuration change proposal in which all
+    /// changes are correct can be applied later.
+    ///
+    /// The proposal approval process details, and even the configuration proposal format, depends
+    /// on the particular implementation.
```
vitvakatu:
```suggestion
    /// on the particular runtime implementation.
```

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,161 @@
...
+    ///
+    /// [identifier]: ../../constant.SUPERVISOR_SERVICE_ID.html
+    fn verify_config(
+        &self,
+        context: TransactionContext,
+        params: Self::Params,
+    ) -> Result<(), ExecutionError>;
+    /// Update service configuration with the given parameters.
+    ///
+    /// The configuration parameters passed to the method are discarded immediately.
```
vitvakatu: Again, `discarded` means "get rid of (someone or something) as no longer useful or desirable". This is obviously not the case. I suggest `The configuration parameters passed to this method are not saved by the dispatcher, hence the service implementation must save them if needed.`

cc @Maria-Nosyk

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// So the service instance should save them by itself if it is important for
+    /// the service business logic.
+    ///
+    /// This method is called then some external conditions occur and thus this happens
```
vitvakatu: I suggest the following:
`This method can be triggered at any time and does not follow the general transaction execution workflow, so the errors returned might be ignored.`

exonum/src/runtime/rust/interfaces.rs
```diff
@@ -0,0 +1,161 @@
...
+    caller
+        .as_service()
+        .and_then(|instance_id| {
+            if instance_id == SUPERVISOR_SERVICE_ID {
+                Some(())
+            } else {
+                None
+            }
+        })
+        .ok_or_else(|| {
```
vitvakatu: Why combinators are used here instead of straightforward `if`?

exonum/src/runtime/rust/interfaces.rs
popzxc: Then I think it'll be better to have a `Enum` or mapping "method_name" => "method_id" returned by service via `service.get_methods()` instead of constants.

vitvakatu: I don't like if the only way to access method IDs would be using some methods (because it has significant impact on performance in EJB), but I don't mind agains a public enum + optional methods for Rust

popzxc: Well, even if methods will be the only mechanism for retrieving such info, it's not the kind of data that you need to refresh often. If that's expensive, you always can call it only once per service/interface instance and cache it.

However, I don't like the idea of designing things in favor of some concrete target implementation. See dependency inversion principle to understand why it's bad.

vitvakatu: Yep, I do not suggest design runtimes in a way that is better for EJB. I just say that methods will involve additional complexity comparing to simple constants. Dealing with such complexity is possible (we know how to cache data), but not free in terms of time, code size and runtime performance.

Actually, do you need mapping "name => id"? Names are not used by user, so some kind of C-style enum would be more suitable here, wouldn't it?

popzxc: Yeah, I don't mind your suggestion, it was just a remark to be kept in mind.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -320,9 +330,19 @@ impl Dispatcher {
...
-    pub(crate) fn before_commit(&self, fork: &mut Fork) {
+    pub(crate) fn before_commit(&mut self, fork: &mut Fork) {
+        let dispatcher_ref = DispatcherRef::new(self);
         for runtime in self.runtimes.values() {
-            runtime.before_commit(self, fork);
+            runtime.before_commit(&dispatcher_ref, fork);
+        }
+        // Execute pending dispatcher actions.
+        for action in dispatcher_ref.take_actions() {
+            let _ = action.execute(self, fork).map_err(|e| {
```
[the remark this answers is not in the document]
alekseysidorov: I established a corresponding task in Jire to review panics and rollbacks in the dispatcher.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -415,25 +425,68 @@ impl Dispatcher {
...
+
+    /// Perform a configuration update with the specified changes.
+    fn update_config(
+        &self,
+        fork: &Fork,
+        caller_instance_id: InstanceId,
+        changes: Vec<ConfigChange>,
+    ) -> Result<(), ExecutionError> {
+        changes.into_iter().try_for_each(|change| match change {
+            ConfigChange::Consensus(config) => {
```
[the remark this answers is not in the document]
alekseysidorov: This is a deep architectural flaw and we can't resolve it by this PR.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -479,6 +537,43 @@ impl std::fmt::Debug for DeployArtifactRequest {
     }
 }
 
+/// Reference to the underlying runtime dispatcher.
+#[derive(Debug)]
+pub struct DispatcherRef<'a> {
+    /// List of dispatcher actions that will be performed after execution finishes.
+    actions: RefCell<Vec<Action>>,
```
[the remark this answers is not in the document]
alekseysidorov: I think that DispatcherRef is unsound name for this entity. But now I can't invent perfect name for it.

exonum/src/runtime/rust/interfaces.rs
alekseysidorov: It can be deduced by the `exonum-derive` or certain codegen tool which whould be generate trait from the protobuf file spec. But it can't be deduced from the regular Rust code.

exonum/src/runtime/rust/service.rs
alekseysidorov: Maybe no, this is just only an experiment in the public API.
But I believe that we must protect service writers from the easy footshots. If you know how I can simplify this code just say.

exonum/src/runtime/rust/service.rs
alekseysidorov: Because I can't distinguish at runtime supervisor from the other services.

exonum/src/runtime/mod.rs
alekseysidorov: Yes.

exonum/src/runtime/rust/service.rs
popzxc: `F` can be `Fn(&Caller) -> bool`, so we get rid of the `T` and `E` and also return just `&Fork` instead of `Result` with tuple.
Also `check` can be renamed to `predicate`, because context is already telling you that `check_caller` performs some kind of check, and `predicate` will be indicating that it's a function.

alekseysidorov: I think that information like caller instance ID or transaction author is very useful as the result of check.

exonum/src/runtime/rust/interfaces.rs
alekseysidorov: I remember that we decided to use infinitive form in such descriptions.

cc @Maria-Nosyk

exonum/src/runtime/mod.rs
vitvakatu: Then it does not make any sense, as in the previous sentence you say that errors are allowed and will be correctly handled by dispatcher.

exonum/src/runtime/rust/service.rs
popzxc: Well, yes, but by introducing such a functionality, the function scope goes far from just `check_caller`. Nothing prevents user from wtiting code like

```rust
if let Some(fork) = context.check_caller(some_predicate) {
    let caller_info = extract_data_from_caller(context.caller());
}
```

IMHO It will be even more straightforward than
```
let (fork, caller_info) = context.check_caller(some_function)?; // Why `check_caller` returns some info?
```

alekseysidorov: In this case extraction if information performs twice.

exonum/src/runtime/mod.rs
alekseysidorov: This is just a recomendation.

vitvakatu: Then the phrase must be `It is recommended to avoid errors returned from the method.`

exonum/src/runtime/rust/interfaces.rs
vitvakatu: There was no such decision, but at least you must remove `before` duplication

exonum/src/runtime/mod.rs
```diff
@@ -205,11 +226,11 @@ pub trait Runtime: Send + Debug + 'static {
...
     /// ['start_service`]: #start_service
     /// ['stop_service`]: #stop_service
-    fn configure_service(
+    fn initialize_service(
         &self,
         fork: &Fork,
-        descriptor: InstanceDescriptor,
-        parameters: Any,
+        instance: InstanceDescriptor,
+        parameters: Vec<u8>,
```
vitvakatu: Is format of `parameters` runtime-specific?

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -415,25 +439,78 @@ impl Dispatcher {
...
+                    let context = CallContext::new(
+                        fork,
+                        &dispatcher_ref,
+                        caller_instance_id,
+                        config.instance_id,
+                    );
+                    ConfigureCall::from(context).apply_config(config.params)
+                });
+
+                match configure_result {
```
vitvakatu: Why the call of `Configure::apply_config` method is different from any other service method? Here you have an entirely duplicated piece of code from `execute`, don't you?

popzxc:
> It is not clear why the configuration of the consensus and configurations of the user services are tightly coupled in a single `ConfigChange` enum, involving code duplication in Dispatcher and tight coupling of the whole configuration module (not Rust "module", but a piece of functionality).

Agree. See discussion at [#discussion_r326216727](https://github.com/exonum/exonum/pull/1467#discussion_r326216727)

I think it will be one of the first things to be changed, but somewhat later.

exonum/src/runtime/dispatcher/mod.rs
alekseysidorov: Yes, but execute method is only for transactions and, secondly, behavior on errors is different.

exonum/src/runtime/mod.rs
alekseysidorov: Format of parameters is a service specific, but I suppose that I may write that parameters should be a correct protobuf message.

PR #1470, opened 2019-09-23
Minor derive keys fixes.

pavel-mukhanov:


cli/src/command/generate_config.rs
```diff
@@ -131,6 +131,8 @@ impl ExonumCommand for GenerateConfig {
         let master_key_path = self
             .master_key_path
             .unwrap_or_default()
+            .canonicalize()
+            .unwrap_or_default()
```
vitvakatu: Shouldn't we return an error in case `canonicalize` fails?

PR #1472, opened 2019-09-24
Fix some of the testkit ignored doctests

ozkriff:
See: https://jira.bf.local/browse/ECR-3275

### DIFF
--- a/testkit/src/lib.rs
+++ b/testkit/src/lib.rs
@@ -16,75 +16,57 @@
 //! and in the same process as the testkit.
 //!
 //! # Example
-//! ```ignore [ECR-3275]
-//! extern crate exonum;
-//! #[macro_use]
-//! extern crate exonum_derive;
-//! #[macro_use]
-//! extern crate exonum_testkit;
-//! extern crate serde_json;
-//! #[macro_use] extern crate serde_derive;
-//! extern crate failure;
-//!
-//! use serde_json::Value;
-//! use std::borrow::Cow;
-//!
-//! use exonum::api::node::public::explorer::{BlocksQuery, BlocksRange, TransactionQuery};
-//! use exonum::blockchain::{Block, Schema, Service, Transaction, TransactionContext,
-//! TransactionSet, ExecutionResult};
-//! use exonum::crypto::{gen_keypair, Hash, PublicKey, SecretKey, CryptoHash};
-//! use exonum::explorer::TransactionInfo;
-//! use exonum::helpers::Height;
-//! use exonum::messages::{Signed, AnyTx, Message};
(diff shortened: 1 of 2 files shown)


testkit/src/lib.rs
```diff
@@ -16,75 +16,57 @@
...
 //! impl Service for TimestampingService {
-//!     fn service_name(&self) -> &str {
-//!         "timestamping"
-//!     }
+//!     fn state_hash(&self, _: InstanceDescriptor, _: &dyn Snapshot) -> Vec<Hash> { vec![] }
+//! }
 //!
-//!     fn state_hash(&self, _: &Snapshot) -> Vec<Hash> {
-//!         Vec::new()
+//! impl From<TimestampingService> for InstanceCollection {
```
popzxc: I'm not sure if this is logically correct to implement such a conversion, especially in the example.
And another thought: since we expect that user will test already implemeted services, maybe hide service implementation from docs (if we'll rename service to ExampleService and tx to ExampleTx, it will be self-explanatory enough IMO)?

testkit/src/lib.rs
```diff
@@ -897,27 +886,29 @@ impl TestKit {
 ///         AfterCommitService { counter: Arc::new(AtomicUsize::default()) }
 ///     }
 ///
+/// #    pub fn new_instance(&self) -> Box<dyn Service> {
```
popzxc: Is this method used anywhere?

ozkriff: yep, it's used by the `ServiceFactory` macro: `service_constructor = "Self::new_instance",`

testkit/src/lib.rs
ozkriff: Removed these From impls, renamed My* to Example*.

> hide service implementation from docs

They seem to be already hidden (at least in the doc-tests that this PR touches), except for the top-level crate example that is supposed (as I see it) to be a minimal but complete example.

testkit/src/lib.rs
```diff
@@ -16,106 +16,86 @@
...
-//!         let tx = TimestampingTransactions::tx_from_raw(raw)?;
-//!         Ok(tx.into())
+//! impl TimestampingInterface for TimestampingService {
+//!     fn timestamp(&self, _: TransactionContext, arg: TxTimestamp) -> Result<(), ExecutionError> {
+//!         Ok(())
 //!     }
 //! }
 //!
 //! fn main() {
 //!     // Create testkit for network with four validators.
```
popzxc: Maybe add a note like "...and add a builtin `timestamping` service with ID=1"?

ozkriff: Done

PR #1473, opened 2019-09-24
Extract supervisor service from core

popzxc:
...and return where it belongs to.

Quick summary:

- `exonum-supervisor` is now a separate crate in the `services` directory.
- `blockchain` does not insert `exonum-supervisor` into services list.
- ...and `exonum-cli` does.
- `testkit` is adapted according to the changes above.
- `supervisor` tests from `testkit` are moved to the `exonum-supervisor` crate.


services/supervisor/src/errors.rs
```diff
@@ -14,7 +14,7 @@
 
 /// Common errors emitted by transactions during execution.
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, IntoExecutionError)]
-#[exonum(crate = "crate")]
+#[exonum()]
```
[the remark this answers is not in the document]
popzxc: I've decided to think about this macro like about some kind of magic that does *something* `¯\_(ツ)_/¯`

Here it ineed seems to be unneccessary.

services/supervisor/src/transactions.rs
```diff
@@ -25,7 +25,7 @@ use crate::{
 use super::{DeployConfirmation, DeployRequest, Error, Schema, StartService, Supervisor};
 
 /// Supervisor service transactions.
-#[exonum_service(crate = "crate")]
+#[exonum_service()]
 pub trait Transactions {
```
vitvakatu: nit: Can we name it `SupervisorInterface`, as `cryptocurrency-advanced` example does?

testkit/src/lib.rs
```diff
@@ -242,6 +243,19 @@ impl TestKit {
             .create()
     }
 
+    /// Returns the list of the Instances enabled by default.
+    ///
+    /// Currently only `Supervisor` service is included.
+    fn default_service_instances() -> Vec<InstanceCollection> {
```
alekseysidorov: In my PR with Configure interface I refuse of default service instances in Testkit. I think that you should completely remove the supervisor dependency in the testkit.

popzxc: Currently it's not really possible.
Let's wait until your PR will be merged then.

cli/src/lib.rs
```diff
@@ -115,7 +116,7 @@ pub struct NodeBuilder {
 impl NodeBuilder {
     /// Creates new builder.
     pub fn new() -> Self {
-        Self::default()
+        Self::default().with_service(Supervisor)
```
alekseysidorov: Are you sure that Supervisor will start with the default instance?

services/supervisor/src/lib.rs
```diff
@@ -29,19 +34,20 @@ use crate::{
...
 mod transactions;
 
 #[derive(Debug, ServiceFactory)]
 #[exonum(
-    crate = "crate",
-    proto_sources = "proto::schema",
+    proto_sources = "proto",
     artifact_name = "exonum-supervisor",
-    implements("transactions::Transactions")
+    implements("transactions::SupervisorInterface")
```
alekseysidorov: Why not always use `Transactions` interface name for the service own transactions?

popzxc: [Try to agree between yourselves](https://github.com/exonum/exonum/pull/1473#discussion_r327947649)

Personally I don't care much about how it's called.

cli/src/lib.rs
popzxc: Good catch.

services/supervisor/src/lib.rs
vitvakatu: It's `<service-name>Interface` in every example. So let's be consistent in either way

PR #1476, opened 2019-09-26
Improve support for additional runtimes in TestKit [ECR-3444]

skletsun:
## Overview

Rust runtime is considered builtin (and always included into the testing blockchain) and any other runtime - additional. Additional runtimes should be created via factories. Also, factories re-create runtimes through testkit's restart.

---

See: https://jira.bf.local/browse/ECR-3444

### Definition of Done


testkit/tests/runtime.rs
```diff
@@ -0,0 +1,226 @@
...
+        assert_eq!(state.deploy_spec, deploy_spec);
+    }
+
+    fn assert_config_params_passed(&self, config_params: Any) {
+        let state = self.state.borrow();
+        assert_eq!(state.config_params, config_params)
+    }
+}
+
+impl Default for RuntimeTester {
```
vitvakatu: nit: Why do you need manual `Default` implementation? `derive` would work

testkit/tests/runtime.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+}
+
+// We assert that:
+//  1) TestRuntime was passed to the testing blockchain
+//  2) Artifact was deployed with correct deploy specification
+//  3) Service was instantiated with correct initialization parameters
+#[test]
+fn test_runtime_factory() {
+    let tester = Arc::new(RuntimeTester::default());
```
vitvakatu: Will `Rc` work?

testkit/src/builder.rs
```diff
@@ -111,6 +111,8 @@ pub struct TestKitBuilder {
     validator_count: Option<u16>,
     service_instances: Vec<InstanceCollection>,
     logger: bool,
+    runtime_factories: Vec<Box<dyn RuntimeFactory>>,
+    instances: Vec<InstanceConfig>,
```
dmitry-timofeev: What's the difference between service_instances (also holding InstanceConfig as a field) and instances? Shall `instances` include all the things stored inside `service_instances`?

testkit/src/builder.rs
```diff
@@ -156,6 +162,17 @@ impl TestKitBuilder {
         self
     }
 
+    /// Adds a runtime factory to the testkit.
+    pub fn with_runtime_factory(
+        mut self,
+        runtime_factory: Box<dyn RuntimeFactory>,
+        instances: impl IntoIterator<Item = InstanceConfig>,
```
vitvakatu: Why we pass instances to method `with_runtime_factory`? Can we modify `with_service` method instead?

testkit/src/builder.rs
vitvakatu: So where is the difference between `service_instances` and `instances`? Why such duplication is needed?

testkit/tests/runtime.rs
skletsun: `derive` doesn't work because of `ArtifactId` that doesn't have `Default` implementation.

testkit/tests/runtime.rs
skletsun: No, `Runtime` should be `Send`, `RuntimeFactory` - `Send` + `Sync`

exonum/src/blockchain/builder.rs
```diff
@@ -171,13 +185,34 @@ impl BlockchainBuilder {
     }
 }
 
+/// DTO for service instance configuration.
```
alekseysidorov: It seems to me that the abbreviation does not look good here.

testkit/tests/runtime.rs
```diff
@@ -0,0 +1,226 @@
...
+            deploy_spec: Default::default(),
+            config_params: Default::default(),
+        };
+        Self {
+            state: RefCell::new(state),
+        }
+    }
+}
+
+// It's safe because in fact we don't leave the single thread scope during testing.
```
alekseysidorov: It looks like a shotgun for footshots.

testkit/tests/runtime.rs
```diff
@@ -0,0 +1,226 @@
...
+struct RuntimeState {
+    deployed_artifact: ArtifactId,
+    deploy_spec: Any,
+    config_params: Any,
+}
+
+// Main purpose is to track and make some assertions on state of runtime.
+#[derive(Debug)]
+struct RuntimeTester {
+    state: RefCell<RuntimeState>,
```
alekseysidorov: I you use `RwLock` here you can avoid unsafe impl.

testkit/tests/runtime.rs
alekseysidorov: So, maybe you should create more meaningfull artifact by default.

testkit/src/builder.rs
alekseysidorov: Even more, I suppose that the Rust instances should also use `RuntimeFactory` in common way like other Runtimes.

testkit/tests/runtime.rs
vitvakatu: I really don't think there should be a default implementation of `ArtifactId`, so this code is okay (though a bit verbose)

exonum/src/blockchain/builder.rs
vitvakatu: Perhaps the reason is that Rust does not have DTOs? I can suggest `Service instance instantiation parameters`

testkit/src/builder.rs
skletsun: In short - because of support of testkit restart. As soon as we remove StoppedTestkit - it will simplify the whole schema.

testkit/src/builder.rs
skletsun: Well, they were different methods initially, then I decided to combine them because every runtime has its own set of instances and this approach prevents a user from forgetting passing instances in addition to runtime factory.

exonum/src/blockchain/builder.rs
skletsun: Reworded

testkit/tests/runtime.rs
skletsun: Well yes, even if it is testing code and it's safe it's better to rewrite it 🙂

testkit/tests/runtime.rs
skletsun: Done.

testkit/src/builder.rs
vitvakatu: How the user can forget to pass services, if this is the main feature of TestKit - to create a blockchain with specified services list?

exonum/src/blockchain/builder.rs
```diff
@@ -171,13 +185,34 @@ impl BlockchainBuilder {
...
+    pub instance_spec: InstanceSpec,
+    /// Artifact deploy specification.
+    pub artifact_spec: Option<Any>,
+    /// Service configuration parameters.
+    pub constructor: Any,
+}
+
+impl InstanceConfig {
+    pub fn new(instance_spec: InstanceSpec, artifact_spec: Option<Any>, constructor: Any) -> Self {
+        InstanceConfig {
```
aleksuss:
```suggestion
        Self {
```
?

testkit/src/builder.rs
```diff
@@ -185,7 +215,25 @@ impl TestKitBuilder {
             .test_network
             .unwrap_or_else(|| TestNetwork::with_our_role(our_validator_id, 1));
         let genesis = network.genesis_config();
-        TestKit::assemble(TemporaryDB::new(), self.service_instances, network, genesis)
+
+        let mut runtimes = vec![];
+        if !self.rust_services.is_empty() {
```
vitvakatu: If I understand correctly, it's impossible to use any runtime without Rust one because Supervisor is absolute need. Why it is possible to create a testkit without Rust runtime enabled then?

skletsun: According to our discussion going to bring back the functionality of always having default Rust runtime.

testkit/src/builder.rs
```diff
@@ -173,6 +191,18 @@ impl TestKitBuilder {
         self
     }
 
+    /// Adds a runtime to the testkit. Must not be used with the Rust runtime as argument in
+    /// conjunction with the `.with_service()` method.
+    pub fn with_runtime(mut self, runtime: impl Into<(u32, Box<dyn Runtime>)>) -> Self {
```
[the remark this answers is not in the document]
skletsun: Going to implement checking for runtime duplicates according to our discussion.

PR #1477, opened 2019-09-26
Map proof protobuf serialization

pavel-mukhanov:
## Overview
Implemented `ProtobufConvert` for `MapProof`.

### Definition of Done


exonum/src/proto/mod.rs
```diff
@@ -325,6 +327,77 @@ where
...
+            })
+            .collect();
+
+        let entries: Vec<OptionalEntry> = self
+            .all_entries_unchecked()
+            .map(|(key, value)| {
+                let mut entry = OptionalEntry::new();
+                let mut buf = vec![0u8; key.size()];
+                key.write(&mut buf);
+                entry.set_key(buf.to_vec()); //TODO: maybe we should use separate field for missing keys.
```
popzxc: Just do not set it. All fields in `proto3` are optional by default.

components/merkledb/src/proof_map_index/proof.rs
```diff
@@ -440,6 +440,28 @@ impl<K, V> MapProof<K, V> {
         debug_assert!(self.proof.windows(2).all(|w| w[0].path < w[1].path));
         self
     }
+
+    /// Creates MapProof from provided `proof` and `entries` vectors. Used to construct proof
+    /// after deserialization.
+    /// TODO: maybe it's better to use builder methods(but I don't think they should be exposed to public).
```
popzxc: I think current approach is better. Builder methods are associated with some `ProofMapIndex`, and deserialized proofs are not.

Just name method `from_raw_parts` (by analogy from `Vec`) to indicate that user must be really sure that he wants to use it.

exonum/src/proto/schema/exonum/proof.proto
```diff
@@ -0,0 +1,34 @@
...
+
+message OptionalEntry {
+    bytes key = 1;
+    bytes value = 2;
+}
+
+message MapProofEntry {
+    bytes proof_path = 1;
+    exonum.Hash hash = 2;
+}
```
popzxc: No endline.

exonum/src/proto/tests.rs
```diff
@@ -326,3 +328,36 @@ fn test_convert_any_message() {
...
+fn assert_deserialized_proof<K, V>(proof: MapProof<K, V>)
+where
+    K: BinaryKey + ObjectHash + fmt::Debug,
+    V: BinaryValue + ObjectHash + fmt::Debug,
+    MapProof<K, V>: ProtobufConvert + PartialEq,
+{
+    let pb = proof.to_pb();
+    let deserialized: MapProof<K, V> = MapProof::from_pb(pb).unwrap();
+    assert_eq!(proof, deserialized);
+    deserialized.check().expect("proof is not valid");
```
popzxc: Maybe add a check that root hashes from original and deserialized proofs are equal?

exonum/src/proto/tests.rs
```diff
@@ -326,3 +328,36 @@ fn test_convert_any_message() {
...
+    }
+
+    let proof = table.get_proof(5);
+    assert_deserialized_proof(proof);
+
+    let proof = table.get_multiproof(5..15);
+    assert_deserialized_proof(proof);
+}
+
+fn assert_deserialized_proof<K, V>(proof: MapProof<K, V>)
```
popzxc: I'd name it `assert_proof_roundtrip`

exonum/src/proto/mod.rs
```diff
@@ -335,6 +337,77 @@ where
     }
 }
 
+impl<K, V> ProtobufConvert for exonum_merkledb::MapProof<K, V>
+where
+    K: BinaryKey,
```
slowli: I'd suggest to restrict `key`s to types implementing `ProtobufConvert` (meaning "any type we know how to (de)serialize to / from Protobuf") instead of `BinaryKey`, and use the corresponding (de)serialization logic. The less proofs expose internal logic of MerkleDB, the better.

exonum/src/proto/mod.rs
```diff
@@ -335,6 +337,77 @@ where
...
+    V: BinaryValue,
+    Vec<(K, Option<V>)>: FromIterator<(<K as ToOwned>::Owned, Option<V>)>,
```
slowli: This is a weird type constraint. Cannot it be simplified to `K: From<<K as ToOwned>::Owned>`, which seems to be logically equivalent to `K: ToOwned<Owned = K>`, or , logically speaking, `K: 'static` (although I'm not quite sure that the compiler will be able to correctly reason about the latter).

exonum/src/proto/mod.rs
```diff
@@ -335,6 +337,77 @@ where
     }
 }
 
+impl<K, V> ProtobufConvert for exonum_merkledb::MapProof<K, V>
+where
+    K: BinaryKey,
+    V: BinaryValue,
```
slowli: I'm not 100% sure whether using `BinaryValue` as a constraint similarly breaches the encapsulation; AFAIK, `BinaryValue` is automatically derived together with `ProtobufConvert`, so it *could* make sense to use `ProtobufConvert` constraint here as well.

exonum/src/proto/mod.rs
```diff
@@ -335,6 +337,77 @@ where
...
+    fn to_pb(&self) -> Self::ProtoStruct {
+        let mut map_proof = MapProof::new();
+
+        let proof: Vec<MapProofEntry> = self
+            .proof_unchecked()
+            .iter()
+            .map(|(p, h)| {
+                let mut entry = MapProofEntry::new();
+                entry.set_hash(h.to_pb());
+                entry.set_proof_path(p.as_bytes().to_vec());
```
slowli: I'd suggest to extract `ProofPath` serialization into a separate place (e.g., by implementing `ProtobufConvert` for this type), and to use the same "compressed" serialization as for hashing:

```proto
message ProofPath {
  uint32 bit_count = 1;
  bytes bits = 2; // minimum necessary length, padded with zeros if necessary
}
```

exonum/src/proto/mod.rs
```diff
@@ -335,6 +337,77 @@ where
...
+                ))
+            })
+            .collect::<Result<Vec<(_, _)>, Error>>()?;
+
+        let entries = pb
+            .get_entries()
+            .iter()
+            .map(|entry| {
+                let key = K::read(entry.get_key());
+                let value = V::from_bytes(Cow::Borrowed(entry.get_value())).ok();
```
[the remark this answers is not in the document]
alekseysidorov: I wonder if the `repeated` field with the single is more fit in this case?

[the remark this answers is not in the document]
slowli: I like `oneof` better from the semantic standpoint - looking at the Protobuf declaration, it's clear what the intent is. From the technical perspective, a `repeated` field looks OK.

components/merkledb/src/proof_map_index/proof.rs
```diff
@@ -440,6 +440,27 @@ impl<K, V> MapProof<K, V> {
         debug_assert!(self.proof.windows(2).all(|w| w[0].path < w[1].path));
         self
     }
+
+    /// Creates MapProof from provided `proof` and `entries` vectors. Used to construct proof
+    /// after deserialization.
+    pub fn from_raw_parts(proof: &[(ProofPath, Hash)], entries: Vec<(K, Option<V>)>) -> Self {
```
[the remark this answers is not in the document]
pavel-mukhanov: I think there is trade off, because to implement `ProtobufConvert` for `MapProof` in `merkledb` crate we need to add proto files and `exonum-build` dependency(to use exonum protobuf types).  
In one hand we need to hide inner proof details, in other hand we don't want to add extra dependencies because IMO `merkledb` should not be depended on protobuf.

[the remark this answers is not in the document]
slowli: Extracting `ProtobufConvert` is orthogonal to the question where to implement `ProtobufConvert` for proofs 🙃 Also, I'd imagine if `ProtobufConvert` *is* extracted, the types that proofs depend on would have `ProtobufConvert` implementations directly in the `exonum-crypto` crate (maybe, gated via a feature, since having Protobuf compiler locally installed is a non-trivial complication, which is unnecessary in some scenarios). Likewise, proof implementations could be gated via a feature.

In general, I don't think that implementing Protobuf (de)serialization is significantly different from implementing the `serde` one.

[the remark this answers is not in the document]
vitvakatu: > where to implement ProtobufConvert for proofs

Due to orphan rules, it's possible only in MerkleDB, isn't it?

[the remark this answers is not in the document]
slowli: ...Or in the crate defining `ProtobufConvert`, as it is done in the PR currently.

PR #1478, opened 2019-09-26
Set address name in index_metadata [ECR-3578]

popzxc:
Somehow it turned out that provided index address name was lost on the `index_metadata` call step.

This PR fixes it and adds a test so it won't happen anymore.

### DIFF
--- a/components/merkledb/src/views/metadata.rs
+++ b/components/merkledb/src/views/metadata.rs
@@ -172,21 +172,30 @@ where
     T: IndexAccess,
     V: BinaryAttribute + Copy + Default,
 {
-    let index_name = index_address.fully_qualified_name();
+    // Actual name.
+    let index_name = index_address.name.clone();
+    // Full name for internal usage.
+    let index_full_name = index_address.fully_qualified_name();
 
     let mut pool = IndexesPool::new(index_access.clone());
-    let (metadata, is_new) = if let Some(metadata) = pool.index_metadata(&index_name) {
+    let (metadata, is_new) = if let Some(metadata) = pool.index_metadata(&index_full_name) {
         assert_eq!(
             metadata.index_type, index_type,
             "Index type does not match specified one"
         );
         (metadata, false)
     } else {
-        (pool.create_index_metadata(&index_name, index_type), true)
+        (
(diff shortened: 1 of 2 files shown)


components/merkledb/src/views/metadata.rs
```diff
@@ -172,21 +172,30 @@ where
...
+    let (metadata, is_new) = if let Some(metadata) = pool.index_metadata(&index_full_name) {
         assert_eq!(
             metadata.index_type, index_type,
             "Index type does not match specified one"
         );
         (metadata, false)
     } else {
-        (pool.create_index_metadata(&index_name, index_type), true)
+        (
+            pool.create_index_metadata(&index_full_name, index_type),
```
aleksuss: What if change signature of `create_index_metadata` to:
```rust
pool.create_index_metadata(&index_name, &index_full_name, index_type) 
```
for doing this assignment `index_address.name = index_name;` there 
???

popzxc: I don't think that it makes sense: `index_name` isn't actually stored in the metadata (only as a part of `full_name`), so it will be weird to store something in structure but not use it.

aleksuss: Then it's ok if so.

components/merkledb/src/views/metadata.rs
```diff
@@ -172,21 +172,30 @@ where
...
+            true,
+        )
     };
 
-    let index_address = metadata.index_address();
-    let index_state = IndexState::new(index_access, index_name, metadata, is_new);
+    let mut index_address = metadata.index_address();
+    // Set index adress name, since metadata itself doesn't know it.
```
aleksuss:
```suggestion
    // Set index address name, since metadata itself doesn't know it.
```

components/merkledb/src/views/tests.rs
```diff
@@ -986,3 +1001,14 @@ fn fork_from_patch() {
     db.merge(fork.into_patch())
         .expect("Fork created from patch should be merged successfully");
 }
+
+//#[should_panic(expected = "Index name must not be empty")]
```
pavel-mukhanov: Unused comment

PR #1479, opened 2019-09-27
Implemented `ProtobufConvert` for `ListProof`.

pavel-mukhanov:
## Overview
Implemented `ProtobufConvert` for `ListProof`.

### Definition of Done

### DIFF
--- a/components/merkledb/src/proof_list_index/proof.rs
+++ b/components/merkledb/src/proof_list_index/proof.rs
@@ -521,6 +521,31 @@ impl<V: BinaryValue> ListProof<V> {
                 }
             })
     }
+
+    /// Provides access to the proof part of the view. Used in serialization.
+    pub fn proof_unchecked(&self) -> Vec<(ProofListKey, Hash)> {
+        self.proof
+            .iter()
+            .map(|entry| (entry.key, entry.hash))
+            .collect()
+    }
+
+    /// Creates `ListProof` from `proof` and `entries` vectors. Used to construct proof
+    /// after deserialization.
+    pub fn from_raw_parts(
+        proof: Vec<(ProofListKey, Hash)>,
+        entries: Vec<(u64, V)>,
+        length: u64,
+    ) -> Self {
+        Self {
+            proof: proof
+                .into_iter()
+                .map(|(key, hash)| HashedEntry::new(key, hash))
+                .collect(),
+            entries,
+            length,
+        }
+    }
 }
 
 /// Version of `ListProof` obtained after verification.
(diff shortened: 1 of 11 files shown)


components/merkledb/src/proof_list_index/proof.rs
```diff
@@ -521,6 +521,31 @@ impl<V: BinaryValue> ListProof<V> {
                 }
             })
     }
+
+    /// Provides access to the proof part of the view. Used in serialization.
+    pub fn proof_unchecked(&self) -> Vec<(ProofListKey, Hash)> {
```
slowli: Same problem here as for `MapProof`s (#1477): because `ProtobufConvert` is not extracted as a separate crate, it's necessary to expose internal structure of MerkleDB. This doesn't look like a good decision in long term.

exonum/src/proto/schema/exonum/proof.proto
```diff
@@ -0,0 +1,39 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+import "helpers.proto";
+
+package exonum.proof;
+
+message ListProof {
+    repeated HashedEntry proof = 1;
+    repeated ListProofEntry entries = 2;
```
slowli: Nit: `ListProofEntry` looks like a non-descriptive name. I'd suggest something like

```proto
message ListProof {
  message HashedEntry { /* ... */ }
  message ValueEntry { /* ... */ }

  repeated HashedEntry proof = 1;
  repeated ValueEntry entries = 2;
}
```

Also, a semi-troll proposition is to use `map<uint64, bytes>` for `entries`. Are `map`s supported by the crate we use?

exonum/src/proto/schema/exonum/proof.proto
```diff
@@ -0,0 +1,39 @@
...
+    uint64 length = 3;
+}
+
+message HashedEntry {
+    ProofListKey key = 1;
+    bytes hash = 2;
```
slowli: Why not `exonum.Hash`?

exonum/src/proto/mod.rs
```diff
@@ -335,6 +339,86 @@ where
...
+        list_proof
+    }
+
+    fn from_pb(pb: Self::ProtoStruct) -> Result<Self, Error> {
+        let proof = pb
+            .get_proof()
+            .iter()
+            .map(|entry| {
+                Ok((
+                    exonum_merkledb::ProofListKey::new(
```
slowli: It is necessary to check ranges for `height` and `index`. The constructor will crash if `index > 1 << 58` with the debug assertions on, and probably can manifest unintended behavior further on if the assertions are off.

exonum/src/proto/schema/exonum/proof.proto
pavel-mukhanov: Yes, it does. Will be added in the new PR.

PR #1482, opened 2019-09-30
Add shutdown method into runtime trait

popzxc:
## Overview

With such a method runtimes other that `Rust` will be able to be stopped gracefully.


popzxc:
I think we should somehow exclude `tests.rs` and `tests/*.rs` from coverage calculation.

PR #1483, opened 2019-10-01
Use strings for protobuf files

popzxc:
Returning reference is not really handy.

### DIFF
--- a/exonum/src/runtime/mod.rs
+++ b/exonum/src/runtime/mod.rs
@@ -352,8 +352,8 @@ pub struct ArtifactProtobufSpec {
 }
 
 impl ArtifactProtobufSpec {
-    pub fn from_str_list(sources_strs: &[(&str, &str)]) -> Self {
-        let sources = sources_strs.iter().map(From::from).collect();
+    pub fn from_str_list(sources_strings: &[(&str, &str)]) -> Self {
+        let sources = sources_strings.iter().map(From::from).collect();
 
         Self { sources }
     }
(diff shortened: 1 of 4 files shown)


exonum/src/runtime/mod.rs
```diff
@@ -323,24 +323,46 @@ where
...
+
+impl Default for ArtifactProtobufSpec {
     /// Create blank artifact information without any proto sources.
     fn default() -> Self {
-        const EMPTY_SOURCES: [(&str, &str); 0] = [];
-
-        Self {
-            sources: EMPTY_SOURCES.as_ref(),
-        }
+        Self { sources: vec![] }
```
alekseysidorov: Why not just derive?

exonum/src/runtime/mod.rs
```diff
@@ -323,24 +323,46 @@ where
...
     /// is the file name, the second one is its content.
     ///
     /// The common interface entry point is always in the `service.proto` file.
-    pub sources: &'a [(&'a str, &'a str)],
+    pub sources: Vec<ProtoSourceFile>,
 }
 
-impl<'a> Default for ArtifactProtobufSpec<'a> {
+impl ArtifactProtobufSpec {
+    pub fn from_str_list(sources_strings: &[(&str, &str)]) -> Self {
```
alekseysidorov: You also may write `From` conversion for this case, or just write `sources_strings.iter().map(From::from).collect();` inside the `quote!` macro.

PR #1485, opened 2019-10-01
Allow to use `Arc<dyn Snapshot>` as IndexAccess

alekseysidorov:
This change allows to write code like following:

```rust
fn schema(&self) -> FooSchema<Arc<dyn Snapshot>> {
    let snapshot = Arc::new(self.snapshot());
    FooSchema::new(snapshot)
}
```


popzxc:
Isn't it a [duplicate](https://github.com/exonum/exonum/pull/1455/files#diff-9b27e06382bf8ba688dc91b7d3271bb1R370)?

PR #1488, opened 2019-10-03
Add support of external runtimes to exonum-cli

popzxc:
Currently we aren't able to create a node with additional runtimes with `exonum-cli`.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -93,7 +93,7 @@ use exonum::{
     blockchain::InstanceCollection,
     exonum_merkledb::{Database, RocksDB},
     node::Node,
-    runtime::rust::ServiceFactory,
+    runtime::{rust::ServiceFactory, Runtime},
 };
 use exonum_supervisor::Supervisor;
 
@@ -111,6 +111,7 @@ pub mod password;
 #[derive(Debug, Default)]
 pub struct NodeBuilder {
     services: Vec<Box<dyn ServiceFactory>>,
+    external_runtimes: Vec<(u32, Box<dyn Runtime>)>,
 }
 
 impl NodeBuilder {
@@ -122,6 +123,14 @@ impl NodeBuilder {
     /// Adds new Rust service to the list of available services.
     pub fn with_service(mut self, service: impl Into<Box<dyn ServiceFactory>>) -> Self {
         self.services.push(service.into());
+        self
+    }
+
+    /// Adds a new Runtime to the list of available runtimes.
+    ///
+    /// Note that you don't have to add a Rust Runtime, since it's included by default.
+    pub fn with_external_runtime(mut self, runtime: impl Into<(u32, Box<dyn Runtime>)>) -> Self {
+        self.external_runtimes.push(runtime.into());
         self
     }
 
@@ -138,6 +147,7 @@ impl NodeBuilder {
             let node_config_path = run_config.node_config_path.to_string_lossy().to_string();
             let node = Node::new(
                 database,
+                self.external_runtimes,
                 self.services.into_iter().map(InstanceCollection::new),
                 run_config.node_config,
                 Some(node_config_path),
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -111,7 +111,6 @@ pub mod password;
 #[derive(Debug, Default)]
 pub struct NodeBuilder {
     services: Vec<Box<dyn ServiceFactory>>,
-    instances: Vec<InstanceCollection>,
     external_runtimes: Vec<(u32, Box<dyn Runtime>)>,
 }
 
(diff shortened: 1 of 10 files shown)


examples/cryptocurrency/examples/demo.rs
```diff
@@ -64,10 +64,14 @@ fn node_config() -> NodeConfig {
 fn main() {
     exonum::helpers::init_logger().unwrap();
 
+    let external_runtimes: Vec<(u32, Box<dyn exonum::runtime::Runtime>)> = vec![];
```
popzxc: I think we should refine `Node` and `Blockchain` creation algorithm and enforce builders usage.

So we will have a much more cleaner code and encapsulate the `Node` internal structure:
```rust
let node = NodeBuilder::new()
    .with_db(TemporaryDB::new()
    .with_service(CryptocurrencyService)
    .finalize(node_config());
```

exonum/src/node/mod.rs
```diff
@@ -909,6 +910,7 @@ impl Node {
     /// Creates node for the given services and node configuration.
     pub fn new(
         database: impl Into<Arc<dyn Database>>,
+        external_runtimes: impl IntoIterator<Item = impl Into<(u32, Box<dyn Runtime>)>>,
```
aleksuss: Just out of curiosity. Is there really reason to use such complicated type rather then `Vec<(u32, Box<dyn Runtime)>` or something like this ?

examples/cryptocurrency/examples/demo.rs
aleksuss: Good idea. 👍🏻

exonum/src/node/mod.rs
popzxc: Well, I thought about it too. We have even a much funnier place: function takes `IntoIter` just to `collect` it into a `vec`, and then (!) call `into_iter` for it to pass into another function: [src](https://github.com/exonum/exonum/blob/dynamic_services/exonum/src/blockchain/builder.rs#L75).

However, I wanted this change to be consistent with surrounding code, so I made it `IntoIter` just as `services`.

I think it's a good topic for further refactoring, but fixing the whole issue is a bit out of scope of this PR.

aleksuss: Ok. Obviously it makes sense for now.

exonum/src/blockchain/builder.rs
```diff
@@ -89,6 +88,17 @@ impl BlockchainBuilder {
         self.with_additional_runtime(runtime)
     }
 
+    pub fn with_external_runtimes(
```
vitvakatu: Now we have 4 (!) methods for builder that are used to add Runtime:
`with_external_runtimes`, `with_additional_runtime`, `with_default_runtime` and `with_rust_runtime`. Don't you think it's somewhat complicated?

popzxc: Yeah, it should be refined too.

popzxc:
Regarding "this change will not bring any benefits yet":

1. I'm working on an additional runtime on my own, and this runtime doesn't require a complex initialization.
2. We have a `sample_runtime` example, which, I think should be refactored to use `exonum-cli`.

So it will bring some benefits.

popzxc:
> Interesting will we have to allow adding of service instance collections directly in the `NodeBuilder` to create own networks with payload at startup.

Yes, I think it's not hard to implement. I'll do it after #1490 will be merged.

popzxc:
As @alekseysidorov requested, I also added a possibility to add builtin instance collections.

So users of `exonum-cli` can deploy&init service at the start of the Exonum node.

cli/src/lib.rs
```diff
@@ -130,6 +132,26 @@ impl NodeBuilder {
...
+    }
+
+    /// Adds a new InstanceCollection to the list of available services.
+    ///
+    /// With that method you can start Exonum with already deployed & running service
+    /// instances.
+    ///
+    /// Note that if you add the `InstanceCollection` manually, you don't need to call
+    /// `with_service`, since `InstanceCollection` already contains a service inside.
+    pub fn with_instance_collection(mut self, instance_collection: InstanceCollection) -> Self {
```
alekseysidorov: I mean that we should provide runtime-agnostic mechanism to add instances at startup.

Something like this:

```rust
struct Instance {
    name: String,
    artifact: ArtifactId,
    constructor: Vec<u8>
}
```

popzxc: Ah. I see. I thought you need it as a temporary solution to run services without `exonum-launcher`.

I'll remove it then.

popzxc: However, runtime-independent deploy at the start is much bigger question, and totally out of the scope of this PR.

PR #1489, opened 2019-10-03
Avoid waiting in the `add_transaction` endpoint [ECR-3222]

alekseysidorov:
This change may improve liveness on highload.


aleksuss:
I've run benchmarks and got good results. No dropped connections at all. I think it could be merged.

PR #1492, opened 2019-10-04
Do start and initialize service at single step [ECR-3222]

alekseysidorov:


popzxc:
Regarding the PR idea: I like it. Currently the service has a Z state between the start and initialize: service is already running, but it cannot do anything, since it's not initalized.
And since those methods are called one after another I don't see a reason why not to combine them.

alekseysidorov:
Yes that is correct. In old interface there was sitiuations where service became in incomplete state between `start_service` and `initialize_service` invocations. That lead to fatal errros in some cases.
This PR change workflow to avoid this situations by design.
In other words now service is in correct state all the time.

examples/sample_runtime/src/main.rs
```diff
@@ -96,59 +114,28 @@ impl Runtime for SampleRuntime {
...
-            SampleService {
-                name: spec.name.clone(),
-                ..SampleService::default()
-            },
-        );
-        println!("Starting service: {:?}", spec);
+    /// Starts an existing `SampleService` instance with the specified ID.
+    fn restart_service(&mut self, spec: &InstanceSpec) -> Result<(), ExecutionError> {
+        let instance = self.start_service(spec)?;
+        println!("Starting service {}: {:?}", spec, instance);
```
aleksuss: Why not log macro ?

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -435,13 +403,13 @@ impl Dispatcher {
         runtime_changes.push(ApiChange::InstanceAdded(instance.id));
     }
 
-    /// Start a new service instance.
+    /// Start a new previously added service instance.
```
aleksuss:
```suggestion
    /// Restart a new previously added service instance.
```

exonum/src/runtime/mod.rs
```diff
@@ -198,49 +198,32 @@ pub trait Runtime: Send + Debug + 'static {
     /// * Ensure that the deployed artifact has the following information, even if it is empty.
     fn artifact_protobuf_spec(&self, id: &ArtifactId) -> Option<ArtifactProtobufSpec>;
 
-    /// Start a new service instance with the given specification.
+    /// Start previously added service instance with the given specification.
```
```suggestion
    /// Restart previously added service instance with the given specification.
```

PR #1494, opened 2019-10-07
Enhancements in Testkit

skletsun:
## Overview

- Rust runtime is considered default and always created while creating Testkit. Other runtimes considered additional and can be added via builder's method `.with_additional_runtime()`.
- Added checking for runtimes duplicates in the builder.
- Various enhancements: refactoring, methods naming, an additional method for adding `SimpleSupervisor` to the builder.
---

### Definition of Done

### DIFF
--- a/services/time/tests/test_exonum_time.rs
+++ b/services/time/tests/test_exonum_time.rs
@@ -98,7 +98,7 @@ fn assert_transaction_result<S: IndexAccess>(
 fn test_exonum_time_service_with_3_validators() {
     let mut testkit = TestKitBuilder::validator()
         .with_validators(3)
-        .with_service(TimeServiceInstance)
+        .with_rust_service(TimeServiceInstance)
         .create();
 
     let validators = testkit.network().validators().to_vec();
@@ -159,7 +159,7 @@ fn test_exonum_time_service_with_3_validators() {
 fn test_exonum_time_service_with_4_validators() {
     let mut testkit = TestKitBuilder::validator()
         .with_validators(4)
-        .with_service(TimeServiceInstance)
+        .with_rust_service(TimeServiceInstance)
         .create();
 
     let validators = testkit.network().validators().to_vec();
@@ -271,7 +271,7 @@ fn test_exonum_time_service_with_4_validators() {
 fn test_exonum_time_service_with_7_validators() {
     let mut testkit = TestKitBuilder::validator()
         .with_validators(7)
-        .with_service(TimeServiceInstance)
+        .with_rust_service(TimeServiceInstance)
         .create();
 
     let validators = testkit.network().validators().to_vec();
@@ -321,7 +321,7 @@ fn test_exonum_time_service_with_7_validators() {
 fn test_mock_provider() {
     let mock_provider = MockTimeProvider::default();
     let mut testkit = TestKitBuilder::validator()
-        .with_service(
+        .with_rust_service(
             InstanceCollection::new(TimeServiceFactory::with_provider(mock_provider.clone()))
                 .with_instance(INSTANCE_ID, INSTANCE_NAME, ()),
         )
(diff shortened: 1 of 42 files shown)


testkit/src/builder.rs
```diff
@@ -117,34 +114,20 @@ pub struct TestKitBuilder {
     our_validator_id: Option<ValidatorId>,
     test_network: Option<TestNetwork>,
     logger: bool,
-    rust_services: Vec<Box<dyn ServiceFactory>>,
-    runtimes: Vec<(u32, Box<dyn Runtime>)>,
+    rust_runtime: RustRuntime,
```
popzxc: Nit: Do we have to distinguish rust runtime on the field level?

testkit/src/builder.rs
```diff
@@ -117,34 +114,20 @@ pub struct TestKitBuilder {
...
     pub fn validator() -> Self {
-        TestKitBuilder {
-            test_network: None,
-            our_validator_id: Some(ValidatorId(0)),
-            logger: false,
-            rust_services: vec![],
-            runtimes: vec![],
-            instances: vec![],
-        }
+        Self::new(true)
```
popzxc: Nit: `Self::new(true)` does not really explain what exactly is going on. Maybe add an enum `TestkitRole` with variants `Validator` and `Auditor`?

testkit/src/builder.rs
```diff
@@ -247,4 +235,22 @@ impl TestKitBuilder {
         let testkit = self.create();
         testkit.run(public_api_address, private_api_address);
     }
+
+    // Creates testkit for validator or auditor node. Adds supervisor service by default.
```
popzxc: > Adds supervisor service by default.

Does it?

testkit/src/builder.rs
skletsun: Yes. If we store it as a trait object like other runtimes we lose the ability to populate it with service factories.

testkit/src/builder.rs
skletsun: Thanks, it's outdated.

testkit/src/builder.rs
skletsun: Well, good point, thanks. But it's rather an internal method that eliminates code duplication for the `validator()` and `auditor()` methods. And now I think why do we need that boolean flag and why not just pass the Option<ValidatorId> instead? Going to implement it that way.

popzxc: Yeah, makes sense.

testkit/src/simple_supervisor/tests.rs
```diff
@@ -425,7 +425,7 @@ fn service_config_discard_fake_supervisor() {
 fn test_configuration_and_rollbacks() {
     let mut testkit = TestKitBuilder::auditor()
         .with_validators(1)
-        .with_service(SimpleSupervisor)
+        .with_simple_supervisor()
```
alekseysidorov: I think that `with_rust_service(SimpleSupervisor)` is enough.

skletsun: Well, I added it because `SimpleSupervisor` is not used outside TestKit and this method simplifies its usage a little. But in fact, it's not such important and can be deleted.

services/time/tests/test_exonum_time.rs
```diff
@@ -368,7 +368,7 @@ fn test_selected_time_less_than_time_in_storage() {
     let mut testkit = TestKitBuilder::validator()
         .with_validators(1)
         .with_rust_service(TimeServiceInstance)
-        .with_simple_supervisor()
+        .with_rust_service(SimpleSupervisor)
```
aleksuss: Do we still need the `.with_simple_supervisor()` method ?

skletsun: I removed it within 5a9f90e

PR #1495, opened 2019-10-08
Implement API endpoints that shows config proposals in Supervisor [ECR-3610]

SergeiMal:
## Overview

---

See: https://jira.bf.local/browse/ECR-3610

### Definition of Done

### DIFF
--- a/services/supervisor/src/api.rs
+++ b/services/supervisor/src/api.rs
@@ -47,9 +47,9 @@ pub trait PublicApi {
     /// Error type for the current API implementation.
     type Error: Fail;
     /// Returns an actual consensus configuration of the blockchain.
-    fn consensus_config(&self, _query: ()) -> Result<ConsensusConfig, Self::Error>;
+    fn consensus_config(&self) -> Result<ConsensusConfig, Self::Error>;
     /// Returns an pending propose config change.
-    fn config_proposal(&self, _query: ()) -> Result<ConfigProposalWithHash, Self::Error>;
+    fn config_proposal(&self) -> Result<ConfigProposalWithHash, Self::Error>;
 }
 
 struct ApiImpl<'a>(&'a ServiceApiState<'a>);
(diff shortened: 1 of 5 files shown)


services/supervisor/src/api.rs
```diff
@@ -34,6 +38,18 @@ pub trait PrivateApi {
     /// Creates and broadcasts the `StartService` transaction, which is signed
     /// by the current node, and returns its hash.
     fn start_service(&self, service: StartService) -> Result<Hash, Self::Error>;
+    /// Creates and broadcasts the `ConfigVote` transaction, which is signed
+    /// by the current node, and returns its hash.
+    fn confirm_config(&self, vote: ConfigVote) -> Result<Hash, Self::Error>;
```
popzxc: What about `create_config_proposal` endpoint?

services/supervisor/src/api.rs
```diff
@@ -59,6 +75,25 @@ impl PrivateApi for ApiImpl<'_> {
...
+
+    fn confirm_config(&self, vote: ConfigVote) -> Result<Hash, Self::Error> {
+        self.broadcast_transaction(vote).map_err(From::from)
+    }
+}
+
+impl PublicApi for ApiImpl<'_> {
+    type Error = api::Error;
+
+    fn consensus_config(&self) -> Result<ConsensusConfig, Self::Error> {
```
popzxc: What about service configs?

services/supervisor/src/api.rs
```diff
@@ -59,6 +75,25 @@ impl PrivateApi for ApiImpl<'_> {
...
+        Ok(CoreSchema::new(self.0.snapshot()).consensus_config())
+    }
+
+    fn config_proposal(&self) -> Result<ConfigProposalWithHash, Self::Error> {
+        Schema::new(self.0.instance.name, self.0.snapshot())
+            .pending_proposal()
+            .get()
+            .ok_or_else(|| Self::Error::NotFound("Pending config not found.".to_string()))
```
popzxc: I don't think that lack of pending proposal should be treated like error. Maybe `Ok(None)`?

services/supervisor/src/api.rs
SergeiMal: It seems that the concrete service have to care about it by itself. Exonum doesn't store this information now.

popzxc: OK then.

PR #1497, opened 2019-10-09
Extract keys module into a separate crate

popzxc:
Since we agreed that demonolithization is one of the goals that we should achieve, I decided to make a small step in that direction.

### DIFF
--- a/components/keys/Cargo.toml
+++ b/components/keys/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "exonum-keys"
+version = "0.12.0"
+edition = "2018"
+authors = ["The Exonum Team <exonum@bitfury.com>"]
+homepage = "https://exonum.com/"
+repository = "https://github.com/exonum/exonum"
+license = "Apache-2.0"
+keywords = ["exonum", "cryptography", "library"]
+categories = ["cryptography", "hashing"]
+description = "Exonum validator keys management."
+
+[dependencies]
+rand = "0.7"
+failure = "0.1.5"
+pwbox = "0.2.1"
+secret-tree = "0.2.0"
+serde = "1.0.10"
+serde_derive = "1.0.64"
+toml = "0.5.0"
+
+exonum-crypto = { version = "0.12.0", path = "../crypto" }
+
+[dev-dependencies]
+tempdir = "0.3.7"
+hex = "0.4"
--- a/testkit/src/network.rs
+++ b/testkit/src/network.rs
@@ -18,9 +18,9 @@ use exonum::{
     blockchain::{ConsensusConfig, ValidatorKeys},
     crypto::{self, PublicKey, SecretKey},
     helpers::{Height, Round, ValidatorId},
-    keys::Keys,
     messages::{Precommit, Propose, Verified},
 };
+use exonum_keys::Keys;
 use exonum_merkledb::ObjectHash;
 
 // TODO Refactor TestNetwork and TestkitBuilder [ECR-3222]
--- a/testkit/src/network.rs
+++ b/testkit/src/network.rs
@@ -18,9 +18,9 @@ use exonum::{
     blockchain::{ConsensusConfig, ValidatorKeys},
     crypto::{self, PublicKey, SecretKey},
     helpers::{Height, Round, ValidatorId},
+    keys::Keys,
     messages::{Precommit, Propose, Verified},
 };
-use exonum_keys::Keys;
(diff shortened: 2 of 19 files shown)


components/keys/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+secret-tree = "0.2.0"
+serde = "1.0.10"
+serde_derive = "1.0.64"
+toml = "0.5.0"
+
+exonum-crypto = { version = "0.12.0", path = "../crypto" }
+
+[dev-dependencies]
+tempdir = "0.3.7"
+hex = "0.4"
```
aleksuss: New line ?

testkit/src/network.rs
```diff
@@ -18,9 +18,9 @@ use exonum::{
     blockchain::{ConsensusConfig, ValidatorKeys},
     crypto::{self, PublicKey, SecretKey},
     helpers::{Height, Round, ValidatorId},
-    keys::Keys,
     messages::{Precommit, Propose, Verified},
 };
+use exonum_keys::Keys;
```
alekseysidorov: Can you reexport `exonum_keys` as `exonum::keys` like `merkledb` or `crypto`?

popzxc: To be honest, I don't really like the idea of re-exporting, since it results in a confusing scenarios:

[uno](https://github.com/exonum/exonum/blob/dynamic_services/cli/src/lib.rs#L94) / [dos](https://github.com/exonum/exonum/blob/dynamic_services/examples/cryptocurrency/src/lib.rs#L43) / [tres](https://github.com/exonum/exonum/blob/dynamic_services/exonum/src/blockchain/tests.rs#L26)

The same thing is done *three* different ways. And exonum is already has plenty of such an inconsistencies.

So I don't see any benifits caused by re-exporting, but instead do see only problems.

Can you please explain what's the motivation lies behind the re-exporting policy?

vitvakatu: All dependencies of Exonum (including MerkleDB and this new crate) are not dependencies of user applications and there is no need in forcing user to handle these additional libraries and their versions in Cargo.toml. Each type which is used in public Exonum API must be available automatically for user. 

Besides clattering Cargo.toml, need in manual dependencies import might also cause interoperability problems between different versions of Exonum and their dependencies.

About your examples - it is definitely a problem, it looks like one of the public reexports of MerkleDB is redundant. Can't say which one exactly - we have problem with `ProtobufConvert` macro when `merkledb` isn't in a scope, I need to check code to give answer. Inconsistency in importing may be solved quite easily - by choosing a method and sticking to it.

> but instead do see only problems.

Can you elaborate if you mean smth other than inconsistency?

popzxc: Yeah, I agree that we should provide users an interface to all the used types.

But for that purpose I'd expect to have a separate module, e.g. `prelude`, which won't only provide re-exports to used crates, but also create a flat import structure (currently users have to build a pretty complex `use` tree just to get things compiled).

And I will be totally fine with such a module.

>Can you elaborate if you mean smth other than inconsistency?

Sort of. Since now we're going to extract several parts of exonum into separate crates, there will be a bigger amount of components (possibly, we wil have more component crates than core modules). And in that scenario, using `exonum::name_of_reexported_crate` inside of the exonum will be sort of confusing -- if you see that, you suppose that it's a module, but actually a separate crate.

As a conclusion:

I'd be happy if we will have a `prelude` module which will export all the required types **for users**, but internally we will use crate names to have an explicit structure of the project.

vitvakatu: I see. I like your idea about `prelude` module. Yep, I was only talking about users. `use exonum::name_of_reexported_crate` is definitely invalid usage inside Exonum itself, it must be used only in user services (e.g. cryptocurrency).

popzxc: As a conclusion: OK, I'll add a re-export for now, but soon I'll create a PR with `prelude` module. Also I suggest to forbid usage of re-exported crates inside exonum itself.

Is that OK for everybody? cc @aleksuss @alekseysidorov

popzxc:
Talking about conversions: Since `KeyPair` is a part of `crypto` crate, those conversions doesn't belong to the scope of this PR.

And btw, conversions from keys to `KeyPair` are [already implemented](https://github.com/exonum/exonum/blob/dynamic_services/components/crypto/src/lib.rs#L513).

If you meant conversions like `(KeyPair, KeyPair) -> Keys` then I think that it's better to construct `Keys` explicitly.

PR #1500, opened 2019-10-15
Encapsulated dispatcher

popzxc:
**Upd 2:** One more entity inroduced: `CommunicationChannel`. This entity encapsulated both dispatcher and mailbox and provides an interface to the mailbox only to supervisor. Eventually, `BlockchainMailbox` and `BlockchainSecretary` have became simlpier.

Supervisor entities are recognized by implementing `SupervisorAccess` empty trait. On the one hand, it means that any service that will decide to implement it, will gain access to the mailbox. But come on, you can not "accidentally" implement trait, and then call `doc(hidden)` method. Services doing so should be concidered malicious and I can't imagine a situation when N validator node administrators will accept deploying of such a service.

**Upd:** Added a [comment](https://github.com/exonum/exonum/pull/1500#issuecomment-542175367) explaining why I decided to "over-engineer" this solution.

**State of the PR**: work in progress, it is not yet well tested. However, the architectural skeleton is strong enough to be criticized. Code has several `TODO`s and emits some warnings that will be resolved soon.

This PR introduces two new entities:

- `BlockchainMailbox`, which combines `DispatcherRef` and `DispatcherSender` responsibilities in one concrete place. `BlockchainMailbox` is created by the core and provided to the runtimes, so every service can request blockchain to perform some kind of actions.
- `BlockchainSecretary`, which is capable of filtering received requests and performing ones from authorized services. Also this entity notifies the core about any actions that were performed during request handling.

New workflow is the following:

1. Blockchain creates a new empty `BlockchainMailbox` before any of the following events: transaction execution, `before_commit` event and `after_commit` event.
2. `BlockchainMailbox` is provided to the runtimes (using dispatcher as a proxy).
3. Runtimes provide a restricted interface to the `BlockchainMailbox` for services: service can request an action, but runtime is capable of setting the correct instance id. Currently the only runtime that supports (and needs) `BlockchainMailbox` is the Rust runtime, which provides restricted access through `*Context` entities.
4. After the event, core creates a `BlockchainSecretary` entity and provides mailbox to it. `BlockchainSecretary` verifies that entity is authorized for request and dispatches requests to the `Dispatcher` (logic here was mainly moved from `impl dispatcher::Action`). If service is not authorized for a request (or error occurs during the request completion), two scenarios are possible: if it's a transaction execution context, the transaction considered failed and rolled back with no further request processing, otherwise (for `before_commit` and `after_commit`) such a request will be simply ignored.
5. After requests processing blockchain core retrieves notifications and can do any follow-up work, e.g. restart API.

Core moments of implementation:

- Currently we have only one service which is authorized for requesting actions (`Supervisor`). However, this can change in the future (e.g. `Supervisor` can be splitted into `ServiceManager` and `Configuration`, or `Dispatcher` can become a service), so the approach used in both `BlockchainMailbox` and `BlockchainSecretary` is generic and flexible for upcoming changes.
- Despite the fact that `Dispatcher` isn't capable of action management anymore, `DispatcherRef` is still needed in the `execute` and `before_commit` for calling other services. I think we can get rid of `Dispatcher` in the interfaces completely, however it's out of the scope of this PR.
- Implementation of `BlockchainMailbox` does not consider services as trusted code, but it consider runtimes to be fair. IMO attempts to be aware of malicious runtimes will result in too much overhead and it just doesn't worth it.


exonum/src/runtime/mailbox/blockchain_secretary.rs
```diff
@@ -0,0 +1,321 @@
...
+            AuthoriziationResult::ShouldSkip
+        }
+    }
+}
+
+/// Internal function encapsulating the check for service
+/// to have sufficient rights to request actions from the blockchain.
+fn is_authorized_for_requests(instance_id: InstanceId) -> bool {
+    // Currently, only Supervisor service is authorized to request changes.
+    instance_id == crate::runtime::SUPERVISOR_INSTANCE_ID
```
popzxc: I think currently it's OK to have that check done in this way. However, since it's encapsulated, this check can be moved to the more appropriate place (e.g. `Blockchain`) easily.

exonum/src/runtime/mailbox/blockchain_secretary.rs
```diff
@@ -0,0 +1,321 @@
...
+}
+
+/// Internal function encapsulating the check for service
+/// to be a part of core api schema.
+fn is_part_of_core_api(instance_spec: &InstanceSpec) -> bool {
```
popzxc: Well, I think it will be nice to restart api only if there are actual api changes, so I decided to make a small step in that direction.

exonum/src/runtime/rust/call_context.rs
```diff
@@ -0,0 +1,110 @@
...
+
+use crate::runtime::{
+    BlockchainMailbox, CallInfo, Caller, DispatcherRef, ExecutionContext, ExecutionError,
+    InstanceId, MethodId,
+};
+
+// TODO Write a full documentation when the interservice communications are fully implemented. [ECR-3493]
+/// Provide a low level context for the call of methods of a different service instance.
+#[derive(Debug)]
+pub struct CallContext<'a> {
```
popzxc: Eventually `CallContext` was in the `mod.rs` despite the fact that it's specific to the Rust runtime.

popzxc:
As a little resolution for private conversation with @alekseysidorov 

It may look that this thing is too over-engineered: why do we need those generic entities, and -- even more -- have them public.

I decided to do things this way as a small step in the direction of making `Rust` runtime optional. With approach introduces in this PR -- ability to create request to abstract blockchain entity, as the environment in which services exist -- it will be possibly later to implement `Supervisor` (and maybe even `Dispatcher`) in `Java` (or any other language), and have an Exonum core with only `Java` runtime.

exonum/src/blockchain/mod.rs
```diff
@@ -243,16 +243,15 @@ impl Blockchain {
...
-                secretary.process_requests_mut(&mut mailbox, &mut dispatcher, &mut fork);
+                let secretary = BlockchainSecretary::new(MailboxContext::NoTx);
 
-                for notification in mailbox.get_notifications() {
-                    // TODO
-                    // unimplemented!();
-                }
+                // We don't care about result, secretary will rollback any failed requests
+                // occurred during processing.
+                let _result =
```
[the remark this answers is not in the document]
popzxc: I've added it to the `BlockchainSecretary`

exonum/src/runtime/mailbox/blockchain_secretary.rs
aleksuss: Interesting. Is it ok to use a doc-string for private method ?

exonum/src/runtime/rust/call_context.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Identifier of the caller service instance.
+    caller: InstanceId,
+    /// Identifier of the called service instance.
+    called: InstanceId,
+    /// The current state of the blockchain.
+    fork: &'a Fork,
+    /// Reference to the communication channel.
+    communication_channel: &'a CommunicationChannel<'a>,
+    /// Depth of call stack.
```
aleksuss: I'd like to see more details here about a purpose of the field.

exonum/src/runtime/mailbox/blockchain_secretary.rs
popzxc: AFAIK rustdoc will not generate docs for private module. So docstring here only indicates that it's a documentation (for reader of sources). However, I will check that.

exonum/src/runtime/rust/call_context.rs
aleksuss: So let's copy-paste this comment in the field description 😄

exonum/src/runtime/mailbox/blockchain_secretary.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// Creates a new `BlockchainSecretary` instance.
+    pub fn new(context: MailboxContext) -> Self {
+        Self { context }
+    }
+
+    /// Processes requests within immutable context.
+    /// Since any failures won't affect the blockchain state, they are simply ignored.
+    pub fn process_requests(&self, mailbox: &mut BlockchainMailbox, dispatcher: &mut Dispatcher) {
+        // During requests processing new requests may appear.
+        loop {
```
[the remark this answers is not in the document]
popzxc: Well, if we will talk about hypothetical situations, I can imagine a scenario when updating config of service A depends on updating of config of service B which depend on updating of config of service A.

But in that scenario such a loop will appear anyway, directly or indirectly (e.g. through recursion), and I don't think that it can be avoided from the side of blockchain.

Talking about correct scenarios, I can not imagine it.

However, I can add an iteration counter (e.g. 4) on which loop will be broken. Again, it's unlikely to be ever achieved, but we can do it for the glory of Defencive Programming.

Will it be ok for you?

exonum/src/runtime/communication_channel.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Creates a new communication channel.
+    pub(crate) fn new(mailbox: &'a BlockchainMailbox, dispatcher: &'a Dispatcher) -> Self {
+        Self {
+            mailbox,
+            dispatcher,
+            phantom: std::marker::PhantomData,
+        }
+    }
+
+    /// Call the corresponding runtime method.
```
dmitry-timofeev: runtime or service method?

dmitry-timofeev: Also, the PR description says the dispatcher is needed to call the other services — but there is also this method. Which is the right one to use?

exonum/src/runtime/communication_channel.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub fn call(
+        &self,
+        context: &ExecutionContext,
```
dmitry-timofeev: Is it correct that ExecutionContext has CommunicationChannel, and
CommunicationChannel#call accepts ExecutionContext (which has CommunicationChannel)?
Are they/must they be equal (channel1.call(context1) is valid iff context1.channel == channel1)?

exonum/src/runtime/communication_channel.rs
```diff
@@ -0,0 +1,92 @@
...
+where
+    T: SupervisorAccess,
+{
+    /// Adds a request to the list of pending actions. These changes will be applied immediately
+    /// before the block commit.
+    ///
+    /// Currently only the supervisor service is allowed to perform this action.
+    /// If any other instance will call this method, the request will be ignored.
+    #[doc(hidden)]
+    pub fn request_action(&self, action: Action, and_then: AfterRequestCompleted) {
```
dmitry-timofeev: I'd clarify when the 'and_then' action is performed (with respect to
   the requesting code, and/or the block processing stages) and from which thread.

exonum/src/runtime/communication_channel.rs
```diff
@@ -0,0 +1,92 @@
...
+    ///
+    /// Only the supervisor service is allowed to perform this action.
+    /// If any other instance will call this method, the request will be ignored.
+    #[doc(hidden)]
+    pub fn update_config(&self, caller_instance_id: InstanceId, changes: Vec<ConfigChange>) {
+        let action = Action::UpdateConfig {
+            caller_instance_id,
+            changes,
+        };
+        self.mailbox.add_request(action, None);
```
dmitry-timofeev: self.request_action(action, None);

exonum/src/runtime/communication_channel.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub(crate) fn new(mailbox: &'a BlockchainMailbox, dispatcher: &'a Dispatcher) -> Self {
+        Self {
+            mailbox,
+            dispatcher,
+            phantom: std::marker::PhantomData,
+        }
+    }
+
+    /// Call the corresponding runtime method.
+    pub fn call(
```
dmitry-timofeev: If there is `call`, shall we also add here `call_async` as a replacement for/extension to ApiSender#submit? It, however, will be rather different as it will produce _num_validators_ transactions when called unconditionally.

exonum/src/runtime/communication_channel.rs
popzxc: Well, `ExecutionContext` has **the reference** to the `CommunicationChannel`, and `call` accepts **the reference** to the `ExecutionContext`. Constructors of the both `ExectionContext` and `CommunicationChannel` aren't exposed to services.
So yes, they must be equal, since end-side willl have only one reference to each structure, and will use it.

exonum/src/runtime/communication_channel.rs
popzxc: >It, however, will be rather different as it will produce num_validators transactions when called unconditionally.

Yep, I think that's the core point. Async transactions currently can be emitted only by users, since they sign them with their personal keys. I think implementing of such a feature will require a really hard R&D work inbefore it can be implemented.

dmitry-timofeev: The thing is currently they can be emitted using ApiSender and signed by the node service key (time service uses this, probably, some configuration service uses this), and the service developers using this [has to be aware of the consequences](https://exonum.com/doc/api/java-binding/0.8.0/com/exonum/binding/core/service/Node.html#submitTransaction\(com.exonum.binding.core.transaction.RawTransaction\)).

exonum/src/runtime/communication_channel.rs
```diff
@@ -0,0 +1,91 @@
...
+            dispatcher,
+            phantom: std::marker::PhantomData,
+        }
+    }
+
+    /// Calls the corresponding service method.
+    pub fn call(
+        &self,
+        context: &ExecutionContext,
+        call_info: &CallInfo,
```
dmitry-timofeev: If this is to be used by services, and service name is considered the primary identifier of a service (and it is definitely easier to use during configuration of a service), why don't we pass service name instead of service id (in CallInfo)? If this method accepts id, then the callee has to either lookup it (how?) or, which is worse in terms of UX, the user will have to configure it with id of the service on which it depends (instead of the name, which the user already knows).

exonum/src/runtime/communication_channel.rs
popzxc: Well, I (from my side) would call that an architectural flaw. We need to develop a safe&consistent interface for that type of actions.

exonum/src/runtime/communication_channel.rs
popzxc: This is to be used by runtimes (since it can provide supervisor extensions, it's dangerous to provide communication channel to the service directly). Runtimes are free to build a handy interface atop of `CommunicationChannel`.

dmitry-timofeev: Ok, but the runtimes themselves also need the handy interface, at least, to translate names to ids (they can't do that on their own in case of cross-lang interop).

popzxc: Well, to be honest, I don't understand the difference between a numeric ID and a string name.

For example, you have a service A that wants to call service B.
Service A has 3 different instances, service B also has 3 different instances.
How instances of service A will know the name of instances of service B that they want to call?

I guess, it will be in the service config. And if you can provide a service instance name to the config, why can't you provide a numeric ID instead?

exonum/src/runtime/rust/service.rs
```diff
@@ -279,23 +284,25 @@ impl<'a> BeforeCommitContext<'a> {
...
-            self.instance.id,
-            dispatcher::Action::UpdateConfig {
-                caller_instance_id: self.instance.id,
-                changes,
-            },
-        )
+    pub fn supervisor_extensions(&self) -> Option<CommunicationChannel<SupervisorAccess>> {
+        if !is_supervisor(self.instance.id) {
+            return None;
+        };
```
slowli: Nit: unnecessary semicolon.

services/supervisor/src/lib.rs
```diff
@@ -99,7 +100,10 @@ impl Service for Supervisor {
                         entry.config_propose
                     );
                     // Perform the application of configs.
-                    context.update_config(entry.config_propose.changes);
+                    let communication_channel = context.supervisor_extensions().unwrap();
```
slowli: Is it possible to cause a panic here? If so, using `expect` with a short description (e.g., `"Blockchain is improperly configured: the `Supervisor` service has incorrect instance ID"`) would be better.

services/supervisor/src/lib.rs
```diff
@@ -150,12 +154,10 @@ impl Service for Supervisor {
                     }
                 };
                 // TODO Rewrite on async await syntax. [ECR-3222]
-                context.dispatcher_channel().request_deploy_artifact(
-                    context.instance.id,
-                    artifact,
-                    spec,
-                    and_then,
-                );
+                let action = Action::StartDeploy { artifact, spec };
```
slowli: Off-topic: Do I understand correctly that each validator triggers a deploy action and broadcasts endorsements of all pending deploys on each block? Is this the expected behavior? Are the repeated endorsement transactions actually broadcast, or is this suppressed somewhere?

exonum/src/runtime/mailbox/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+use crate::runtime::{ArtifactId, ConfigChange, InstanceId, InstanceSpec};
+
+/// Optional callback to be called after request is completed by the blockchain core.
+pub type AfterRequestCompleted = Option<Box<dyn FnOnce() + 'static>>;
+
+/// An interface for runtimes to interact with the Exonum blockchain core.
+///
+/// All the requests added to the mailbox will be processed by the core `Blockchain` structure.
+///
+/// **Important note:** All the requests received after the transaction execution are considered
```
slowli: > received after transaction execution

You mean *during* execution, right?

exonum/src/runtime/mailbox/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+/// **the part of execution process**. So, if service requests blockchain to perform some action
+/// and an error occurs during the request processing, the transaction will be treated as failed
+/// and, as a result, rolled back.
+///
+/// **Policy on request failures:**
```
slowli: This is terribly complicated. Just sayin'.

exonum/src/runtime/mailbox/blockchain_secretary.rs
```diff
@@ -0,0 +1,258 @@
...
+            Action::StartDeploy { artifact, spec } => {
+                // Request the dispatcher to start deploy process and wait until it completed.
+                // Please note that it doesn't mean completion of the deployment process,
+                // because after that artifact should be registered as deployed.
+
+                // TODO: Get rid of `wait` here.
+                dispatcher.deploy_artifact(artifact, spec).wait()
+            }
+
+            // Mutable action.
```
slowli: The whole concept of mutable actions (i.e., actions requiring a fork, which, as far as I can see, stems from dispatcher having its own `Schema`) looks like a sign of flawed design.

exonum/src/runtime/mailbox/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        requests.into_inner()
+    }
+
+    // TODO: currently blockchain doesn't read notifications, because services started
+    // on start of the node aren't added into notifications list.
+    // This should be fixed and notifications mechanism should be used instead of
+    // `dispatcher.notify_api_changes`.
+    #[allow(dead_code)]
+    /// Consumes a mailbox, receiving the notifications about performed actions.
+    pub fn get_notifications(self) -> Vec<Notification> {
```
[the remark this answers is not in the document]
popzxc: That's dead for a purpose, I will add an use of this method in the (literally) next PR (scope of which is making dispatcher not responsible for collecting API changes).

exonum/src/runtime/mailbox/mod.rs
popzxc: Do you have a proposal how to make it simpler, given the fact that requests will be processed when context in which a request was created is already gone?

services/supervisor/src/lib.rs
popzxc: Yes, you understand that (almost) correctly.
It is *somewhat* suppressed: deployment is currently applied synchronously right after request (it's in `after_commit` and request is processed after `after_commit` event), so every validator sends a confirmation in the next block, and thus it will be filtered in the `filter` above.

One piece of flawed code causes other piece of flawed code to work correctly. Checkmate, adepts of clean programming :)

Talking seriously, yeah, it's a breach. And the current logic of deploy-register-init process in Supervisor should be refined as well.

exonum/src/runtime/mailbox/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+    },
+}
+
+/// Enum denoting a request to perform a change in the Exonum blockchain structure.
+#[derive(Debug, Clone)]
+pub enum Action {
+    /// Request to start artifact deployment process.
+    StartDeploy { artifact: ArtifactId, spec: Vec<u8> },
+    /// Request to register the deployed artifact in the blockchain.
+    /// Make sure that you successfully complete the deploy artifact procedure.
```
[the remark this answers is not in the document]
popzxc: To be honest, I didn't understand that either. But decided to leave this comment as-is until @alekseysidorov will be able to explain it correctly.

exonum/src/runtime/mailbox/blockchain_secretary.rs
popzxc: I don't think that we can easily get rid of the dispatcher schema right now.

exonum/src/runtime/mailbox/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+mod blockchain_secretary;
+
+pub use blockchain_secretary::{BlockchainSecretary, MailboxContext};
+
+use std::cell::RefCell;
+
+use crate::runtime::{ArtifactId, ConfigChange, InstanceId, InstanceSpec};
+
+/// Optional callback to be called after request is completed by the blockchain core.
+pub type AfterRequestCompleted = Option<Box<dyn FnOnce() + 'static>>;
```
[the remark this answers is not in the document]
popzxc: Synchronous approach won't work good for heavy-weight artifacts in the future, since it will block core main thread.

Regarding node restarts: I think it's okay for now, since if node will be restarted during the deployment, it won't be deployed anyway (since it won't be registered in the dispatcher schema) and overall deploy will also fail. The whole deploy system currently is not designed to be bulletproof. That's a problem, yeah, but to resolve it we should do a research first (which will create a near-complete list of existing flaws and propose a generic algorithm to deal with them).

exonum/src/runtime/dispatcher/error.rs
```diff
@@ -48,6 +48,10 @@ pub enum Error {
     UnauthorizedCaller = 11,
     /// Malformed arguments for calling a service interface method.
     MalformedArguments = 12,
+    /// Action requested at the inappropriate time.
+    /// E.g. action that requires mutable access to the database, requested
+    /// in the `after_commit`.
+    InappropriateTimeForAction = 13,
```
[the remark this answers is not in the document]
popzxc: We either allow this or make interfaces bigger (currently we have 3 methods in service *Context structures of the Rust runtime, with type-checking it will become 6 (one immutable and one mutable method)).

As an alternative, we can `panic` if action is requested at the inappropriate time. Would it be ok?

exonum/src/runtime/communication_channel.rs
dmitry-timofeev: You-the-user? Doesn't the framework assign the numeric ids?

popzxc: Yep, "you" == "the user".
And yep, numeric IDs are assigned by the framework.

dmitry-timofeev: Then I see no reason to be user-administrator-hostile and require specifying a second identifier which the framework already knows; or user-service-developer-hostile and require the service to read it from _somewhere_. Administrators supply a name when they start service B. The name is the easiest way to reference B in the config for A.

popzxc: OK. Do I understand it correctly that you suggest to include a service name into the `CallInfo` structure which is a part of every transaction?
Or do you suggest to call the service through `communication_channel` in the different way than it'll be done through usual transaction?

dmitry-timofeev: I don't suggest including in the transaction, but changing the signature of this method to use name instead of id.

popzxc: Then again, why is it not "hostile" to send a numeric ID in the transaction and it is "hostile" to use numeric ID here?

dmitry-timofeev: Both are. For txs, IIRC, that is done for storage efficiency. For calls, which happen in a completely different context, I don't see a reason.

dmitry-timofeev: Too bad you can't see the worse UX.

@aleksuss , would you please design the inter-service communication in core to have a reasonable way for a Runtime implementation to perform calls by service name?

aleksuss: @dmitry-timofeev Ok. We'll discuss it.

popzxc:
Closing since #1509 is superior to this.

PR #1505, opened 2019-10-18
Proto optimization [ECR-3472]

skletsun:
## Overview

Attempt to optimize sharing Exonum protobuf specification with external clients according to the discussion in corresponding ticket.
- .proto files updated with Java specific package (does not afect non-Java builds)
- Added script for sharing public info of specification to a specific repo.

**UPD:**
As discussed:
- the proto files synchronization script has been moved to the `exonum-proto-sources` repo
- the `java_package` option has been changed from `com.exonum.binding.messages` to `com.exonum.core.messages` for all shared proto files.
---

See: https://jira.bf.local/browse/ECR-3472

### Definition of Done


exonum/src/proto/push_proto_files.sh
```diff
@@ -0,0 +1,63 @@
...
+cp -v ${COMPONENTS_DIR}/proto/src/proto/common.proto ${PROTO_REPO_TMP_DIR}
+# Crypto stuff
+cp -v ${COMPONENTS_DIR}/crypto/src/proto/schema/types.proto ${PROTO_REPO_TMP_DIR}
+
+header "ADDING PROTO FILES TO THE REPO"
+
+cd ${PROTO_REPO_TMP_DIR}
+
+# Create a new branch for changes and push
+BRANCH_NAME=${BRANCH_RPEFIX}-$(date "+%Y.%m.%d-%H%M%S")
```
vitvakatu: Do we ever need minutes and seconds here? I'd rather expect to see a version number (supplied by the user of the script)

skletsun: Actually, the initial idea was to use that resulting number as something unique (akin to version) in order to be able to quickly create a branch without syncing to version numbers (and branch then can safely be deleted after merging the PR). What do we really need the exact version for? It even may remain unchanged between the revisions of exonum.

components/crypto/src/proto/schema/types.proto
```diff
@@ -16,6 +16,8 @@ syntax = "proto3";
 
 package exonum.crypto;
 
+option java_package = "com.exonum.binding.messages";
```
popzxc: I don't really like idea of making framework know something about client code.

exonum/src/proto/push_proto_files.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
popzxc: Should this script be a part of core repository?

aleksuss: Good point 👍🏻

components/crypto/src/proto/schema/types.proto
dmitry-timofeev: It is not the client code, it is (generated) framework code.

popzxc: IMHO from side of framework everything that depends on it is the client code. We can rephrase it like "I don't really like idea of making base entity know something about child entity"

skletsun: > making base entity know something about child entity

In fact, the base entity does not lean on this piece of information, it's kinda annotation for the protobuf compiler _in case_ entities are _built for Java_.
I believe there are only two real options for configuring this for Java: specifying it in a `.proto` file via `option` or patching it (looks like there is no way specifying it via option for cmd line `protoc` or maven plugin). Personally I like more the first approach - you only need to do this once per platform (and this does not affect others).

exonum/src/proto/push_proto_files.sh
skletsun: According to the discussion in the [ECR-3472](https://jira.bf.local/browse/ECR-3472) if we consider the core repo a single source of changes for proto files this looks more reasonable - the core team knows when the proto files format gets stabilized and when it's time to update it for others (without exposing intermediate API to the public for example).

popzxc: As for me, it makes more sense to have this script in the repo with updated proto files and invoke it on each release, so this script will extract proto files from `exonum`, commit it to the exported-protos repo and create a tag denoting the release of `exonum` for which files were extracted.

exonum/src/proto/push_proto_files.sh
```diff
@@ -0,0 +1,62 @@
...
+# prints a section header
+function header() {
+    local title=$1
+    local rest="========================================================================"
+    echo
+    echo "===[ ${title} ]${rest:${#title}}"
+    echo
+}
+
+PROTO_REPO_URI="https://github.com/exonum/exonum-proto.git"
```
popzxc: `exonum-proto` repo name collides with an `exonum` component [`exonum-proto`](https://github.com/exonum/exonum/blob/dynamic_services/components/proto/Cargo.toml).

exonum/src/proto/push_proto_files.sh
skletsun: Well, makes sense if proto files remain within the single repo.

exonum/src/proto/push_proto_files.sh
skletsun: Is this really that bad? I mean, if I get things right this can only mislead the one who's interesting in the sources of the `exonum-proto` component. But is it really possible scenario WRT the sources are already in (almost) single place? If so - should it be renamed to, for instance, `exonum-protobuf`? I feel like just `exonum/proto` is also not an option.

popzxc: `exonum-proto-sources` maybe?

skletsun: Renamed

components/crypto/src/proto/schema/types.proto
alekseysidorov: The need for such directives indicates on design flaws in protobuf.

exonum/src/proto/push_proto_files.sh
alekseysidorov: I think that you can automatically add `option java_package` during this script execution.

dmitry-timofeev: > you can automatically add option java_package during this script 

I don't think it is sensible to introduce this complexity in the shell script. As these messages are considered to be public, they must be used *as is* by all the other runtimes to ensure compatibility (Java, Python, C#) — and acquire all the options needed. If they don't, some will eventually end up with incompatible classes for `exonum.Hash` in different packages.

popzxc: Options do not affect the serialization algorithm.

alekseysidorov: I think that each runtime should fix flaws in protobuf design by itself to keep proto files clear.

aleksuss:
I suppose we should discuss it inside.

exonum/src/proto/push_proto_files.sh
dmitry-timofeev: > Options do not affect the serialization algorithm.

On the wire != source compatibility.

> to keep proto files clear.

To the point of being useless?

popzxc: Okay, to make it clear, let's take a look at the extreme:
If some client X of the Exonum (let's make him a cash-bringing client to discard an option to ignore) will have an esoteric backend implemented in his personal langauge Y which requires a custom Y-specific metadata in the proto file, would we add this metadata (important *only* for that client) to the core repository (meant for the wide audience)?

popzxc: >To the point of being useless?
I don't think that sarcasm can be helpful in the discussion.

dmitry-timofeev: Let's consider your esoteric language when it becomes realistic.

The protobuf design requires the codegeneration options to be specified in the file. Whether @alekseysidorov likes it or not, that's the way it is designed — to have a single source of the data description and compatible generated code. If the core decides to use protobuf, then it has to follow its design principles and guidelines to make it usable. To (a) ensure on-the wire *and* source compatibility; (b) have easy and reliable configuration in the downstream projects, the source files must have options to enable codegeneration in the [target languages](https://github.com/protocolbuffers/protobuf/blob/5b232b8ecbce13286be09e703997e887ae0d464d/src/google/protobuf/timestamp.proto#L35-L41). Currently the only other target language is Java — hence the Java options.

PR #1508, opened 2019-10-21
Remove validator_id method from AfterCommitContext

popzxc:
## Overview

---

### Definition of Done

### DIFF
--- a/exonum/src/helpers/validator.rs
+++ b/exonum/src/helpers/validator.rs
@@ -1,0 +1,26 @@
+// Copyright 2019 The Exonum Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use exonum_crypto::PublicKey;
+use exonum_merkledb::Snapshot;
+
+use crate::{blockchain::Schema as CoreSchema, helpers::ValidatorId};
+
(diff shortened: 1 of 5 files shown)


exonum/src/helpers/validator.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use crate::{blockchain::Schema as CoreSchema, helpers::ValidatorId};
+
+/// Attempts to find a `ValidatorId` of current node.
+/// Returns `None` if node is an auditor.
+pub fn validator_id(snapshot: &dyn Snapshot, service_public_key: PublicKey) -> Option<ValidatorId> {
+    CoreSchema::new(snapshot)
+        .consensus_config()
+        .find_validator(|validator_keys| service_public_key == validator_keys.service_key)
+}
```
aleksuss: Do you think it's a good idea to add an extra module for one function only ? I'd add this function into `blockchain::Schema` e. g. Thoughts ?

popzxc: Hm. I just didn't want to make more dependencies of `blockchain::Schema` module. Here usage of `CoreSchema` is somewhat encapsulated.

If we will move it there, then we will obtain a reverse situation: importing core schema just to get the validator id.

However, if you think that it will be better, I can do it.

popzxc: And by the way I think it doesn't really matter, since we're planning to remove it completely anyway.

aleksuss: Ok. I've just thought the same. Let's stay as is.

exonum/src/helpers/validator.rs
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use exonum_crypto::PublicKey;
+use exonum_merkledb::Snapshot;
+
+use crate::{blockchain::Schema as CoreSchema, helpers::ValidatorId};
+
+/// Attempts to find a `ValidatorId` of current node.
```
dmitry-timofeev: Not of current, apparently.

PR #1509, opened 2019-10-21
Make dispatcher mostly syncrhonous

slowli:
As inspired by discussion in #1500, I've attempted to made a fully synchronous version of the dispatcher (i.e., one that does not process service starts, artifact deploys, etc. out of band).

Highlights:

- Making the dispatcher synchronous is achieved by encapsulating runtimes into reference-counted pointers. Since a dispatcher instance is encapsulated within a `Blockchain`, this means using `Arc<_>`, which in turn led to some additional `Sync` constraints (e.g., for `Runtime`) and, correspondingly, more complex data types. This shouldn't impact performance *too* much, since the access to the guarded vars is uncontested; but I didn't check the performance impact.
- I believe that it is possible (and desirable) to remove `Dispatcher` from `Blockchain`; `Blockchain` should be a cloneable pointer to shared blockchain resources (e.g., DB), while `Dispatcher`, like `Fork`s, should reside in a single consensus thread. This is currently blocked by `Dispatcher` being used to generate HTTP APIs for the services. I believe that it makes sense to make the HTTP API the implementation detail of the Rust runtime and inject service endpoints into it.
- Contexts for transaction execution and `before_commit` hook are now expressed with a single type, which also implements "child call" functionality. On creating a child context, we check that the `instance_id` we attempt to call exists, and return an error otherwise. I believe we can (and should) check the interface implementation likewise, but I didn't implement this so far.
- Config changes and `Configure` interface are fully moved out from the core crate into the supervisor crate. Dispatcher schema is not (yet) removed, but this looks like the primary target for further work.


cli/src/lib.rs
```diff
@@ -111,7 +111,7 @@ pub mod password;
 #[derive(Debug, Default)]
 pub struct NodeBuilder {
     services: Vec<Box<dyn ServiceFactory>>,
-    external_runtimes: Vec<(u32, Box<dyn Runtime>)>,
+    external_runtimes: Vec<(u32, Arc<dyn Runtime>)>,
```
slowli: It seems possible to retain `Box<dyn Runtime>` in most places since a universal `Box` -> `Arc` transform exists.

components/derive/src/exonum_service.rs
```diff
@@ -151,9 +151,7 @@ impl ExonumService {
                 .iter()
                 .map(|ServiceMethodDescriptor { arg_type, id, .. }| {
                     quote! {
-                        impl #cr::runtime::rust::Transaction for #arg_type {
-                            type Service = &'static dyn #trait_name;
```
slowli: Not sure what is the purpose of the associated type. I've changed it since in the supervisor, both the simple and advanced implementations use the same transaction type; in general, it seems counter-productive not to allow this.

components/derive/src/service_factory.rs
```diff
@@ -134,15 +134,17 @@ impl ServiceFactory {
...
-            )
+        if let Some(ref proto_sources_mod) = self.proto_sources {
+            quote! {
+                #cr::runtime::ArtifactProtobufSpec::from(
+                    #proto_sources_mod::PROTO_SOURCES.as_ref(),
+                )
+            }
+        } else {
+            quote! {
+                #cr::runtime::ArtifactProtobufSpec { sources: vec![] }
```
slowli: Not quite sure if this makes sense.

exonum/src/runtime/mod.rs
```diff
@@ -414,16 +420,16 @@ impl Caller {
...
     pub caller: Caller,
     /// Identifier of the service interface required for the call. Keep in mind that this field in
     /// fact is a part of unfinished "interfaces feature" and will be replaced in future releases.
     /// At the moment this field can only contains a core interfaces like `Configure` and
     /// always empty for the common the service interfaces.
     pub interface_name: &'a str,
-    /// Reference to the underlying runtime dispatcher.
-    dispatcher: &'a DispatcherRef<'a>,
+    /// Reference to the communication channel.
+    pub dispatcher: &'a mut Dispatcher,
```
popzxc: It's a bit strage how the idea of original PR was to encapsulate the dispatcher, and now it's even more public.

exonum/src/runtime/mod.rs
```diff
@@ -251,17 +246,22 @@ pub trait Runtime: Send + Debug + 'static {
...
     ///
-    /// * The order of services during invocation of this method must be the same for each node.
-    /// In other words, the order of the runtime services must be the same for each node.
+    /// - Calls are ordered by increasing `instance_id`.
+    /// - Each `before_commit` call is isolated with a separate checkpoint. A call that returns
+    ///   an error will be rolled back.
     ///
     /// # Policy on Panics
     ///
     /// * Catch each kind of panics except for `FatalError` and write
```
dmitry-timofeev: If the core becomes responsible for roll backs, I don't think this method shall catch panics.

exonum/src/runtime/mod.rs
```diff
@@ -251,17 +246,22 @@ pub trait Runtime: Send + Debug + 'static {
...
     /// them into the log.
     /// * If panic occurs, the runtime rolls back the changes in the fork.
```
dmitry-timofeev: Conflicts with the guarantees — the core rolls back, not the runtime.

exonum/src/runtime/mod.rs
```diff
@@ -251,17 +246,22 @@ pub trait Runtime: Send + Debug + 'static {
 
     /// Calls `before_commit` for all the services stored in the runtime.
```
dmitry-timofeev: The core calls this method for each service instance started in this runtime, doesn't it?

exonum/src/runtime/mod.rs
```diff
@@ -414,16 +420,16 @@ impl Caller {
 pub struct ExecutionContext<'a> {
     /// The current state of the blockchain. It includes the new, not-yet-committed, changes to
     /// the database made by the previous transactions already executed in this block.
-    pub fork: &'a Fork,
+    pub fork: &'a mut Fork,
```
popzxc: If I remember correctly, previously we needed `&mut Fork` because Rust runtime did manage `flush` and `rollback` calls in the `before_commit` method. Do we still need it for some reason?

exonum/src/runtime/mod.rs
```diff
@@ -432,15 +438,63 @@ impl<'a> ExecutionContext<'a> {
...
+            return Err((kind, msg).into());
+        }
+
+        let runtime = self.dispatcher.runtime_for_service(call_info.instance_id)?;
+        let reborrowed = ExecutionContext {
+            fork: &mut *self.fork,
+            caller: self.caller,
+            interface_name,
+            dispatcher: &mut *self.dispatcher,
+            call_stack_depth: self.call_stack_depth,
```
popzxc: Shouldn't `call_stack_depth` be increased?

services/supervisor/src/proto_structures.rs
```diff
@@ -72,6 +99,54 @@ pub struct ConfigPropose {
...
+
+    /// Creates a new proposal which activates at the specified height.
+    pub fn actual_from(height: Height) -> Self {
+        Self {
+            actual_from: height,
+            changes: Vec::default(),
+        }
+    }
+
+    pub fn consensus_config(mut self, config: ConsensusConfig) -> Self {
```
aleksuss: A method description is missing.

services/supervisor/src/proto_structures.rs
```diff
@@ -72,6 +99,54 @@ pub struct ConfigPropose {
...
+        self.changes.push(ConfigChange::Consensus(config));
+        self
+    }
+
+    pub fn service_config(mut self, instance_id: InstanceId, config: impl BinaryValue) -> Self {
```
aleksuss: Same here.

exonum/src/runtime/mod.rs
slowli: Yeah, I didn't update docs yet.

exonum/src/runtime/rust/service.rs
```diff
@@ -131,9 +130,7 @@ where
 }
 
 /// Transaction specification for a specific service interface.
-pub trait Transaction: BinaryValue {
-    /// Service interface associated with the given transaction.
-    type Service;
+pub trait Transaction<Svc: ?Sized>: BinaryValue {
```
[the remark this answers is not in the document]
slowli: Short for `Service`.

exonum/src/runtime/mod.rs
slowli: My oversight; this field can be made private right away. There is still a `&mut Dispatcher` ref in `Runtime::after_commit`, though.

exonum/src/runtime/mod.rs
slowli: Yes, it is necessary for supervisor extensions; specifically, for isolating application of config updates.

exonum/src/runtime/mod.rs
slowli: I think not; this code is a copy-paste of the `reborrow()` method (with the exception of overriden `interface_name`), which is sort of like `&mut *r` for `r: &mut T`, only for owned data. In other words, the only purpose of this code is to mutably reborrow the inner contents of the context (i.e., `fork` and `dispatcher`) for some time.

slowli:
@popzxc 

> It seems to me that with that requirement all the runtimes are sentenced to have RwLock<Inner> inside (both SampleRuntime and RustRuntime do it already) and be aware of possible deadlocks. Or am I wrong?

Yes, if a runtime mutates its content in response to some methods, then it needs to have interior mutability (e.g., via `Mutex` / `RwLock`). With the approach taken in this PR, there is no way around this (we need to have mutable borrow of `Dispatcher`, which contains all runtimes, while at the same time borrowing from an indefinite number of runtimes, some potentially several times). It could be possible to wrap runtimes in `RwLock` within the dispatcher, but this would incur overhead for *all* runtimes. And it wouldn't solve all problems anyway (e.g., services within the Rust runtime need to be similarly wrapped in smart pointers). I have not studied [`ReentrantMutex`](https://docs.rs/parking_lot/0.9.0/parking_lot/type.ReentrantMutex.html) yet; it could work (given that all processing is performed on a single thread) and simplify things.

Also, I think I've found a serious flaw in the current implementation, and both #1500 and this PR. What happens if there are two block proposals with conflicting service starts? (Deployments are processed within `after_commit`, which works, although currently there are no restrictions to attempt to start a deployment elsewhere.) Presently, changes are propagated to the runtimes right away (or, in #1500, after some time, but still before we know for certain that the block is committed), so such a situation will potentially break stuff. I believe we've discussed this a long time ago, but none of implementations allows for such a situation. I see at least 2 ways to fix this:

- As we've discussed, store information about "provisionally" started services in the fork, near the fork, or in the dispatcher / runtimes (provided that a fork has unique ID, which will be later provided in `after_commit`).
- This can be difficult, so an easy cop-out is to allow to start services (and, more generally, to mutate runtimes) in the `after_commit` hook only. This doesn't seem to bring significant problems for the current implementation of the supervisor, since the service start is sort of unpredictable anyway. The toy implementation is more affected, but since it's used for testing only, this doesn't look like a significant problem.

cc @alekseysidorov @aleksuss

popzxc:
>What happens if there are two block proposals with conflicting service starts? 

I wonder if it's possible to write a test case for such a scenario.

exonum/src/runtime/rust/call_context.rs
```diff
@@ -0,0 +1,198 @@
...
+                .dispatcher
+                .service_name(self.instance_id)
+                .unwrap(),
+        }
+    }
+
+    /// Returns the validator ID if the transaction author is a validator.
+    pub fn validator_id(&self) -> Option<ValidatorId> {
+        // TODO Perhaps we should optimize this method [ECR-3222]
+        self.caller().author().and_then(|author| {
```
vitvakatu: conflicts with https://github.com/exonum/exonum/pull/1508

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -213,143 +211,120 @@ impl Dispatcher {
...
     }
 
-    /// Call the corresponding runtime method.
+    /// Looks up a runtime by its identifier.
+    pub(crate) fn runtime_by_id(&self, id: u32) -> Option<&dyn Runtime> {
+        self.runtimes.get(&id).map(AsRef::as_ref)
+    }
+
+    /// Looks up the runtime for the specified service instance. Returns a reference to
+    /// the runtime, or an error if the service with the sepcified instance ID does not exist.
```
aleksuss:
```suggestion
    /// the runtime, or an error if the service with the specified instance ID does not exist.
```

exonum/src/runtime/mod.rs
```diff
@@ -198,37 +198,42 @@ pub trait Runtime: Send + Debug + 'static {
...
+    /// may freely discard the instantiated service instance after completing this method. To
+    /// add the service to the runtime permanently, `add_service()` is called.
     ///
     /// # Policy on Panics
     ///
     /// * Catch each kind of panics except for `FatalError` and convert
     /// them into `ExecutionError`.
-    /// * If panic occurs, the runtime must ensure that it is in a consistent state.
-    fn restart_service(&mut self, spec: &InstanceSpec) -> Result<(), ExecutionError>;
+    fn start_adding_service(
```
vitvakatu: `start_adding_service` doesn't seem like a really good name for the method. I'd suggest `try_adding_service`.

exonum/src/runtime/mod.rs
```diff
@@ -198,37 +198,42 @@ pub trait Runtime: Send + Debug + 'static {
     /// * Ensure that the deployed artifact has the following information, even if it is empty.
     fn artifact_protobuf_spec(&self, id: &ArtifactId) -> Option<ArtifactProtobufSpec>;
 
-    /// Restart previously added service instance with the given specification.
+    /// Runs the constructor of a new service instance with the given specification
+    /// and initial configuration.
+    ///
+    /// The service is not guaranteed to be added to the runtime at this point, so the runtime
+    /// may freely discard the instantiated service instance after completing this method. To
```
vitvakatu: Why `may` is used here? Is implementation of Runtime possible, which performs service instantiation in `start_adding_service` and does nothing in `start_service`?

vitvakatu: What instantiation of the service leads to changes in the database state? Should these changes be reverted in case of error pr panic?

exonum/src/runtime/mod.rs
```diff
@@ -198,37 +198,42 @@ pub trait Runtime: Send + Debug + 'static {
...
+    /// Permanently adds a service to the runtime.
     ///
-    /// The configuration parameters passed to the method are discarded immediately.
-    /// So the service instance should save them by itself if it is important for
-    /// the service business logic.
+    /// It is guaranteed that `start_adding_service()` was called with the same `spec` earlier
+    /// and returned `Ok(())`. Note that this previous call may have happened indefinite time ago;
+    /// indeed, the method is called for all services on the node startup.
     ///
     /// # Policy on Panics
```
vitvakatu: Not really a policy on panics, but rather on errors. Should runtime implementation catch all panics or will Core do it itself?

exonum/src/runtime/mod.rs
```diff
@@ -238,40 +243,45 @@ pub trait Runtime: Send + Debug + 'static {
...
-    ///
-    /// # Policy on Panics
-    ///
-    /// * Catch each kind of panics except for `FatalError` and write
-    /// them into the log.
-    /// * If panic occurs, the runtime rolls back the changes in the fork.
-    fn before_commit(&self, dispatcher: &DispatcherRef, fork: &mut Fork);
+    /// - Each `before_commit` call is isolated with a separate checkpoint. A call that returns
+    ///   an error will be rolled back.
+    /// - Calls are globally ordered by increasing `instance_id`. This should be considered
```
vitvakatu: Why should we mention this implementation detail in the docs?

exonum/src/runtime/mod.rs
slowli: IMO, `try_adding_service` implies that adding is complete if the "normal" result is returned. This is not the case here; a call to `start_adding_service()` is not necessarily followed by adding the service "permanently" to the runtime.

services/supervisor/src/simple/mod.rs
```diff
@@ -14,33 +14,31 @@
...
 #[exonum(
-    proto_sources = "proto::schema",
-    artifact_name = "exonum-testkit-supervisor",
-    implements("Transactions")
+    proto_sources = "crate::proto",
+    artifact_name = "simple-supervisor",
+    implements("SupervisorInterface")
 )]
-pub struct SimpleSupervisor;
+pub struct Supervisor;
```
popzxc: Do we actually need this to be a part of `supervisor`? I thought it only exists for test purposes.

exonum/src/runtime/rust/call_context.rs
```diff
@@ -0,0 +1,200 @@
...
+        self.inner.fork
+    }
+
+    /// Returns the initiator of the actual transaction execution.
+    pub fn caller(&self) -> &Caller {
+        &self.inner.caller
+    }
+
+    /// Accesses dispatcher information.
+    pub fn dispatcher(&self) -> Schema<&Fork> {
```
popzxc: Since the return value is not `Dispatcher`, I guess `dispatcher_info` (or something similar) will be more suitable name.

services/supervisor/src/simple/mod.rs
popzxc: By the way I don't really like this rename, since now we have to remember current directory to understand that it's not a main supervisor, but a simple one. IMO it's a bit confusing.

exonum/src/runtime/mod.rs
```diff
@@ -174,9 +172,34 @@ impl From<RuntimeIdentifier> for u32 {
...
+    /// called *exactly once* during `Runtime` lifetime.
+    ///
+    /// The default implementation does nothing.
+    fn initialize(&mut self, blockchain: &Blockchain) {}
+
+    /// Notifies the runtime that the dispatcher has completed re-initialization after
+    /// node restart, including restoring the deployed artifacts / started service instances
+    /// for all runtimes.
+    ///
+    /// This method is called *no more than once* during `Runtime` lifetime. It is called iff
```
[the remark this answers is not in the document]
slowli: "iff" means "if and only if", just shorter.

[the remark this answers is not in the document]
vitvakatu: It's not a typo, it means `if and only if`

exonum/src/runtime/rust/call_context.rs
```diff
@@ -0,0 +1,200 @@
...
+        F: Fn(&Caller) -> Option<T>,
+    {
+        // TODO Think about returning structure with the named fields instead of unnamed tuple
+        // to make code more clear. [ECR-3222]
+        predicate(&self.inner.caller).map(|result| (result, &*self.inner.fork))
+    }
+
+    /// Isolates execution of the provided closure.
+    ///
+    /// This method should be used with extreme care, since it subverts the usual rules
```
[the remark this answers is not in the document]
slowli: I wholeheartedly agree; the only saving grace is that this is internal / hidden method not intended for general use. I *think* the version in which `fork.flush()` is called before the provided closure is more understandable, but even it is counterintuitive w.r.t. changes performed before and after the calls. I wonder if deferred call execution a la Golang would make more sense; I'm willing to prototype it, but I didn't have the time yet.

exonum/src/runtime/rust/call_context.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// of transaction roll-back provided by Exonum. Namely:
+    ///
+    /// - If the execution of the closure is successful, all changes to the blockchain state
```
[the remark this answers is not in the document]
slowli: This feature is also known as multi-layered (fork) checkpoints. We've discussed it previously (since it makes sense for inter-service communication), but no implementation thus far has appeared 🤷‍♂

services/supervisor/src/lib.rs
```diff
@@ -34,19 +29,72 @@ use exonum::{
...
+
+        ConfigChange::Service(config) => {
+            log::trace!(
+                "Updating service instance configuration, instance ID is {}",
+                config.instance_id
+            );
+
+            let configure_result = context.isolate(|mut context| {
+                context
+                    .interface::<ConfigureCall>(config.instance_id)?
```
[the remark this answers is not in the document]
slowli: > every service

Every service *that wants to be configured via supervisor*, yes. I don't see why this is a problem, but it would be easy to move `Configure`-related stuff to a separate crate.

exonum/src/runtime/mod.rs
```diff
@@ -436,15 +476,89 @@ impl<'a> ExecutionContext<'a> {
...
+    /// (i.e., does not process transactions or the `before_commit` hook)
+    /// until the block built on top of the provided `fork` is committed.
+    ///
+    /// This method should be called for the exact context passed to the runtime.
+    pub(crate) fn start_adding_service(
+        &mut self,
+        spec: InstanceSpec,
+        constructor: impl BinaryValue,
+    ) -> Result<(), ExecutionError> {
+        debug_assert!(spec.validate().is_ok(), "{:?}", spec.validate());
```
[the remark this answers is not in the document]
slowli: This part was already there. It looks weird to me as well, but I didn't investigate the reasoning behind it (yet?).

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -30,93 +32,177 @@ impl<T: IndexAccess> Schema<T> {
...
+            || self.pending_service_instances().contains(&spec.name)
+        {
             return Err(Error::ServiceNameExists);
         }
         // Checks that instance identifier doesn't exist.
-        if self.service_instance_ids().contains(&spec.id) {
+        if self.service_instance_ids().contains(&spec.id)
+            || self.pending_instance_ids().contains(&spec.id)
+        {
             return Err(Error::ServiceIdExists);
```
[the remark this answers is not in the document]
slowli: This check was already there (with the exception of pending instances), so I didn't change it. I'm not 100% sure that dispatcher should be responsible for assigning IDs (note that all other schema functionality is there purely for persistence across node restarts and between block execution and commitment). Thus, I'd leave this check for now.

exonum/src/runtime/rust/mod.rs
```diff
@@ -341,64 +366,56 @@ impl Runtime for RustRuntime {
...
-        &self,
-        dispatcher: &DispatcherSender,
-        snapshot: &dyn Snapshot,
-        service_keypair: &(PublicKey, SecretKey),
-        tx_sender: &ApiSender,
-    ) {
+    fn after_commit(&mut self, snapshot: &dyn Snapshot, mailbox: &mut Mailbox) {
+        self.on_resume();
+
+        // By convention, services don't handle `after_commit()` on the genesis block.
```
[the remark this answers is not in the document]
popzxc: Yeah, makes sense.

services/supervisor/src/lib.rs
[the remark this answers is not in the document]
popzxc: >I don't see why this is a problem
I think it can become a problem on a cross-runtime border: e.g. configurable Java service with a Rust supervisor.

So yeah, IMHO it's better to keep the interfaces separate from the implementation.

services/supervisor/src/simple/mod.rs
slowli: I've moved it from testkit as to not create a corresponding dependency (I think it's not cyclical, since testkit is a *dev* dependency of the supervisor, but I didn't check). Would you say that if I implemented more "real-world" auth (e.g., by one of validators), `SimpleSupervisor` would make more sense outside of tests?

exonum/src/runtime/rust/mod.rs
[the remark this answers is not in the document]
slowli: BTW, it seems possible to call `after_commit` for Rust services on genesis block safely. [As coded now](https://github.com/exonum/exonum/blob/ddf55717fa4d94840750fba6f40787b60f0c6b41/exonum/src/blockchain/mod.rs#L227-L236), `after_commit` on the genesis block is called without merging information about the block, which may lead to a panic, e.g., if the node accesses information about blockchain height from the core schema. Now that I think about it, it's a pretty weird choice.

exonum/src/runtime/types.rs
```diff
@@ -316,24 +323,32 @@ impl Display for InstanceSpec {
...
 
-/// Configuration parameters of the certain service instance.
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, ProtobufConvert)]
-#[protobuf_convert(source = "schema::runtime::ServiceConfig")]
-pub struct ServiceConfig {
-    /// Corresponding service instance ID.
-    pub instance_id: InstanceId,
-    /// Raw bytes representation of service configuration parameters.
-    pub params: Vec<u8>,
+/// Allows to query a service instance by either of the two identifiers.
```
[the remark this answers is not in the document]
slowli: If you're talking about [this TODO](https://github.com/exonum/exonum/blob/ddf55717fa4d94840750fba6f40787b60f0c6b41/exonum/src/runtime/dispatcher/mod.rs#L361), it's quite easy to fix by introducing the corresponding lookup table into `Dispatcher`. So I'd regard speed an implementation detail at this point.

services/supervisor/src/simple/mod.rs
popzxc: Yeah, I think so.

services/supervisor/src/lib.rs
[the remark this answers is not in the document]
slowli: Can we defer this question for later?

services/supervisor/src/lib.rs
```diff
@@ -94,70 +138,63 @@ impl Service for Supervisor {
...
                         "New configuration has been accepted: {:?}",
                         entry.config_propose
                     );
                     // Perform the application of configs.
-                    context.update_config(entry.config_propose.changes);
+                    update_configs(&mut context, entry.config_propose.changes);
+
                     // Remove config from proposals.
+                    let schema = Schema::new(context.instance().name, context.fork());
                     schema.pending_proposal().remove();
```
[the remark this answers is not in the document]
slowli: Yes, it is. This is checked in `test_service_config_discard_single_apply_error` / `test_service_config_discard_single_apply_panic` in `tests/config.rs`. If you comment out this line, these tests fail (well, actually they don't right now since they use too permissive blockchain height, but it's trivial to fix).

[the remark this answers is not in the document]
popzxc: Hm. If I understand them correctly, they're checking applying the *service* config. And I'm talking about [`assert` in `update_configs`](https://github.com/exonum/exonum/pull/1509/files#diff-7a5cd928f4e93e0b1eceb23242deba18R74), which corresponds to the *consensus* config update.

services/supervisor/src/lib.rs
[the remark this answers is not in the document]
popzxc: Sure.

exonum/src/blockchain/mod.rs
```diff
@@ -234,7 +234,10 @@ impl BlockchainMut {
                 instance_config.constructor,
             )?;
         }
-        self.dispatcher.after_commit(&fork);
+        // We need to activate services before calling `create_patch()`; unlike all other blocks,
+        // initial services are considered immediately active in the genesis block, i.e.,
+        // their state should be included into `patch` created below.
+        self.dispatcher.commit_block(&fork);
```
slowli: IMO, this split is sort of cringe, but trying to apply usual rules (services are activated _after_ the block is committed) leads to some problems:

- Lots of broken tests (e.g., in the sandbox)
- More importantly, with this change the hash of the genesis block would not depend on the included services! This brings us back to the genesis block having real transactions (as in: transactions to deploy some artifacts and initialize some services).

exonum/src/runtime/mod.rs
```diff
@@ -253,10 +253,22 @@ pub trait Runtime: Send + fmt::Debug + 'static {
     /// and returned `Ok(())`. The results of the call (i.e., changes to the blockchain state)
     /// are guaranteed to be persisted from the call.
     ///
-    /// Note that a call to `start_adding_service()` may have happened indefinite time ago;
+    /// A call to `start_adding_service()` may have happened indefinite time ago;
     /// indeed, `commit_service()` is called for all services on the node startup. Likewise,
     /// `commit_service()` may be called an indefinite number of times for the same instance.
     ///
+    /// `snapshot` is the storage snapshot at the latest height when the method is called:
```
slowli: This level of complexity is not quite comfortable, so it could make sense to remove the `snapshot` arg altogether (it isn't used by the Rust runtime). OTOH, I can imagine use cases where `snapshot` is useful (e.g., a potential "on node start" hook for Rust services).

examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -82,21 +82,21 @@ pub struct CreateWallet {
 #[exonum_service]
 pub trait CryptocurrencyInterface {
     /// Transfers `amount` of the currency from one wallet to another.
-    fn transfer(&self, ctx: TransactionContext, arg: Transfer) -> Result<(), Error>;
+    fn transfer(&self, ctx: CallContext, arg: Transfer) -> Result<(), Error>;
```
slowli: `CallContext` / `ExecutionContext` can use `CallContext<'_>` syntax as per 2018 edition guidelines.

examples/sample_runtime/src/main.rs
```diff
@@ -176,37 +182,32 @@ impl Runtime for SampleRuntime {
...
+        _context: ExecutionContext,
+        _id: InstanceId,
+    ) -> Result<(), ExecutionError> {
+        Ok(())
     }
+
+    fn after_commit(&mut self, _snapshot: &dyn Snapshot, _mailbox: &mut Mailbox) {}
 }
 
 impl From<SampleRuntime> for (u32, Box<dyn Runtime>) {
```
slowli: IMO, this conversion looks too generic / non-obvious. It could make sense to introduce a special-purpose trait like

```rust
trait RuntimeWithId: Runtime {
    const ID: u32;
}
```

exonum/src/api/manager.rs
```diff
@@ -100,35 +102,69 @@ impl ApiManager {
...
+        self.api_runtime_addresses.insert(addr, msg.config);
     }
 }
 
-impl Message for StartServer {
-    type Result = ();
+/// Updates user-provided endpoints, restarting all HTTP service managed by the addressed
+/// `ApiManager`.
+#[derive(Debug, Clone)]
+pub struct UpdateEndpoints {
```
slowli: In the case there are several producers of endpoints (e.g., several runtimes), this struct needs to contain the ID of the changed part in order not to erase parts corresponding to other producers. Seems easy to implement, but I'm not sure whether this is necessary right now (we don't have other producers besides `RustRuntime`).

exonum/src/api/manager.rs
```diff
@@ -100,35 +102,69 @@ impl ApiManager {
...
+        info!(
+            "Handling server start: {:?} (attempt #{})",
+            msg.config,
+            msg.attempt + 1
+        );
+        let addr = match self.start_server(msg.config.clone()) {
+            Ok(addr) => addr,
+            Err(e) => {
+                warn!("Error handling service start {:?}: {}", msg.config, e);
+                if msg.attempt == Self::RETRY_ATTEMPTS {
```
slowli: `Self::RETRY_ATTEMPTS` and `Self::RETRY_INTERVAL` could be parameterized.

exonum/src/blockchain/schema.rs
```diff
@@ -367,7 +364,6 @@ impl IndexOwner {
 #[repr(u16)]
 enum IndexTag {
     Core = 0,
-    Dispatcher = 1,
```
slowli: Some cleanup should be welcome. In general, this API looks too difficult; I'd try to simplify it.

exonum/src/node/mod.rs
```diff
@@ -1075,23 +1089,12 @@ impl Node {
...
-                    Event::Internal(InternalEvent::RestartApi) => self.system_runtime.restart(),
-                    event => self.inner.handle_event(event),
-                }
-            }
-        }
-
         let connect_message = self.state().our_connect_message().clone();
         let connect_list = self.state().connect_list().clone();
+
+        self.api_runtime_config
```
slowli: Restarting HTTP servers should be tested. I've run into some problems with server restarts on my laptop with Windows 10; not sure whether this is Windows problem or something more generic.

exonum/src/runtime/mod.rs
```diff
@@ -375,11 +414,15 @@ pub enum Caller {
...
     },
+
     /// Method is invoked during the method execution of a different service.
     Service {
         /// Identifier of the service instance which invoked this method.
         instance_id: InstanceId,
     },
+
+    /// Method is invoked by the `before_commit` hook.
+    BeforeCommit,
```
slowli: It could make sense to generalize this to `Blockchain` (as proposed by @alekseysidorov), or to split into `BeforeCommit`, `GenesisBlock`, etc.

exonum/src/runtime/rust/call_context.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /// Returns the initiator of the actual transaction execution.
+    pub fn caller(&self) -> &Caller {
+        &self.inner.caller
+    }
+
+    /// Accesses dispatcher information.
+    pub fn dispatcher(&self) -> Schema<&Fork> {
+        Schema::new(self.fork())
```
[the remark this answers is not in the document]
slowli: Tried `self.fork().as_ref()`; it doesn't work because `Fork::as_ref()` does not capture non-flushed changes in the fork. I think this is a problem with MerkleDB APIs, specifically, with lack of split between read-only methods and full-access methods. I believe we've already discussed this design flaw.

> no service will be able to screw the dispatcher schema up?

It will be possible as long as the schema constructor is public, *or* if there is no split of read-only / full access methods of the indexes / schemas as described above. It could make sense to restrict write access to the schema of the service being executed to alleviate similar concerns once and for all, but it is a separate task.

exonum/src/runtime/rust/call_context.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub fn verify_caller<F, T>(&self, predicate: F) -> Option<(T, &Fork)>
+    where
+        F: Fn(&Caller) -> Option<T>,
+    {
+        // TODO Think about returning structure with the named fields instead of unnamed tuple
+        // to make code more clear. [ECR-3222]
+        predicate(&self.inner.caller).map(|result| (result, &*self.inner.fork))
+    }
+
+    /// Isolates execution of the provided closure.
```
slowli: Should be tested. Also, it should be decided if this method calls `fork.flush()` in the beginning or not. In the (slightly) longer term, a feasibility study / spec / PoC regarding deferred execution could make sense.

testkit/src/lib.rs
```diff
@@ -657,23 +682,25 @@ impl TestKit {
 
     fn run(mut self, public_api_address: SocketAddr, private_api_address: SocketAddr) {
```
slowli: Not tested.

examples/sample_runtime/src/main.rs
popzxc: Same for `impl From<ServiceName> for InstanceCollection`.

exonum/src/node/mod.rs
popzxc: We should also consider splitting "Core API" and "Service API" into several independent entities. I've [somewhat tested](https://github.com/popzxc/exonum-python-backend/blob/master/python_runtime/exonum_runtime/runtime/runtime.py#L113) this approach in the Python runtime MVP and it seems to work just fine.

With that approach we will not have to restart HTTP server on each instance added.

exonum/src/api/backends/actix.rs
```diff
@@ -34,15 +35,14 @@ use std::{
...
     thread::{self, JoinHandle},
 };
 
+use crate::api::manager::UpdateEndpoints;
 use crate::api::{
-    error::Error as ApiError,
-    manager::{ApiManager, RestartServer},
-    ApiAccess, ApiAggregator, ApiBackend, ApiScope, ExtendApiBackend, FutureResult, Immutable,
-    Mutable, NamedWith, Result,
+    self, manager::ApiManager, ApiAccess, ApiAggregator, ApiBackend, ApiScope, ExtendApiBackend,
```
aleksuss:
```suggestion
    self, manager::{ApiManager, UpdateEndpoints}, ApiAccess, ApiAggregator, ApiBackend, ApiScope, ExtendApiBackend,
```
This change allows to get rid of 42 line.

exonum/src/runtime/dispatcher/schema.rs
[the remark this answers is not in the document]
alekseysidorov: Old implementation was not be consistent in this question. And I think that we should solve this issue neat in the separate PR.
Maybe we should just add TODOs for inconsistent pieces of code like this.

exonum/src/runtime/mod.rs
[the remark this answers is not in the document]
alekseysidorov: Cool idiom, but how common is it? And should it be written in capital letters?

exonum/src/runtime/mod.rs
```diff
@@ -238,62 +293,49 @@ pub trait Runtime: Send + Debug + 'static {
...
+    /// # Guarantees
     ///
-    /// * Catch each kind of panics except for `FatalError` and write
-    /// them into the log.
-    fn after_commit(
+    /// - Each `before_commit` call is isolated with a separate checkpoint. A call that returns
+    ///   an error will be rolled back.
+    /// - Ordering of calls among service instances is not specified, but is guaranteed
+    ///   to be the same for all nodes.
+    fn before_commit(
```
alekseysidorov: I think that `on_resume` and `before_commit`, `after_commit` sounds a little bit inconsistent together.

exonum/src/runtime/mod.rs
```diff
@@ -303,15 +345,12 @@ pub trait Runtime: Send + Debug + 'static {
     /// Invoking of this callback is guaranteed to be the last operation for the runtime.
     /// Since this method is a part of shutdown process, runtimes can perform blocking and
     /// heavy operations here if needed.
-    fn shutdown(&self) {}
+    fn shutdown(&mut self) {}
```
alekseysidorov: Maybe `on_shutdown` is more good name in according of `on_resume`?

exonum/src/runtime/mod.rs
alekseysidorov: imo: If we split this options, it will sounds like a provocation to consider these cases as completely different.

exonum/src/runtime/mod.rs
```diff
@@ -410,6 +453,17 @@ impl Caller {
             None
         }
     }
+
+    /// Verify that the caller of this method is supervisor service.
```
alekseysidorov: Do we have an explanation in the documentation that only supervisor service has `0` identifier? And thus supervisor is the special kind of services.
Also marker trait for supervisor from #1500 is better idea imo.

exonum/src/runtime/rust/call_context.rs
alekseysidorov: Can you add TODO comment with this description?

exonum/src/runtime/rust/call_context.rs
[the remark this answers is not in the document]
alekseysidorov: Yeah, the public method that should be used with extreme care is scary.

exonum/src/runtime/rust/call_context.rs
slowli: I'll add Jira tasks for this and other unresolved questions / tech debts.

exonum/benches/criterion/block.rs
```diff
@@ -88,18 +88,15 @@ fn create_blockchain(
...
-        db,
-        external_runtimes,
-        services,
-        genesis_config,
-        service_keypair,
-        ApiSender::new(mpsc::channel(0).0),
-        mpsc::channel(0).0,
-    )
+    let api_sender = ApiSender::new(mpsc::channel(0).0);
+    Blockchain::new(db, service_keypair, api_sender)
```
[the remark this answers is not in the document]
slowli: > Only `BlockchainBuilder` should create an instance of `BlockchainMut`.

Isn't this the case right now? Or would you like to replace `Blockchain::into_mut()` with `BlockchainBuilder::new(Blockchain, ConsensusConfig)`? I'm not particularly against this, although I think that the current interface looks good enough.

exonum/src/node/mod.rs
```diff
@@ -793,6 +794,11 @@ impl ApiSender {
         ApiSender(inner)
     }
 
+    /// Creates a dummy sender which cannot send messages.
+    pub fn closed() -> Self {
```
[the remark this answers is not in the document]
slowli: > Is this method useful outside of testing purposes?

No, probably not; but the purpose of the method is clearly stated in the doc comment.

> May you mark it as `cfg(test)`?

It is used in `Blockchain::build_for_tests()`, which is not `cfg(test)` since it has uses outside of the crate (e.g., in the `explorer` example / integration test). Do you think this method pollutes the public interface of `ApiSender`?

exonum/src/runtime/mod.rs
slowli: The problem with any client-enforced method (like forcing the service to implement a marker trait) is that it's trivial to circumvent and therefore difficult to reason about. I'm for this method provided that the dispatcher knows the correspondence between services and implemented interfaces (which seems a good idea in general) and has a marker interface for dispatcher (more generally, for privileged services).

exonum/benches/criterion/block.rs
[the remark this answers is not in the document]
slowli: Also, consider that we may want to split logic of `BlockchainBuilder` into building runtimes and building the genesis block; see ECR-3750.

exonum/src/runtime/mod.rs
slowli: IMO, the naming is different on purpose, although I'm not 100% sure. When we run `Runtime::shutdown`, we *command* the runtime to shut down; there is no implied time dependencies between `shutdown` and other commands (other than that it runs after all other commands). When `Runtime::on_resume` is called, this has first and foremost *signalling* purpose; `Dispatcher` signals to the runtime that it has done restoring the state. It is not implied (at least currently) that the runtime should resume its state within the call.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -213,150 +218,125 @@ impl Dispatcher {
...
+    /// This is not checked by the consensus algorithm as usual.
+    pub(crate) fn commit_block(&mut self, fork: &mut Fork) {
+        // If the fork is dirty, `snapshot` will be outdated, which can trip
+        // `Runtime::start_service()` calls.
+        fork.flush();
+        let snapshot = fork.as_ref();
+        let mut schema = Schema::new(&*fork);
+
+        // Deploy pending artifacts.
+        let mut artifacts = schema.pending_artifacts();
```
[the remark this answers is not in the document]
slowli: Maybe, although I'm not quite sure what should happen if a `drain()` iterator is dropped before clearing all elements. AFAIK, the standard library admits memory leaks in this case 🙃 Also, consider that removing elements one by one may occupy much more in `Fork` than a `clear` instruction.

exonum/benches/criterion/block.rs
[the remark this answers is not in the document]
alekseysidorov: Yeah proposal from task looks better than my, so I think that the current code is good enough.

PR #1510, opened 2019-10-23
Fix serde imports in exonum-cli

popzxc:
A big question: how the hell CI was green? O_o


vitvakatu:
Well, don't know why CI is green, but initially it was okay:
https://github.com/exonum/exonum/pull/1496/files#diff-474f9d160624516e523edbad6af0815fL18

PR #1511, opened 2019-10-23
Add includes to proto-sources

pavel-mukhanov:
TBD


components/build/src/lib.rs
```diff
@@ -147,7 +149,16 @@ fn generate_mod_rs<P: AsRef<Path>, Q: AsRef<Path>>(
...
         })
     };
-    let proto_files = include_proto_files(proto_files);
+
+    // To avoid cases where input sources are also added as includes, use only
+    // unique paths.
+    let proto_files: Vec<PathBuf> = [proto_files, includes]
+        .concat()
+        .into_iter()
+        .unique()
```
popzxc: I think you can change signature of `include_proto_files` to `IntoIter<..>` and use a `HashSet` here (so we can get rid of `itertools` dependency). `unique()` [uses a set with clones of added items](https://docs.rs/itertools/0.8.0/itertools/trait.Itertools.html#method.unique) inside anyway.

aleksuss: Agree with @popzxc. One more extra dependency for a one method. It's not worth it. Let's do it with `HashSet`.

components/build/src/lib.rs
```diff
@@ -152,13 +152,9 @@ fn generate_mod_rs<P: AsRef<Path>, Q: AsRef<Path>>(
 
     // To avoid cases where input sources are also added as includes, use only
     // unique paths.
-    let proto_files: Vec<PathBuf> = [proto_files, includes]
-        .concat()
-        .into_iter()
-        .unique()
-        .collect();
+    let proto_files: HashSet<PathBuf> = [proto_files, includes].concat().into_iter().collect();
```
popzxc: Also btw you can use `proto_files.iter().chain(includes.iter()).collect()` to avoid allocation for concatenated slice.

exonum-dictionary.txt
```diff
@@ -64,6 +64,7 @@ hasher
 healthcheck
 idempotence
 interlayer
+itertools
```
vitvakatu: nit: no more needed?

CHANGELOG.md
```diff
@@ -37,6 +37,15 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   - Removed obsolete `TestKit::blockchain_mut` method and `TestKit::blockchain`
   now returns value instead of reference.
 
+- `system/v1/proto-sources` endpoint has been divided into two endpoints. (#1511)
+
+  - `system/v1/proto-sources/core` - returns core proto files.
+
+  - `system/v1/proto-sources/service` - returns services specific proto files.
+
+  - Endpoint format is also has been changed and now it returns include files together
```
vitvakatu:
```suggestion
  - Endpoint format has been changed and now it returns include files together
```

CHANGELOG.md
```diff
@@ -37,6 +37,15 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   - Removed obsolete `TestKit::blockchain_mut` method and `TestKit::blockchain`
   now returns value instead of reference.
 
+- `system/v1/proto-sources` endpoint has been divided into two endpoints. (#1511)
```
popzxc: I think placeholder should be kept the most-bottom one entry until it's replaced with actual changes description.

exonum/src/runtime/mod.rs
```diff
@@ -341,13 +341,17 @@ pub struct ArtifactProtobufSpec {
     ///
     /// The common interface entry point is always in the `service.proto` file.
     pub sources: Vec<ProtoSourceFile>,
+    /// List of service's proto include files.
```
vitvakatu: What is the difference between include files and sources? E.g. I have `service.proto`, `my-awesome-helpers.proto` and `my-data-types.proto`. `my-awesome-helpers.proto` is used both in `service.proto` and `my-data-types.proto`, while `service.proto` also includes `my-data-types.proto`. What are includes and what are sources here?
How the user should communicate this to the runtime?

exonum/src/runtime/mod.rs
```diff
@@ -341,13 +341,17 @@ pub struct ArtifactProtobufSpec {
...
+    pub includes: Vec<ProtoSourceFile>,
 }
 
-impl From<&[(&str, &str)]> for ArtifactProtobufSpec {
-    fn from(sources_strings: &[(&str, &str)]) -> Self {
-        let sources = sources_strings.iter().map(From::from).collect();
+#[allow(clippy::type_complexity)]
```
popzxc: Maybe not allow, but create a type alias for `(&str, &str)`?

exonum/src/runtime/mod.rs
vitvakatu: nit: documentation of `sources` field is invalid, as it mentions `tuple`

pavel-mukhanov: The difference is important when you compile proto files. To complie service.proto you need to
provide `service.proto` itself and two include files `awesome-helpers.proto` and `my-data-types.proto`.
> How the user should communicate this to the runtime?
`ProtobufGenerate` will create appropriate constants in service's generated proto_mod.rs

exonum/src/api/node/public/system.rs
```diff
@@ -188,7 +216,22 @@ impl SystemApi {
...
             .handle_list_services_info("v1/services", api_scope)
-            .handle_proto_source("v1/proto-sources", api_scope);
+            .handle_proto_source_core("v1/proto-sources/core", api_scope)
+            .handle_proto_source_service("v1/proto-sources/service", api_scope);
         api_scope
     }
 }
+
+fn filter_exonum_sources(files: Vec<ProtoSourceFile>) -> Vec<ProtoSourceFile> {
+    let sources = EXONUM_PROTO_SOURCES
```
popzxc: Do I understand it correctly that `EXONUM_INCLUDES` will not be filtered?

pavel-mukhanov: They filtered during the build phase.

popzxc: Well... That's kinda not obvious. Includes for exonum are filtered from service includes during build phase, but exonum files itself are not filtered. Do I understand that correctly?

CHANGELOG.md
```diff
@@ -30,13 +30,15 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
     ```
 
-- Placeholder for changes due to dynamic services (#9999)
-
   - `StoppedTestKit::resume` accepts list of runtimes instead of a list of services.
 
   - Removed obsolete `TestKit::blockchain_mut` method and `TestKit::blockchain`
   now returns value instead of reference.
 
+- `system/v1/proto-sources` now returns all required proto files. (#1511)
```
popzxc: I'm not sure however if we need this. We're restored old behavior, aren't we?

PR #1514, opened 2019-10-25
Use enum to represent ErrorKind [ECR-3717]

ozkriff:
## Overview

Use an enum in the proto files to avoid the manual bookkeeping of constants in java/python.

---

See: https://jira.bf.local/browse/ECR-3717

### DIFF
--- a/exonum/src/proto/schema/exonum/runtime.proto
+++ b/exonum/src/proto/schema/exonum/runtime.proto
@@ -44,10 +44,17 @@ message InstanceSpec {
   ArtifactId artifact = 3;
 }
 
+enum ErrorKind {
+  PANIC = 0;
+  DISPATCHER = 1;
+  RUNTIME = 2;
+  SERVICE = 3;
+}
+
 /// Result of unsuccessful runtime execution.
 message ExecutionError {
   /// The kind of error that indicates in which module the error occurred.
-  uint32 kind = 1;
+  ErrorKind kind = 1;
   /// User defined error code that can have different meanings for the different
   /// error kinds.
   uint32 code = 2;
--- a/exonum/src/proto/schema/exonum/runtime.proto
+++ b/exonum/src/proto/schema/exonum/runtime.proto
@@ -44,7 +44,7 @@ message InstanceSpec {
   ArtifactId artifact = 3;
 }
 
-/// The kind of ExecutionError.
+// The kind of ExecutionError.
 enum ErrorKind {
   PANIC = 0;
   DISPATCHER = 1;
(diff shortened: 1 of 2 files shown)


exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -44,10 +44,17 @@ message InstanceSpec {
   ArtifactId artifact = 3;
 }
 
+enum ErrorKind {
```
aleksuss: Add description please.

exonum/src/runtime/error.rs
```diff
@@ -74,25 +74,24 @@ impl ErrorKind {
         ErrorKind::Service { code: code.into() }
     }
 
-    fn into_raw(self) -> (u8, u8) {
+    fn into_raw(self) -> (runtime::ErrorKind, u8) {
```
popzxc: Maybe change it to `runtime_proto` to explicitly mark that it's not the `runtime` mod in which `error.rs` lies in?

exonum/src/runtime/error.rs
```diff
@@ -74,25 +74,24 @@ impl ErrorKind {
...
+            ErrorKind::Panic => (runtime::ErrorKind::PANIC, 0),
+            ErrorKind::Dispatcher { code } => (runtime::ErrorKind::DISPATCHER, code),
+            ErrorKind::Runtime { code } => (runtime::ErrorKind::RUNTIME, code),
+            ErrorKind::Service { code } => (runtime::ErrorKind::SERVICE, code),
         }
     }
 
-    fn from_raw(kind: u8, code: u8) -> Result<Self, failure::Error> {
+    fn from_raw(kind: runtime::ErrorKind, code: u8) -> Result<Self, failure::Error> {
         match kind {
```
popzxc: I guess now we can write
```rust
let kind = match raw_kind { .. };
Ok(kind)
```

exonum/src/proto/schema/exonum/runtime.proto
dmitry-timofeev: It would be helpful as it will be inserted in the generated classes which may be used in user code: ECR-3731

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -44,10 +44,18 @@ message InstanceSpec {
   ArtifactId artifact = 3;
 }
 
+/// The kind of ExecutionError.
```
dmitry-timofeev: Protobuf uses double slash. The 3rd will appear in the doc.

PR #1516, opened 2019-10-30
Integration test python lib

popzxc:
This PR provides a library for writing integration tests for Exonum, and two examples of usage (one based on `exonum-python-client` and one based on `exonum-launcher`).

What's in the scope of this PR:
- Providing a handy interface for writing integration tests.
- Providing several examples to simplify library learning.

What's out of the scope of this PR:
- Adding integration tests to the CI.
- Creating a comprehensive integration test set.
- Publishing library to the `pypi`.
- Creating documentation.


test-suite/exonum-py-tests/exonum_tests/deploy.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class DeployTest(unittest.TestCase):
+    """Tests for Exonum deploy process."""
+
+    def test_deploy_run_dev(self):
+        """Tests the deploy mechanism in run-dev mode."""
+        with run_dev_node("exonum-cryptocurrency-advanced") as network:
+            # Wait some time for node to start
+            time.sleep(5)
```
popzxc: To be honest, I really don't like this approach (it's at the same time error-prone and not reliable), but I don't know if we have a correct way to check if node has completed initialization (including connecting to other peers and starting the consensus) and ready for interaction.

aleksuss: We will be able to get a consensus status via `/healthcheck` when we will fix `ECR-2931`, won't we ?

popzxc: Yeah, I think so (however, test for healtcheck endpoint itself won't be able to use that method :upside_down_face:).

aleksuss: Sure 😄 But for other tasks.

test-suite/exonum-py-tests/exonum_tests/api.py
```diff
@@ -0,0 +1,29 @@
...
+from exonum_client import ExonumClient
+
+from suite import run_4_nodes, assert_processes_exited_successfully
+
+
+class ApiTest(unittest.TestCase):
+    """Tests for Exonum API."""
+
+    def test_health_info(self):
+        """Tests the `health_info` endpoint."""
```
aleksuss: Did you mean `healthcheck` ?
https://github.com/exonum/exonum/blob/master/exonum/src/api/node/public/system.rs#L153
```suggestion
        """Tests the `healthcheck` endpoint."""
```

test-suite/exonum-py-tests/suite/common.py
```diff
@@ -0,0 +1,89 @@
...
+
+from typing import List, Dict, Any
+import unittest
+
+from suite import ExonumNetwork, ProcessOutput, ProcessExitResult
+
+
+def run_dev_node(application: str) -> ExonumNetwork:
+    """Starts a single node in the run-dev mode and returns
+    `ExonumNetwork object with the running node.
```
aleksuss:
```suggestion
    `ExonumNetwork` object with the running node.
```

test-suite/exonum-py-tests/suite/common.py
```diff
@@ -0,0 +1,89 @@
...
+        network.finalize(i, public_api_address, private_api_address)
+        available_api_port += 2
+
+    for i in range(nodes_amount):
+        network.run_node(i)
+
+    return network
+
+
+def run_4_nodes(application: str) -> ExonumNetwork:
```
aleksuss: Digit in the function name ? Rare case. I'd write a `run_four_nodes`. But not principle.

popzxc: Well, I think that it's shorter, doesn't affect readability and `4` is easier to see that `four`, but I don't insist on it either.

test-suite/exonum-py-tests/suite/temp_dir.py
```diff
@@ -0,0 +1,25 @@
...
+        """Initializes a TempDir object."""
+        temp_dir = tempfile.mkdtemp(prefix="exonum_test_suite_")
+
+        return cls(temp_dir)
+
+    def path(self) -> str:
+        """Returns the path of the temporary dir."""
+        return self.temp_dir
+
+    def remove(self) -> None:
```
aleksuss: Just out of curiosity. Do we really need the explicit method `remove` ? Isn't enough a `__exit__` method ?

popzxc: With that method user will be able to use `TempDir` without context manager, if he will ever want to, just like you can open a file either with
```python
with open(filename) as some_file:
    do_magic(some_file)
```

or manually:
```python
some_file = open(filename)
do_magic(some_file)
some_file.close()
```

It is usually good idea to support both workflows, since we aren't always able to use context managers (e.g. if object is a class field).

test-suite/exonum-py-tests/suite/process_manager.py
```diff
@@ -0,0 +1,84 @@
...
+    stdout: str
+    stderr: str
+
+
+class ProcessManager:
+    """ProcessManager is the entity capable of the running the process
+    in the separate thread, joining it and collectiong outputs."""
+
+    def __init__(self, command: str):
+        self._thread_handle = Thread(target=self._start_process)
```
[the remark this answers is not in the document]
popzxc: Well, yeah, but since we are launching the `exonum` in the shell, a separate process will be created anyway. This thread is only a wrapper around the launched process, and I don't think it should be a process itself. For example, here we do share some state between spawned thread and original one (`self._output`) and GIL is our best friend to not bother ourselves with synchronization.

test-suite/exonum-py-tests/suite/common.py
```diff
@@ -0,0 +1,89 @@
...
+    """Creates and runs a network with N validators and return an
+    `ExonumNetwork` object with it."""
+
+    address = "127.0.0.1:{}"
+
+    # Assign peer ports starting from 6331.
+    available_peer_port = 6331
+
+    # Assign API ports starting from 8080.
+    available_api_port = 8080
```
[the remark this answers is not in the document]
popzxc: Good point, but I think it'd be an overkill for test library (which will be running either on developers machines (and it's somewhat a convention to assign ports starting from 8080 in development process) or in CI (and having those ports busy on CI will mean an error in the CI configuration)).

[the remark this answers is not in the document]
Parfyonator: OK, I agree )

test-suite/exonum-py-tests/suite/process_manager.py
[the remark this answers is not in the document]
Parfyonator: Ouh, OK. I seem to forget about `self._output`. `self._start_process` was a bit misleading. Maybe a `self._start_thread` instead?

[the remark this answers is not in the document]
popzxc: It's a bit complicated, since we're starting a thread which starts a process via subprocess `¯\_(ツ)_/¯`

I'm not sure about the naming, but I think since this entity is called `ProcessManager` and it's all about the spawned process and its outputs, I think that process at least should be mentioned.

How about `_spawn_subprocess` or `_spawn_subprocess_in_thread`?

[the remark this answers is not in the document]
popzxc: Or even simplier: just `_start`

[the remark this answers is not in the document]
Parfyonator: I guess `_start_subprocess` will be fine.

PR #1520, opened 2019-11-01
SimpleSupervisor: Verify that config proposal is sent by validator [ECR-3742]

popzxc:
## Overview

This PR enhances the `SimpleSupervisor` with check for proposal author to be a validator (as in `Supervisor`). For node administrator workflow is not the same as for `Supervisor`: request with config proposal should be sent through private API endpoint.

However, this is the only check performed, there is no need to send request by all validators, and there is no need to send confirmations as well.

---

See: https://jira.bf.local/browse/ECR-3742

### DIFF
--- a/services/supervisor/src/simple/api.rs
+++ b/services/supervisor/src/simple/api.rs
@@ -41,8 +41,8 @@ impl<'a> ApiImpl<'a> {
         &self,
         transaction: impl Transaction<dyn SimpleSupervisorInterface>,
     ) -> Result<Hash, failure::Error> {
-        let keypair = self.0.service_keypair;
-        let signed = transaction.sign(self.0.instance.id, keypair.0, &keypair.1);
+        let (pub_key, sec_key) = self.0.service_keypair;
+        let signed = transaction.sign(self.0.instance.id, *pub_key, sec_key);
 
         let hash = signed.object_hash();
         self.0.sender().broadcast_transaction(signed)?;
--- a/services/supervisor/src/simple/api.rs
+++ b/services/supervisor/src/simple/api.rs
@@ -1,0 +1,67 @@
+// Copyright 2019 The Exonum Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use exonum::{
+    crypto::Hash,
+    runtime::{
+        api::{self, ServiceApiBuilder, ServiceApiState},
+        rust::Transaction,
+    },
+};
+use exonum_merkledb::ObjectHash;
+use failure::Fail;
+
+use crate::{simple::SimpleSupervisorInterface, ConfigPropose};
+
+/// Private API specification of the simple supervisor service.
+pub trait PrivateApi {
+    /// Error type for the current API implementation.
+    type Error: Fail;
+
+    /// Creates and broadcasts the `ConfigPropose` transaction, which is signed
+    /// by the current node, and returns its hash.
+    fn propose_config(&self, proposal: ConfigPropose) -> Result<Hash, Self::Error>;
+}
+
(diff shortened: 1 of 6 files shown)


services/supervisor/src/simple/api.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+struct ApiImpl<'a>(&'a ServiceApiState<'a>);
+
+impl<'a> ApiImpl<'a> {
+    fn broadcast_transaction(
+        &self,
+        transaction: impl Transaction<dyn SimpleSupervisorInterface>,
+    ) -> Result<Hash, failure::Error> {
+        let keypair = self.0.service_keypair;
```
aleksuss: Why don't you unpack the tuple `keypair` ? IMO, It's more obvious and readable.
```suggestion
        let (pub_key, sec_key) = self.0.service_keypair;
```

services/supervisor/src/simple/mod.rs
```diff
@@ -69,6 +72,10 @@ impl SimpleSupervisorInterface for SimpleSupervisor {
             .verify_caller(Caller::as_transaction)
             .ok_or(DispatcherError::UnauthorizedCaller)?;
 
+        // Verify that transaction author is validator.
+        let author = context.caller().author().unwrap();
```
aleksuss: `unwrap()` is really unsafe here ?

services/supervisor/tests/simple.rs
```diff
@@ -30,6 +31,15 @@ use exonum_supervisor::{
     ConfigPropose, Configure,
 };
 
+pub fn sign_config_propose_transaction(
+    testkit: &TestKit,
+    config: ConfigPropose,
+    initiator_id: ValidatorId,
+) -> Verified<AnyTx> {
+    let keys = &testkit.validator(initiator_id).service_keypair();
```
aleksuss: I just like to unpack tuples 😄

services/supervisor/tests/simple.rs
```diff
@@ -266,13 +297,17 @@ fn discard_errored_service_config_change() {
...
-    testkit.create_block_with_transaction(
-        ConfigPropose::actual_from(cfg_change_height)
-            .service_config(ConfigChangeService::INSTANCE_ID, params.clone())
-            .service_config(ConfigChangeService::INSTANCE_ID, "error".to_owned())
-            .consensus_config(new_consensus_config)
-            .sign_for_simple_supervisor(),
-    );
+    let initiator_id = testkit.network().us().validator_id().unwrap();
+    let config_propose = ConfigPropose::actual_from(cfg_change_height)
+        .service_config(ConfigChangeService::INSTANCE_ID, params.clone())
```
aleksuss: Needless clone ?

services/supervisor/src/simple/mod.rs
popzxc: Safe. We've checked `as_transaction` several lines above.

slowli: Nit: Why not use `context.caller().author().ok_or(DispatcherError::UnauthorizedCaller)` above?

services/supervisor/tests/simple.rs
```diff
@@ -587,3 +660,88 @@ fn several_service_config_changes() {
...
+
+    let old_validators = testkit.network().validators();
+
+    // Sign request by an auditor.
+    let keys = &testkit.network().us().service_keypair();
+    let config_propose = ConfigPropose::actual_from(cfg_change_height)
+        .consensus_config(new_consensus_config.clone())
+        .sign_for_simple_supervisor(keys.0, &keys.1);
+
+    testkit.create_block_with_transaction(config_propose);
```
slowli: Nit: It could make sense to check that the transaction has errored here, and to verify the error code.

services/supervisor/tests/simple.rs
```diff
@@ -154,11 +169,15 @@ fn exclude_us_from_validators() {
...
-    testkit.create_block_with_transaction(
-        ConfigPropose::actual_from(cfg_change_height)
-            .consensus_config(new_consensus_config.clone())
-            .sign_for_simple_supervisor(),
-    );
+    let initiator_id = testkit.network().us().validator_id().unwrap();
+    let config_propose = ConfigPropose::actual_from(cfg_change_height)
+        .consensus_config(new_consensus_config.clone());
+
+    testkit.create_block_with_transaction(sign_config_propose_transaction(
```
slowli: Nit: Maybe, it could make sense to implement a variation of `sign_config_propose_transaction` in which `validator_id` is set as `testkit.network().us().validator_id().unwrap()`. This would reduce the amount of repeated code a little bit.

PR #1523, opened 2019-11-04
Differentiated access to database [ECR-3760] [ECR-3768] [ECR-3769]

slowli:
## Overview

This PR introduces read-only access to forks, differentiates read-only and full access to indexes, and maps these changes to the blockchain access API.

### Bikeshedding targets

- Names of methods, e.g., in `merkledb::AccessExt` (`list` or `get_list`? `ensure_map` or `get_or_create_map` or something else entirely?) and `BlockchainData`.
- Whether `AccessExt` makes sense in general (vs having public index constructors).
- Whether it was a correct decision to remove references to indexes and `ObjectAccess`.
- Lazy vs eager creation of indexes in schemas, its impact on performance and UX.

### More important stuff

- How much does this break EJB?

### Definition of Done


exonum/src/blockchain/mod.rs
```diff
@@ -287,15 +293,15 @@ impl BlockchainMut {
         }
 
         // Get tx & state hash.
-        let schema = Schema::new(&fork);
+        let schema = Schema::get_unchecked(&fork);
```
popzxc: `Schema::get_unchecked` is called 30 times within that PR, and `Schema::get` is only 4. I don't know why, but it just don't feel right.

exonum/src/blockchain/mod.rs
```diff
@@ -287,15 +293,15 @@ impl BlockchainMut {
...
         let state_hash = {
             let mut sum_table = schema.state_hash_aggregator();
             // Clear old state hash.
             sum_table.clear();
             // Collect all state hashes.
             let state_hashes = self
                 .dispatcher
-                .state_hash(fork.as_ref())
+                .state_hash(fork.snapshot_with_flushed_changes())
```
popzxc: Is there any situation when user wants the snapshot to be without flushed changes?

services/supervisor/src/simple/schema.rs
```diff
@@ -14,25 +14,37 @@
...
-pub struct Schema<T: IndexAccess> {
-    access: T,
+const NOT_INITIALIZED: &str = "Supervisor schema is not initialized";
+
+pub struct Schema<T: AccessExt> {
+    pub config_propose: Entry<T::Base, ConfigPropose>,
 }
 
-impl<T: IndexAccess> Schema<T> {
+impl<T: AccessExt> Schema<T> {
```
slowli: It could make sense to restrict service schema constructors to `Prefixed<_>` in order to minimize possible misuse, like `SomeSchema::new(&testkit.snapshot())`.

examples/cryptocurrency-advanced/backend/src/schema.rs
```diff
@@ -14,94 +14,106 @@
...
+    T::Base: IndexAccessMut,
+{
+    pub(crate) fn initialize(access: T) -> Self {
+        Self {
+            wallets: access.ensure_proof_map("wallets"),
+            access,
+        }
+    }
+
+    fn ensure_wallet_history(&self, public_key: &PublicKey) -> ProofListIndex<T::Base, Hash> {
```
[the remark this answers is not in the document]
slowli: > it can be a structure field (when should it? when should it not?)

Top-level / singular indexes can be fields in the schema; grouped indexes cannot, at least with the current set of abstractions.

> what does `None` mean? a DB failure, forgotten initialization, empty index, something else?

`None` means no index in the database.

> why can't we always call `ensure_*` method?

Because `ensure_*` is only defined for writeable `AccessExt` implementations (e.g., `&Fork` or `Prefixed<&Fork>`). Readonly implementations naturally cannot perform write operations on the storage, such as creating indexes. Hence, I'd suggest to view `ensure_*` methods of `AccessExt` as ones that *ensure* that an index is there (these methods are idempotent, they do not create indexes on each call). Other methods *get* an index handle from the storage and return `None` if the index is not in the storage. Essentially, `access.ensure_something(addr)` is a shortcut for `access.something(addr).unwrap_or_else(|| Something::create(addr))`, except for now `Something::create` constructors are not publicly exposed.

examples/cryptocurrency-advanced/backend/src/schema.rs
```diff
@@ -14,94 +14,106 @@
...
+        vec![self.wallets.object_hash()]
+    }
+}
+
+impl<T> Schema<T>
+where
+    T: AccessExt,
+    T::Base: IndexAccessMut,
+{
+    pub(crate) fn initialize(access: T) -> Self {
```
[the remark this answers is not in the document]
slowli: Why? Services are still stateless, unless you view blockchain data as a part of service state (in which case, services are stateful regardless of this PR).

I'd say that creating indexes explicitly is much clearer than magicking them out of nowhere on the first use. Take, for example, the following code which works with the MerkleDB from `dynamic_services` branch:

```rust
let db = TemporaryDB::new();
let snapshot = db.snapshot();

let mut list: ProofListIndex<_, u32> = ProofListIndex::new("test", &snapshot);
list.push(1); // This is fine
// list.push(2); // This is no longer fine
// list.extend(vec![4, 5]); // This panics, too
assert_eq!(list.len(), 1);
assert!(list.iter().collect::<Vec<_>>().is_empty()); // Oh hell yeah
```

examples/cryptocurrency-advanced/backend/src/schema.rs
```diff
@@ -14,94 +14,106 @@
...
-where
-    T: IndexAccess,
-{
+impl<T: AccessExt> Schema<T> {
     /// Creates a new schema from the database view.
-    pub fn new(service_name: &'a str, access: T) -> Self {
-        Schema {
+    pub fn new(access: T) -> Self {
+        Self {
+            wallets: access.proof_map("wallets").expect(NOT_INITIALIZED),
```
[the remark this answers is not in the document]
slowli: They can (potentially) use a derive macro for schemas, which is out of the scope of this PR.

examples/cryptocurrency-advanced/backend/src/schema.rs
```diff
@@ -14,94 +14,106 @@
...
     /// Returns the state hash of cryptocurrency service.
     pub fn state_hash(&self) -> Vec<Hash> {
-        vec![self.wallets().object_hash()]
+        vec![self.wallets.object_hash()]
+    }
+}
+
+impl<T> Schema<T>
+where
+    T: AccessExt,
```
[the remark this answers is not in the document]
slowli: > was not really good

What's the reasoning for it? Note that there is a single method (or field) to access an index from the schema, which was no the case in the previous differentiated iteration.

`ensure_*` methods are only required during initialization, and so can be called exactly once for singular indexes from a single place (`Schema::initialize()`). For indexes in index families the situation is admittedly worse, but I don't think it could be made better without reverting to the old concept of index families (i.e., the entire family is constructed at the same time).

exonum/src/blockchain/mod.rs
```diff
@@ -413,6 +415,40 @@ impl BlockchainMut {
...
+    pub fn add_transactions_into_pool(
+        &mut self,
+        // ^-- mutable reference taken for future compatibility.
+        transactions: impl IntoIterator<Item = Verified<AnyTx>>,
+    ) {
+        Self::add_transactions_into_db_pool(self.inner.db.as_ref(), transactions);
+    }
+
+    /// Same as `add_transactions_into_pool()`, but accepting a database handle instead
+    /// of the `BlockchainMut` instance. Beware that accesses to database need to be synchronized
```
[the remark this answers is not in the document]
slowli: Yeah right, as if having full access to all schemas from all contexts isn't 🙃

examples/cryptocurrency-advanced/backend/src/lib.rs
```diff
@@ -47,11 +51,16 @@ pub const INITIAL_BALANCE: u64 = 100;
 pub struct CryptocurrencyService;
 
 impl Service for CryptocurrencyService {
-    fn wire_api(&self, builder: &mut ServiceApiBuilder) {
-        CryptocurrencyApi.wire(builder);
+    fn initialize(&self, context: CallContext<'_>, _params: Vec<u8>) -> Result<(), ExecutionError> {
+        Schema::initialize(context.service_data());
```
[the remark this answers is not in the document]
slowli: > even writing the tests won't help you there

Why do you think this is the case? For example, it's enough to comment out this line for *all* `cryptocurrency-advanced` tests to fail. And they all fail with "Cryptocurrency schema is not initialized" message, too!

Also, I'm sorry, but "It's pretty easy to forget to initialize the schema" reads like "It's pretty easy to forget to create your tables in RDBMS". I'd say (a) it's not that easy given the right mindset (indexes = RDMBS tables) and (b) the consequences are immediate and obvious. You can no longer materialize an index from thin air, so when the only option API gives you is `Option<Index>`, you should think why this `Option` should or should not be `Some(_)`.

Regarding automated initialization: it could be possible to implement a more typed version of `Service` trait for Rust services, which would know about schema (e.g., have it as an associated type), so that the schema could be initialized automatically. This is outside the scope of this PR.

examples/cryptocurrency/src/lib.rs
```diff
@@ -120,6 +117,18 @@ pub mod schema {
             vec![]
         }
     }
+
+    impl<T> CurrencySchema<T>
```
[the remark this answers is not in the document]
slowli: It could be replaced with `impl CurrencySchema<Prefixed<&Fork>>` at the cost of loss of generalization.

exonum/src/blockchain/mod.rs
slowli: The idea is that `Fork` consists of 3 parts: an underlying `Snapshot`, flushed changes and unflushed changes. This method captures first 2 parts, but not the third one. I suppose `snapshot_without_unflushed_changes()` name could convey this more clearly.

examples/cryptocurrency-advanced/backend/src/lib.rs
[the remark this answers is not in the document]
popzxc: Well, okay, I was wrong about the tests, but as I told: it's just a first glance and not the actual review. To understand that commenting this line will break the tests, you have to compile the Exonum in your head, since this fact relies on the surrounding context.

Your analogy is good, but I'd compare it with raw pointers in the C++: of course you have to init it within constructor, and of course consequences are immediate and obvious if you'll forget to do it. But somehow programs still do fail because of the null pointer dereference. And Rust somehow allows users not to worry about it.

tl;dr: I really love keeping things as simple and stupid as possible, and IMO it's too low-level for a public framework interface. I would really like the situation when users won't have to think why API gives you an `Option`, so they'll be able to focus on the business logic.

examples/cryptocurrency-advanced/backend/src/schema.rs
```diff
@@ -14,94 +14,106 @@
...
 
 use crate::{wallet::Wallet, INITIAL_BALANCE};
 
+const NOT_INITIALIZED: &str = "Cryptocurrency schema is not initialized";
+
 /// Database schema for the cryptocurrency.
 #[derive(Debug)]
-pub struct Schema<'a, T> {
-    service_name: &'a str,
+pub struct Schema<T: AccessExt> {
```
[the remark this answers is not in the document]
slowli: I called the trait `AccessExt` because it initially was purely an extension trait for `IndexAccess`. This is no longer the case (`Prefixed<_>` implements `AccessExt`, but not `IndexAccess`; I've tried this way btw, but it turned ugly really fast), so the name may indeed be confusing. `Access` could probably be a good name (I'd rename `IndexAccess` to something like `RawAccess` or `ViewAccess` as well).

components/merkledb/examples/blockchain.rs
```diff
@@ -84,7 +84,7 @@ struct Schema<T: Access> {
     pub transactions: MapIndex<T::Base, Hash, Transaction>,
     pub blocks: ListIndex<T::Base, Hash>,
     pub wallets: ProofMapIndex<T::Base, PublicKey, Wallet>,
-    access: T,
+    pub wallet_history: Group<T, PublicKey, ProofListIndex<T::Base, Hash>>,
```
slowli: @popzxc Does this form for index groups looks better? In the same vein, it could be possible to implement `Lazy<T, Index<T, ..>>` wrapper, but I'm not sure whether it makes sense.

exonum/src/blockchain/mod.rs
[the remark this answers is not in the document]
slowli: I think in general public functions with `doc(hidden)` should be used at caller's risk. There is even an explicit comment next to the `doc(hidden)` attribute not to use this method anywhere.

components/merkledb/examples/blockchain.rs
popzxc: Hm... Yeah, I think it's better, but nevertheless the overall verbosity of the construction (double template, `T`, `T::Base`) looks scary. On the other hand, this can be overrided with type aliases I guess.

exonum/src/blockchain/mod.rs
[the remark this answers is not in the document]
popzxc: Agree, that was not really even a nitpick, more like thinking out loud.
I don't mind having such a function until it's under `doc(hidden)` protection.
Maybe just add a note for developers so we won't make it public accidently.

slowli:
I've reverted back to the old implicit construction of indexes, which I believe was the main contention point. Correspondingly, `ensure_*` method names are replaced with `get_*`, and the main target for bikeshedding is the naming of the `Restore` trait.

I had no luck simplifying bounds for mutable `Access` so far. Something like this is easy:

```rust
pub trait AccessMut<'a>: Access<Base = &'a Fork> {}
// there are no other `Base`s for practical mutable access other than `&Fork`

impl<'a, T> AccessMut<'a> for T where T: Access<Base = &'a Fork> {}
```

...but this still necessitates a bogus lifetime argument for the trait:

```rust
impl<'a, T: AccessMut<'a>> Schema<T> { /* ... */ }
// Not much better than
impl<'a, T: Access<Base = &'a Fork>> Schema<T> { /* ... */ }
```

components/merkledb/benches/storage.rs
```diff
@@ -76,7 +76,7 @@ fn plain_map_index_insert(b: &mut Bencher, len: usize) {
         |(db, data)| {
             let fork = db.fork();
             {
-                let mut table = MapIndex::new(NAME, &fork);
+                let mut table = fork.as_ref().ensure_map(NAME);
```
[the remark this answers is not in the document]
slowli: `as_ref()` required because `Fork` doesn't implement `RawAccess`, `&Fork` does (which, in turn, is due to the fact that the associated type `ChangeSet` in `RawAccess` needs to be generic in order to allow `Fork` implementation). It is possible to write `(&fork).blah()` instead, but IMO `fork.as_ref().blah()` reads better.

slowli:
Here are local benchmark results for different schema patterns (see the `schema_patterns` bench in the `merkledb` crate):

```text
schema_patterns/eager   time:   [867.38 ms 870.42 ms 873.58 ms]
                        thrpt:  [11.447 Kelem/s 11.489 Kelem/s 11.529 Kelem/s]
schema_patterns/lazy    time:   [845.06 ms 848.09 ms 852.33 ms]
                        thrpt:  [11.732 Kelem/s 11.791 Kelem/s 11.833 Kelem/s]
schema_patterns/wrapper time:   [850.91 ms 856.59 ms 868.55 ms]
                        thrpt:  [11.513 Kelem/s 11.674 Kelem/s 11.752 Kelem/s
```

So, a schema with fields and `Lazy<_>` cold indexes is the fastest, although only marginally compared to an old-style wrapper schema. A schema with eagerly created fields is slower by ~2.5%.

components/merkledb/src/db.rs
```diff
@@ -614,6 +685,19 @@ impl Fork {
...
+    /// Returns snapshot that also captures flushed changes in the fork,
+    /// but does not capture unflushed changes.
+    pub fn snapshot_without_unflushed_changes(&self) -> &dyn Snapshot {
+        &self.patch
+    }
+
+    /// Returns a readonly wrapper around the fork. Indices created based on the readonly
+    /// version cannot be modified; on the other hand, it is possible to have multiple
+    /// copies of an index at the same time.
+    pub fn readonly(&self) -> ReadonlyFork<'_> {
```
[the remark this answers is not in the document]
slowli: Using `ToReadonly` requires importing it into the scope; using a builtin method does not 🙃

components/merkledb/src/db.rs
```diff
@@ -642,26 +726,92 @@ impl<'a> IndexAccess for &'a Fork {
...
+/// let mut list = readonly.get_list("list");
+/// list.push(1_u32); // No `push` method for `ListIndex<ReadonlyFork, u32>`!
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct ReadonlyFork<'a>(&'a Fork);
+
+impl<'a> ToReadonly for &'a Fork {
+    type Readonly = ReadonlyFork<'a>;
+
+    fn to_readonly(&self) -> Self::Readonly {
```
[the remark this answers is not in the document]
slowli: I'm not sure, `ToReadonly` looks like a conversion trait, so `to_*` naming looks natural to me (or maybe `as_readonly`, [since conversion is cheap](https://doc.rust-lang.org/1.0.0/style/style/naming/conversions.html) for supported types).

components/merkledb/src/db.rs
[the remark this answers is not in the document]
aleksuss: What about to rename to `read_only` ?

components/merkledb/src/db.rs
```diff
@@ -642,26 +726,92 @@ impl<'a> IndexAccess for &'a Fork {
...
+/// error (sort of like attempting both an exclusive and a shared borrow from a `RefCell`
+/// or `RwLock`).
+///
+/// # Examples
+///
+/// ```
+/// # use exonum_merkledb::{access::AccessExt, Database, ReadonlyFork, TemporaryDB};
+/// let db = TemporaryDB::new();
+/// let fork = db.fork();
+/// fork.as_ref().get_list("list").push(1_u32);
```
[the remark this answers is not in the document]
slowli: Would `(&fork).blah()` be better? `fork.as_ref()` looks better IMO, but I won't argue with either of the variants. Other choices:

- Make an additional extension trait which takes `<T: Access + Copy>` by value
- Make `Access::get_or_create_view` take `self` by value; correspondingly, make `AccessExt` (and `Restore`) methods take `self` by value, too. Note that this is already *de facto* the case for `RawAccess` implementations, but I'm slightly worried about performance degradation for `Prefixed`. I'll benchmark this choice and if there is no noticeable degradation, will use it.

components/merkledb/src/access/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+///
+/// impl<T: Access> Schema<T> {
+///     fn get_some_data(&self) -> Option<u64> {
+///         Some(self.list.get(0)? + self.map.get(&"foo".to_owned())?)
+///     }
+/// }
+/// ```
+pub trait Access: Clone {
+    /// Raw access serving as the basis for created indices.
+    type Base: RawAccess;
```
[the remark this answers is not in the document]
slowli: It's long though, and there is seemingly no way around in writing it for writeable methods of schemas... (other than implementing methods for a specific `Access` instantiation, e.g., `Prefixed<&Fork>`).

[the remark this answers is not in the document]
popzxc: Yeah, I don't happy about it either. But IMO `Base` doesn't really explain things to user (I had to dig into the sources to understand what it means).

components/merkledb/src/db.rs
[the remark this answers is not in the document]
slowli: `std` lib [uses `readonly` spelling](https://doc.rust-lang.org/std/fs/struct.Permissions.html#method.readonly).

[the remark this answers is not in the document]
popzxc: I like `readonly` more too.

components/merkledb/src/db.rs
```diff
@@ -642,26 +726,92 @@ impl<'a> IndexAccess for &'a Fork {
...
-        &self.patch
+/// Readonly wrapper for a `Fork`.
+///
+/// This wrapper allows to read from index state from the fork
+/// in a type-safe manner (it is impossible to accidentally modify data in the index), and
+/// without encountering runtime errors when attempting to concurrently get the same index
+/// more than once.
+///
+/// Since the wrapper borrows the `Fork` immutably, it is still possible to access indexes
+/// in the fork directly. In this scenario, the caller should be careful that `ReadonlyFork`
```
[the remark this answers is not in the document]
slowli: Well, it could be possible for `ReadonlyFork` to wrap `&mut Fork` instead of `&Fork`, but that would defeat the purpose of `ReadonlyFork`. In Exonum, we *want* to have concurrent access to different parts of the database via the same `Fork`: full access to the data of the executing service, and readonly access to the core / dispatcher and other services.

components/merkledb/src/access/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        actual: IndexType,
+    },
+
+    /// Custom error.
+    #[fail(display = "{}", _0)]
+    Custom(#[fail(cause)] Error),
+}
+
+/// Restores an object from the database.
+pub trait Restore<T: Access>: Sized {
```
alekseysidorov: In my opinion, such name can lead to confusion. What actually "restore" means: restore broken index, restore access to the index or something else?

components/merkledb/src/db.rs
```diff
@@ -337,28 +387,49 @@ pub enum Change {
...
+///
+/// To enable immutable / shared references to indexes, you may use [`readonly`] method:
+///
 /// ```
+/// # use exonum_merkledb::{access::AccessExt, TemporaryDB, ListIndex, Database};
+/// let db = TemporaryDB::new();
+/// let fork = db.fork();
+/// fork.as_ref().get_list::<_, u8>("index").extend(vec![1, 2, 3]);
+///
+/// let readonly = fork.readonly();
```
alekseysidorov: What about naming like `RwLock::read`? 

```rust
let readonly = fork.read();
```

Or 

```rust
let rw = fork.write();
```

To replace `fork.as_ref()`?

components/merkledb/src/db.rs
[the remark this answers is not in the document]
alekseysidorov: I'd suggest to use `read` name. And rename trait to `Read` or something like `FooRead`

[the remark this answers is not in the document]
slowli: There is `Read` trait in std, which has substantially different semantics. IMO, `Read` implies that you can read something from the implementation, which is not the case here; the trait performs a *conversion*, so IMO, naming `ToReadonly::to_readonly` or `AsReadonly::as_readonly` conveys the meaning best.

components/merkledb/examples/blockchain.rs
```diff
@@ -1,11 +1,13 @@
+use failure::Error;
```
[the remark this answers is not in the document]
slowli: We can change our approach in a separate PR. MerkleDB already uses `failure` with or without this PR, so removing it is outside the scope.

components/merkledb/src/access/mod.rs
slowli: Would something like `FromAccess` be acceptable?

components/merkledb/src/entry.rs
```diff
@@ -31,125 +31,31 @@ use crate::{
...
-    /// let index: Entry<_, u8> = Entry::new(name, &snapshot);
-    /// ```
-    pub fn new<S: Into<String>>(index_name: S, view: T) -> Self {
-        let (base, state) = IndexBuilder::new(view)
-            .index_type(IndexType::Entry)
-            .index_name(index_name)
-            .build::<()>();
-
+    fn new(view: ViewWithMetadata<T>) -> Self {
+        let (base, _) = view.into_parts::<()>();
```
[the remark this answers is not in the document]
slowli: Done.

components/merkledb/src/db.rs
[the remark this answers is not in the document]
slowli: I've implemented the approach with taking `self` by value in several traits.

components/merkledb/src/db.rs
```diff
@@ -255,6 +302,9 @@ impl WorkingPatch {
             let changes = changes.unwrap_or_else(|| {
                 panic!("changes are still borrowed at address {:?}", address);
             });
+            let changes = Rc::try_unwrap(changes).unwrap_or_else(|_| {
```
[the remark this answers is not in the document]
slowli: I've commented the point of these checks; they check mutable / immutable borrows respectively.

components/merkledb/examples/blockchain.rs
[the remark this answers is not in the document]
alekseysidorov: Yes of course, this task is not important at the moment

components/merkledb/src/db.rs
```diff
@@ -209,51 +228,102 @@ impl WorkingPatch {
...
+    /// are mutably borrowed.
+    fn clone_view_changes(&self, address: &IndexAddress) -> Rc<ViewChanges> {
+        let mut changes = self.changes.borrow_mut();
+        changes
+            .entry(address.clone())
+            .or_insert_with(|| Some(Rc::new(ViewChanges::new())))
+            .as_ref()
+            .unwrap_or_else(|| {
+                panic!(
+                    "Attempting to borrow {:?} immutably while it's borrowed mutably",
```
popzxc: This comment looks weird. We've borrowed changes mutably several lines above, who exactly tries to borrow them immutable here?

components/merkledb/benches/schema_patterns.rs
```diff
@@ -0,0 +1,327 @@
...
+            }
+        }
+
+        if transaction.value % COLD_CHANCE == 0 {
+            self.other_cold_index.insert(transaction.value);
+        }
+    }
+}
+
+enum EagerStyle {}
```
aleksuss: Just out of curiosity. Should it be a exactly enum rather than unit-like struct ?

components/merkledb/src/views/metadata.rs
```diff
@@ -12,34 +12,45 @@
...
+///
 /// `IndexType` is used for type checking indexes when they are created/accessed.
 #[derive(Debug, Copy, Clone, PartialEq, Primitive, Serialize, Deserialize)]
 #[repr(u32)]
 pub enum IndexType {
+    /// Non-merkelized map index.
     Map = 1,
+    /// Non-merkelized list index.
     List = 2,
+    /// Singleton entry.
```
popzxc: I'm not sure if "signleton" is applicable here, since we can have multiple different entries.

components/merkledb/src/views/metadata.rs
```diff
@@ -12,34 +12,45 @@
...
+    /// Sparse list index.
     SparseList = 6,
+    /// Merkelized list index.
     ProofList = 7,
+    /// Merkelized map index.
     ProofMap = 8,
+
+    /// Unknown index type.
+    #[doc(hidden)]
     Unknown = 255,
```
popzxc: Off-topic: do we need it at all?

components/merkledb/src/views/metadata.rs
```diff
@@ -212,135 +261,128 @@ impl<T: IndexAccess> IndexesPool<T> {
...
+    fn set_len(&mut self, len: u64) {
+        self.0.put_or_forget(&(), len);
     }
 
     fn create_index_metadata<V>(
         &mut self,
         index_name: &[u8],
         index_type: IndexType,
-    ) -> IndexMetadata<V>
+    ) -> (IndexMetadata<V>, bool)
```
popzxc: Nit: I think it deserves an enum: `Phantom`/`Real` is more readable than `true`/`false` (especially when you realize that `true` means phantom index).

components/merkledb/src/views/metadata.rs
```diff
@@ -212,135 +261,128 @@ impl<T: IndexAccess> IndexesPool<T> {
...
+        let this = Self {
+            view: View::new(index_access, index_address),
+            metadata,
+            index_full_name,
+            is_phantom,
+        };
+        if real_index_type == index_type {
+            Ok(this)
+        } else {
+            Err(this)
```
popzxc: I guess an explanation (maybe in docstring) what `Err(this)` does mean and why if it's an error we still return the value would be nice.

components/merkledb/benches/schema_patterns.rs
slowli: Probably should be `struct EagerStyle(!)` as per [clippy suggestion](https://rust-lang.github.io/rust-clippy/master/index.html#empty_enum); don't know why clippy missed this.

components/merkledb/src/views/mod.rs
```diff
@@ -76,191 +81,80 @@ impl ChangeSet for ChangesRef<'_> {
...
+///     }
+/// }
 /// ```
-#[derive(Debug)]
-pub struct IndexBuilder<T> {
-    index_access: T,
-    address: IndexAddress,
-    index_type: IndexType,
-}
+pub trait RawAccessMut: RawAccess {}
```
popzxc: Unlike the `RawAccess`, `RawAccessMut` is used almost everywhere, so the mutable/immutable code looks [asymmetric](https://github.com/exonum/exonum/pull/1523/files#diff-ee26286bb122f945e5926d1f6f710e23R18): for immutable access we use `Access`, but for mutable it's `RawAccessMut`. Maybe make this pair `Access`/`AccessMut`?

(or it even can be `MutAccess` to be closer to the plain english)

components/merkledb/src/views/mod.rs
```diff
@@ -389,8 +291,10 @@ fn key_bytes<K: BinaryKey + ?Sized>(key: &K) -> Vec<u8> {
     concat_keys!(key)
 }
 
-impl<T: IndexAccess> View<T> {
-    ///TODO: add documentation
+impl<T: RawAccess> View<T> {
+    /// Creates a new view for an index with the specified address.
+    #[doc(hidden)]
+    // ^-- This should be `pub(crate)`, but is currently used by the testkit to revert blocks.
```
popzxc: I guess a TODO with ECR should be places here?

components/merkledb/src/views/mod.rs
```diff
@@ -525,8 +429,12 @@ impl<T: IndexAccess> View<T> {
         }
     }
 
-    /// Inserts a key-value pair into the fork.
-    pub fn put<K, V>(&mut self, key: &K, value: V)
+    /// Crutch to be able to create metadata for indexes not present in the storage.
```
popzxc: I think that comments like this are useful for developers but not really appropriate for documentation.

exonum/src/blockchain/schema.rs
```diff
@@ -95,38 +95,35 @@ impl TxLocation {
 /// Indices defined by this schema are present in the blockchain regardless of
 /// the deployed services and store general-purpose information, such as
 /// committed transactions.
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 pub struct Schema<T> {
     access: T,
```
popzxc: Nit: I think a small note for developers why we don't use a field-based schema here will be nice.

components/merkledb/benches/schema_patterns.rs
```diff
@@ -0,0 +1,327 @@
...
+    cold_group: Group<T, u64, ProofListIndex<T::Base, u64>>,
+    // Accessed once per ~`COLD_DIVISOR` transactions.
+    other_cold_index: KeySetIndex<T::Base, u64>,
+}
+
+impl<T: Access> EagerSchema<T> {
+    fn new(access: T) -> Self {
+        Self {
+            transactions: Restore::restore(access.clone(), "transactions".into()).unwrap(),
+            hot_index: Restore::restore(access.clone(), "hot_index".into()).unwrap(),
```
popzxc: `access.clone()` is used really wide. Maybe make it `Copy`?

components/merkledb/src/db.rs
```diff
@@ -614,6 +705,19 @@ impl Fork {
     pub fn working_patch(&self) -> &WorkingPatch {
         &self.working_patch
     }
+
+    /// Returns snapshot that also captures flushed changes in the fork,
+    /// but does not capture unflushed changes.
+    pub fn snapshot_without_unflushed_changes(&self) -> &dyn Snapshot {
```
aleksuss: What about `snapshot_with_flushed_changes` ? Weird name as well but a little bit shorter 🙃

slowli: It was initially named that way; see https://github.com/exonum/exonum/pull/1523#discussion_r342094796.

components/merkledb/src/views/mod.rs
slowli: The `put_or_forget` method is crate-specific though, so the comment is *already* targeting MerkleDB devs. Would you like me to remove this comment? Make it a simple comment rather than a doc comment?

components/merkledb/benches/schema_patterns.rs
slowli: `Prefixed` implements `Access` and is not `Copy`... See, e.g., its usage in `SnapshotExt` or `BlockchainData` - the prefix may be determined dynamically, so it is seemingly impossible to have `Prefixed` contain `&str` instead of `Cow<str>`. The only way I see it could work is to require a string service name (instead of more generic `InstanceQuery`) when accessing service data. Would this be OK?

components/merkledb/src/views/mod.rs
popzxc: Yeah, simple comment will be better IMO. However, that's completely not critical.

components/merkledb/src/views/mod.rs
slowli: `*Mut` suffix follows the naming conventions from std, such as `AsMut` and `BorrowMut`. Regarding `AccessMut`, I'm not opposed to the idea in principle, but `Access` / `RawAccessMut` are not *intended* to be symmetric: `RawAccessMut` is a mutable counterpart to `RawAccess` and thus concerns *indexes*, whereas `Access` concerns more high-level storage access.

popzxc: I understand that, but not sure if users should even think about it. IMHO for users MerkleDB should provide a set of simple abstractions that "just work", without need to understand inner mechanics.

components/merkledb/benches/storage.rs
```diff
@@ -93,7 +93,7 @@ fn plain_map_index_with_family_insert(b: &mut Bencher<'_>, len: usize) {
         |(db, data)| {
             let fork = db.fork();
             {
-                let mut table = MapIndex::new_in_family(NAME, FAMILY, &fork);
+                let mut table = fork.get_map((NAME, FAMILY));
```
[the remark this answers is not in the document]
slowli: This syntax is useful for testing. Also I think that using `get_group` everywhere would require type hints for index types.

components/merkledb/benches/schema_patterns.rs
popzxc: Hm. If `Prefixed` is the only reason that stops us here, maybe implement `Copy` for it manually?

components/merkledb/benches/schema_patterns.rs
```diff
@@ -0,0 +1,329 @@
...
+    cold_index: Lazy<T, ProofMapIndex<T::Base, u64, Hash>>,
+    // groups are already lazy
+    cold_group: Group<T, u64, ProofListIndex<T::Base, u64>>,
+    other_cold_index: Lazy<T, KeySetIndex<T::Base, u64>>,
+}
+
+impl<T: Access> LazySchema<T> {
+    fn new(access: T) -> Self {
+        Self {
+            transactions: FromAccess::from_access(access.clone(), "transactions".into()).unwrap(),
```
alekseysidorov: One little thought: what if we'd introduced a little helper method to reduce visual code complexity?

slowli: I had in mind that a derive macro would take care of this. (Now that I think of it, it could make sense to implement the derive macro for `FromAccess` and have a method `FromAccess::<T>::create_here(T)` with a default implementation.) Does it make sense to defer this as a separate clean-up task?

components/merkledb/benches/schema_patterns.rs
slowli: Not sure what you mean. Admittedly, I'm not very knowledgeable in the details of low-level memory management, but wouldn't implementing `Copy` for a type containing `String` lead to double memory frees?

slowli: Also, I've been thinking about an approach in which `Prefixed` is replaced with *root objects*, as defined in the original MerkleDB spec. This would require substantially more changes in the MerkleDB interface though, which would hinder our release plans.

popzxc: >Not sure what you mean

Ah, indeed. I forgot that `Copy` trait cannot be implemented using some workaround. Sad.

>Also, I've been thinking about an approach in which Prefixed is replaced with root objects, as defined in the original MerkleDB spec. This would require substantially more changes in the MerkleDB interface though, which would hinder our release plans.

Yeah, I think it can wait.

PR #1525, opened 2019-11-05
Make protobuf artifacts implementation detail of Rust runtime [ECR-3776]

alekseysidorov:


exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -177,19 +174,7 @@ impl Dispatcher {
...
-                    .and_then(move |sources| {
-                        future_state
-                            .artifact_sources
-                            .write()
-                            .unwrap()
-                            .insert(artifact, sources);
-                        Ok(())
-                    });
-            Either::A(task)
+            Either::A(runtime.deploy_artifact(artifact.clone(), payload))
```
popzxc: Do we still need an `Either` here?

exonum/src/runtime/rust/api.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+fn filter_exonum_proto_sources(
+    files: Vec<ProtoSourceFile>,
+    exonum_sources: Vec<ProtoSourceFile>,
+) -> Vec<ProtoSourceFile> {
+    files
+        .iter()
+        .cloned()
+        .filter(|file| !exonum_sources.contains(file))
```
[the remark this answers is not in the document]
alekseysidorov: When we will work on the proto sources specification we return to this question.
But at the moment, for example, python protobuf implementation just ignores package name in the proto files and has an other flaws in implementation.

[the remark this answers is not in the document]
slowli: So... Could it be possible to perform filtering once when the files is provided by the service vs every time an endpoint is called? I don't think we should be *too* serious about DoS opportunities at this stage, but if we can avoid one, we should go for it.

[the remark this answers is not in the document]
alekseysidorov: Sounds reasonable.

exonum/src/runtime/rust/api.rs
```diff
@@ -116,6 +116,19 @@ pub fn endpoints(runtime: &RustRuntime) -> impl IntoIterator<Item = (String, Api
...
     let exonum_sources = exonum_proto_sources();
+    // Cache filtered sources to avoid expensive operations in the endpoint handler.
+    let filtered_sources = artifact_proto_sources
+        .clone()
+        .into_iter()
+        .map(|(artifact_id, sources)| {
+            let mut proto = sources.sources;
+            proto.extend(filter_exonum_proto_sources(
+                sources.includes,
+                exonum_sources.clone(),
```
slowli: Shouldn't it be possible to use a reference here to avoid cloning?

exonum/src/runtime/rust/api.rs
```diff
@@ -116,6 +116,19 @@ pub fn endpoints(runtime: &RustRuntime) -> impl IntoIterator<Item = (String, Api
         })
         .collect::<HashMap<_, _>>();
     let exonum_sources = exonum_proto_sources();
+    // Cache filtered sources to avoid expensive operations in the endpoint handler.
+    let filtered_sources = artifact_proto_sources
+        .clone()
```
slowli: Cloning seems to be unnecessary (unless I'm missing something, `artifact_proto_sources` isn't used anywhere else). Thus, it could make sense to combine creation of `filtered_sources` with creation of `artifact_proto_sources`.

exonum/src/runtime/rust/api.rs
```diff
@@ -0,0 +1,155 @@
...
+        })
+        .collect::<HashMap<_, _>>();
+
+    let mut builder = ApiBuilder::new();
+    builder
+        .public_scope()
+        // This endpoint returns list of protobuf source files of the specified artifact,
+        // otherwise it returns source files of Exonum itself.
+        .endpoint("proto-sources", {
+            move |query: ProtoSourcesQuery| {
```
aleksuss: Unnecessary but it would be good to define a return type of closure, I suppose

```suggestion
            move |query: ProtoSourcesQuery| -> Result<Vec<ProtoSourceFile>, api::Error> {
```

exonum/src/runtime/rust/api.rs
aleksuss: Works without `clone()`.

PR #1526, opened 2019-11-05
Sending an empty POST request to /shutdown endpoint doesn't work [ECR-3756]

SergeiMal:
## Overview

---

See: https://jira.bf.local/browse/ECR-3756

### Definition of Done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,7 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum
 
-- `system/v1/shutdown` endpoint doesn't requre json body with "null". (#)
+- `system/v1/shutdown` endpoint doesn't requre json body with "null". (#1526)
 
 #### exonum-merkledb
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,7 +82,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 #### exonum
 
-- `system/v1/shutdown` endpoint doesn't require json body with "null". (#1526)
+- `system/v1/shutdown` endpoint has been modified and now accepts empty POST
+  requests. (#1526)
 
 #### exonum-merkledb
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Internal improvements
 
+#### exonum
+
+- `system/v1/shutdown` endpoint doesn't requre json body with "null". (#)
+
 #### exonum-merkledb
 
 - `ProofListIndex::extend()` method has been refactored, leading to up to 10x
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -80,6 +80,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 ### Internal improvements
 
+#### exonum
+
+- `system/v1/shutdown` endpoint doesn't requre json body with "null". (#1526)
```
```suggestion
- `system/v1/shutdown` endpoint now accepts empty POST requests. (#1526)
```

aleksuss:
@SergeiMal you can see an example here https://github.com/exonum/exonum/blob/dynamic_services/exonum/src/api/node/public/explorer.rs#L281

aleksuss:
Looks good as aa97b51665a17941399cd9237b19af322937cee1. @ivan-ochc check it out please.

exonum/src/api/node/private/mod.rs
```diff
@@ -192,11 +194,30 @@ impl SystemApi {
...
                 .send_external_message(ExternalMessage::Shutdown)
                 .map_err(ApiError::from)
-        });
+        };
+
+        let index = move |_request: HttpRequest| -> BackendFutureResponse {
+            let handler = handler.clone();
+
+            let future = Ok(handler())
+                .and_then(|_| Ok(HttpResponse::Ok().json(())))
```
popzxc: Can we return `Ok(HttpResponse::Ok().json(()))` in the `handler` and get rid of this `and_then`?

IMHO It'd be better in the form of the

```rust
let handler = handler.clone();
let future = handler().into_future();
Box::new(future)
```

And btw I'd split it into several lines to avoid this parentheses chaos:
```rust
let ok_response = HttpResponse::Ok().json(());
Ok(ok_response)
```

exonum/src/api/node/private/mod.rs
```diff
@@ -17,16 +17,18 @@
 //! Private API includes requests that are available only to the blockchain
 //! administrators, e.g. view the list of services on the current node.
 
-use std::{collections::HashMap, net::SocketAddr};
+use std::{collections::HashMap, net::SocketAddr, sync::Arc};
 
 use crate::{
-    api::{ApiScope, Error as ApiError},
+    api::backends::actix::{FutureResponse as BackendFutureResponse, RawHandler, RequestHandler},
```
popzxc: I'd add those backend-specific imports inside the `handle_shutdown` function with a comment that we're doing it to support empty request which is not easy in the generic approach, so it will be harder to misuse those features (and as a result get a completely backend-dependent code).

exonum/src/api/node/private/mod.rs
```diff
@@ -191,12 +189,37 @@ impl SystemApi {
     }
 
     fn handle_shutdown(self, name: &'static str, api_scope: &mut ApiScope) -> Self {
+        // These uses need to provide realisation of the support of empty request
```
popzxc:
```suggestion
        // These backend-dependent uses are needed to provide realization of the support of empty request
```

PR #1527, opened 2019-11-06
Rename exonum-proto-derive to exonum-protobuf-convert

pavel-mukhanov:
Since `exonum-proto-derive` doesn't have any references to exonum(except some examples in docs) we decided to rename it to `protobuf-convert` .


CHANGELOG.md
```diff
@@ -42,9 +42,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Introduced a new crate `exonum-proto`. Trait `ProtobufConvert` is moved
   to this crate. (#1496)
 
-#### exonum-proto-derive
+#### protobuf-convert
```
aleksuss: A name of the crate should be `exonum-protobuf-convert`, shouldn't it ?

components/proto/Cargo.toml
```diff
@@ -20,7 +20,7 @@ bit-vec = "0.6.1"
 serde = "1.0.101"
 serde_derive = "1.0.101"
 exonum-derive = { path = "../derive", version = "0.12.0"}
-exonum-proto-derive = { path = "../proto-derive", version = "0.12.0"}
+protobuf-convert = { version = "0.12.0", path = "../protobuf-convert"}
```
aleksuss: Look above. `exonum-derive` not just `derive`.

CHANGELOG.md
pavel-mukhanov: We discussed with @alekseysidorov, maybe it's possible to create general purpose crate not linked to exonum. To replace https://crates.io/crates/protobuf-convert

aleksuss: If so then it should be out of `exonum` repo.

PR #1531, opened 2019-11-06
ProofMapIndex variants for hashed and raw keys [ECR-3777]

pavel-mukhanov:
New variant of the `ProofMapIndex` have been introduced - `RawProofMapIndex`. It is used for keys that maps directly to `ProofPath`, for example `Hash` and  `PublicKey`. By default `ProofMapIndex` is used for keys that implements `ObjectsHash`.
Also removed `ObjectHash` implementation for [u8; 32].


components/merkledb/src/proof_map_index/tests.rs
```diff
@@ -83,11 +85,46 @@ fn generate_random_data_keys<R: Rng>(
...
+make_tests! { test_remove_trivial { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_remove_simple { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_remove_reverse { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_clear { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_fuzz_insert { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_proof_in_empty_tree { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_multiproof_in_empty_tree { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_proof_in_single_node_tree { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_multiproof_in_single_node_tree { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_proof_in_complex_tree { (raw, Raw), (hashed, Hashed) } }
```
slowli: This test should fail for hashed `[u8; 32]` because it is specifically written with the assumption that the index keys are not transformed. Test does not fail because `[u8; 32]` has [identity `ObjectHash` implementation](https://github.com/exonum/exonum/blob/7547e85fe077a548c73caef574bee43e379ccb13/components/merkledb/src/hash.rs#L220), so in fact all `Raw` / `Hashed` test couples test the exact same thing 🙃 The `ObjectHash` implementation seems bogus to me; suggest to remove it.

CHANGELOG.md
```diff
@@ -78,6 +78,11 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `ProtobufConvert` has been implemented for `MapProof`. (#1512)
 
+- New variants of the `ProofMapIndex` has been introduced. First variant
```
skletsun: Nit: has -> have

components/merkledb/src/proof_map_index/key.rs
```diff
@@ -57,6 +57,73 @@ fn reset_bits(value: &mut u8, pos: u16) {
...
 }
 
+/// A trait that defines a subset of storage key types which are suitable for use with
+/// `ProofMapIndex`.
+///
+/// The size of the keys must be exactly [`PROOF_MAP_KEY_SIZE`] bytes and the keys must have
+/// a uniform distribution.
+///
+/// [`PROOF_MAP_KEY_SIZE`]: constant.PROOF_MAP_KEY_SIZE.html
+pub trait ProofMapKey
```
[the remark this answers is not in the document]
pavel-mukhanov: Forgot to remove it.

components/merkledb/src/proof_map_index/key.rs
```diff
@@ -57,6 +57,73 @@ fn reset_bits(value: &mut u8, pos: u16) {
...
+    /// The buffer is guaranteed to have size [`PROOF_MAP_KEY_SIZE`].
+    ///
+    /// [`PROOF_MAP_KEY_SIZE`]: constant.PROOF_MAP_KEY_SIZE.html
+    fn write_key(&self, to: &mut [u8]);
+
+    /// Reads this key from the buffer.
+    fn read_key(from: &[u8]) -> Self::Output;
+}
+
+#[allow(clippy::empty_enum)]
```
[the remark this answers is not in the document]
pavel-mukhanov: It's strange that clippy suggests this because `!` is currenty experimental.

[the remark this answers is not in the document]
popzxc: Why not use just empty struct (`struct Hashed;`)?

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -202,11 +209,20 @@ impl BinaryAttribute for Option<ProofPath> {
     }
 }
 
-impl<T, K, V> ProofMapIndex<T, K, V>
+/// Hashed variant of the `ProofMapIndex`, useful for arbitrary keys that implement
+/// `ObjectHash` trait.
+pub type HashedProofMap<T, K, V> = ProofMapIndex<T, K, V, Hashed>;
```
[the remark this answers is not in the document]
pavel-mukhanov: Hm, its strange but complier could infer type in the case below:
```rust
let mut table: ProofMapIndex<&Fork, Hash, Vec<u8>> = ProofMapIndex::new(NAME, &storage);
```
But couldn't infer when other type params is not specified.

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -202,11 +209,20 @@ impl BinaryAttribute for Option<ProofPath> {
...
+
+/// Raw variant of the `ProofMapIndex`, useful for keys that mapped directly to
+/// `ProofPath`. For example `Hash` and `PublicKey`.
+pub type RawProofMap<T, K, V> = ProofMapIndex<T, K, V, Raw>;
```
popzxc: I don't like `RawProofMap` name. "Raw" is usually treated like something low-level, but here it means only the key hashing approach.

IMHO, it'll be better to call current `ProofMapIndex` a `RawProofMapIndex` (since it requires us to provide a `Style` parameter), and this one (as covering most of the cases) to rename to the `ProofMapIndex`.

As a result, for situations where key should be hashed, a `ProofMapIndex` will be used (which is consistent and familiar), and for situations where key is already "hashed" it will become a `HashedProofMapIndex`.

components/merkledb/src/proof_map_index/mod.rs
popzxc: I don't like inconsistency of all the indices having "*Index" postfix and "ProofMap" going out of this convention.

components/merkledb/src/proof_map_index/key.rs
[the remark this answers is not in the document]
pavel-mukhanov: I've added empty struct as you suggest, but I haven't commit it yet.

components/merkledb/tests/common/mod.rs
```diff
@@ -19,18 +19,54 @@ use proptest::test_runner::TestCaseResult;
...
+        Self(buf)
+    }
+}
+
 pub trait FromFork {
     fn from_fork(fork: Rc<Fork>) -> Self;
     fn clear(&mut self);
 }
 
+#[allow(dead_code)]
```
popzxc: Why?

pavel-mukhanov: As far as I understand every module in external tests directory links with merkledb lib separately.  Not all code in `common.rs` used in all tests and that's why i need to mark it with `dead_code`.

popzxc: Maybe then add a comment, so developer who'll see it in the future won't attempt to remove it?

components/merkledb/src/proof_map_index/mod.rs
pavel-mukhanov: Agree, but my variant is slightly different. I've changed 
```
ProofMapIndex -> ProofMapIndexBase
HashedProofMap -> ProofMapIndex
RawProofMap -> RawProofMapIndex
```

components/merkledb/src/proof_map_index/tests.rs
```diff
@@ -112,21 +149,23 @@ fn test_map_methods() {
     index.clear();
 
     assert!(!index.contains(&[2; 32]));
-    assert!(!index.contains(&[3; 32]));
+    assert!(!index.contains(&[3; 32]))
```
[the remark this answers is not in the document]
pavel-mukhanov: No reason, reverted it back.

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -825,7 +845,7 @@ where
     /// assert!(!index.contains(&hash));
     /// ```
     pub fn remove(&mut self, key: &K) {
-        let proof_path = ProofPath::new(key);
+        let proof_path = Style::transform_key(key);
```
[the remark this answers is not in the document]
pavel-mukhanov: As @slowli pointed out, I've removed `transform_key` from `MerklePatriciaTree` and changed
`create_proof` signature to pair of `(ProofPath, K)`.

components/merkledb/src/proof_map_index/tests.rs
```diff
@@ -83,11 +85,46 @@ fn generate_random_data_keys<R: Rng>(
     (0..len).map(kv_generator).collect::<Vec<_>>()
 }
 
-#[test]
-fn test_map_methods() {
+macro_rules! make_tests {
```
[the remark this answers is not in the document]
pavel-mukhanov: I've removed `make_tests` macro.

components/merkledb/src/proof_map_index/tests.rs
```diff
@@ -683,11 +751,13 @@ fn test_build_multiproof_in_single_node_tree() {
     check_map_multiproof(&proof, keys, &table);
 }
 
-#[test]
-fn test_build_proof_in_complex_tree() {
+fn test_build_proof_in_complex_tree<S>()
+where
+    S: KeyTransform<[u8; 32]>,
```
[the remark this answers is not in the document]
pavel-mukhanov: I've implemented your variant and IMO it's looks better than separate methods.

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -987,7 +1008,7 @@ where
     V: BinaryValue + fmt::Debug,
 {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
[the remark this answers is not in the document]
pavel-mukhanov: Good idea, but it's not clear how to do that since style is type parameter.

components/merkledb/src/proof_map_index/mod.rs
popzxc: Yeah, that's OK.

components/merkledb/src/proof_map_index/tests.rs
```diff
@@ -83,11 +85,46 @@ fn generate_random_data_keys<R: Rng>(
...
+make_tests! { test_build_multiproof_in_empty_tree { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_proof_in_single_node_tree { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_multiproof_in_single_node_tree { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_proof_in_complex_tree { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_build_multiproof_simple { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_fuzz_delete { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_fuzz_insert_after_delete { (raw, Raw), (hashed, Hashed) } }
+make_tests! { test_iter { (raw, Raw), (hashed, Hashed) } }
+
+fn test_map_methods<S>()
```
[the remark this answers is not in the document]
pavel-mukhanov: Good point, created task for that(ECR-3825), because I think it's out of scope of this PR.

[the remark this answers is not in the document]
popzxc: Sure

components/merkledb/src/proof_map_index/key.rs
```diff
@@ -57,6 +57,42 @@ fn reset_bits(value: &mut u8, pos: u16) {
     *value &= reset_bits_mask;
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Hashed;
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Raw;
+
+/// Trait defines key transforming function used to transform key to `ProofPath`.
```
alekseysidorov: What do you think about to use `ToProofPath` name?

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -202,11 +197,27 @@ impl BinaryAttribute for Option<ProofPath> {
...
+/// `ProofMapIndex` requires that keys implement the [`BinaryKey`] trait and
+/// values implement the [`BinaryValue`] trait.
+///
+/// [`BinaryKey`]: ../trait.BinaryKey.html
+/// [`BinaryValue`]: ../trait.BinaryValue.html
+pub type ProofMapIndex<T, K, V> = ProofMapIndexBase<T, K, V, Hashed>;
+
+/// Raw variant of the `ProofMapIndex`, useful for keys that mapped directly to
+/// `ProofPath`. For example `Hash` and `PublicKey`.
+pub type RawProofMapIndex<T, K, V> = ProofMapIndexBase<T, K, V, Raw>;
```
alekseysidorov: It may be clearest to use `Transparent` prefix instead of `Raw`.

components/merkledb/tests/proof_map_index.rs
```diff
@@ -41,7 +41,8 @@ use std::{
     ops::{Range, RangeInclusive},
 };
 
-use exonum_merkledb::{BinaryValue, ObjectHash};
+mod common;
+use crate::common::Key;
```
popzxc: Nit: Usually we place module definitions after `use` section.

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -67,30 +69,23 @@ where
...
-/// keys.
-///
-/// `ProofMapIndex` implements a Merkle Patricia tree, storing values as leaves.
-/// `ProofMapIndex` requires that keys implement the [`BinaryKey`] trait and
-/// values implement the [`BinaryValue`] trait.
-///
-/// [`BinaryKey`]: ../trait.BinaryKey.html
-/// [`BinaryValue`]: ../trait.BinaryValue.html
-pub struct ProofMapIndex<T: IndexAccess, K, V> {
+#[doc(hidden)]
```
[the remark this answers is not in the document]
pavel-mukhanov: Not possible because we can't leak private or crate-visible type.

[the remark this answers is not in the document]
popzxc: Ah, sure.

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -754,11 +773,7 @@ where
                 if i < proof_path.len() {
                     let mut branch = BranchNode::empty();
                     branch.set_child(proof_path.bit(i), &proof_path.suffix(i), &leaf_hash);
-                    branch.set_child(
-                        prefix_path.bit(i),
-                        &prefix_path.suffix(i),
-                        &prefix_data.object_hash(),
-                    );
+                    branch.set_child(prefix_path.bit(i), &prefix_path.suffix(i), &prefix_data);
```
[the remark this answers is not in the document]
pavel-mukhanov: Calling `object_hash` here is redundant, because `prefix_data` has type `Hash`.  In other places we use hash directly.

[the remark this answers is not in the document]
popzxc: Got it.

components/merkledb/src/proof_map_index/mod.rs
pavel-mukhanov: Not sure about this, `Raw` is much shorter.

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -67,30 +68,23 @@ where
...
-///
-/// `ProofMapIndex` implements a Merkle Patricia tree, storing values as leaves.
-/// `ProofMapIndex` requires that keys implement the [`BinaryKey`] trait and
-/// values implement the [`BinaryValue`] trait.
-///
-/// [`BinaryKey`]: ../trait.BinaryKey.html
-/// [`BinaryValue`]: ../trait.BinaryValue.html
-pub struct ProofMapIndex<T: IndexAccess, K, V> {
+#[doc(hidden)]
+pub struct ProofMapIndexBase<T: IndexAccess, K, V, KeyMode: ToProofPath<K>> {
```
slowli: I still think that having base a separate type rather than defining `ProofMapIndex` with a default `KeyMode` is unwarranted (could be solved later, I suppose). Right now, it could be fixed by introducing separate constructors for default / any `KeyMode`; with #1523 merged, it would be solved by having 2 separate methods in `AccessExt` for `ProofMapIndex`es with hashed / raw keys.

components/merkledb/src/proof_map_index/proof.rs
```diff
@@ -260,16 +262,25 @@ impl<K, V> OptionalEntry<K, V> {
...
+    #[serde(skip)]
+    _key_mode: PhantomData<KeyMode>,
 }
 
+/// Variant of [`MapProof`] with raw keys, in other words with keys that mapped
+/// directly to [`ProofPath`].
+///
+/// [`MapProof`]: struct.MapProof.html
+/// [`ProofPath`]: struct.ProofPath.html
+pub type RawMapProof<K, V> = MapProof<K, V, Raw>;
```
slowli: Nit / subjective: Is this type alias really necessary? IMO, `MapProof<K, V, Raw>` is just as readable.

components/merkledb/src/proof_map_index/proof_builder.rs
```diff
@@ -158,22 +161,23 @@ impl ContourNode {
 /// implement `BuildProof` as well.
 ///
 /// [`MerklePatriciaTree`]: trait.MerklePatriciaTree.html
-pub trait BuildProof<K, V> {
+pub trait BuildProof<K, V, KeyMode> {
     /// Creates a proof of existence / absence for a single key.
-    fn create_proof(&self, key: K) -> MapProof<K, V>;
+    fn create_proof(&self, searched_path: ProofPath, key: K) -> MapProof<K, V, KeyMode>;
```
slowli: Since `BuildProof` is parametric by `KeyMode`, converting keys to paths outside of it seems an overkill. This is especially true for the `create_multiproof` method, which may involve reallocation (or at least it seemingly does now, if the compiler doesn't eliminate it).

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -430,10 +447,14 @@ where
...
     /// ```
-    pub fn get_multiproof<KI>(&self, keys: KI) -> MapProof<K, V>
+    pub fn get_multiproof<KI>(&self, keys: KI) -> MapProof<K, V, KeyMode>
     where
         KI: IntoIterator<Item = K>,
     {
+        let keys = keys
+            .into_iter()
+            .map(|key| (KeyMode::transform_key(&key), key))
+            .collect::<Vec<_>>();
```
slowli: Collecting into a `Vec` seems unnecessary.

components/merkledb/tests/map.rs
```diff
@@ -142,8 +143,8 @@ fn compare_map(map: &MapIndex<Rc<Fork>, u8, i32>, ref_map: &HashMap<u8, i32>) ->
 }
 
 fn compare_proof_map(
```
slowli: Off-topic: The `compare_proof_list_to_vec` test at the bottom of the file is named incorrectly (the correct name should be something like `compare_proof_map_to_hash_map`). Could you fix it?

components/merkledb/tests/common/mod.rs
```diff
@@ -19,18 +19,61 @@ use proptest::test_runner::TestCaseResult;
...
+use exonum_crypto::{hash, Hash, HASH_SIZE};
+use exonum_merkledb::{BinaryKey, Database, Fork, ObjectHash, TemporaryDB};
 
 // Max size of the generated sequence of actions.
+//
+// Due external tests running mechanism this file is linked
+// separately with other modules in tests directory. This
+// constant is used in all of then except `proof_map_index`, that's
+// why it marked with `dead_code`.
+#[allow(dead_code)]
```
slowli: Nit: it make sense to move `Key` to a separate module in order to not force `#[allow(dead_code)]`.

components/merkledb/tests/proof_map_index.rs
```diff
@@ -141,35 +155,41 @@ fn index_data(
     btree_map(array::uniform32(key_bytes), any::<u64>(), sizes)
 }
 
+fn data_for_absent(key_bytes: RangeInclusive<u8>) -> impl Strategy<Value = Vec<Key>> {
+    vec(array::uniform32(key_bytes), 20).prop_map(|key| vec![Key(*key.get(0).unwrap())])
```
slowli: This transform is seemingly incorrect: why are all keys ignored except for the first one?

components/merkledb/tests/proof_map_index.rs
```diff
@@ -68,15 +71,16 @@ where
     Ok(())
 }
 
-fn check_map_multiproof<T, K, V>(
-    proof: &MapProof<K, V>,
+fn check_map_multiproof<T, K, V, S>(
+    proof: &MapProof<K, V, S>,
     keys: BTreeSet<&K>,
-    table: &ProofMapIndex<T, K, V>,
+    table: &ProofMapIndexBase<T, K, V, S>,
```
slowli: This test sort of relies on the assumption that keys are not transformed. Hence, there are special test cases for keys with limited key space (`TestParams.key_bytes` field), which (if the keys are not transformed) leads to a particular structure of the Merkle Patricia tree. I'd suggest to use untransformed keys (e.g., `Hash`) in this test.

components/merkledb/tests/map.rs
pavel-mukhanov: Yes, sure.

examples/timestamping/backend/src/api.rs
```diff
@@ -45,7 +45,8 @@ pub struct TimestampProof {
     /// Actual state hashes of the timestamping service with their proofs.
     pub state_proof: MapProof<IndexCoordinates, Hash>,
     /// Actual state of the timestamping database with proofs.
-    pub timestamp_proof: MapProof<Hash, TimestampEntry>,
+    //TODO: revert change to Raw after refactoring
```
slowli: What's wrong with the current form of the proof?

examples/cryptocurrency-advanced/backend/src/api.rs
```diff
@@ -37,7 +37,8 @@ pub struct WalletProof {
     /// Proof of the whole database table.
     pub to_table: MapProof<IndexCoordinates, Hash>,
     /// Proof of the specific wallet in this table.
-    pub to_wallet: MapProof<PublicKey, Wallet>,
+    //TODO: revert change to Raw
```
slowli: Will this task be performed separately? Also, not sure why it is necessary. (Do we consider `RawProofMapIndex` too difficult for a tutorial / demo?)

components/merkledb/tests/proof_map_index.rs
```diff
@@ -142,35 +154,42 @@ fn index_data(
     btree_map(array::uniform32(key_bytes), any::<u64>(), sizes)
 }
 
+fn data_for_absent(key_bytes: RangeInclusive<u8>) -> impl Strategy<Value = Vec<Key>> {
```
slowli: Nit: `absent_keys` could be a better name.

components/merkledb/tests/proof_map_index.rs
```diff
@@ -142,35 +154,42 @@ fn index_data(
...
+    vec(array::uniform32(key_bytes), 20)
```
slowli: Nit: `vec(array::uniform32(key_bytes).prop_map(Key), 20)` is shorter and more understandable.

components/merkledb/tests/proof_map_index.rs
```diff
@@ -142,35 +154,42 @@ fn index_data(
...
+) -> impl Strategy<Value = (Vec<Key>, Data)> {
     index_data(key_bytes, sizes)
         .prop_flat_map(|data| (vec(0..data.len(), data.len() / 5), Just(data)))
         .prop_map(|(indexes, data)| {
             // Note that keys may coincide; this is intentional.
-            let keys: Vec<_> = indexes
+            let keys: Vec<Key> = indexes
                 .into_iter()
                 .map(|i| *data.keys().nth(i).unwrap())
+                .map(|key| key.into())
```
slowli: Nit: Shouldn't it be possible to use `.map(Key)` here?

components/merkledb/tests/proof_map_index.rs
```diff
@@ -223,7 +242,7 @@ impl TestParams {
         let data_strategy = index_data(self.key_bytes(), self.index_sizes());
         proptest!(self.config(), |(key in key_strategy, data in data_strategy)| {
```
slowli: Nit: it could make sense to add `.prop_map(Key)` to `key_strategy` variables.

CHANGELOG.md
```diff
@@ -84,6 +84,11 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `ProtobufConvert` has been implemented for `MapProof`. (#1512)
 
+- New variant of the `ProofMapIndex` have been introduced - `RawProofMapIndex`.
+  It is used for keys that maps directly to `ProofPath`, for example `Hash` and
+  `PublicKey`. By default `ProofMapIndex` is used for keys that implements
+  `ObjectsHash`. (#1531)
```
slowli: Do I understand correctly that both key types result in the same result for `Hash` keys? It may make sense to mention it and possibly introduce an alias for `ProofMapIndex<T, Hash, V>` (like `HashMapIndex<T, V>`) to make this more transparent. (Although it's up to discussion whether this actually simplifies things.)

examples/timestamping/backend/src/api.rs
pavel-mukhanov: Nothing is wrong, forgot to remove this todo.

examples/cryptocurrency-advanced/backend/src/api.rs
pavel-mukhanov: No, it's debug artefact, forgot to remove it.

components/merkledb/src/proof_map_index/mod.rs
```diff
@@ -858,27 +868,33 @@ where
     }
 }
 
-impl<T, K, V> fmt::Debug for ProofMapIndex<T, K, V>
+impl<T, K, V, KeyMode> fmt::Debug for ProofMapIndex<T, K, V, KeyMode>
 where
     T: RawAccess,
     K: BinaryKey + ObjectHash,
```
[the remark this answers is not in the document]
pavel-mukhanov: Yes, bacause `BuildProof` requires it.

CHANGELOG.md
```diff
@@ -84,6 +84,15 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `ProtobufConvert` has been implemented for `MapProof`. (#1512)
 
+- New variant of the `ProofMapIndex` have been introduced - `RawProofMapIndex`.
+  It is used for keys that maps directly to `ProofPath`, for example `Hash` and
+  `PublicKey`. (#1531)
+
+  - By default `ProofMapIndex` is used for keys that implements `ObjectsHash`.
```
```suggestion
  - By default `ProofMapIndex` is used for keys that implements `ObjectHash`.
```

components/merkledb/src/access/extensions.rs
```diff
@@ -115,6 +116,40 @@ pub trait AccessExt: Access {
...
+    /// // Hashed variant for keys implementing `ObjectHash`.
+    /// let hashed_map: ProofMapIndex<&Fork, u32, u32, Hashed> = fork.get_gen_proof_map("hashed");
+    ///
+    /// // Raw variant for keys that maps directly to `ProofPath`.
+    /// let raw_map: ProofMapIndex<&Fork, PublicKey, u32, Raw> = fork.get_gen_proof_map("raw");
+    /// ```
+    /// # Panics
+    ///
+    /// If the index exists, but is not a Merkelized map.
+    fn get_gen_proof_map<I, K, V, KeyMode>(
```
popzxc: Maybe name it `get_generic_proof_map`?

components/merkledb/src/access/extensions.rs
```diff
@@ -115,6 +116,40 @@ pub trait AccessExt: Access {
...
+    /// proof_map_index::{Raw, Hashed}};
+    /// use exonum_crypto::PublicKey;
+    ///
+    /// let db = TemporaryDB::new();
+    /// let fork = db.fork();
+    ///
+    /// // Hashed variant for keys implementing `ObjectHash`.
+    /// let hashed_map: ProofMapIndex<&Fork, u32, u32, Hashed> = fork.get_gen_proof_map("hashed");
+    ///
+    /// // Raw variant for keys that maps directly to `ProofPath`.
```
popzxc:
```suggestion
    /// // Raw variant for keys that can be mapped directly to `ProofPath`.
```

components/merkledb/src/access/extensions.rs
popzxc: And, by the way, do we need it at all? Maybe it'll be better to have 2 methods: `get_proof_map` and `get_raw_proof_map`?

components/merkledb/src/proof_map_index/key.rs
```diff
@@ -57,6 +57,44 @@ fn reset_bits(value: &mut u8, pos: u16) {
     *value &= reset_bits_mask;
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+/// Hashed variant of proof map key.
```
popzxc: Nit: usually we place doc-strings above the derive and proc-macro section.

components/merkledb/src/proof_map_index/key.rs
```diff
@@ -57,6 +57,44 @@ fn reset_bits(value: &mut u8, pos: u16) {
...
+pub struct Hashed;
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+/// Raw variant of proof map key.
+pub struct Raw;
+
+/// Trait defines key transforming function used to transform key to `ProofPath`.
```
popzxc:
```suggestion
/// Trait defining key transforming function used to transform key to `ProofPath`.
```

components/merkledb/src/proof_map_index/mod.rs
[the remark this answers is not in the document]
popzxc: Hm. And why does it require it? Shouldn't it use `KeyMode::transform` instead?

components/merkledb/src/access/extensions.rs
pavel-mukhanov: Method `get_raw_proof_map` has been used only once in tests. And also in new `schema` implementations we doesn't use these methods and use `FromAccess::from_acces` instead.

popzxc: Well, we [have](https://github.com/exonum/exonum/blob/dynamic_services/exonum/src/blockchain/schema.rs) places where `get_index_type` methods are used (and generally the same will apply to any situation when eager initialization doesn't fit ones needs), so IMHO it'll be better to have `get_*` methods for any type of index, just to be consistent.

popzxc: We have a method `get_proof_map` (and I don't think that it should be removed), so `get_generic_proof_map` is actually needed only for a raw variant.

pavel-mukhanov: `get_generic_proof_map` is also needed for tests and that why I've created `TestsExt` trait, but it seems to be a bad idea.

slowli: The request to refactor the code was mine, so I'll chime in.

> `get_generic_proof_map` is actually needed only for a raw variant.

Not really, since there are no restrictions on implementing custom key styles externally (should there be ones?). I'd argue that schema-like code will either use `FromAccess` (through a derive macro once it's implemented), or have code like this:

```rust
struct Schema<T>(T);

impl<T: Access> Schema<T> {
    pub fn some_map(&self) -> RawProofMapIndex<T::Base, PublicKey, u64> {
        self.0.clone().get_generic_proof_map("some_map")
        // the return type is known, so there's no difference between
        // `generic` and `raw` indexes from the UX perspective.
    }
}
```

popzxc: Not sure if I understand. Why it can't be `self.0.clone().get_raw_proof_map("some_map")` if the return type is already `RawProofMapIndex`?

PR #1540, opened 2019-11-13
Deploy workflow

slowli:
This PR describes artifact deployment in greater detail and adds some tests covering deployment.


exonum/src/runtime/dispatcher/tests.rs
```diff
@@ -568,3 +573,330 @@ fn test_shutdown() {
...
+
+#[test]
+fn failed_deployment_with_node_restart() {
+    let db = Arc::new(TemporaryDB::new());
+    let runtime = DeploymentRuntime::default();
+    let db_ = Arc::clone(&db);
+    let runtime_ = runtime.clone();
+    // The node should stop after unsuccessful sync deployment.
+    panic::catch_unwind(|| test_failed_deployment(db_, runtime_, "recoverable_after_restart"))
+        .unwrap_err();
```
popzxc: Maybe `expect_err` for better description if this code won't panic?

exonum/src/runtime/dispatcher/tests.rs
```diff
@@ -568,3 +573,330 @@ fn test_shutdown() {
...
+    let mut dispatcher = DispatcherBuilder::new()
+        .with_runtime(2, runtime.clone())
+        .finalize(&blockchain);
+
+    let mut fork = db.fork();
+    create_genesis_block(&mut dispatcher, &mut fork);
+    db.merge_sync(fork.into_patch()).unwrap();
+
+    // Queue an artifact for deployment.
+    let artifact = ArtifactId {
```
popzxc: Nit: code from
```rust
// Queue an artifact for deployment.
```

till

```
db.merge_sync(fork.into_patch()).unwrap();
```

is almost copy-pasted several times. Maybe it worth moving into a separate function?

exonum/src/runtime/mod.rs
```diff
@@ -41,19 +41,40 @@
...
-//! 3. Once the artifact is deployed, it is possible to instantiate a corresponding service.
+//! 3. For each node, an artifact may be deployed either asynchronously or synchronously /
+//!   in a blocking manner. The supervisor usually first commands a node to deploy the artifact
+//!   asynchronously via [`Mailbox`] once the decision to start deployment is reached
+//!   by the blockchain administrators.
+//!
+//! 4. Async deployment usually has a deadline and/or success and failure conditions. As an example,
+//!   the supervisor may collect confirmations from the validator nodes that have
+//!   successfully deployed the artifact, and once all the validator nodes have sent
+//!   their confirmations, the artifact is *committed*.
```
dmitry-timofeev: As there is no 'undeploy' operation in case when some nodes failed to deploy the artifact, is that correct that the runtimes do not know if and when the artifact is actually committed and just rely on the (?) dispatcher issuing `start_adding_service` only if the conditions are met?

What would happen if
1. Most nodes deployed an artifact correctly, but a couple of nodes failed (e.g., the admins forgot to put the artifact where it belongs). As a result, the runtimes on the 'good' nodes have it deployed; on the 'bad' — don't. The runtimes on the 'good' don't know that it was not committed.

2. The administrators attempt a re-deploy. This time, as the 'good' nodes already have this artifact, the operation fails there (just because of an illegal attempt to deploy it twice); and succeeds on the 'bad' ones.

exonum/src/runtime/mod.rs
```diff
@@ -294,10 +333,12 @@ pub trait Runtime: Send + fmt::Debug + 'static {
     ///
     /// # Return value
     ///
-    /// Any error or panic returned from this method should be considered fatal. There are edge
-    /// cases where the returned error does not stop the enclosing process (e.g.,
-    /// if several alternative initial service configurations are tried), but as a rule of thumb,
-    /// a `Runtime` should not return an error or panic here unless it wants the node to stop forever.
+    /// An error or panic returned from this method will not be processed and will lead
+    /// to the node stopping. A runtime should only return an error / panic if the error is local
```
dmitry-timofeev: Am I right that the Java runtime that performs _some_ user and framework code (= same on all nodes and hopefully mostly deterministic — we can't say that of all the framework and library code) that is not perfomed in `start_adding_service` (but is not considered likely to fail) violates this restriction? But respects

> The runtime should commit long-term resources for the service after a `commit_service()` call.

Is that fine given small/unknown likelyhood? Or shall we attempt to do everything in `start_adding_services` and then just flip a switch (that would complicate things with clean ups in after_commit, and, if we need that, a restriction of API operation of a newly added service before commit_service; oh, and with no `start_adding_` for restarts, it will have to operate differently for newly added and restarted services)?

slowli: > Am I right that the Java runtime [...] violates this restriction?

There is no 100% non-violable restriction here, hence the use of "should" instead of "must" as per [RFC 2119](https://tools.ietf.org/html/rfc2119). The balance between delaying resource commitment and potentially ending up with the non-functional blockchain is up to the runtime. Subjectively, I don't think it's reasonable to assume that the probability of the service developer screwing up HTTP API (but not screwing up anything else) is high enough to complicate workflow because of it.

dmitry-timofeev: Thanks, I agree it is reasonable to not complicate the workflow for highly unlikely scenarious (also given the user tests their code with testkit).

exonum/src/runtime/mod.rs
dmitry-timofeev: So, "consensus-agreed result" = failure -> deploy of this particular artifact failed in the whole network, but 'good' runtimes will _retain_ the artifact as deployed till _restarted_ -> the redeploy of the same artifact is impossible without restart, is that correct?

I think a deploy error on some but all nodes is a realistic use-case (but the one we can probably agree to ignore in the upcoming release). Would a 'Runtime#unloadArtifact' operation, that the dispatcher invokes when the operation failed in terms of network but completed successfully on this node (i.e., if I remember the current policy correctly, failed on _any_ single node), help? Or `Runtime#unloadIfPresent` if we'd like the dispatcher to not have local data on whether the local deployment was successful — but it might already have that.

dmitry-timofeev: > the redeploy of the same artifact is impossible without restart

@slowli , @aleksuss Is this limitation planned to be addressed?

aleksuss: AFAIK yes, we plan it.

PR #1541, opened 2019-11-14
Extract genesis config and improve adding builtin services [ECR-3750]

skletsun:
## Overview

- Genesis config is extracted  into separated entity.
- Improved adding builtin services.

**UPD**: Current status (since [c7abbea]):

- [x] Add ability to specify default instances for `ServiceFactory`.
- [x] Remove `InstanceConfig`.
- [x] Fix majourity of relatively small comments.
- [x] Refactor `Node::new()`.
- [x] `WellKnownRuntime` and the rest of tasks from ECR-3744.
- [ ] Get rid of `InstanceCollection`.

---

See: https://jira.bf.local/browse/ECR-3750, https://jira.bf.local/browse/ECR-3690.

### Definition of Done


exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,135 @@ impl ValidateInput for ConsensusConfig {
...
+    /// Consensus config.
+    consensus_config: ConsensusConfig,
+    /// Artifacts specifications for builtin services.
+    artifacts: Vec<ArtifactSpec>,
+    /// Instances of builtin services.
+    builtin_instances: Vec<ConfiguredInstanceSpec>,
+}
+
+#[derive(Debug)]
+pub struct InstanceConfig {
```
alekseysidorov: Imo name is a little bit weird, what do you think about `ArtifactConfig` or `ArtifactDeployment` or something else?

@popzxc any suggestions?

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,135 @@ impl ValidateInput for ConsensusConfig {
...
+    ) -> Self {
+        assert_eq!(self.artifact_spec.artifact, instance_spec.artifact);
+        self.instances.push(ConfiguredInstanceSpec {
+            instance_spec,
+            constructor: constructor.into_bytes(),
+        });
+        self
+    }
+
+    pub fn from_spec(instance_spec: InstanceSpec, constructor: impl BinaryValue) -> Self {
```
alekseysidorov: By this method there is no way to set deploy args, is this a correct approach?

skletsun: Yes. The main intention of this method was using in tests for converting in tests. I wanted to reflect this in the documentation, but forgot. Now I see that this method is now such useful, so probably it will be better to get rid of it.

exonum/src/blockchain/config.rs
skletsun: Yes, you're welcomed to propose your options in new structures naming because I also not quite happy with current naming.

popzxc: `ArtifactBootstrap`?

exonum/src/blockchain/builder.rs
```diff
@@ -84,15 +63,6 @@ impl BlockchainBuilder {
         self
     }
```
skletsun: @alekseysidorov @popzxc what about `.with_additional_runtime()` -> `.with_runtime()`?

skletsun:
There are some minor tasks left (like write docs, rename structs), but the PR, in general, is ready for review.

exonum/src/helpers/mod.rs
```diff
@@ -156,6 +162,29 @@ pub fn byzantine_quorum(total: usize) -> usize {
     total * 2 / 3 + 1
 }
 
+/// Creates and initializes RustRuntime and GenesisConfig with information from collection of InstanceCollection.
+pub fn create_rust_runtime_and_genesis_config(
```
popzxc: Not sure about this combination. I think that genesis config and rust runtime with its instances should be created independently.

exonum/src/helpers/mod.rs
```diff
@@ -156,6 +162,29 @@ pub fn byzantine_quorum(total: usize) -> usize {
...
+    api_notifier: mpsc::Sender<UpdateEndpoints>,
+    consensus_config: ConsensusConfig,
+    instances: impl IntoIterator<Item = InstanceCollection>,
+) -> (RustRuntime, GenesisConfig) {
+    let (rust_runtime, genesis_config_builder) = instances.into_iter().fold(
```
popzxc: IMHO that construction is too complex for `fold`, I suggest to split it into several lines and use a loop.

exonum/src/blockchain/mod.rs
```diff
@@ -226,24 +231,33 @@ impl BlockchainMut {
...
 
         // Add service instances.
-        for instance_config in initial_services {
-            self.dispatcher.add_builtin_service(
-                &mut fork,
-                instance_config.instance_spec,
-                instance_config.artifact_spec.unwrap_or_default(),
-                instance_config.constructor,
-            )?;
+        for ConfiguredInstanceSpec {
```
popzxc: We don't really often (if do at all) use unpacking in exonum, so this code looks a bit inconsistent. Could it be avoided (here and in other places)?

exonum/src/blockchain/mod.rs
```diff
@@ -226,24 +231,33 @@ impl BlockchainMut {
...
+        } = genesis_config;
+        consensus_config.validate()?;
         let mut fork = self.fork();
-        Schema::new(&fork).consensus_config_entry().set(config);
+        Schema::new(&fork)
+            .consensus_config_entry()
+            .set(consensus_config);
+
+        for ArtifactSpec { artifact, payload } in artifacts {
+            if !self.dispatcher.is_artifact_deployed(&artifact) {
```
popzxc: Is it possible to have deployed artifact before creating the genesis block?

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+    pub fn with_builtin_instances(
+        self,
+        instance_configs: impl IntoIterator<Item = InstanceConfig>,
+    ) -> Self {
+        instance_configs
+            .into_iter()
+            .fold(self, |s, instance| s.with_builtin_instance(instance))
+    }
+
+    pub fn build(self) -> GenesisConfig {
```
popzxc: I'd suggest `finalize`.

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+    pub fn with_consensus_config(consensus_config: ConsensusConfig) -> Self {
+        Self {
+            consensus_config,
+            artifacts: vec![],
+            builtin_instances: vec![],
+        }
+    }
+
+    pub fn with_builtin_instance(mut self, instance_config: InstanceConfig) -> Self {
+        if !instance_config.instances.is_empty() {
```
popzxc: I'd suggest to add `with_artifact` method to avoid implicit artifacts creation.

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+        }
+        self.builtin_instances.extend(instance_config.instances);
+        self
+    }
+
+    pub fn with_builtin_instances(
+        self,
+        instance_configs: impl IntoIterator<Item = InstanceConfig>,
+    ) -> Self {
+        instance_configs
```
popzxc: If artifact addition will be done in the separate method, this method could be simplified.

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+            instances: vec![],
+        }
+    }
+
+    pub fn with_instance(
+        mut self,
+        instance_spec: InstanceSpec,
+        constructor: impl BinaryValue,
+    ) -> Self {
+        assert_eq!(self.artifact_spec.artifact, instance_spec.artifact);
```
popzxc: `debug_assert`?

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+    Hash,
+    Serialize,
+    Deserialize,
+    ProtobufConvert,
+    BinaryValue,
+    ObjectHash,
+)]
+#[protobuf_convert(source = "runtime::ConfiguredInstanceSpec")]
+// TODO: Think about better name and/or substitution the `InstanceSpec`
+pub struct ConfiguredInstanceSpec {
```
popzxc: `InstanceInitParameters`?

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+
+    /// Artifacts specification of builtin services.
+    pub artifacts: Vec<ArtifactSpec>, // TODO: HashSet? Duplications? Perhaps check in builder?
+
+    /// List of the privileged services with the configuration parameters that are created directly
+    /// in the genesis block.
+    pub builtin_instances: Vec<ConfiguredInstanceSpec>,
+}
+
+/// A wrapper for `InstanceSpec` that also stores a constructor argument for the service instance.
```
popzxc: I'd rephrase this docstring as "Data required for service instance initialization".

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+/// Used in `GenesisConfig`.
```
popzxc: Not sure if documentation should denote where some entity is used.

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+)]
+#[protobuf_convert(source = "runtime::GenesisConfig")]
+pub struct GenesisConfig {
+    /// Blockchain configuration used to create the genesis block.
+    pub consensus_config: ConsensusConfig,
+
+    /// Artifacts specification of builtin services.
+    pub artifacts: Vec<ArtifactSpec>, // TODO: HashSet? Duplications? Perhaps check in builder?
+
+    /// List of the privileged services with the configuration parameters that are created directly
```
popzxc: Why are they priveleged?

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,131 @@ impl ValidateInput for ConsensusConfig {
...
+    BinaryValue,
+    ObjectHash,
+)]
+#[protobuf_convert(source = "runtime::GenesisConfig")]
+pub struct GenesisConfig {
+    /// Blockchain configuration used to create the genesis block.
+    pub consensus_config: ConsensusConfig,
+
+    /// Artifacts specification of builtin services.
+    pub artifacts: Vec<ArtifactSpec>, // TODO: HashSet? Duplications? Perhaps check in builder?
```
popzxc: I think `Vec` is ok.

Regarding checks: you can impement a `validate` method similar to one in `consensus_config` which will check that (and call `consensus_config.validate()` as well).

exonum/src/blockchain/builder.rs
```diff
@@ -177,29 +121,48 @@ impl InstanceCollection {
...
+        let instance_config = ConfiguredInstanceSpec {
+            instance_spec: spec,
+            constructor: params.into_bytes(),
+        };
         self.instances.push(instance_config);
         self
     }
 }
 
+impl From<InstanceCollection> for (Box<dyn ServiceFactory>, InstanceConfig) {
```
popzxc: Seems like `InstanceCollection` now exists only to collect factories and instances together, and then split them back. Do we need it at all?

exonum/src/blockchain/builder.rs
vitvakatu: What about merging `with_runtime` and `with_external_runtime`?

exonum/src/blockchain/builder.rs
skletsun: I also noticed that, but didn't end up with decision if its worth refactoring within this PR.

exonum/src/blockchain/config.rs
skletsun: Sounds better, I'd even shorten it - `InstanceInitParams` (currently we have shortened `...Spec`).

exonum/src/helpers/mod.rs
skletsun: I also don't like it much either, but for the moment entities of different kind (service factory and wrapper of `InstanceSpec`) comes from single source (InstanceCollection).

popzxc: I believe it's one more argument for removing the `InstanceCollection` abstraction then.

exonum/src/helpers/mod.rs
popzxc: Yeah, for simple cases `fold` is better, I agree.

exonum/src/blockchain/config.rs
skletsun: Just had a quick overview of existing builders within the project - the majority has got `.build()` as a final method. "Irregular" are: `DispatcherBuiler` - `.finalize()`, `TestKitBuilder` - `.create()`,  `[Args|Node]Builder` - `.run()` (and some specific API-related builders that can produce different entities in result). While it's obvious for the last kind of builders, personally I don't see a big difference between `.build()` and `.finalize()`. Maybe `.build()` is more expected for a `WhateverBuilder` entity?

exonum/src/blockchain/mod.rs
skletsun: Well, perhaps I overused unpacking somewhere in code 🙂. But I'd like to keep it in loops because it saves us a temporary variable[s].

exonum/src/blockchain/config.rs
popzxc: Yeah, agree. Dunno why I suggested it ^_^"

examples/sample_runtime/src/main.rs
```diff
@@ -273,12 +275,15 @@ fn main() {
     println!("Creating blockchain with additional runtime...");
     // Create a blockchain with the Rust runtime and our additional runtime.
     let blockchain_base = Blockchain::new(db, service_keypair.clone(), api_sender.clone());
-    let blockchain = BlockchainBuilder::new(blockchain_base, genesis)
-        .with_rust_runtime(
-            channel.endpoints.0.clone(),
-            vec![InstanceCollection::from(Supervisor)],
-        )
-        .with_additional_runtime(SampleRuntime::default())
+    let (factory, cfg) = InstanceCollection::from(Supervisor).into();
```
slowli: Not a fan of this transform. IMO, something like this would be more readable:

```rust
let rust_runtime =
    RustRuntime::new(channel.endpoints.0.clone()).with_factory(Supervisor);
let genesis_config = GenesisConfigBuilder::with_consensus_config(genesis)
    .with_service(Supervisor::default_instance())
    .build();
```

exonum/benches/criterion/block.rs
```diff
@@ -47,14 +47,15 @@ use std::{collections::BTreeMap, iter, sync::Arc};
...
+        ConsensusConfig, InstanceCollection, ValidatorKeys,
     },
     crypto::{self, Hash, PublicKey, SecretKey},
-    helpers::{Height, ValidatorId},
+    helpers::{create_rust_runtime_and_genesis_config, Height, ValidatorId},
     messages::{AnyTx, Verified},
     node::ApiSender,
-    runtime::SnapshotExt,
+    runtime::rust::RustRuntime,
+    SnapshotExt,
```
slowli: ???

exonum/src/blockchain/builder.rs
```diff
@@ -79,20 +58,11 @@ impl BlockchainBuilder {
     }
 
     /// Add an additional runtime with the specified identifier.
-    pub fn with_additional_runtime(mut self, runtime: impl Into<(u32, Box<dyn Runtime>)>) -> Self {
+    pub fn with_runtime(mut self, runtime: impl Into<(u32, Box<dyn Runtime>)>) -> Self {
```
slowli: Could you replace `Into<(u32, Box<dyn Runtime>)>` with a more specific trait? E.g.,

```rust
trait WellKnownRuntime: Runtime {
    const ID: u32;
}
```

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,130 @@ impl ValidateInput for ConsensusConfig {
...
+impl GenesisConfigBuilder {
+    pub fn with_consensus_config(consensus_config: ConsensusConfig) -> Self {
+        Self {
+            consensus_config,
+            artifacts: vec![],
+            builtin_instances: vec![],
+        }
+    }
+
+    pub fn with_builtin_instance(mut self, instance_config: InstanceConfig) -> Self {
```
slowli: Nit: As far as I understand, right now this method adds *several* instances in general. How often is creating multiple instances of the same service will be performed? (I'd wager, not very.) Thus, it could make sense to make this method add a single service instance, which is presumably returned in a conventional way (e.g., via `Factory::default_instance()` as I've commented above).

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,130 @@ impl ValidateInput for ConsensusConfig {
...
+        if !instance_config.instances.is_empty() {
+            self.artifacts.push(instance_config.artifact_spec);
+        }
+        self.builtin_instances.extend(instance_config.instances);
+        self
+    }
+
+    pub fn with_builtin_instances(
```
slowli: I'd suggest to remove this method; its uses can be trivially simplified to use `with_builtin_instance`.

exonum/src/blockchain/config.rs
slowli: I'd suggest to add this method, but for a different reason than suggested by @popzxc: it may make sense to deploy artifacts on the genesis block without creating corresponding instances.

exonum/src/blockchain/config.rs
```diff
@@ -295,6 +296,130 @@ impl ValidateInput for ConsensusConfig {
...
+    /// Constructor argument for specific `InstanceSpec`.
+    pub constructor: Vec<u8>,
+}
+
+#[derive(Debug)]
+pub struct GenesisConfigBuilder {
+    /// Consensus config.
+    consensus_config: ConsensusConfig,
+    /// Artifacts specifications for builtin services.
+    artifacts: Vec<ArtifactSpec>,
```
slowli: It could make sense to verify (or ensure) that `artifacts` are unique, e.g., by using `HashMap<ArtifactId, Vec<u8>>` here. Similarly, it could make sense to verify uniqueness of instances during the building process.

exonum/src/blockchain/mod.rs
```diff
@@ -226,24 +231,28 @@ impl BlockchainMut {
...
 
         // Add service instances.
-        for instance_config in initial_services {
-            self.dispatcher.add_builtin_service(
-                &mut fork,
-                instance_config.instance_spec,
-                instance_config.artifact_spec.unwrap_or_default(),
-                instance_config.constructor,
-            )?;
+        for InstanceInitParams {
```
slowli: Nit: Multi-line destructuring harms readability, IMO.

exonum/src/blockchain/config.rs
skletsun: I took as-is from the `BlockchainBuilder` 😌. I guess the privilege is that they're deployed and started without corresponding transactions. Perhaps this could be reworded as "List of the services with its configuration parameters that are created directly in the genesis block".

exonum/src/node/mod.rs
```diff
@@ -950,8 +951,14 @@ impl Node {
             node_cfg.service_keypair(),
             ApiSender::new(channel.api_requests.0.clone()),
         );
-        let blockchain = BlockchainBuilder::new(blockchain, node_cfg.consensus.clone())
-            .with_rust_runtime(channel.endpoints.0.clone(), services)
+        let (rust_runtime, genesis_config) = create_rust_runtime_and_genesis_config(
```
slowli: I'd suggest to refactor `Node::new` for clarity. Ideally, it should accept `GenesisConfig` directly or via a field in `NodeConfig` (since the consensus config is already there). Likewise, service factories should be accepted instead of `InstanceCollection`s.

exonum/src/runtime/types.rs
```diff
@@ -227,7 +227,18 @@ impl FromStr for ArtifactId {
 }
 
 /// Exhaustive artifact specification.
-#[derive(Debug, Clone, PartialEq, Eq, Hash, ProtobufConvert, BinaryValue, ObjectHash)]
+#[derive(
```
slowli: Off-topic: Is it possible to do something to `rustfmt` for it to accept multi-line `derive`s like

```rust
#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
#[derive(ProtobufConvert, BinaryValue, ObjectHash)]
struct Foo { /* ... */ }
```

IMO, this would look significantly more readable.

exonum/src/blockchain/mod.rs
skletsun: I agree, but that's what rustfmt applies to this specific part of code...

exonum/src/runtime/types.rs
popzxc: Unfortunately, no. We tried to split derives as you suggest, but `rustfmt` anyway prefer
this
very
beautiful
style.

exonum/benches/criterion/block.rs
skletsun: Fixed.

exonum/src/runtime/types.rs
slowli: [This](https://rust-lang.github.io/rustfmt/#merge_derives) seems to be relevant.

exonum/src/blockchain/config.rs
alekseysidorov: With clear error message?

exonum/src/blockchain/config.rs
alekseysidorov: At the moment "privileged" is obsolete term in documentation. We have switched to the different approach where only "supervisor" service is privileged.

exonum/src/blockchain/builder.rs
alekseysidorov: Anyway, I'd create corresponding task and write TODO comment in code in order not to forget about it.

exonum/tests/explorer/blockchain/mod.rs
```diff
@@ -122,10 +125,15 @@ pub fn create_blockchain() -> BlockchainMut {
...
 
-    let services =
-        vec![InstanceCollection::new(MyService).with_instance(SERVICE_ID, "my-service", ())];
-    BlockchainBuilder::new(blockchain, config.consensus)
-        .with_rust_runtime(mpsc::channel(1).0, services)
+    let (factory, cfg) = InstanceCollection::new(MyService)
+        .with_instance(SERVICE_ID, "my-service", ())
+        .into();
+    let rust_runtime = RustRuntime::new(mpsc::channel(1).0).with_available_service(factory);
+    let genesis_config = GenesisConfigBuilder::with_consensus_config(config.consensus.clone())
```
alekseysidorov: Is Clone needs here?

exonum/src/node/mod.rs
```diff
@@ -950,8 +951,14 @@ impl Node {
...
+            channel.endpoints.0.clone(),
+            node_cfg.consensus.clone(),
+            services,
+        );
+
+        let blockchain = BlockchainBuilder::new(blockchain, genesis_config)
+            .with_runtime(rust_runtime)
             .with_external_runtimes(external_runtimes)
```
alekseysidorov: This method looks like an redundant.

skletsun: Going to get rid of it during the `Node::new()` refactoring.

exonum/src/node/mod.rs
skletsun: Should we even pass service factories here? Wouldn't it be better to pass `RustRuntime` populated with factories along with other runtimes instead?

examples/timestamping/backend/tests/api.rs
```diff
@@ -43,18 +46,21 @@ const SERVICE_NAME: &str = "my-timestamping";
...
     let mock_provider = MockTimeProvider::new(SystemTime::now().into());
+    let time_service = TimeServiceFactory::with_provider(mock_provider.clone());
+    let timestamping = TimestampingService;
     let mut testkit = TestKitBuilder::validator()
-        .with_rust_service(
-            InstanceCollection::new(TimeServiceFactory::with_provider(mock_provider.clone()))
-                .with_instance(TIME_SERVICE_ID, TIME_SERVICE_NAME, ()),
-        )
-        .with_rust_service(InstanceCollection::new(TimestampingService).with_instance(
+        .with_artifact(time_service.get_artifact(), ())
```
alekseysidorov: `get_instance` and `get_artifact` names is not so clear for me. `get_instance` for example, looks more like instance constructor but not like instance specification getter.

exonum/src/node/mod.rs
```diff
@@ -1232,14 +1241,23 @@ mod tests {
         let db = Arc::from(Box::new(TemporaryDB::new()) as Box<dyn Database>) as Arc<dyn Database>;
         let node_cfg = helpers::generate_testnet_config(1, 16_500)[0].clone();
 
-        let services = vec![InstanceCollection::new(TestService).with_instance(
-            SERVICE_ID,
-            "test-service",
-            (),
-        )];
+        let service = TestService;
```
alekseysidorov: This code snippet looks too low level, imo. You don't use power of type system to avoid some kind of boiler-plate code.

exonum/src/runtime/rust/service.rs
```diff
@@ -137,6 +138,43 @@ where
...
+        id: InstanceId,
+        name: impl Into<String>,
+        constructor: impl BinaryValue,
+    ) -> InstanceInitParams {
+        InstanceInitParams::new(id, name.into(), self.get_artifact(), constructor)
+    }
+
+    /// Returns service's ArtifactId.
+    fn get_artifact(&self) -> ArtifactId {
+        self.artifact_id().clone().into()
```
alekseysidorov: Why not to use `artifact_id().clone()` directly?

exonum/src/runtime/rust/service.rs
```diff
@@ -137,6 +138,43 @@ where
     }
 }
 
+/// Simplifies creation of supplementary entities for `ServiceFactory`.
+pub trait InstanceInfoProvider: ServiceFactory {
```
alekseysidorov: Also in my opinion this name is a little bit unclear. And it seems that you can use `ServiceFactory` trait directly, this would free the user from having to write an empty implementation of this trait.

exonum/src/node/mod.rs
skletsun: I'd say it became more explicit according to previous review comments.

examples/timestamping/backend/tests/api.rs
skletsun: Yeap I was puzzled with naming. I agree that for the instance it's more close to `create_` than `get_`, but for artifact semantically it just converts internal rust artifact to `ArtifactId`. `.get_instance()` -> `create_instance()` and left `.get_artifact()` as is?

exonum/src/runtime/rust/service.rs
skletsun: It's used by the default `DefaultInstance` implementation and manually in tests.

exonum/src/runtime/rust/service.rs
skletsun: Yeap, I've already implemented `InstanceInfoProvider` for `ServiceFactory`, so no need for manual implementations.
About naming - I'm always open for suggestions 🙂

test-suite/testkit/src/builder.rs
```diff
@@ -172,13 +178,20 @@ impl TestKitBuilder {
...
-        self.instances.extend(instances);
+    pub fn with_rust_service(mut self, service: impl Into<Box<dyn ServiceFactory>>) -> Self {
+        let service = service.into();
+        self.rust_runtime = self.rust_runtime.with_factory(service);
         self
     }
 
+    /// Adds a Rust service that has default instance configuration to the testkit. Corresponding
+    /// artifact and default instance are added implicitly.
+    pub fn with_rust_service_default(self, service: impl DefaultInstance) -> Self {
```
slowli: Nit: `with_default_rust_service` looks like a better name.

alekseysidorov: I'd suggest to use `builtin` instead of `default`.

exonum/src/blockchain/mod.rs
```diff
@@ -23,7 +23,7 @@ pub use crate::runtime::{
 
 pub use self::{
     block::{Block, BlockProof},
-    builder::{BlockchainBuilder, InstanceCollection, InstanceConfig},
+    builder::{BlockchainBuilder, InstanceCollection},
```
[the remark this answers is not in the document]
skletsun: Yes, I'd suggest to remove it in separate PR in order to (1) not to blow up current PR more and (2) review it sooner.

test-suite/testkit/src/builder.rs
skletsun: I feel like `with_builtin_rust_service` causes `DefaultInstance` -> `BuiltinInstance` (as @popzxc  suggested). On the one hand, `Default...` communicates that it provides default values for a service. But on the other hand, the actual implementation is about builtin services and probably the meaning of "builtin service" is more close to blockchain in this case than "default service". So, `builtin...` sounds better.

exonum/src/runtime/mod.rs
```diff
@@ -438,6 +438,30 @@ impl<T: Runtime> From<T> for Box<dyn Runtime> {
...
 }
 
+/// Specifies system identifier for a [`Runtime`].
+pub trait WellKnownRuntime: Runtime {
+    /// Identifier of this runtime.
+    const ID: u32;
+}
+
+#[derive(Debug)]
+/// Instance of [`Runtime`] with corresponding ID.
```
skletsun: For the potential question of why I introduced this entity - implementing `WellKnownRuntime` for `(u32, Box<dyn Runtime>)` breaks the Rust's orphan rule.

examples/sample_runtime/src/main.rs
```diff
@@ -265,20 +267,24 @@ fn main() {
...
     // Create a blockchain with the Rust runtime and our additional runtime.
     let blockchain_base = Blockchain::new(db, service_keypair.clone(), api_sender.clone());
-    let blockchain = BlockchainBuilder::new(blockchain_base, genesis)
-        .with_rust_runtime(
-            channel.endpoints.0.clone(),
-            vec![InstanceCollection::from(SimpleSupervisor::new())],
-        )
-        .with_additional_runtime(SampleRuntime::default())
+    let supervisor_service = SimpleSupervisor::new();
+    let genesis_config = GenesisConfigBuilder::with_consensus_config(consensus_config)
```
aleksuss: I suppose `GenesisConfigBuilder::new().with_consensus_config(consensus_config)` would be more idiomatic ?

popzxc: Is there any condition when we don't want to call `with_consensus_config`? I suggest to just rename this method to `new`, so it'll be `GensisConfigBuilder::new(consensus_config)`.

popzxc: However maybe it can be later.

skletsun: Well, perhaps it makes sense for [this](https://github.com/exonum/exonum/pull/1541#discussion_r352083134)

popzxc: I think that since it's not a common scenario, it's better to have a separate constructor like `without_consensus_config`.

exonum/src/blockchain/config.rs
```diff
@@ -294,6 +295,124 @@ impl ValidateInput for ConsensusConfig {
...
+pub struct InstanceInitParams {
+    /// Wrapped `InstanceSpec`.
+    pub instance_spec: InstanceSpec,
+    /// Constructor argument for specific `InstanceSpec`.
+    pub constructor: Vec<u8>,
+}
+
+impl InstanceInitParams {
+    /// Generic constructor.
+    pub fn new(
```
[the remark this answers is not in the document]
skletsun: Avoided unit value while implementing @slowli suggestions.

exonum/src/runtime/rust/service.rs
```diff
@@ -137,6 +138,43 @@ where
...
+    /// Returns service's ArtifactId.
+    fn get_artifact(&self) -> ArtifactId {
+        self.artifact_id().clone().into()
+    }
+}
+
+impl<T> InstanceInfoProvider for T where T: ServiceFactory {}
+
+/// Provides default instance configuration parameters for `ServiceFactory`.
+pub trait DefaultInstance: InstanceInfoProvider {
```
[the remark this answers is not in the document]
skletsun: Renamed

exonum/src/blockchain/config.rs
```diff
@@ -294,6 +295,124 @@ impl ValidateInput for ConsensusConfig {
...
+        Self {
+            consensus_config,
+            artifacts: HashMap::new(),
+            builtin_instances: vec![],
+        }
+    }
+
+    /// Adds an artifact with corresponding deploy argument. Does nothing in case artifact with
+    /// given id is already added.
+    pub fn with_artifact(mut self, artifact: ArtifactId, deploy_spec: impl BinaryValue) -> Self {
```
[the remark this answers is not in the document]
skletsun: Implemented it via (1) passing here `impl Into<ArtifactSpec>` and (2) `impl<T: Into<ArtifactId>> From<T> for ArtifactSpec`.

test-suite/testkit/src/builder.rs
```diff
@@ -129,7 +134,8 @@ pub struct TestKitBuilder {
     rust_runtime: RustRuntime,
     api_notifier_channel: ApiNotifierChannel,
     additional_runtimes: HashMap<u32, Box<dyn Runtime>>,
-    instances: Vec<InstanceConfig>,
+    instances: Vec<InstanceInitParams>,
+    artifacts: HashMap<ArtifactId, Vec<u8>>,
```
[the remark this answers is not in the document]
skletsun: AFAIR, while we discussed the extraction of `GenesisConfig` changing of the public interface of `Blockchain` and `Testkit` was only mentioned, so I got it as we keep interface for `TestKitBuilder` simple with no explicit `GenesisConfig` preparation. I suggest doing this within separate PR as this similar to the upcoming refactoring of `SandboxBuilder` (in the context of getting rid of `InstanceConfig`).

exonum/tests/explorer/blockchain/mod.rs
```diff
@@ -126,10 +126,14 @@ pub fn create_blockchain() -> BlockchainMut {
         ApiSender::closed(),
     );
 
-    let services =
-        vec![InstanceCollection::new(MyService).with_instance(SERVICE_ID, "my-service", ())];
-    BlockchainBuilder::new(blockchain, config.consensus)
-        .with_rust_runtime(mpsc::channel(1).0, services)
+    let my_service = MyService;
+    let genesis_config = GenesisConfigBuilder::with_consensus_config(config.consensus)
+        .with_artifact(my_service.get_artifact(), ())
```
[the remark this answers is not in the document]
skletsun: Implemented by passing `impl Into<ArtifactSpec>` with the corresponding conversion implementations.

exonum/src/runtime/mod.rs
```diff
@@ -438,6 +438,30 @@ impl<T: Runtime> From<T> for Box<dyn Runtime> {
...
+pub struct RuntimeInstance {
```
slowli: Not sure that exposing `RuntimeInstance` APIs publicly is optimal (or necessary); I'd suggest to investigate this later.

CHANGELOG.md
```diff
@@ -52,6 +52,13 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Services can now use `BlockchainData` and `SnapshotExt` types to access data
   from the blockchain in a more structured manner. (#1523)
 
+- `GenesisConfig` is extracted into separate entity. `BlockchainBuilder`, `Node`
+ and `Testkit` explicitly accepts it during creation. (#1541)
+
+- Added `BuiltinInstance` trait for declaration of builtin services. (#1541)
+
+- Placeholder for changes due to dynamic services (#9999)
```
popzxc: This line should be removed.

test-suite/testkit/tests/interfaces/services.rs
```diff
@@ -148,9 +148,9 @@ impl DepositInterface for DepositService {
     }
 }
 
-impl DefaultInstance for DepositService {
-    const DEFAULT_INSTANCE_ID: u32 = Self::ID;
-    const DEFAULT_INSTANCE_NAME: &'static str = "deposit";
+impl BuiltinInstance for DepositService {
```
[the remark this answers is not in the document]
skletsun: Let's vote? `DefaultInstance` 👍 , `BuiltinInstance` 👎

test-suite/testkit/src/builder.rs
```diff
@@ -203,29 +203,35 @@ impl TestKitBuilder {
     /// # Panics
     ///
     /// - Panics if builder's instance already contains specified runtime.
-    pub fn with_additional_runtime(mut self, runtime: impl Into<(u32, Box<dyn Runtime>)>) -> Self {
-        let (id, runtime) = runtime.into();
-        if id == RustRuntime::ID as u32 || self.additional_runtimes.contains_key(&id) {
-            panic!("TestkitBuilder already contains runtime with id {}", id);
+    pub fn with_additional_runtime(mut self, runtime: impl Into<RuntimeInstance>) -> Self {
```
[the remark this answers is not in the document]
skletsun: Good point. I implemented it in analogue to the `BlockchainBuilder.with_runtime()` which has to accept `RuntimeInstance` as well, but for this case `impl WellKnownRuntime` looks better.

test-suite/testkit/src/lib.rs
```diff
@@ -856,10 +858,7 @@ impl StoppedTestKit {
     /// (which is also what may happen with real Exonum apps).
     ///
     /// This method will not add the default Rust runtime, so you must do this explicitly.
-    pub fn resume(
-        self,
-        runtimes: impl IntoIterator<Item = impl Into<(u32, Box<dyn Runtime>)>>,
-    ) -> TestKit {
+    pub fn resume(self, runtimes: impl IntoIterator<Item = impl Into<RuntimeInstance>>) -> TestKit {
```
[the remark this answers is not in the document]
skletsun: Well, in fact, it is used exactly for this situation. More presize - in tests there are a few places where we pass `RustRuntime` this way.

test-suite/testkit/src/builder.rs
```diff
@@ -133,9 +133,9 @@ pub struct TestKitBuilder {
     logger: bool,
     rust_runtime: RustRuntime,
     api_notifier_channel: ApiNotifierChannel,
-    additional_runtimes: HashMap<u32, Box<dyn Runtime>>,
+    additional_runtimes: Vec<RuntimeInstance>,
     instances: Vec<InstanceInitParams>,
-    artifacts: HashMap<ArtifactId, Vec<u8>>,
+    artifacts: HashSet<ArtifactSpec>,
```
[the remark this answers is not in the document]
skletsun: Thanks for pointing out, corrected.

test-suite/testkit/tests/interfaces/services.rs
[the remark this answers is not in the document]
skletsun: Renamed it back...

test-suite/testkit/src/builder.rs
```diff
@@ -203,29 +203,35 @@ impl TestKitBuilder {
...
 
     /// Adds an artifact with corresponding deploy argument. Does nothing in case artifact with
     /// given id is already added.
-    pub fn with_artifact(mut self, artifact: ArtifactId, deploy_spec: impl BinaryValue) -> Self {
-        self.artifacts
-            .entry(artifact)
-            .or_insert_with(|| deploy_spec.into_bytes());
+    pub fn with_artifact(mut self, artifact: impl Into<ArtifactSpec>) -> Self {
+        let artifact = artifact.into();
+        if !self.artifacts.contains(&artifact) {
```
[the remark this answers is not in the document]
skletsun: Re-wrote it to be actual again.

exonum/src/helpers/mod.rs
skletsun: Mentioned nit finished tasks in TODOs and in ECR-3913.

exonum/src/blockchain/builder.rs
skletsun: ECR-3913

exonum/src/runtime/types.rs
skletsun: Strange. I remember I left comment here but it disappeared somehow...
Once more: I tried suggested approach, but it didn't help:
- `true` is default and it aligns averything horizontally but seems after some amount of elements it starts doing it vertically.
- If I set it to `false` and split big derives into two horizontal lines - they remain horizontally aligned but everything else becomes vertically aligned.

popzxc: I think we'll investigate it later, don't bother yourself.

exonum/src/runtime/types.rs
```diff
@@ -238,6 +258,25 @@ pub struct ArtifactSpec {
...
+    /// Generic constructor.
+    pub fn new(artifact: ArtifactId, deploy_spec: impl BinaryValue) -> Self {
+        Self {
+            artifact,
+            payload: deploy_spec.into_bytes(),
+        }
+    }
+}
+
+impl<T: Into<ArtifactId>> From<T> for ArtifactSpec {
```
popzxc: This conversion seems to be very error-prone, since it hides the fact of setting payload to empty value. 
Service having no payload is not an invariant, but `From` conversion makes it feel like it is.

I suggest it to be a method with the name explaining it unambiguously, e.g. `new_without_spec` (and @slowli originally [suggested](https://github.com/exonum/exonum/pull/1541#discussion_r352093418) it to be a method, and his suggestion still makes sense).

exonum/src/runtime/types.rs
```diff
@@ -324,6 +371,15 @@ impl Display for InstanceSpec {
     }
 }
 
+impl From<InstanceSpec> for InstanceInitParams {
+    fn from(spec: InstanceSpec) -> Self {
```
popzxc: Same here.

exonum/src/blockchain/mod.rs
```diff
@@ -220,24 +223,24 @@ impl BlockchainMut {
...
         let mut fork = self.fork();
         // Write genesis configuration to the blockchain.
-        Schema::new(&fork).consensus_config_entry().set(config);
+        Schema::new(&fork)
+            .consensus_config_entry()
+            .set(genesis_config.consensus_config);
+
+        for ArtifactSpec { artifact, payload } in genesis_config.artifacts {
+            Dispatcher::commit_artifact(&fork, artifact.clone(), payload.clone())?;
+            if !self.dispatcher.is_artifact_deployed(&artifact) {
```
alekseysidorov: Possible that this check is redundant, can we deploy artifact before this method call?

skletsun: Most probably it is - the uniqueness for an artifact is checked during the addition of artifact to the genesis config.

popzxc: I think we'll fix it later.

skletsun: Removed

PR #1544, opened 2019-11-15
Remove atty dependency

popzxc:
Seems that we don't use it anymore.


popzxc:
^ This change surely decreases coverage, yeah. We need to do something with codecov, for real =_=

PR #1545, opened 2019-11-15
Write full and meaningful description of the IndexCoordinates [ECR-3824]

alekseysidorov:


exonum/src/blockchain/schema.rs
```diff
@@ -335,46 +335,46 @@ where
...
-/// Describes the membership for the index.
+/// Describes the origin of the schema.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub enum IndexOwner {
-    /// This index is a part of the core schema.
+pub enum SchemaOrigin {
+    /// This is a Core schema.
     Core,
-    /// This index is a part of the runtime schema with the specified ID.
+    /// This is a runtime schema with the specified ID.
```
popzxc:
```suggestion
    /// Schema belongs to the runtime with the specified ID.
```

exonum/src/blockchain/schema.rs
```diff
@@ -335,46 +335,46 @@ where
...
     Runtime(u32),
-    /// This index is a part of some service schema with the specified ID.
+    /// This is a service schema with the specified instance ID.
     Service(InstanceId),
 }
 
-impl IndexOwner {
-    /// Creates index coordinate for the current owner.
+impl SchemaOrigin {
+    /// Computes the index coordinates inside the schema.
```
popzxc:
```suggestion
    /// Computes coordinates for a given schema index.
```

exonum/src/blockchain/schema.rs
```diff
@@ -335,46 +335,46 @@ where
...
     pub fn coordinate_for(self, index_id: u16) -> IndexCoordinates {
         IndexCoordinates::new(self, index_id)
     }
 
-    /// Returns the corresponding tag.
-    fn tag(self) -> IndexTag {
+    /// Returns the corresponding origin binary value.
```
popzxc: Maybe "the origin label" instead of "binary value"? Usually "binary value" is associated with byte juggling, which is not exactly the case here.

exonum/src/blockchain/schema.rs
```diff
@@ -335,46 +335,46 @@ where
...
-            IndexOwner::Service { .. } => IndexTag::Service,
+            SchemaOrigin::Core => Origin::Core,
+            SchemaOrigin::Runtime { .. } => Origin::Runtime,
+            SchemaOrigin::Service { .. } => Origin::Service,
         }
     }
 
-    /// Returns the corresponding group id.
-    fn group_id(self) -> u32 {
+    /// Returns the corresponding schema ID.
```
popzxc: Hm. Not sure about this. "Schema ID" can be considered an unique identifier of the schema, but returned value here will be the same for core, rust runtime and service with id 0. Maybe name it `local_schema_id` and state in the docstring that returned value is ID local to the origin?

exonum/src/blockchain/schema.rs
```diff
@@ -384,40 +384,41 @@ enum IndexTag {
...
-    pub fn owner(self) -> IndexOwner {
-        match self.tag {
-            0 => IndexOwner::Core,
-            2 => IndexOwner::Runtime(self.group_id),
-            3 => IndexOwner::Service(self.group_id),
+    /// Returns a schema origin for this index.
+    pub fn schema_origin(self) -> SchemaOrigin {
+        match self.origin {
+            0 => SchemaOrigin::Core,
+            2 => SchemaOrigin::Runtime(self.schema_id),
```
popzxc: Off-topic: interesting, where did we lose `1`?

exonum/src/blockchain/schema.rs
```diff
@@ -384,40 +384,41 @@ enum IndexTag {
...
+            origin: schema_origin.origin() as u16,
+            schema_id: schema_origin.schema_id(),
             index_id,
         }
     }
 
     // TODO Write more meaningful description [ECR-3824]
-    /// For the given index owner, returns a list of the index coordinates that match the corresponding
-    /// object hashes of the indices.
+    /// For the given index owner, returns a list of the index coordinates that match the
```
Maria-Nosyk:
```suggestion
    /// For the given schema origin, returns a list of the index coordinates that match the
```

exonum/src/blockchain/schema.rs
```diff
@@ -384,40 +384,41 @@ enum IndexTag {
...
+    /// corresponding object hashes of the indices.
```
Maria-Nosyk: if the "object hash" means the "hash of the index", I would delete the word "object" not to multiple different names for the same notion
```suggestion
    /// corresponding hashes of the indices.
```

slowli:
Just FYI: I'm currently working on automated aggregation of the MerkleDB state (https://github.com/exonum/exonum/compare/dynamic_services...slowli:auto-state-hash, see also ECR-3843); I hope it will be done in the nearest future (although probably not by 0.13 release). The idea is to have fully automated aggregation of all top-level `ProofMap` / `ProofList` indexes (+ a subset of `Entry`s), with zero config from service / core, and a nice API to extract proofs from the aggregator (not in the branch yet, but it should probably look similar to `SnapshotExt` methods). Thus, `IndexCoordinates` API may not live for very long.

Currently the branch is at funposting stage, so it's way too early to review. (E.g., I've changed `Patch` intrinsics, although not completely randomly: I want to update the aggregator automatically in `Fork::into_patch`, which sort of necessitates structuring of `Patch` changes by indexes rather than by column families. As a side-effect, some benches indicate that the changes have resulted in 5-10% performance boost.)

exonum/src/blockchain/schema.rs
alekseysidorov: '1' is an origin of the `Dispatcher` schema.

popzxc: I guess it should be mentioned somewhere.

aleksuss: By the way, it will be added here after implementing this issue ECR-3810

PR #1547, opened 2019-11-18
Remove metrics module

popzxc:
It seems to be unfinished and not really used.


popzxc:
Metrics were mentioned in the changelog, so probably we should mention their removal.

PR #1550, opened 2019-11-18
Remove isolate method [ECR-3820]

popzxc:
We either apply the config with all the changes, or don't apply it at all.
So the `isolate` method is not required anymore.

Slippery moments:
- We still have to panic to rollback changes (will be fixed once we'll be able to return an error in `before_commit`).
- If panic/error occurs during config applying, the proposal is not removed (well, in fact it's removed but then rolled back). However I've added a check in the config proposal tx for it, so it's not a big problem I guess.

### DIFF
--- a/services/supervisor/src/lib.rs
+++ b/services/supervisor/src/lib.rs
@@ -60,23 +60,9 @@ pub const fn supervisor_name() -> &'static str {
 const NOT_SUPERVISOR_MSG: &str = "`Supervisor` is installed as a non-privileged service. \
                                   For correct operation, `Supervisor` needs to have numeric ID 0.";
 
-/// Applies configuration changes, isolating each of them with by using `Fork` checkpoints.
-///
-/// # Safety
-///
-/// This function should be used with extreme care. It makes the following assumptions:
-///
-/// - The function must be called at the end of the transaction or `before_commit` execution.
-///   If the transaction errors / panics afterwards, the changes to the configs will not
-///   be rolled back.
-/// - No changes to the blockchain state should be introduced before the call to this function.
-///   Any changes that are introduced will be committed regardless of the execution status,
-///   or the status of application of any config change. This is if the execution wasn't interrupted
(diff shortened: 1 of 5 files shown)


services/supervisor/src/transactions.rs
```diff
@@ -169,17 +169,23 @@ where
...
+            // We have a proposal, check that it's actual.
+            if current_height < proposal.config_propose.actual_from {
+                return Err(Error::ConfigProposeExists.into());
+            } else {
+                // Proposal is outdated but was not removed (e.g. because of the panic
+                // during config applying), clean it.
+                schema.pending_proposal.remove();
+            }
         }
+        drop(schema);
```
aleksuss: What about additional scope ? The question is: what is more idiomatic approach here? @alekseysidorov what do you think ? I'd avoid explicit call of `drop`.

popzxc: I'm not a fan of it either, but since `drop` was [already introduced](https://github.com/exonum/exonum/blob/dynamic_services/services/supervisor/src/lib.rs#L209) I decided to match the style.

aleksuss: Ok. I got it.

services/supervisor/src/lib.rs
```diff
@@ -75,60 +75,57 @@ const NOT_SUPERVISOR_MSG: &str = "`Supervisor` is installed as a non-privileged
...
-                    e
+            ConfigChange::Service(config) => {
+                log::trace!(
+                    "Updating service instance configuration, instance ID is {}",
+                    config.instance_id
                 );
+
+                let configure_result = context
+                    .interface::<ConfigureCall<'_>>(config.instance_id)
+                    .expect("Obtaining Configure interface failed")
```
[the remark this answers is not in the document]
popzxc: Yep, the logic were based on that fact. Agree, I'll add a comment.

PR #1552, opened 2019-11-18
Assign instance IDs in the Supervisor [ECR-3746]

popzxc:
This PR moves instance ID assignment logic from Dispatcher to the Supervisor.

I decided to not remove `MAX_BUILTIN_INSTANCE_ID` constant in this PR, but created a task in Jira for it (ECR-3851).


services/supervisor/src/schema.rs
```diff
@@ -80,4 +85,22 @@ impl Schema<Prefixed<'_, &Fork>> {
         let new_configuration_number = self.configuration_number.get().unwrap_or(0) + 1;
         self.configuration_number.set(new_configuration_number);
     }
+
+    /// Assign unique identifier for an instance.
+    pub(crate) fn assign_instance_id(&mut self) -> InstanceId {
+        let id = self
+            .vacant_instance_id
+            .get()
+            .unwrap_or(MAX_BUILTIN_INSTANCE_ID);
```
aleksuss: What about to rename into `FIRST_DEPLOYED_INSTANCE_ID` ? As I can see builtin instance can't get id 1024.

services/supervisor/src/lib.rs
```diff
@@ -45,6 +45,14 @@ mod proto_structures;
 mod schema;
 mod transactions;
 
+/// Max instance identifier for builtin service.
```
aleksuss:
```suggestion
/// Instance identifier for first deployed service.
```

services/supervisor/src/schema.rs
popzxc: It'll get removed soon anyway, so I don't think it matters (see ECR-3851) :)

services/supervisor/src/schema.rs
```diff
@@ -80,4 +85,22 @@ impl Schema<Prefixed<'_, &Fork>> {
...
+        id
+    }
+}
+
+/// Creates an index given its name and access object.
+fn construct<'a, T: Access, U: FromAccess<Prefixed<'a, T>>>(
+    access: &Prefixed<'a, T>,
+    index_name: &str,
+) -> U {
+    FromAccess::from_access(access.clone(), index_name.into()).unwrap()
```
pavel-mukhanov: I wonder if this unwrap is really safe?

popzxc: Error is returned if the index exists but has a different type. Since this function is private and used only in the schema constructor (thus types are correct by design), this unwrap is indeed safe.

PR #1553, opened 2019-11-18
Automatic state aggregation in MerkleDB [ECR-3843]

slowli:
## Overview

This PR introduces fully automated state hash aggregation in MerkleDB, which replaces the corresponding app-level mechanisms used previously.

Not done yet / wishlist:

- Proptests ensuring that state aggregator is indeed up to date in `Patch`
- Proptests for rollback patches (not directly related to the PR subject)
- Benchmarks (preliminary, there's <5% performance boost for the `merkledb::transactions` benchmark)
- Trimming public API of MerkleDB some more

### Definition of Done


components/merkledb/src/proof_entry.rs
```diff
@@ -0,0 +1,296 @@
...
+/// You can add an element to this index and check whether it exists. A value
+/// should implement [`BinaryValue`] and [`ObjectHash`] traits. Unlike [`Entry`],
+/// hashed entries are eagerly hashed and may participate in [state aggregation].
+///
+/// [`BinaryValue`]: trait.BinaryValue.html
+/// [`ObjectHash`]: trait.ObjectHash.html
+/// [`Entry`]: struct.Entry.html
+/// [state aggregation]: index.html#state-aggregation
+#[derive(Debug)]
+pub struct ProofEntry<T: RawAccess, V> {
```
slowli: It could be possible to customize `Entry` instead:

```rust
struct Entry<T: RawAccess, V, Type: EntryType> { /* ...*/ }
```

...in order to minimize the amount of copy-pasted code.

exonum/src/runtime/blockchain_data.rs
```diff
@@ -68,6 +68,33 @@ impl<'a, T: RawAccess + AsReadonly> BlockchainData<'a, T> {
...
 }
 
+impl BlockchainData<'_, &dyn Snapshot> {
+    /// Returns a proof for a Merkelized index with the specified name
+    /// in the currently executing service.
+    ///
+    /// # Return value
+    ///
+    /// If the index does not exist or is not Merkelized, returns `None`. Note that this may
+    /// occur before the index is accessed the first time, which, depending on the service logic,
```
slowli: One of drawbacks of implicit index initialization.

exonum/src/sandbox/consensus/basic.rs
```diff
@@ -185,34 +185,36 @@ fn test_query_state_hash() {
...
+        let proof = proof.check_against_hash(state_hash).unwrap();
         assert_ne!(configs_rh, Hash::zero());
         assert_eq!(
-            proof.entries().collect::<Vec<_>>(),
-            vec![(&configs_key, &configs_rh)]
+            proof
+                .entries()
+                .map(|(k, v)| (k.as_str(), v))
+                .collect::<Vec<_>>(),
+            vec![("core.consensus.config", &configs_rh)]
```
slowli: Why is the index not named `"core.consensus_config"`? Looks weird.

components/merkledb/src/views/metadata.rs
```diff
@@ -237,27 +271,36 @@ where
...
 /// Persistent pool used to store indexes metadata in the database.
 /// Pool size is used as an identifier of newly created indexes.
-struct IndexesPool<T: RawAccess>(View<T>);
+pub(super) struct IndexesPool<T: RawAccess>(View<T>);
 
 impl<T: RawAccess> IndexesPool<T> {
-    fn new(index_access: T) -> Self {
-        Self(View::new(index_access, INDEXES_POOL_NAME))
+    pub(super) fn new(index_access: T) -> Self {
+        Self(View::new(
```
popzxc: Nit: I'd move `View` creation to the separate variable.

components/merkledb/src/views/metadata.rs
```diff
@@ -283,7 +326,8 @@ impl<T: RawAccess> IndexesPool<T> {
     {
         let len = self.len();
         let metadata = IndexMetadata {
-            identifier: len,
+            identifier: len + 1,
+            // ^-- Identifier should be non-zero to translate to a correct id in `ResolvedRef`
```
popzxc: Why do we need it btw?

components/merkledb/src/views/system_info.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub fn new(access: T) -> Self {
+        SystemInfo(access)
+    }
+
+    /// Returns the total number of indexes in the storage. This information is always up to date
+    /// (even for `Fork`s).
+    ///
+    /// System-defined indexes (e.g., `state_aggregator`) are *included* into this count.
+    pub fn index_count(&self) -> u64 {
+        IndexesPool::new(self.0.clone()).len() - 1
```
popzxc: This looks like a hack, to be honest.

examples/cryptocurrency-advanced/backend/src/lib.rs
```diff
@@ -48,8 +48,11 @@ pub const INITIAL_BALANCE: u64 = 100;
 pub struct CryptocurrencyService;
 
 impl Service for CryptocurrencyService {
-    fn state_hash(&self, data: BlockchainData<&dyn Snapshot>) -> Vec<Hash> {
-        Schema::new(data.for_executing_service()).state_hash()
+    fn initialize(&self, context: CallContext<'_>, _params: Vec<u8>) -> Result<(), ExecutionError> {
+        // Initialize indexes. Not doing this may lead to errors in HTTP API, since it relies on
+        // `wallets` indexes being initialized for returning corresponding proofs.
+        Schema::new(context.service_data());
```
popzxc: Just a thought. Maybe it worth adding a type `Schema` to the `Service` trait and have a default implementation of `Service::initialize` which'll call `Service::Schema::new(context.service_data())`?

In theory it should reduce amount of boilerplate and protect those fellows who'll forget to do it.

alekseysidorov: This will not work because traits with associated types are not object safe. We have to create separated trait and then convert it to the `Box<dyn Service>`.

popzxc: Indeed.

components/merkledb/src/backends/rocksdb.rs
```diff
@@ -92,54 +93,97 @@ impl RocksDB {
...
     ) -> crate::Result<()> {
-        let snapshot = self.snapshot();
-        let mut iterator = snapshot.iter(cf_name, prefix);
-        while let Some((key, ..)) = iterator.next() {
-            if !key.starts_with(prefix) {
-                break;
-            }
+        let snapshot = self.typed_snapshot();
+        let mut iterator = snapshot.typed_iter(resolved, &[]);
+        while iterator.next().is_some() {
```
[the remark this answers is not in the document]
slowli: Not likely; the `Iterator` trait, which `RocksdbIterator` implements, is defined in the `db` module and is not a standard iterator.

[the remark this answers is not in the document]
popzxc: Ah, I was confused by name. OK then.

components/merkledb/src/views/address.rs
```diff
@@ -0,0 +1,207 @@
...
+
+pub fn key_bytes<K: BinaryKey + ?Sized>(key: &K) -> Vec<u8> {
+    concat_keys!(key)
+}
+
+/// Represents the address of an index in the database.
+///
+/// An address has a string *name* and an optional byte *key*. An index is uniquely identified
+/// by its address. Different addresses correspond to different indexes. Addresses with the same
+/// name and differing prefixes are said to belong to the same *group* (see also [`Group`]). Groups
```
alekseysidorov: I suppose that prefix in this context a little bit misleading. What it means? Is this an address suffix or prefix for index key?

components/merkledb/src/views/metadata.rs
```diff
@@ -294,6 +339,45 @@ impl<T: RawAccess> IndexesPool<T> {
     }
 }
 
+/// Obtains `object_hash` for an aggregated index.
+pub fn get_object_hash<T: RawAccess>(access: T, addr: ResolvedAddress) -> Hash {
+    use crate::{ObjectHash, ProofListIndex, ProofMapIndex};
+
+    let index_full_name = addr.name.as_bytes().to_vec();
```
alekseysidorov: Is it necessary to do allocation here?

components/merkledb/src/views/system_info.rs
```diff
@@ -0,0 +1,247 @@
...
+///     snapshot.get_proof_list::<_, u32>("list").object_hash()
+/// );
+///
+/// // It is possible to prove that an index has a specific state
+/// // given `state_hash`:
+/// let proof = aggregator.get_proof("list".to_owned());
+/// proof.check_against_hash(state_hash).unwrap();
+/// ```
+#[derive(Debug, Clone, Copy)]
+pub struct SystemInfo<T>(T);
```
alekseysidorov: I'd suggest to rename to `SystemSchema` or even somethinkg like `RootSchema`.

exonum/src/blockchain/block.rs
```diff
@@ -119,6 +121,19 @@ pub struct BlockProof {
     pub precommits: Vec<Verified<Precommit>>,
 }
 
+/// Proof of authenticity for a single index within the database.
```
alekseysidorov: I wonder if we implement protobuf serialization for this kind of proof and also by analogy implement `GroupIndexProof`, can we cover all use cases of proofs?

components/merkledb/src/entry.rs
```diff
@@ -193,33 +191,3 @@ where
         previous
     }
 }
-
-impl<T, V> ObjectHash for Entry<T, V>
```
popzxc: Can't add the comment to the corresponding line, but I think that `ObjectHash` restriction should be removed from

```rust
impl<T, V> Entry<T, V>
where
    T: RawAccessMut,
    V: BinaryValue + ObjectHash,
```

By the way it's funny that we had no tests for this.

components/merkledb/src/proof_entry.rs
popzxc: Hm. The structure is called `Proof*`, but we can't get a proof for it. On the one hand, the intention is clear and specularity with other indices is good; but on the other hand it isn't *literally* the proof entry.

components/merkledb/src/views/address.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// let addr = IndexAddress::from_root("foo");
+    /// let prefixed = addr.prepend_name("prefix");
+    /// assert_eq!(prefixed.name(), "prefix.foo");
+    /// ```
+    pub fn prepend_name(self, prefix: &str) -> Self {
+        let name = if self.name.is_empty() {
+            prefix.to_owned()
+        } else {
+            // Because `concat` is faster than `format!("...")` in all cases.
+            [prefix, ".", self.name()].concat()
```
popzxc: Can we use [`std::concat`](https://doc.rust-lang.org/std/macro.concat.html) here? Then I think it will look more natural and the comment will not be required anymore.

components/merkledb/src/views/address.rs
```diff
@@ -0,0 +1,207 @@
...
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct ResolvedAddress {
+    /// Name of the column family where the view is stored.
+    pub name: String,
+    /// Index identifier within a column family.
+    pub id: Option<NonZeroU64>,
+}
+
+// This conversion is only useful for tests, since all user-created indexes should have an ID set.
+#[cfg(test)]
```
popzxc: Nit/bikeshedding: I'd put `cfg(test)` conversion below `impl ResolvedAddress`, since it's auxiliary.

components/merkledb/src/views/metadata.rs
```diff
@@ -284,7 +328,8 @@ impl<T: RawAccess> IndexesPool<T> {
     {
         let len = self.len();
         let metadata = IndexMetadata {
-            identifier: len,
+            identifier: len + 1,
+            // ^-- Identifier should be non-zero to translate to a correct id in `ResolvedAddress`
```
popzxc: I still don't quire understand what causes this requirement?

components/merkledb/src/views/metadata.rs
```diff
@@ -309,17 +393,26 @@ where
...
+        Self::get_or_create_unchecked(index_access, index_address, index_type)
+    }
+
+    /// Gets an index with the specified address and type. Unlike `get_or_create`, this method
+    /// does not check if the name of the index is reserved.
+    ///
+    /// # Safety
+    ///
+    /// This method should only be used to create system indexes within this crate.
+    pub(crate) fn get_or_create_unchecked(
```
popzxc: Can visibility of this method be reduced to `pub(super)` just in case?

exonum/src/runtime/mod.rs
```diff
@@ -221,9 +221,8 @@ impl From<RuntimeIdentifier> for u32 {
 /// - `execute`
 /// - `before_commit`
 /// - `start_adding_service`
-/// - `state_hashes`
 ///
-/// All these methods except for `state_hashes` should also produce the same changes
+/// All these methods should also produce the same changes
```
popzxc: Nit: inconsistent line lengths.

components/merkledb/src/views/metadata.rs
slowli: It's used later to construct `ViewWithMetadata`.

components/merkledb/src/backends/rocksdb.rs
```diff
@@ -180,38 +234,51 @@ impl Snapshot for RocksDBSnapshot {
...
+
+        let (key, value) = self.iter.next()?;
+        if let Some(ref prefix) = self.prefix {
+            if &key[..8] != prefix {
+                self.ended = true;
+                return None;
+            }
         }
+
+        self.key = Some(key);
```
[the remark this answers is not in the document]
slowli: How would it be initialized in this case? `Option<Box<T>>` [has the same memory layout](https://doc.rust-lang.org/std/option/index.html#options-and-pointers-nullable-pointers) as `Box<T>`. (The `None` variant corresponds to a null pointer in ~plebian~ less memory-safe programming languages.) Thus, I don't think the change would lead to performance improvement, and I'd argue that the current implementation is understandable.

components/merkledb/src/views/address.rs
slowli: `concat!` concatenates literals, not values; it cannot be used with variables.

components/merkledb/src/views/metadata.rs
slowli: `NonZeroU64` is used as the type of the identifier, together with `None` option reserved for non-index views (e.g., `__METADATA__`). Hence, the index ID should be, well, non-zero.

components/merkledb/src/proof_entry.rs
slowli: I used similarity with `ProofListIndex` / `ProofMapIndex` as the inspiration. Do you have a better naming in mind?

components/merkledb/src/views/address.rs
popzxc: Ah, I see.

components/merkledb/src/backends/rocksdb.rs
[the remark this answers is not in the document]
popzxc: Didn't know about layout, then I think current approach is fine.

components/merkledb/src/proof_entry.rs
popzxc: Nah, I don't think it's that important. Just was thinking out loud.

PR #1557, opened 2019-11-20
Remove max-builtin-id constant [ECR-3851]

popzxc:
**Update:** As @slowli suggested, entry with max builtin instance id was removed and instead public access to the running service list was gained in the Dispatcher.

To ensure that method with `pub` visibility won't leak a mutable access to the dispatcher schema, this method exists only for `ReadonlyFork<'_>` specialization.

---

This PR removes `MAX_BUILTIN_INSTANCE_ID`, introducing the following approach:
- During initialization, `Dispatcher` checks builtin IDs, finding the highest one, and storing it in the entry.
- Within the first dynamic service start, `Supervisor` checks that entry of `Dispatcher`, and assigns the ID as the next to the max builtin ID, initializing its own entry `vacant_instance_id`.
- Since that, IDs are assigned sequentially using `vacant_instance_id` entry.

Such an approach supports sparse builtin instance ID spaces, which may be helpful, i.e., when user has three builtin instances with IDs 0, 1 and 2, but then decides to remove instance with ID 1 from code -- nothing will break then.


exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -360,6 +365,17 @@ impl Dispatcher {
         );
         Ok(())
     }
+
+    /// Checks if provided instance ID is greater than the currently stored one
+    /// and updates it if needed.
+    fn update_max_builtin_id(&self, fork: &Fork, instance_id: InstanceId) {
```
aleksuss: Don't we need to return `Restult<(), Error>` here and return `Err(_)` if `current_max_id` is `None` ?

popzxc: Nope. `None` means that field was not initialized. Here we use that `Option` implements `PartialEq`: `Some(val)` is always greater than `None`, so at the first iteration (when field is `None`) it will be initialized correctly.

services/supervisor/src/lib.rs
```diff
@@ -123,6 +116,37 @@ fn update_configs(context: &mut CallContext<'_>, changes: Vec<ConfigChange>) ->
     Ok(())
 }
 
+/// Assigns the instance ID for a new service, initializing the schema `vacant_instance_id`
+/// entry if needed.
+fn assign_instance_id(context: &mut CallContext<'_>) -> InstanceId {
```
slowli: Nit: is `&mut CallContext` (instead of `&CallContext`) really needed here?

popzxc:
>IMO, a more comprehensive approach would be to allow to iterate over service instances given the dispatcher schema. In this case, the supervisor would just do this when first requested to deploy a service, and select a starting ID correspondingly.

I thought it will be nice to keep Supervisor know as less as possible about the running services. Its responsibility is to **start** services, and with the possibility to iterate through services with a handy interface, it will be easy to misuse this interface later and get a service excessively tied to the core.

Of course, even now supervisor *can* iterate through running services, but it will look ugly and probably make author of the code think if he's doing something wrong.

>Alternatively, the supervisor just could try IDs iteratively via DispatcherSchema::get_instance. In this case, we wouldn't be afraid of sparse identifiers (and, probably, several supervisor services).

Yeah, I considered that variant as well, but I'm not sure about it. Currently we have to worry about builtin IDs only once, which is intuitive, but with iterating approach the logic related to the builtin IDs will become the part of the overall ID assignment logic. I thought it's less elegant, sort of.

popzxc:
Other sort-of related thought I had when decided to just skip sparse spaces:
if user requests a start of two instances within one request, they usually think that assigned IDs will be consecutive. Of course, we don't provide such a guarantee, but nevertheless user can be confused if all of a sudden they'll get one service with ID 1 and one with ID 99.

popzxc:
Description of the PR is updated to match the new state of PR.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -119,6 +119,14 @@ impl<T: Access> Schema<T> {
     }
 }
 
+// `ReadonlyFork` specialization to ensure that we won't leak mutable schema access.
+impl Schema<ReadonlyFork<'_>> {
+    /// Readonly set of launched service instances.
+    pub fn running_instances(&self) -> MapIndex<ReadonlyFork<'_>, String, InstanceSpec> {
```
slowli: Hm, I thought about literally exposing an iterator:

```rust
pub fn services<'s>(&'s self) -> impl Iterator<Item = InstanceSpec> + 's {
    // ...
}
```

The problem with this approach is that it's seemingly impossible given MerkleDB limitations (it is impossible to have an iterator without a corresponding index, although logically they are independent - the lifetime of the iterator is tied to the lifetime of a snapshot / fork).

An alternative is to use

```rust
impl<T> Schema<T>
where
    T: Access + AsReadonly<Readonly = T>,
{
    // ...
}
```

This seemingly works, although you would need to add `impl AsReadonly for ReadonlyFork` (the implementation is trivial).

popzxc: >Hm, I thought about literally exposing an iterator

I thought about it too, but this approach is worse if some service will want to lookup concrete service: it will have to enjoy `O(N)` approach instead of `O(1)`.

>This seemingly works, although you would need to add impl AsReadonly for ReadonlyFork (the implementation is trivial).

Like it, will do.

popzxc: Done.

services/supervisor/src/lib.rs
```diff
@@ -123,6 +116,40 @@ fn update_configs(context: &mut CallContext<'_>, changes: Vec<ConfigChange>) ->
...
+            let new_instance_id = builtin_instances
+                .values()
+                .max_by_key(|spec| spec.id)
+                .map(|spec| spec.id)
+                .unwrap_or(SUPERVISOR_INSTANCE_ID)
+                + 1;
+
+            // We're going to use ID obtained above, so the vacant ID is next to it.
+            let vacant_instance_id = new_instance_id + 1;
+            schema.initialize_vacant_instance_id_entry(vacant_instance_id);
```
slowli: Micro-nit / bikeshedding: I don't really see a purpose to extract a single assignment into a method...

PR #1558, opened 2019-11-20
Harsher naming restrictions in MerkleDB APIs [ECR-3834]

pavel-mukhanov:
Deny using dot in prefixes and service names.


components/merkledb/src/validation.rs
```diff
@@ -12,26 +23,43 @@ pub fn is_valid_index_name<S: AsRef<str>>(name: S) -> bool {
 /// Check that latin1 character is allowed in index name.
 /// Only these combination of symbols are allowed:
 ///
-/// `[0..9]`, `[a-z]`, `[A-Z]`, `_`, `-`, `.`
+/// `[0..9]`, `[a-z]`, `[A-Z]`, `_`, `-`
```
popzxc: Well, dot **is** a valid latin1 character, so now this function is not really correct. I'd suggest rename it to the `is_allowed_index_name_char` or something like this.

components/merkledb/src/validation.rs
```diff
@@ -12,26 +23,43 @@ pub fn is_valid_index_name<S: AsRef<str>>(name: S) -> bool {
...
 pub fn is_allowed_latin1_char(c: u8) -> bool {
     match c {
-        48..=57   // 0..9
+        48..=57     // 0..9
```
popzxc: Hm. Maybe use byte character literals? `b'0'..=b'9'` is much more self-explanatory.
Same applies for `is_valid_index_name`: `b'.'` is better than 46 IMO.

components/merkledb/src/validation.rs
```diff
@@ -1,7 +1,18 @@
...
 pub fn is_valid_index_name<S: AsRef<str>>(name: S) -> bool {
+    name.as_ref()
+        .as_bytes()
+        .iter()
+        .copied()
+        .all(|c| is_allowed_latin1_char(c) || c == 46)
+}
+
+/// Validate index name prefix, it shouldn't contain the dot.
+pub fn is_valid_prefix<S: AsRef<str>>(name: S) -> bool {
```
popzxc: AFAICS, this is used not only for prefix. Maybe something like `is_valid_index_name_component` here and `is_valid_index_full_name` instead of `is_valid_index_name`?

components/merkledb/src/validation.rs
```diff
@@ -12,26 +23,43 @@ pub fn is_valid_index_name<S: AsRef<str>>(name: S) -> bool {
...
+        | 45        // -
         | 95        // _
         => true,
         _ => false,
     }
 }
 
 /// Calls the `is_valid_name` function with the given name and panics if it returns `false`.
-pub(crate) fn assert_valid_name<S: AsRef<str>>(name: S) {
+pub(crate) fn assert_index_valid_name<S: AsRef<str> + fmt::Debug>(name: S) {
```
popzxc: Nit: a private helper HOF taking a predicate can be introduced here to avoid code duplication.

exonum/src/runtime/types.rs
```diff
@@ -167,9 +167,9 @@ impl ArtifactId {
 
     /// Check that the artifact name contains only allowed characters and is not empty.
     fn is_valid_name(name: impl AsRef<[u8]>) -> bool {
-        // Extended version of `exonum_merkledb::is_valid_name` that also allows ':`.
+        // Extended version of `exonum_merkledb::is_valid_name` that also allows '.' and ':'.
```
popzxc: Is dot allowed for artifact names?

exonum/src/runtime/types.rs
```diff
@@ -369,11 +371,11 @@ fn parse_artifact_id_incorrect_layout() {
         ("ava:123", "invalid digit found in string"),
         (
             "123:I am a service!",
-            "Artifact name contains an illegal character",
+            "Artifact name(I am a service!) contains an illegal character",
```
popzxc: Error message will look better like `Artifact name ("I am a service!") contains an illegal character`

exonum/src/runtime/types.rs
```diff
@@ -292,8 +294,8 @@ impl InstanceSpec {
             "Service instance name should not be empty"
         );
         ensure!(
-            is_valid_index_name(name),
-            "Service instance name contains illegal character, use only: a-zA-Z0-9 and one of _-."
+            is_valid_prefix(name),
+            "Service instance name contains illegal character, use only: a-zA-Z0-9 and one of _-"
```
popzxc: There is no test for this new restriction.

exonum/src/runtime/types.rs
pavel-mukhanov: Based on task description for artifact names yes, for services not.

exonum/src/runtime/types.rs
```diff
@@ -167,22 +167,24 @@ impl ArtifactId {
 
     /// Check that the artifact name contains only allowed characters and is not empty.
```
aleksuss:
```suggestion
    /// Checks that the artifact name contains only allowed characters and is not empty.
```

exonum/src/runtime/types.rs
```diff
@@ -167,22 +167,24 @@ impl ArtifactId {
...
+            46 | 58 => true,
+            c => is_allowed_index_name_char(c),
         })
     }
 }
 
 impl ValidateInput for ArtifactId {
     type Error = failure::Error;
 
+    /// Check that the artifact name contains only allowed characters and is not empty.
```
aleksuss:
```suggestion
    /// Checks that the artifact name contains only allowed characters and is not empty.
```

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,57 @@
 //! Validation helpers for index names.
 
+use core::fmt;
+
 /// Validate index name.
```
aleksuss:
```suggestion
/// Validates index name.
```

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,57 @@
...
     name.as_ref()
         .as_bytes()
         .iter()
         .copied()
-        .all(is_allowed_latin1_char)
+        .all(|c| is_allowed_index_name_char(c) || c == 46)
 }
 
-/// Check that latin1 character is allowed in index name.
+/// Validate index name prefix, it shouldn't contain the dot.
```
aleksuss:
```suggestion
/// Validates index name prefix, it shouldn't contain the dot.
```

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,57 @@
...
+pub fn is_valid_index_name_component<S: AsRef<str>>(name: S) -> bool {
+    name.as_ref()
+        .as_bytes()
+        .iter()
+        .copied()
+        .all(is_allowed_index_name_char)
+}
+
+/// Check that character is allowed in index name.
```
aleksuss:
```suggestion
/// Checks that character is allowed in index name.
```

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,57 @@
...
-    if !is_valid_index_name(name) {
-        panic!("Wrong characters using in name. Use: a-zA-Z0-9 and _-.");
+    if !predicate(name) {
+        panic!("Wrong characters using in name ({:?}). {}", name, desc);
     }
 }
+
+/// Calls the `is_valid_index_full_name` function with the given name
+/// and panics if it returns `false`.
+pub(crate) fn assert_index_valid_full_name<S: AsRef<str> + Copy + fmt::Debug>(name: S) {
```
slowli: Nit: Since this and the following methods are crate-specific, could they be simplified to accept `name: &str`?

exonum/src/runtime/types.rs
```diff
@@ -165,24 +165,26 @@ impl ArtifactId {
...
 
-    /// Check that the artifact name contains only allowed characters and is not empty.
+    /// Checks that the artifact name contains only allowed characters and is not empty.
     fn is_valid_name(name: impl AsRef<[u8]>) -> bool {
-        // Extended version of `exonum_merkledb::is_valid_name` that also allows ':`.
+        // Extended version of `exonum_merkledb::is_valid_name` that also allows '.' and ':'.
         name.as_ref().iter().all(|&c| match c {
-            58 => true,
-            c => is_allowed_latin1_char(c),
+            46 | 58 => true,
```
slowli: Nit: It's possible to use `b'.' | b':'` here, right?

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,57 @@
...
+    assert_valid_name(name, is_valid_index_full_name, "Use: a-zA-Z0-9 and _-.")
+}
+
+/// Calls the `is_valid_index_name_component` function with the given
+/// name and panics if it returns `false`.
+pub(crate) fn assert_valid_index_name_component<S: AsRef<str> + Copy + fmt::Debug>(name: S) {
```
slowli: Nit: I'd name this `assert_valid_name_component`, since the method isn't always used for indexes (e.g., it's used for `Group`).

components/merkledb/src/group.rs
```diff
@@ -51,6 +52,7 @@ where
     I: FromAccess<T>,
 {
     fn from_access(access: T, addr: IndexAddress) -> Result<Self, AccessError> {
+        assert_valid_index_name_component(&addr.name);
```
slowli: Why is naming checked for component here? It should be possible to create a group with a name like `foo.bar`, isn't it?

components/merkledb/src/views/metadata.rs
```diff
@@ -318,7 +318,7 @@ where
         index_address: &IndexAddress,
         index_type: IndexType,
     ) -> Result<Self, Self> {
-        assert_valid_name(index_address.name());
+        assert_index_valid_full_name(index_address.name());
```
slowli: I don't know whether this makes sense within this PR (IMO, it does), but it could make sense to refactor this constructor to return `Result<Self, AccessError>`, and add `AccessError` variants for an invalid name and for names used by the system (which could be defined as a name starting with `__`).

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,57 @@
...
-/// Validate index name.
-pub fn is_valid_index_name<S: AsRef<str>>(name: S) -> bool {
+use core::fmt;
+
+/// Validates index name.
+pub fn is_valid_index_full_name<S: AsRef<str>>(name: S) -> bool {
     name.as_ref()
         .as_bytes()
         .iter()
         .copied()
```
slowli: `.copied()` could be avoided since there is an explicit closure later anyway.

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,57 @@
...
-        .all(is_allowed_latin1_char)
+        .all(|c| is_allowed_index_name_char(c) || c == 46)
```
slowli: Nit: It's possible to use `b'.'` here, right?

CHANGELOG.md
```diff
@@ -63,6 +65,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Protobuf serialization for crypto types are now implemented in `exonum-crypto`
   crate (#1496).
 
+#### exonum-merkledb
+
+- Dot symbol is not allowed in index names and group prefixes anymore . (#1558)
```
slowli: This looks like a bogus description. The dot symbol is not allowed in:

- `Prefixed` accesses
- index / group / component suffixes **in the `FromAccess` derive macro** (which is implemented in #1562; I'll update it accordingly)

Indexes, groups, etc. created manually should still allow dots in their names.

components/derive/src/service_factory.rs
```diff
@@ -39,7 +33,10 @@ fn is_allowed_latin1_char(c: u8) -> bool {
 ///
 /// `[0..9]`, `[a-z]`, `[A-Z]`, `_`, `-`, `.`, ':'
```
alekseysidorov: Is this comment stil actual?

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,57 @@
 //! Validation helpers for index names.
 
-/// Validate index name.
-pub fn is_valid_index_name<S: AsRef<str>>(name: S) -> bool {
+use core::fmt;
```
alekseysidorov: Why not `std::fmt`?

components/merkledb/src/validation.rs
alekseysidorov: Please describe which symbol has this code.

components/merkledb/src/validation.rs
```diff
@@ -55,14 +55,14 @@ where
 {
     if name.as_ref().is_empty() {
         Err("Index name must not be empty".into())
-    } else if !predicate(name) {
+    } else if predicate(name) {
```
popzxc: Ok case between error cases looks weird.

pavel-mukhanov: Yes, but clippy doesn't agree. I think it better to turn it off.

popzxc: maybe 
```rust
if name.as_ref().is_empty() {
        Err(...)
} else if !predicate(name) {
        Err(...)
} else {
        Ok(...)
}
```

it looks more natural.

pavel-mukhanov: It was previous variant, I've changed it due to clippy complains. Will revert it back and turn off clippy.

components/merkledb/src/access/mod.rs
```diff
@@ -143,6 +135,14 @@ pub enum AccessErrorKind {
         actual: IndexType,
     },
 
+    /// Index has invalid name.
+    #[fail(display = "{}", _0)]
+    InvalidIndexName(String),
+
+    /// System index has invalid name.
+    #[fail(display = "{}", _0)]
+    InvalidSystemIndexName(String),
```
slowli: This variant should correspond to the case when client code attempts to access a system view / index (e.g., "\_\_METADATA\_\_"). Correspondingly, it should not contain an error message, but rather be an empty variant.

components/merkledb/src/proof_map_index/key.rs
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This lint is triggered
```
popzxc: I'd add an explanation from @slowli [comment](https://github.com/exonum/exonum/pull/1558/files/430f5ff361563109d703d7778a7777c35e4aa6ec..a563a313f4d7520a2b60c3ce047edef007500133#r349547834) for future readers.

components/merkledb/src/lib.rs
```diff
@@ -194,3 +194,7 @@ pub mod proto;
 
 #[cfg(feature = "with-protobuf")]
 use exonum_proto::ProtobufConvert;
+
+#[cfg(test)]
+#[macro_use]
+extern crate assert_matches;
```
slowli: Why is this declaration required? In the 2018 edition, macros could be exported from a crate just like other items.

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,128 @@
...
+#[allow(clippy::if_not_else)]
+fn check_valid_name<F>(
+    addr: IndexAddress,
+    predicate: F,
+    allowed_chars: String,
+) -> Result<(), AccessError>
+where
+    F: Fn(String) -> bool,
+{
+    let name = addr.name.clone();
```
slowli: Why is cloning necessary here? I'd rather that `name` be cloned exactly if it's needed (i.e., if it's returned within an error).

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,128 @@
...
+        assert_matches!(e.kind, AccessErrorKind::EmptyName);
+        let e = ListIndex::<_, u32>::from_access(&fork, "__METADATA__".into()).unwrap_err();
+        assert_matches!(e.kind, AccessErrorKind::ReservedName);
+        // spell-checker:disable
+        let e = ListIndex::<_, u32>::from_access(
+            &fork,
+            "\u{441}\u{43f}\u{438}\u{441}\u{43e}\u{43a}".into(),
+        )
+        .unwrap_err();
+        assert_matches!(e.kind, AccessErrorKind::InvalidCharsInName { .. } );
```
slowli: Nit: Redundant space before `)`.

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,128 @@
...
+    addr: IndexAddress,
+    predicate: F,
+    allowed_chars: String,
+) -> Result<(), AccessError>
+where
+    F: Fn(String) -> bool,
+{
+    let name = addr.name.clone();
+
+    if name.starts_with("__") {
```
slowli: ⚠️ System checks are valid only for full names. For example, `__METADATA__` is a reserved name, but `cryptocurrency.__meta__` is not. In fact, I'd argue something like `__meta__` or `__version__` is a good place to store metadata / version for service schemas.

components/merkledb/src/views/metadata.rs
```diff
@@ -332,19 +333,25 @@ where
...
         });
         let real_index_type = metadata.index_type;
-        let mut index_address = metadata.index_address();
+        let mut index_address_from_metadata = metadata.index_address();
         // Set index address name, since metadata itself doesn't know it.
-        index_address.name = index_name;
+        index_address_from_metadata.name = index_name;
         let this = Self {
-            view: View::new(index_access, index_address),
+            view: View::new(index_access, index_address_from_metadata.clone()),
```
slowli: Nit: Is cloning necessary here?

components/merkledb/src/views/tests.rs
```diff
@@ -1000,7 +1001,7 @@ fn valid_name_for_url() {
 }
 
 #[test]
-#[should_panic(expected = "Wrong characters using in name")]
+#[should_panic(expected = "InvalidCharsInName")]
```
slowli: Not the best way to report panics to users. It could make sense to replace

```rust
SomeIndex::from_access(self, addr.into()).unwrap()
```

in `AccessExt` with something more readable, e.g.

```rust
SomeIndex::from_access(self, addr.into()).unwrap_or_else(|e| {
    panic!("MerkleDB error: {}", e);
})
```

Could be done in a separate PR, I guess.

components/merkledb/src/access/mod.rs
```diff
@@ -143,6 +135,27 @@ pub enum AccessErrorKind {
...
+    #[fail(display = "Index name is reserved")]
+    ReservedName,
+
+    /// Index name is empty.
+    #[fail(display = "Index name must not be empty")]
+    EmptyName,
+
+    /// Index contains invalid characters.
+    #[fail(
+        display = "Invalid characters using in name ({}). Use {}",
```
slowli: Nit: Probably should be "Invalid characters **used** in name".

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,128 @@
...
-        panic!("Index name must not be empty")
+// Allow because it's looks more readable.
+#[allow(clippy::if_not_else)]
+fn check_valid_name<F>(
+    addr: IndexAddress,
+    predicate: F,
+    allowed_chars: String,
+) -> Result<(), AccessError>
+where
+    F: Fn(String) -> bool,
```
slowli: Why does a predicate require `String` rather than `&str`? Doesn't look like the string should be consumed by it.

components/merkledb/src/views/tests.rs
pavel-mukhanov: Sure, will do.

components/merkledb/src/validation.rs
```diff
@@ -1,37 +1,128 @@
...
+            },
+            addr,
+        })
+    } else {
+        Ok(())
     }
+}
+
+/// Calls the `is_valid_index_full_name` function with the given index address.
+pub(crate) fn check_index_valid_full_name(addr: &IndexAddress) -> Result<(), AccessError> {
```
popzxc: We have `is_valid_index_full_name` and `check_index_valid_full_name` which are expected to be symmetric, but in fact it's possible to make the first function return `true` and the second one return `Err(..)` for the same input. Is that correct?

pavel-mukhanov: Yes, they can return different results and I think it's better to rename `is_valid_index_full_name`.

PR #1560, opened 2019-11-21
Update changelog [ECR-3516]

popzxc:
Hooray, I guess?

Once this gets merged, feel free to turn on your "please-update-a-changelog"-o-meter.

[A link to the rendered version of `CHANGELOG.md`](https://github.com/popzxc/exonum/blob/update-changelog/CHANGELOG.md)

This update schema:
- This is not a complete migration guide / documentation. Only a short description was added with links to the docs/examples.
- Missing changelog entries for non-dymanic-services changes were added.
- For dynamic services, a four-level description is given:
  1. A short overview.
  2. A short migration guide (not really a guide, more of "where you can find the info" section).
  3. A list of important PRs within implementing the feature.
  4. A list of ALL PRs related to this feature. "Garbage" PRs (e.g. "merge master") are filtered out. However, PRs from `3` are included.
- __Important:__ Inserted links are adapted for the release. They refer to the 0.13 release artifacts, so they don't work now, but will become correct on the release.

PR #1562, opened 2019-11-21
Implement `FromAccess` macro [ECR-3832]

slowli:
## Overview

This PR implements derive macro for `FromAccess`.

---

See: https://jira.bf.local/browse/ECR-3832

### Definition of Done


components/derive/src/db_traits.rs
```diff
@@ -116,3 +120,310 @@ pub fn impl_object_hash(input: TokenStream) -> TokenStream {
...
+            impl #impl_generics #tr<#access_ident> for #name #ty_generics #where_clause {
+                #from_access_fn
+                #from_root_fn
+            }
+        };
+        tokens.extend(expanded);
+
+        if self.attrs.is_schema() {
+            let expanded = quote! {
+                impl #impl_generics #name #ty_generics #where_clause {
```
alekseysidorov: Perhaps we might to introduce a new trait like `FromAccess` but for constructing schemas like `IntoSchema` to avoid of ad-hoc `new` method implementation. In my opinion this approach is more generic and a little bit closer to `derive` philoshophy.

alekseysidorov: Or even more: we can use `FromAccess` for both cases.

```rust

impl<T> FromAccess for Schema<T> { ... }

// and

impl<T, A> FromAccess for SomeIndex<T> for (T, A) where A: Into<IndexAddress> { .. }
```

slowli: Then this trait would need to be imported into the scopes that want to use `Schema::new()`. It's possible to introduce `new()` in `FromAccess` or in an extension trait to `FromAccess`. Would any of this make sense to you? 🙃

Also, regarding `(T, A) where A: Into<IndexAddress>`: does this assume removing `from_access` method from `FromAccess`? I'm not sure this is going to work with the transparent representation, or would be composable in general.

aleksuss:
@slowli I suppose it'd be good to add an entry to CHANGELOG ?

popzxc:
I think it'd better be done once #1560 will be merged.

PR #1578, opened 2019-11-26
Explicit panic invocation in from_access

pavel-mukhanov:
Invoke `panic!` to make errors more verbose.

### DIFF
--- a/components/merkledb/src/access/extensions.rs
+++ b/components/merkledb/src/access/extensions.rs
@@ -46,7 +46,9 @@ pub trait AccessExt: Access {
         I: FromAccess<Self>,
     {
         // We know that `Group` implementation of `Restore` never fails
-        Group::from_access(self, IndexAddress::with_root(name)).unwrap()
+        Group::from_access(self, IndexAddress::with_root(name)).unwrap_or_else(|e| {
+            panic!("MerkleDB error: {}", e);
+        })
     }
 
     /// Gets an entry index with the specified address.
@@ -59,7 +61,9 @@ pub trait AccessExt: Access {
         I: Into<IndexAddress>,
         V: BinaryValue,
     {
-        Entry::from_access(self, addr.into()).unwrap()
+        Entry::from_access(self, addr.into()).unwrap_or_else(|e| {
+            panic!("MerkleDB error: {}", e);
+        })
     }
 
     /// Gets a list index with the specified address.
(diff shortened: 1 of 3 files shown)


components/merkledb/src/access/extensions.rs
```diff
@@ -46,7 +46,9 @@ pub trait AccessExt: Access {
         I: FromAccess<Self>,
     {
         // We know that `Group` implementation of `Restore` never fails
-        Group::from_access(self, IndexAddress::with_root(name)).unwrap()
+        Group::from_access(self, IndexAddress::with_root(name)).unwrap_or_else(|e| {
```
popzxc: I believe that since we don't use `format!(..)`, `expect` is sufficient here, since it contains the error message ([1](https://doc.rust-lang.org/src/core/result.rs.html#957-962), [2](https://doc.rust-lang.org/src/core/result.rs.html#1164)) as well.

pavel-mukhanov: `expect` would trigger clippy's [expect_fun_call](https://rust-lang.github.io/rust-clippy/master/#expect_fun_call) lint.

popzxc: Why? It gets triggered if `format` is used, and here it will be just `.expect("MerkleDB error")` (as I mentioned, error message will be added automatically.

pavel-mukhanov: You are right, changed to expect.

slowli: `expect` uses a `Debug` representation of an error though, while we might want a `Display` one.

popzxc: Do we? I thought that for `panic` (which is not meant to be seen by end-side users) should contain `Debug` representation. It's sort of similar to the python `__repr__`/`__str__`: the first is meant to be seen by a programmer (and should be unambiguous), and the latter is meant for users.

slowli: I consider service developers to belong to the "users" category. This is debatable, but I'd argue that a service developer seeing an error like

```text
Error accessing `currency.wallets`: Wrong index type: expected ProofMapIndex, but got MapIndex
```

is more likely to understand / fix it than if she sees 

```text
AccessError { addr: IndexAddress { name: "currency.wallets", bytes: None }, kind: WrongIndexType { expected: ProofMapIndex, actual: MapIndex } }
```

...especially if we're talking about Java developers.

popzxc: Well, makes sense. I'm not against it.

pavel-mukhanov: 👌

PR #1565, opened 2019-11-22
Improve transaction broadcast [ECR-3817]

slowli:
## Overview

This PR unifies transaction broadcast for HTTP API and `after_commit` methods in Rust services.

---

See: https://jira.bf.local/browse/ECR-3817

### Definition of Done

- [ ] There are no TODOs left in the merged code
- [ ] Change is covered by automated tests
- [ ] Benchmark results are attached (if applicable)
- [ ] The [coding guidelines] are followed
- [ ] Public API has proper documentation
- [ ] Changelog is updated if needed (in case of notable or breaking changes)
- [ ] The continuous integration build passes

[coding guidelines]: https://github.com/exonum/exonum/blob/master/CONTRIBUTING.md#conventions

PR #1579, opened 2019-11-26
Check service ID of transaction before accepting to mempool [ECR-3816]

popzxc:
**New changes:**

- Many consensus-related methods of `NodeHandler` are made private.
- Consensus rules are now forbid adding transactions for which `BlockchainMut::check_tx` failed to the block. In particular, that means the following:
  1. Invalid transaction hashes are stored into temporary in-memory set for node to be able to lookup for it without storing it in the persistent memory.
  2. If block propose contains transaction known as invalid, node won't send prevote for it.
  3. If majority of the nodes will decide to approve the block with transaction known as invalid, node will panic (because it's either node implementation bug, or majority of nodes work incorrectly).
  4. If node lagging behind receives a committed block with the incorrect transaction, it panics as well.
- Sandbox tests are added to cover cases 3 and 4 from the list above.

-----

This PR introduces a quasi-new method `check_tx`, which performs some checks on tx before accepting it to the mempool.

Currently, the only check there is correctness of the destination, but later it (hopefully) will be extended with more mature validation.

Since we sort-of had this functionality in 0.12, I guess changelog shouldn't be updated here.


exonum/src/blockchain/mod.rs
```diff
@@ -449,6 +449,15 @@ impl BlockchainMut {
             .expect("Cannot update transaction pool");
     }
 
+    /// Performs several shallow checks that transaction is correct.
+    ///
+    /// Returned `Ok(())` value doesn't necessarily mean that transaction is correct and will be
+    /// executed successfully, but returned `Err(..)` value means that this transaction is
+    /// **obviously** incorrect and should be declined as early as possible.
+    pub(crate) fn check_tx(&self, tx: &Verified<AnyTx>) -> Result<(), ExecutionError> {
```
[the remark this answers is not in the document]
popzxc: `is_processable` doesn't seem good to me, since this transaction can be **not** processable even with `Ok(())` returned from this method.
And since this method will be extended in the future, I believe we'll be able to figure out a better name once it'll happen.

slowli:
- Shall the checks be expanded to cover the transactions submitted to the testkit via `create_block_*` methods (e.g., to end the corresponding method with a panic)?

- Are blocks with transactions from non-existing services incorrect as per consensus rules now? Ideally, a test or two should be added to the sandbox on whether a node can process such a block.

exonum/src/blockchain/mod.rs
```diff
@@ -449,6 +449,15 @@ impl BlockchainMut {
             .expect("Cannot update transaction pool");
     }
 
+    /// Performs several shallow checks that transaction is correct.
```
aleksuss:
```suggestion
    /// Performs that the transaction corresponds to existing instance of the service.
```
???

popzxc: As it said in the body of the `Dispatcher::check_tx`: "Currently the only check is that destination service exists, but later functionality of this method can be extended.".

Since this method just redirects a request, I believe it shouldn't assume exact checks that will happen.

And regarding the same doc-comment on the `Dispatcher::check_tx` -- I believe it's better to update its description once it will be stabilized.

aleksuss: Ok. But there is no guarantee that the behaviour will be changed in the future. 🙃

popzxc: At least task in Jira proposes it. Anyway, it worth nothing to update description (and the name of the function) to match the nature of the function later, if we'll decide to leave things as-is.

popzxc:
PR description was updated to match the current state.

exonum/src/node/consensus.rs
```diff
@@ -613,16 +680,31 @@ impl NodeHandler {
...
-    pub fn handle_incoming_tx(&mut self, msg: Verified<AnyTx>) {
+    pub(crate) fn handle_incoming_tx(&mut self, msg: Verified<AnyTx>) {
         trace!("Handle incoming transaction");
+        // Check transaction before handling it to avoid the broadcasting.
+        // Reaction for an incorrect tx here and in `handle_tx` is different:
+        // here we simply ignore it without notifying the network, but if the
+        // transaction is received from the network, we have to deal with it
+        // and maybe even panic (if most of nodes will approve a block with such a
+        // transaction).
+        if let Err(error) = self.blockchain.check_tx(&msg) {
```
slowli: Off-topic: I wonder if it makes sense to avoid this check in favor of checks in the places where the external transactions are accepted. For example, it could lead to more informative errors in the explorer's `POST transactions` endpoint. Note that per #1565, Rust services can no longer send transactions to any recipient service other than themselves, although it's not true (yet?) for the Java runtime, and there is a possibility to use `Blockchain::sender()` directly.

exonum/src/node/state.rs
```diff
@@ -92,6 +92,16 @@ pub struct State {
...
     tx_cache: BTreeMap<Hash, Verified<AnyTx>>,
+
+    // An in-memory set of transaction hashes, rejected by a node
+    // within block.
+    //
+    // Those transactions are stored to be known if some node will propose a block
+    // with one of them, so node could lookup for it.
+    //
+    // This set is cleared every block.
+    invalid_txs: HashSet<Hash>,
```
slowli: This may be a vector for DoS attacks by memory exhaustion. However, if implementing a more complex solution (e.g., using an LRU cache with the bounded size) is significantly more complex, I'd go with the simpler solution for now and improve it incrementally (we have a "DoS resistance" milestone).

exonum/src/sandbox/consensus/round_details.rs
```diff
@@ -622,6 +627,84 @@ fn handle_precommit_different_block_hash() {
...
+    );
+    sandbox.send(
+        sandbox.public_key(ValidatorId(1)),
+        &make_request_prevote_from_precommit(&sandbox, precommit_1.as_ref()),
+    );
+    sandbox.recv(&propose);
+    sandbox.recv(&incorrect_tx);
+
+    // In normal conditions, here we should sent prevote, but since we consider
+    // the propose to be incorrect, we won't do it.
```
slowli: Is absence of prevote checked automatically? (In `recv`?) Could this check be made explicit?

exonum/src/sandbox/consensus/round_details.rs
```diff
@@ -1631,6 +1714,37 @@ fn handle_tx_ignore_existing_tx_in_blockchain() {
...
+    // add rounds & become leader
+    sandbox.add_time(Duration::from_millis(sandbox.current_round_timeout()));
+    assert!(sandbox.is_leader());
+
+    // Create correct tx, and then sign with the wrong destination.
+    let (pk, sk) = gen_keypair();
+    let incorrect_tx = gen_unverified_timestamping_tx().sign(TimestampingService::ID + 1, pk, &sk);
+
+    // Receive that message.
+    sandbox.recv(&incorrect_tx);
```
slowli: Nit: Is it possible to check that the incorrect transaction is not in the mempool here?

exonum/src/node/consensus.rs
```diff
@@ -613,16 +680,31 @@ impl NodeHandler {
     /// Handles external boxed transaction. Additionally transaction will be broadcast to the
     /// Node's peers.
     #[cfg_attr(feature = "cargo-clippy", allow(clippy::needless_pass_by_value))]
-    pub fn handle_incoming_tx(&mut self, msg: Verified<AnyTx>) {
+    pub(crate) fn handle_incoming_tx(&mut self, msg: Verified<AnyTx>) {
```
slowli: Nit / future work: Could make sense to have a sandbox test for an invalid transaction.

exonum/src/node/consensus.rs
```diff
@@ -238,12 +238,50 @@ impl NodeHandler {
...
+    fn check_propose_and_broadcast_precommit(
+        &mut self,
+        round: Round,
+        propose_hash: Hash,
+        block_hash: Hash,
+    ) {
+        // Do not send a prevote if propose contains incorrect transactions.
+        let propose_state = self.state.propose(&propose_hash).unwrap();
+        if propose_state.has_invalid_txs() {
+            warn!("Denying sending a precommit for a propose with incorrect transactions");
```
slowli: Nit / future work: A sandbox test covering this case would be appreciated.

exonum/src/node/state.rs
```diff
@@ -904,6 +938,8 @@ impl State {
                 // saving a propose to the cache. Think about making this approach less error-prone.
                 // (ECR-1635)
                 is_saved: true,
+                // We expect ourself not to produce invalid proposes.
+                valid: true,
```
slowli: Nit: `is_valid` would look more in line with `is_saved`.

test-suite/testkit/src/lib.rs
```diff
@@ -430,10 +430,15 @@ impl TestKit {
...
         let snapshot = self.snapshot();
         let schema = snapshot.for_core();
-        let uncommitted_txs = transactions.into_iter().filter(|tx| {
-            !schema.transactions().contains(&tx.object_hash())
-                || schema.transactions_pool().contains(&tx.object_hash())
-        });
+        let uncommitted_txs: Vec<_> = transactions
+            .into_iter()
+            .filter(|tx| {
+                self.check_tx(&tx);
```
slowli: Nit / future work: Could make sense to add a couple `TestKit` tests covering an invalid transaction in `create_block_*` arguments, and in `add_tx`.

exonum/src/node/consensus.rs
```diff
@@ -345,7 +383,7 @@ impl NodeHandler {
 
     /// Locks to the propose by calling `lock`. This function is called when node receives
     /// +2/3 pre-votes.
-    pub fn handle_majority_prevotes(&mut self, prevote_round: Round, propose_hash: Hash) {
+    fn handle_majority_prevotes(&mut self, prevote_round: Round, propose_hash: Hash) {
```
slowli: Seems that there can't be +2/3 `Prevote`s (or even +1/3) for an invalid `Propose` because a correctly operating validator must not send `Prevote`s for invalid proposals (which is ensured by `check_propose_and_broadcast_prevote`). Thus, could it make sense to panic here similarly to `handle_majority_precommits`?

exonum/src/sandbox/consensus/round_details.rs
```diff
@@ -622,6 +627,84 @@ fn handle_precommit_different_block_hash() {
...
+    sandbox.send(
+        sandbox.public_key(ValidatorId(1)),
+        &make_request_propose_from_precommit(&sandbox, precommit_1.as_ref()),
+    );
+    sandbox.send(
+        sandbox.public_key(ValidatorId(1)),
+        &make_request_prevote_from_precommit(&sandbox, precommit_1.as_ref()),
+    );
+    sandbox.recv(&propose);
+    sandbox.recv(&incorrect_tx);
```
slowli: Nit / future work: Could make sense to test a case when `incorrect_tx` is received before the `propose`.

exonum/src/sandbox/consensus/block_request.rs
```diff
@@ -218,6 +223,108 @@ fn handle_block_response_with_unknown_tx() {
     ));
 }
 
+/// - should **NOT** process block if tx is incorrect
+/// idea of test is:
+/// - getting Status from other node with later height, send BlockRequest to this node
+/// - receive BlockResponse with unknown tx A
```
slowli: Nit / future work: Could make sense to add a case when the transaction is known to be incorrect to the node.

exonum/src/node/consensus.rs
```diff
@@ -546,24 +595,42 @@ impl NodeHandler {
...
         let snapshot = self.blockchain.snapshot();
         let schema = Schema::new(&snapshot);
-
         if contains_transaction(&hash, &schema.transactions(), self.state.tx_cache()) {
             bail!("Received already processed transaction, hash {:?}", hash)
         }
 
-        // TODO We have to check transaction correctness.
+        if let Err(e) = self.blockchain.check_tx(&msg) {
+            // Store transaction as invalid to know it if it'll be included into a proposal.
```
slowli: It could make sense to add in the comment that marking a transaction as invalid **must** occur before calling `check_incomplete_proposes`, since the latter uses `invalid_txs` to recalculate the validity of proposals.

exonum/src/sandbox/consensus/round_details.rs
popzxc: Somewhat, yes. If I understand things correctly, `sandbox` forbids the situation when node sends a message, but `sandbox.send` with this message is not called. Not sure how to make that more explicit, since it's sort of essential concept of the `sandbox`.

popzxc: To be precise, `recv` contains `self.check_unexpected_message();` check. We can move it out, but then we'll have to check all the sandbox test not to rely on this (which will take a lot of time I think).

exonum/src/node/consensus.rs
```diff
@@ -238,12 +238,50 @@ impl NodeHandler {
...
 
+    /// Checks if propose is correct (doesn't contain invalid transactions), and then
+    /// broadcasts a prevote for this propose.
+    ///
+    /// Returns `true` if majority of prevotes is achieved, and returns `false` otherwise.
+    fn check_propose_and_broadcast_prevote(&mut self, round: Round, propose_hash: Hash) -> bool {
+        // Do not send a prevote if propose contains incorrect transactions.
+        let propose_state = self.state.propose(&propose_hash).unwrap();
+        if propose_state.has_invalid_txs() {
+            warn!("Denying sending a prevote for a propose with incorrect transactions");
```
alekseysidorov: I wonder if we need to save this incorrect proposal for the futher analysis?

exonum/src/node/consensus.rs
```diff
@@ -546,24 +595,42 @@ impl NodeHandler {
 
     /// Checks if the transaction is new and adds it to the pool. This may trigger an expedited
     /// `Propose` timeout on this node if transaction count in the pool goes over the threshold.
-    pub fn handle_tx(&mut self, msg: Verified<AnyTx>) -> Result<(), failure::Error> {
+    ///
+    /// Before adding a transaction into pool, this method calls `BlockchainMut::check_tx` to
+    /// ensure that transaction passes at least basic checks. If `BlockchainMut::check_tx` fails,
+    /// transaction will be considered invalid and not stored to the pool (instead, its hash will
+    /// be stored in the temporary invalid messages set, so we will be able to detect a block/propose
```
alekseysidorov: Which is lifetime of this temporary set of messages?

exonum/src/sandbox/timestamping.rs
```diff
@@ -93,16 +93,23 @@ impl TimestampingTxGenerator {
...
 impl Iterator for TimestampingTxGenerator {
     type Item = Verified<AnyTx>;
 
     fn next(&mut self) -> Option<Verified<AnyTx>> {
-        let mut data = vec![0; self.data_size];
-        self.rand.fill_bytes(&mut data);
-        let mut buf = TimestampTx::new();
-        buf.set_data(data);
-        Some(buf.sign(TimestampingService::ID, self.public_key, &self.secret_key))
+        let tx = self.create_unverified();
```
alekseysidorov: I'd suggest to change name to `gen_tx_payload`.

popzxc:
Benchmarking results of the `block` benches compared to master (requested by @aleksuss): [link](https://paste.ofcode.org/q5xzQ5NQhwPginr9y6aJNZ).

tl;dr: Nothing interesting.

Of course, those benchmarks don't include sending incorrect txs, but for rare occasions their impact won't be sufficient; and within normal workflow we don't expect most of transactions to be incorrect. More than that, externally-retrieved transactions are getting declined without broadcasting. Implementing a bench for the correctly setted-up scenario (emulate a byzantine node, which'll send wrong proposals internally) can take a while, and I'm not sure whether it's worth it.

exonum/src/sandbox/consensus/round_details.rs
slowli: Nah, the current approach is fine.

slowli:
> I'm not sure whether it's worth it

Not worth it in the short term; may become worth it in the longer term once "DoS resistance" epic starts getting developed.

exonum/src/node/consensus.rs
popzxc: Not sure about this. Can you explain use cases?

popzxc:
Most of review suggestions are implemented with exception for:

- Replacing `HashSet` with RLU cache (added TODO with DoS resistance ECR);
- Moving incorrect tx check to the explorer (created ECR-3907);
- Adding more tests for invalid txs processing (created ECR-3908).

exonum/src/sandbox/consensus/round_details.rs
popzxc: Well, it's not added to the pool at `recv` step, but to the tx cache. Added a check for that.

exonum/src/node/consensus.rs
```diff
@@ -389,6 +391,16 @@ impl NodeHandler {
         // Lock to propose
         if self.state.locked_round() < prevote_round && self.state.propose(&propose_hash).is_some()
```
slowli: Micro-nit: I think `if let Some(propose_state) = self.state.propose(&propose_hash)` would be more idiomatic.

PR #1580, opened 2019-11-26
Update examples readmes according to 0.13

ivan-ochc:


vitvakatu:
It isn't user-friendly, as they must create a config file themselves. I think the example config files must be provided.

examples/timestamping/timestamping.yaml
```diff
@@ -0,0 +1,37 @@
...
+deadline_height: 10000
+supervisor_mode: "simple"
+
+artifacts:
+  time:
+    runtime: rust
+    name: "exonum-time:0.12.0"
+  timestamping:
+    runtime: rust
+    name: "exonum-timestamping:0.0.0"
```
alekseysidorov: `exonum-cryptocurrency-advanced` has version `0.12.0`, but `exonum-timestamping` has `0.0.0` that looks a little bit inconsistent.

examples/cryptocurrency-advanced/README.md
```diff
@@ -117,6 +118,12 @@ exonum-cryptocurrency-advanced run --node-config example/3/node.toml --db-path e
 exonum-cryptocurrency-advanced run --node-config example/4/node.toml --db-path example/4/db --public-api-address 0.0.0.0:8203 --master-key-pass pass
 ```
 
+Deploy cryptocurrency-advanced service:
+
+```sh
+python3 -m exonum_launcher -i ../cryptocurrency_advanced.yaml
```
aleksuss: I suppose you need to describe the process of `exonum_launcher` module installing.

examples/timestamping/timestamping.yaml
vitvakatu: It is not config problem, but rather timestamping example problem. The version must be changed there as well

examples/cryptocurrency-advanced/README.md
ivan-ochc: There is a link to exonum-launcher above in the "Be sure you installed necessary packages:" section.

examples/cryptocurrency-advanced/README.md
```diff
@@ -117,6 +118,19 @@ exonum-cryptocurrency-advanced run --node-config example/3/node.toml --db-path e
 exonum-cryptocurrency-advanced run --node-config example/4/node.toml --db-path example/4/db --public-api-address 0.0.0.0:8203 --master-key-pass pass
 ```
 
+Before service deploy make sure that you have pure python implementation of protobuf:
```
vitvakatu: Interesting. @popzxc what is the reason? Shouldn't it be specified in the `exonum_launcher`?

ivan-ochc: It has been specified in exonum-python-client.

vitvakatu: I see now. Unfortunate problem :(

popzxc: >what is the reason?

The reason is extremely "good" protobuf implementation for python: it ignores the `package` field set in files and thinks that two files with the same name are always belong to the same descriptor (and we have plenty of those, e.g. `service.proto`). The intended way (I asked about it in the protobuf repo) is to compile files with the same name providing the folders name as well, so they eventually will become different names.

Pure python implementation softens those restrictions, but they've said that they want to implement this behaviour for pure python library as well :cold_sweat: 

I'm currently thinking what can we do about it, but for now using the pure python implementation is the simplest approach we can get.

examples/timestamping/README.md
```diff
@@ -111,6 +112,19 @@ exonum-timestamping run --node-config example/3/node.toml --db-path example/3/db
...
 ```
 
+Before service deploy make sure that you have pure python implementation of protobuf:
+
+```sh
+pip uninstall protobuf
+pip install --no-binary=protobuf protobuf
+```
+
+Deploy cryptocurrency-advanced service.
```
aleksuss:
```suggestion
Deploy timestamping service.
```

PR #1581, opened 2019-11-28
Clarify exonum-build docs

pavel-mukhanov:
## Overview

- Added crate level doc.
- Method `frequently_used` has been removed.
- Input directory is always added to includes to reduce boilerplate
  code

### DIFF
--- a/components/build/src/lib.rs
+++ b/components/build/src/lib.rs
@@ -13,6 +13,19 @@
 // limitations under the License.
 
 //! This crate simplifies writing build.rs for exonum and exonum services.
+//!
+//! In build.rs
+//! ```
+//! # use exonum_build::ProtobufGenerator;
+//! #
+//! # fn main() {
+//! #   ProtobufGenerator::with_mod_name("protobuf_mod.rs")
+//! #       .with_input_dir("src/proto")
+//! #       .add_path("src/proto")
+//! #       .with_frequently_used()
+//! #       .generate();
+//! # }
+//! ```
 
 #![deny(unsafe_code, bare_trait_objects)]
 #![warn(missing_docs, missing_debug_implementations)]
@@ -31,6 +44,15 @@ use std::{
 };
 
 /// Enum represents various sources of protobuf files.
+///
+/// Crypto sources contains all necessary crypto types used in services and system proto-files.
+/// These types are `Hash`, `PublicKey` and `Signature`.
+///
+/// Exonum sources contains types used in core and in system services such as supervisor.
+///
+/// Common sources currently contains only `BitVec` type used core consesus messages.
+///
+///
 #[derive(Debug, Copy, Clone)]
 pub enum ProtoSources<'a> {
     /// Path to exonum core protobuf files.
--- a/components/build/src/lib.rs
+++ b/components/build/src/lib.rs
@@ -12,24 +12,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! This crate simplifies writing build.rs for exonum and exonum services.
+//! This crate simplifies writing build.rs for Exonum and Exonum services.
 //!
 //! Since protobuf is exonum default serialization format, build.rs is mostly used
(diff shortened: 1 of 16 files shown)


components/build/src/lib.rs
```diff
@@ -13,6 +13,26 @@
 // limitations under the License.
 
 //! This crate simplifies writing build.rs for exonum and exonum services.
+//!
+//! Since protobuf is exonum default serialization format, build.rs is mostly used
+//! to compile protobuf files and generate corresponding code. This code is used later in
+//! exonum core and services.
+//!
+//! All you need to do is call `ProtobufGenerator` with required params, for example see
```
vitvakatu:
```suggestion
//! All you need to do is to call `ProtobufGenerator` with required params, for an example see
```

components/build/src/lib.rs
```diff
@@ -13,6 +13,26 @@
 // limitations under the License.
 
 //! This crate simplifies writing build.rs for exonum and exonum services.
+//!
+//! Since protobuf is exonum default serialization format, build.rs is mostly used
+//! to compile protobuf files and generate corresponding code. This code is used later in
```
vitvakatu:
```suggestion
//! to compile protobuf files and generate a corresponding code. This code is used later in
```

components/build/src/lib.rs
```diff
@@ -13,6 +13,26 @@
...
+//! exonum core and services.
```
vitvakatu:
```suggestion
//! the Exonum core and services.
```

components/build/src/lib.rs
```diff
@@ -13,6 +13,26 @@
 // limitations under the License.
 
 //! This crate simplifies writing build.rs for exonum and exonum services.
```
vitvakatu:
```suggestion
//! This crate simplifies writing build.rs for Exonum and Exonum services.
```

components/build/src/lib.rs
```diff
@@ -13,6 +13,26 @@
...
+//!
+//! Since protobuf is exonum default serialization format, build.rs is mostly used
+//! to compile protobuf files and generate corresponding code. This code is used later in
+//! exonum core and services.
+//!
+//! All you need to do is call `ProtobufGenerator` with required params, for example see
+//! [`ProtobufGenerator`] docs.
+//!
+//! There are three predefined include paths available to use with `ProtobufGenerator`.
```
vitvakatu:
```suggestion
//! There are three predefined sets of protobuf sources available for use.
```

components/build/src/lib.rs
```diff
@@ -13,6 +13,26 @@
...
+//! See [`ProtoSources`].
+//!
+//! - Crypto sources: contains all necessary crypto types used in services and system proto-files.
```
vitvakatu:
```suggestion
//! - Crypto sources: contains all the necessary crypto types used in services and system proto-files.
```

vitvakatu: What are system proto-files?

components/build/src/lib.rs
```diff
@@ -13,6 +13,26 @@
...
+//! These types are `Hash`, `PublicKey` and `Signature`.
+//!
+//! - Exonum sources: contains types used in core and in system services such as supervisor.
+//!
+//! - Common sources: currently contains only `BitVec` type and its used in core consensus messages.
```
vitvakatu:
```suggestion
//! - Common sources: currently contains only `BitVec` type and it is used in the core consensus messages.
```

components/build/src/lib.rs
```diff
@@ -228,9 +238,12 @@ impl<'a> ProtobufGenerator<'a> {
         }
     }
 
-    /// Directory containing input protobuf files.
+    /// Directory containing input protobuf files. This directory is also
+    /// added to includes because in most cases proto files in input directory depends on
```
vitvakatu: There are no `includes` for user, so this must be altered. Documentation of `add_path` must also be expanded. It's not obvious why can't I call `with_input_dir` multiple times for every my directory.

PR #1582, opened 2019-11-29
Update configuration change example [ECR-3894].

pavel-mukhanov:
Update configuration change example and move it to `exonum-supervisor` crate.


services/supervisor/examples/configuration_change.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+impl ConfigChangeInterface for ConfigChangeService {}
+
+// To allow service change its configuration we need to implement `Configure` trait.
+impl Configure for ConfigChangeService {
+    type Params = String;
+
+    fn verify_config(
```
popzxc: Maybe add something like `println!("Verify config called with params {}", params)` here and in `apply_config` to make this example more example-ish?

PR #1585, opened 2019-11-29
Improve `ExecutionError` [ECR-3902] [ECR-3819]

slowli:
## Overview

This PR improves documentation for execution errors.

Bikeshedding:

- Are "checked" / "unchecked" good terms?

---

See: https://jira.bf.local/browse/ECR-3902

### Definition of Done


exonum/src/runtime/error.rs
```diff
@@ -26,57 +26,112 @@ use crate::{
...
+/// ## `Runtime` errors
+///
+/// Use `Runtime` kind if the error has occurred in the runtime code and it makes sense to report
+/// the error to the users. A primary example here is artifact deployment: if deployment has failed,
+/// a `Runtime` error can provide more details about the cause.
+///
+/// ## Panics
+///
+/// Panic in the Rust wrapper of the runtime if a fundamental runtime invariant is broken and
+/// continuing node operation is impossible. A panic will not be caught and will lead
```
slowli: Note that this is not precisely correct in the current `master` (panics in the runtime code are caught for transactions and `before_commit`s), but will become true once `catch_panic` calls are moved to the Rust runtime as per ECR-3819.

exonum/src/runtime/error.rs
```diff
@@ -26,57 +26,112 @@ use crate::{
...
+///
+/// Use `Service` kind if the error has occurred in the service code and it makes sense to notify
+/// users about the error cause and/or its precise kind. These errors are generally raised
+/// if the input data (e.g., the transaction payload) violate certain invariants imposed by the service.
+/// For example, a `Service` error can be raised if the sender of a transfer transaction
+/// in the token service does not have sufficient amount of tokens.
+///
+/// ## `Unchecked` errors
+///
+/// Use `Unchecked` kind if the error has occurred in the service code, and:
```
popzxc: "...and one of the following is correct:"?

exonum/src/runtime/error.rs
```diff
@@ -26,57 +26,112 @@ use crate::{
...
+///
+/// - It is caused by the environment (e.g., out-of-memory)
+/// - It should never occur during normal execution (e.g., out-of-bounds indexing, null pointer)
```
popzxc: "null pointer dereference"?

exonum/src/runtime/error.rs
```diff
@@ -26,57 +26,112 @@ use crate::{
...
+/// - It otherwise makes no sense to report to users in detail.
```
popzxc: It's not really clear what exactly meant here.

exonum/src/runtime/error.rs
```diff
@@ -26,57 +26,112 @@ use crate::{
...
+/// Use `Dispatcher` kind if the error has occurred while dispatching the request (i.e., *not*
+/// in the client code). See [`DispatcherError`] for more details.
+///
+/// ## `Runtime` errors
+///
+/// Use `Runtime` kind if the error has occurred in the runtime code and it makes sense to report
+/// the error to the users. A primary example here is artifact deployment: if deployment has failed,
+/// a `Runtime` error can provide more details about the cause.
+///
+/// ## Panics
```
popzxc: I believe the header level should be different and maybe renamed to "Policy on panics", since it's not really the `ErrorKind`.

exonum/src/runtime/error.rs
```diff
@@ -26,57 +26,112 @@ use crate::{
...
+/// but whether it makes sense heavily depends on the use case.
+///
+/// ## `Dispatcher` errors
+///
+/// Use `Dispatcher` kind if the error has occurred while dispatching the request (i.e., *not*
+/// in the client code). See [`DispatcherError`] for more details.
+///
+/// ## `Runtime` errors
+///
+/// Use `Runtime` kind if the error has occurred in the runtime code and it makes sense to report
```
popzxc: Maybe note that *recoverable* error occurred?

exonum/src/runtime/error.rs
```diff
@@ -26,57 +26,112 @@ use crate::{
...
-    Panic,
-    /// An error in dispatcher during the execution occurred.
+    /// An unchecked error that has occurred in the service code.
+    ///
+    /// Unlike [`Service`](#variant.Service) errors, unchecked errors do not have a user-defined code.
+    /// Thus, it may be impossible for users to figure out the precise cause of the error;
+    /// they can only use the accompanying error description.
+    Unchecked,
+
+    /// An error in the dispatcher code. For example, the method with the specified ID
```
popzxc: This is not quite correct: i.e. `DispatcherError::UnauthorizedCaller` can occur in the service code and it'll be perfectly fine.

exonum/src/runtime/error.rs
```diff
@@ -26,57 +26,112 @@ use crate::{
...
-        /// Error code, available values ​​can be found in the [description] of the dispatcher's errors.
+        /// Error code. Available values can be found in the [description] of dispatcher errors.
         ///
-        /// [description]: ../dispatcher/error/enum.Error.html
+        /// [description]: ../enum.DispatcherError.html
         code: u8,
     },
-    /// An error in the runtime occurred.
+
+    /// An error in the runtime logic. For example, the runtime could not deploy an artifact.
```
popzxc: Is inability of the artifact deployment really an error in the runtime logic?

exonum/src/runtime/error.rs
```diff
@@ -141,22 +196,22 @@ impl Display for ErrorKind {
...
-///     /// Incorrect transaction payload.
-///     IncorrectPayload = 2,
+///     /// Incorrect deployment specification format.
+///     MalformedDeploySpec = 1,
+///     /// Could not compile the artifact.
+///     CompilationFail = 2,
 /// }
 /// ```
-#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Fail)]
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Fail, BinaryValue)]
```
popzxc: Regarding `BinaryValue`: is  `ExecutionError` really meant to be stored in the database? I thought that description is only needed for logging/reporting, and in the database itself we only store the kind.

slowli:
I'm expanding the scope of this PR to include refactoring of `ExecutionError`; please, don't merge.

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -52,7 +52,7 @@ message InstanceSpec {
 
 // The kind of ExecutionError.
 enum ErrorKind {
-  PANIC = 0;
+  UNCHECKED = 0;
```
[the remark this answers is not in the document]
alekseysidorov: I'd suggest to use `unexpected`.

exonum/src/runtime/error.rs
alekseysidorov: We stores description of errors in database but doesn't use them in the object hash.

popzxc: I understand that, but nevertheless we can (can we?) avoid it to optimize storage usage.

slowli:
The PR is not complete yet, but it has taken shape. Ideally, I would want to completely remove ability to create errors in the service code in other ways than using `CallContext` methods, since it provides maximum forward compatibility and hides implementation details. However, this may be outside the scope of *this* PR.

What's more easy to accomplish and should be done within this PR is adding the info on the call site to *all* error types. Right now, `instance_id` is only present for service errors, but it can (and I feel, should) be added for other error types, because for all of them, it is possible to tell during which call an error has occurred. This info will include instance ID, and, depending on the call type, may include method ID and the interface name, or a constructor / `before_commit` tag. The exception here is `Runtime::deploy_artifact`; right now, the errors this method returns are not handled in any way, so they could theoretically be replaced with `failure::Error` without any problems. I feel a better solution would be to have a separate type for "async" errors, though, or just make the call info optional.

exonum/src/runtime/error.rs
```diff
@@ -522,16 +1159,127 @@ mod tests {
...
+            kind: ErrorKind::Unexpected,
+            description: StringMatch::Any,
+            runtime_id: None,
+            instance_id: None,
+            call_type: None,
+        };
+        assert_eq!(error, matcher);
+
+        // Check various description types.
+        matcher.description = StringMatch::Exact("Panic!".to_owned());
```
alekseysidorov: You can simplify code like this by the `vec![(StringMatch::Exact("Panic!".to_owned()), true)]`

exonum/benches/criterion/block.rs
```diff
@@ -429,31 +430,11 @@ mod foreign_interface_call {
...
-                    let arg: SelfTx = exonum_merkledb::BinaryValue::from_bytes(bytes)
-                        .map_err(DispatcherError::malformed_arguments)?;
-                    self.timestamp(ctx, arg)
-                }
-                _ => Err(DispatcherError::NoSuchMethod).map_err(From::from),
-            }
-        }
-    }
-
     #[derive(Debug)]
```
alekseysidorov: I implemented this interface by hand to avoid creation `impl Transaction<dyn ForeingInterface> for SelfTx`. 
And I think that we should add attributes to `exonum_interface` to skip creation of this implementation for the specified methods or interfaces.

exonum/src/blockchain/mod.rs
```diff
@@ -345,27 +342,34 @@ impl BlockchainMut {
...
+                log::error!(
+                    "{:?} transaction at {:?} resulted in unchecked error: {:?}",
+                    transaction,
+                    height,
+                    err
+                );
+            } else {
+                // Checked transaction errors are a regular occurrence. So logging the
+                // whole transaction body is an overkill: the body can be relatively big.
+                log::info!(
```
alekseysidorov: Why not trace level? I think that a lot of failed transactions can cause node freezes.

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -79,6 +79,31 @@ message ExecutionError {
...
+    CallSite call_site = 6;
+    google.protobuf.Empty no_call_site = 7;
+  }
+}
+
+message CallSite {
+  enum Type {
+    METHOD = 0;
+    CONSTRUCTOR = 1;
+    BEFORE_COMMIT = 2;
```
alekseysidorov: Hm... we should not forget to rename this variant to `AFTER_TRANSACTIONS`.

exonum/src/runtime/dispatcher/error.rs
```diff
@@ -51,23 +53,25 @@ pub enum Error {
...
-    /// Create an `UnauthorizedCaller` error with the specified error message.
-    pub fn unauthorized_caller(msg: impl Display) -> ExecutionError {
-        (Error::UnauthorizedCaller, msg).into()
+    /// Creates a `MalformedArguments` error with the user-provided details.
+    pub fn malformed_arguments(details: impl Display) -> ExecutionError {
+        let description = format!(
+            "Malformed arguments for calling a service interface method: {}",
+            details
+        );
+        Error::MalformedArguments.with_description(description)
```
alekseysidorov: Is this information about context available also by the `call_site` of error?

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -245,15 +255,36 @@ impl Dispatcher {
...
+                    .set_call_site(|| CallSite {
+                        instance_id,
+                        call_type: CallType::BeforeCommit,
+                    });
                 fork.rollback();
+                let height = CoreSchema::new(&*fork).height().next();
+
+                if err.kind() == ErrorKind::Unexpected {
+                    log::error!(
+                        "`before_commit` for service {} at {:?} resulted in unchecked error: {:?}",
```
alekseysidorov: Hm, `unchecked` instead of `unexpected`.

exonum/benches/criterion/block.rs
slowli: Personally, I think we should refine stubs and use them everywhere (see ECR-3910). 

AFAIK, the earlier definition of `Transaction` featured `Service` as an associated type rather than a type param, hence, using the same transaction argument in several interfaces was *impossible*; now, it's just *difficult*.

exonum/src/blockchain/mod.rs
slowli: `info` [is used in `master` right now](https://github.com/exonum/exonum/blob/882e9c867354331ffb2867752ad1d6cc533b602b/exonum/src/blockchain/mod.rs#L373). Would you like to decrease the priority for `before` / `after_transactions` errors as well?

popzxc: IMHO `info` level is fine. Those fellows who don't want node freezes usually set the log level to `warning`/`error`, since normal node reports aren't usually that interesting.

exonum/src/runtime/error.rs
slowli: I think this improvement would increase the amount of code 🙃

PR #1587, opened 2019-12-02
Configurable supervisor [ECR-3829]

popzxc:
This PR replaces generic supervisor mode approach with the configurable one.

Now `Supervisor` accepts `SupervisorConfig` as `initialize` parameters, and also implements `Configure` interface.

`SupervisorConfig` is implemented in such a way that it can be easily extended if needed.


vitvakatu:
Well, it solves the problem of having [too much code](https://github.com/exonum/exonum-java-binding/pull/1249) in the EJB, so I definitely like this whole PR :D

I wish I needed to check on your plans before doing the same thing in EJB.

vitvakatu:
However, starting decentralized supervisor without proper support from node configuration system (cli crate) is rather complicated:
1. Start a node with working simple supervisor service
2. Send configuration proposal to alter the supervisor mode.
3. Wait for some time until all the machinery works and mode is actually altered (and also check for that somehow)

What do you think about adding an optional `supervisor_mode` parameter to the `generate-template` step of the node configuration process in Core? This parameter would be useful for both Core and EJB users, as it allows to alter supervisor mode without recompilation and without complicated configuration proposals.

popzxc:
I think it's completely reasonable, but out of the scope of this PR. Feel free to create a Jira task and assign me to it.

vitvakatu:
Want to hear opinions of @slowli, @alekseysidorov and @aleksuss first, and then I can do it myself (fortunately I already have this functionality implemented).

CHANGELOG.md
```diff
@@ -11,12 +11,21 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+
 ### New features
 
 #### exonum
 
 - `before_transactions` hook for services was introduced. (#1577)
 
+#### exonum-supervisor
+
+- `Supervisor` service now can has initial configuration and implements
```
```suggestion
- `Supervisor` service now can have initial configuration and implements
```

services/supervisor/src/api.rs
```diff
@@ -93,6 +96,15 @@ impl PrivateApi for ApiImpl<'_> {
 
         Ok(configuration_number)
     }
+
+    fn supervisor_config(&self) -> Result<SupervisorConfig, Self::Error> {
+        let config = Schema::new(self.0.service_data())
+            .configuration
+            .get()
+            .expect("Supervisor entity was not configured; unable to load configuration");
```
aleksuss: Why not to return `Err` ? Do we really need to panic here ?

popzxc: Yes. Comment from `lib.rs`:
```
            // Incorrect config is a critical error for both the service and the
            // blockchain itself: supervisor can't operate not configured, and the
            // blockchain isn't expected to work without supervisor.
```

Achieving such a condition means a critical and unrecoverable error.

aleksuss: Ok. But what about user ? :) Why don't answer him about this case ? If user makes REST API request and there is no config... so as I understand he never knows about it because a node will crash, right ? IMO it's not good from user experience side.

popzxc: Well, look: this condition isn't achievable in any imaginable scenario -- we require config to be added within initialization and we validate that it can be parsed. There is no valid way to remove the config that once been set (it's only possible to change it to other valid config). So, if, for some reason, node with supervisor have started successfully, but the config was not set, it means that something gone terribly wrong, and the UX is the least of all possible problems at the moment.

popzxc: In other words, this `except` is in fact completely safe. Scenario when this check will panic is (almost) impossible, and should never happen.

services/supervisor/src/lib.rs
```diff
@@ -71,24 +71,27 @@ pub use self::{
...
 
 use exonum::{
+    blockchain::config::InstanceInitParams,
     crypto::Hash,
     runtime::{
         rust::{
-            api::ServiceApiBuilder, AfterCommitContext, Broadcaster, CallContext, DefaultInstance,
-            Service,
+            api::ServiceApiBuilder, AfterCommitContext, Broadcaster, CallContext, Service,
+            ServiceFactory as ServiceFactoryTrait,
```
slowli: Nit: Shouldn't it be possible to use `ServiceFactory as _`? Or was this introduced after Rust 1.36?

services/supervisor/src/lib.rs
```diff
@@ -213,53 +204,66 @@ fn assign_instance_id(context: &CallContext<'_>) -> InstanceId {
...
-where
-    Mode: mode::SupervisorMode,
-{
+impl Service for Supervisor {
+    fn initialize(&self, context: CallContext<'_>, params: Vec<u8>) -> Result<(), ExecutionError> {
+        use std::borrow::Cow;
+
+        // Load configuration from bytes and store it.
+        let config = SupervisorConfig::from_bytes(Cow::from(&params)).unwrap_or_else(|err| {
+            // Incorrect config is a critical error for both the service and the
```
slowli: I don't think we should think about future-proofing (namely, dynamically added / multiple supervisors in this case), but I want to draw your attention that an error in a builtin service constructor [bubbles up](https://github.com/exonum/exonum/blob/88a782fef9d32676e483033d19acc8f37ecf1c28/exonum/src/blockchain/mod.rs#L243). Thus, panicking seems sort of redundant.

services/supervisor/src/lib.rs
```diff
@@ -285,6 +289,10 @@ where
 
     fn after_transactions(&self, mut context: CallContext<'_>) {
         let mut schema = Schema::new(context.service_data());
+        let configuration = schema
```
slowli: Nit: It could make sense to extract the getter into a schema method.

services/supervisor/tests/supervisor/supervisor_config.rs
```diff
@@ -0,0 +1,173 @@
...
+    // Check for simple mode.
+    let testkit = TestKitBuilder::validator()
+        .with_rust_service(Supervisor)
+        .with_artifact(Supervisor.artifact_id())
+        .with_instance(Supervisor::builtin_instance(Supervisor::simple_config()))
+        .create();
+    assert_supervisor_config(
+        &testkit,
+        SupervisorConfig {
+            mode: mode::Mode::Simple,
```
slowli: Nit: Is module prefixing really warranted here? Looks weird, IMO. Also, shouldn't it be possible to use `Supervisor::*_config()` methods throughout the tests?

services/supervisor/tests/supervisor/main.rs
```diff
@@ -1024,7 +1043,11 @@ fn test_id_assignment_sparse() {
     // Create testkit with builtin instance with ID 100.
     let mut testkit = TestKitBuilder::validator()
         .with_logger()
-        .with_default_rust_service(DecentralizedSupervisor::new())
+        .with_rust_service(Supervisor)
```
slowli: These 4 lines look like an fault of `TestKitBuilder` interface, but I'm not sure how it could be improved. Declaring methods for creating `InstanceInitParams` for a simple and decentralized supervisor - i.e., using something like `.with_instance(Supervisor::decentralized())` - could shorten things a bit.

.travis.yml
```diff
@@ -132,7 +132,8 @@ jobs:
   # Unit tests (including test runs for benchmarks).
   - name: unit-test
     script:
-    - cargo test --all --benches
+    - cargo test --all
+    - cargo test --benches
```
slowli: This will run lib tests again in addition to bench tests. AFAIK, the way to deal with this behavior of `cargo test` is to run `cargo test --all --lib --tests --benches` (i.e., everything except for doc tests) and separately `cargo test --all --doc` (`--doc` cannot be combined with other test categories).

popzxc: Hmm... Maybe then just `cargo bench --lib` instead of `cargo test --benches`? Or it will do something different?

aleksuss: I think `cargo bench --lib` runs benchmarks rather then just tests it.

PR #1590, opened 2019-12-03
Middleware service [ECR-3892]

slowli:
## Overview

This PR introduces the *middleware* service, which exposes some frequent use cases for composability of Exonum transactions:

- Checked calls, which check the version of the called service and don't perform the call if it's not expected
- Batch transactions

Possible future additions in this PR:

- HTTP endpoint to find out an identifier of a service by its version requirement
- "Unicast": transaction addressed to a service by its version requirement (rather than by ID) if there is exactly one matching service deployed ("anycast" and "multicast" can be defined by analogy, but I'm struggling to imagine practical use cases for them)

---

See: https://jira.bf.local/browse/ECR-3892


services/utils/src/transactions.rs
```diff
@@ -0,0 +1,191 @@
...
+        context
+            .call_context(instance_id, ChildAuthorization::Fallthrough)?
+            .call("", arg.inner.call_info.method_id, arg.inner.arguments)
+    }
+
+    fn batch(&self, mut context: CallContext<'_>, arg: Batch) -> Result<(), ExecutionError> {
+        for call in arg.inner {
+            context
+                .call_context(call.call_info.instance_id, ChildAuthorization::Fallthrough)?
+                .call("", call.call_info.method_id, call.arguments)?;
```
popzxc: I guess a TODO should be here to use the interface name once it'll become a part of `AnyTx`.

services/utils/tests/inc/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+#[exonum_interface]
+pub trait IncInterface {
+    fn increment(&self, context: CallContext<'_>, arg: Inc) -> Result<(), ExecutionError>;
+}
+
+#[derive(Clone, Default, Debug, ServiceDispatcher)]
+#[service_dispatcher(implements("IncInterface"))]
+pub struct IncService;
+
+impl Service for IncService {
```
popzxc: It seems that `IncService` is already implemented several times in different places. Maybe it makes sense to create a separate crate for it that can be used as a dev-dependency?

slowli: These versions have slightly different functionality though. For example, the version implemented here retains information about transaction callers, which is needed to check that the fall-through auth works properly.

popzxc: Ah, sure.

slowli:
@popzxc Regarding naming, what about "middleware" or something like that?

popzxc:
> @popzxc Regarding naming, what about "middleware" or something like that?

Totally like it.

components/derive/src/service_factory.rs
```diff
@@ -120,15 +113,20 @@ impl ToTokens for ServiceFactory {
...
         let service_constructor = self.service_constructor();
         let (impl_generics, ty_generics, where_clause) = self.generics.split_for_impl();
 
         let expanded = quote! {
             impl #impl_generics #cr::runtime::rust::ServiceFactory for #name #ty_generics #where_clause {
-                fn artifact_id(&self) -> #cr::runtime::rust::RustArtifactId {
-                    #artifact_id.parse().unwrap()
+                fn artifact_id(&self) -> #cr::runtime::ArtifactId {
+                    #cr::runtime::ArtifactId {
+                        runtime_id: 0,
```
popzxc: Nit: I think it's better to use a dedicated constant here: `#cr::runtime::RuntimeIdentifier::Rust as u32`.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -102,43 +101,37 @@ impl<T: AsReadonly> Schema<T> {
...
-        // Add artifact to pending artifacts queue.
-        self.pending_artifacts().push(spec.clone());
         // Add artifact to registry with pending status.
-        let artifact_name = spec.artifact.name.clone();
         self.artifacts().put(
-            &artifact_name,
+            &artifact,
             ArtifactState {
-                spec,
+                deploy_spec: payload,
```
popzxc: Nit: `payload` argument can be renamed to `deploy_spec`.

exonum/src/runtime/types.rs
```diff
@@ -164,14 +180,30 @@ impl ArtifactId {
...
+}
 
-    /// Checks that the artifact name contains only allowed characters and is not empty.
-    fn is_valid_name(name: impl AsRef<[u8]>) -> bool {
-        // Extended version of `exonum_merkledb::is_valid_name` that also allows '.' and ':'.
-        name.as_ref().iter().all(|&c| match c {
-            b'.' | b':' => true,
-            c => is_allowed_index_name_char(c),
-        })
+impl ProtobufConvert for ArtifactId {
```
alekseysidorov: You can use the convenient protobuf convert attribute `with` to reduce amount of boilerplate code.

exonum/src/runtime/rust/runtime_api.rs
```diff
@@ -14,14 +14,17 @@
 
 //! Rust runtime specific API endpoints.
 
+use semver::Version;
```
popzxc: With upcoming `rust` runtime extraction in mind, it could make sense to provide `semver::Version` as a public re-export from core.

exonum/src/runtime/rust/runtime_api.rs
```diff
@@ -135,8 +145,12 @@ pub fn endpoints(runtime: &RustRuntime) -> impl IntoIterator<Item = (String, Api
         // otherwise it returns source files of Exonum itself.
         .endpoint("proto-sources", {
             move |query: ProtoSourcesQuery| -> Result<Vec<ProtoSourceFile>, api::Error> {
-                if let Some(artifact_id) = query.artifact {
-                    let artifact_id = artifact_id.parse::<RustArtifactId>()?;
+                if let ProtoSourcesQuery::Artifact { name, version } = query {
+                    let artifact_id = ArtifactId {
```
popzxc: Bikeshedding: Maybe provide a `new` method for `ArtifactId` taking `RuntimeIdentifier` as an argument to avoid casts everywhere it's being constructed?

services/middleware/src/transactions.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Appends a call into the batch.
+    pub fn with_call(mut self, call: AnyTx) -> Self {
+        self.inner.push(call);
+        self
+    }
+}
+
+/// Transactional interface of the utilities service.
+#[exonum_interface]
+pub trait UtilsInterface {
```
alekseysidorov: The `UtilsInterface` sounds a little bit obscure. What do you think about `CallUtils` name for this interface?

exonum/src/runtime/rust/tests.rs
```diff
@@ -574,10 +600,129 @@ fn state_aggregation() {
...
+    assert_ne!(new_artifact, artifact);
+    assert!(schema.get_artifact(&new_artifact).is_none());
+    Dispatcher::commit_artifact(&fork, new_artifact.clone(), vec![]).unwrap();
+    commit_block(&mut blockchain, fork);
+
+    // ...and a service based on the new artifact.
+    let mut fork = blockchain.fork();
+    let spec = InstanceSpec {
+        artifact: new_artifact.clone(),
+        id: SERVICE_INSTANCE_ID + 1,
```
popzxc: `SERVICE_INSTANCE_ID + 1` and `instance_id += 1` look a bit fragile and hacky, maybe introduce a separate constant for it?

exonum/src/runtime/rust/tests.rs
```diff
@@ -574,10 +600,129 @@ fn state_aggregation() {
...
+
+    {
+        let idx_name = format!("{}.method_a_entry", SERVICE_INSTANCE_NAME);
+        let entry = fork.get_proof_entry(idx_name.as_str());
+        assert_eq!(entry.get(), Some(11));
+        let entry = fork.get_proof_entry::<_, u64>("new_service.method_a_entry");
+        assert!(!entry.exists());
+    }
+
+    call_info.method_id = 1;
```
popzxc: A constant for method id will be nice to have as well.

services/middleware/src/transactions.rs
```diff
@@ -0,0 +1,192 @@
...
+        let instance_id = arg.inner.call_info.instance_id;
+        let dispatcher_schema = context.data().for_dispatcher();
+        let state = dispatcher_schema
+            .get_instance(instance_id)
+            .ok_or(Error::NoService)?;
+        if state.status != InstanceStatus::Active {
+            return Err(Error::ServiceIsNotActive.into());
+        }
+
+        let artifact = &state.spec.artifact;
```
alekseysidorov: Is this approach assume that the service interfaces completely change with every change in the version of the artifact?

exonum/src/runtime/types.rs
```diff
@@ -111,34 +117,42 @@ impl AnyTx {
...
+/// ```
+///
+/// where:
+///
+/// - `runtime_id` is a [runtime identifier],
+/// - `artifact_name` is the name of the artifact
+/// - `version` is the artifact semantic version
 ///
-/// Artifact name contains only the following characters: `a-zA-Z0-9` and one of `_-.:`.
+/// Artifact name may contain the following characters: `a-zA-Z0-9` and `_-.`.
```
popzxc: Fun fact: we're using regex-like ranges here, but in the second part (`_-.`) is follows the range pattern as well. To be a valid within regex, it should be `_.-`.

exonum/src/runtime/types.rs
```diff
@@ -182,8 +214,8 @@ impl ValidateInput for ArtifactId {
     fn validate(&self) -> Result<(), Self::Error> {
         ensure!(!self.name.is_empty(), "Artifact name should not be empty");
         ensure!(
-            Self::is_valid_name(&self.name),
-            "Artifact name ({}) contains an illegal character, use only: a-zA-Z0-9 and one of _-.:",
+            is_valid_identifier(&self.name),
+            "Artifact name ({}) contains an illegal character, use only: a-zA-Z0-9 and one of _-.",
```
popzxc: Nit: I think "one of" should be removed here too.

services/middleware/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+keywords = ["exonum", "batch", "versioning"]
+categories = ["cryptography"]
+description = "Composability utilities for Exonum, packaged as a service"
+
+[badges]
+travis-ci = { repository = "exonum/exonum" }
+
+[dependencies]
+exonum = { version = "0.13.0-rc.2", path = "../../exonum" }
+exonum-crypto = { version = "0.13.0-rc.2", path = "../../components/crypto" }
```
popzxc: Since `exonum` provides public reexports for `crypto` and `merkledb`, can we avoid having them as dependencies?

services/middleware/README.md
```diff
@@ -0,0 +1,29 @@
...
+allows to batch transactions in order to execute the batch atomically,
+or to check the version of the service before performing a call to it.
+
+## Usage
+
+Include `exonum-middleware-service` as a dependency in your `Cargo.toml`:
+
+```toml
+[dependencies]
+exonum = "0.13"
```
popzxc: If I understand things correctly, release candidates aren't considered a full-weight releases, so the soft version requirement won't work:

```sh
cargo update
    Updating crates.io index
error: failed to select a version for the requirement `exonum = "^0.13"`
  candidate versions found which didn't match: 0.12.1, 0.12.0, 0.11.2, ...
  location searched: crates.io index
required by package `proj v0.1.0 (/Users/user/workspace/temp/proj)`
```

For a release candidate, we **must** include the full version.

services/middleware/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+#[derive(Debug, ServiceDispatcher, ServiceFactory)]
+#[service_dispatcher(implements("UtilsInterface"))]
+#[service_factory(proto_sources = "proto")]
+pub struct MiddlewareService;
+
+impl Service for MiddlewareService {}
+
+impl DefaultInstance for MiddlewareService {
+    const INSTANCE_ID: InstanceId = 1;
+    const INSTANCE_NAME: &'static str = "utils";
```
popzxc: I think the instance name should be updated as well.

services/middleware/src/transactions.rs
```diff
@@ -0,0 +1,192 @@
...
+use semver::VersionReq;
+use serde_derive::*;
+
+use crate::{proto, MiddlewareService};
+
+/// Errors of the `MiddlewareService`.
+#[derive(Debug, Clone, Copy, ExecutionFail)]
+pub enum Error {
+    /// The service instance targeted by the checked call does not exist.
+    NoService = 0,
```
popzxc: Interesting side effect: the direct call to the non-existent service will be considered malformed early and won't be stored, but if it was correctly proxied through `middleware`, then the `CheckedCall` transaction will be persisted.

services/middleware/src/transactions.rs
```diff
@@ -0,0 +1,192 @@
...
+pub struct CheckedCall {
+    /// Expected name of the artifact.
+    pub artifact_name: String,
+    /// Version requirement(s) on the artifact.
+    pub artifact_version: VersionReq,
+    /// The call contents.
+    pub inner: AnyTx,
+}
+
+impl ProtobufConvert for CheckedCall {
```
popzxc: Can we implement `ProtobufConvert` for `VersionReq` and just derive it for `CheckedCall` then?

services/middleware/src/transactions.rs
```diff
@@ -0,0 +1,192 @@
...
+        Ok(Self {
+            artifact_name,
+            artifact_version,
+            inner,
+        })
+    }
+}
+
+#[test]
+fn checked_call_in_json() {
```
popzxc: Nit: I'd suggest to move tests to the end of the file.

services/middleware/src/transactions.rs
```diff
@@ -0,0 +1,192 @@
...
+        if arg.artifact_name != artifact.name {
+            return Err(Error::ArtifactMismatch.into());
+        }
+        if !arg.artifact_version.matches(&artifact.version) {
+            return Err(Error::VersionMismatch.into());
+        }
+
+        context
+            .call_context(instance_id, ChildAuthorization::Fallthrough)?
+            .call("", arg.inner.call_info.method_id, arg.inner.arguments)
```
popzxc: Should we place a TODO here to update this place once interface names will become a part of the `AnyTx`?

services/middleware/Cargo.toml
alekseysidorov: This dependencies are needed in the `exonum-derive` macro.

services/middleware/src/transactions.rs
slowli: Do you mean in the `exonum-proto` crate? Otherwise, the implementation is impossible as per Rust orphaning rules.

services/middleware/Cargo.toml
popzxc: Can you please create an issue for fixing it? IMO it's too indirect and hacky.

services/middleware/src/transactions.rs
popzxc: Indeed. Then I think it doesn't worth it.

alekseysidorov: But we may use this attribute to kill boiler-plate anyway.
https://github.com/witnet/protobuf-convert#overriding-conversion-rules

services/middleware/Cargo.toml
slowli: Huh, it turns out that removing these dependencies makes the build script fail, of all things:

```text
thread 'main' panicked at 'Failed to get exonum crypto protobuf path: NotPresent'
...
  11: exonum_build::get_exonum_protobuf_crypto_files_path
  12: exonum_build::ProtoSources::path
```

@aleksuss Is this a known problem? Should I add a task for this issue?

services/middleware/src/transactions.rs
slowli: Yes, that's a downside of this approach. It could be potentially dealt with by the in-pool transaction verification, but this feature is not in the nearest roadmap.

services/middleware/Cargo.toml
aleksuss: Interesting. Create issue please.

services/middleware/src/transactions.rs
slowli: The caller may specify the version requirement in the form he is comfortable with. For example, he may specify the version requirement as `^1.1.3`, in which case any semver-compatible version will match (e.g., `1.1.5`, `1.3.0` or `1.5.9`). If the caller is brave enough, he may specify a more wide range of versions, e.g., `>=1.1.3, <3`. Or he may be more conservative and specify `~1.1.3` (which means `>=1.1.3, <1.2.0`).

What is expected from the service developers is that artifact versioning reflects the public interfaces of the service (i.e., at least transactions), so that semver requirements make sense.

services/middleware/src/transactions.rs
slowli: I've forgotten to rename it to `MiddlewareInterface` 🤦‍♂

services/middleware/Cargo.toml
vitvakatu: @skletsun looks like familiar problem.

services/middleware/src/proto/service.proto
```diff
@@ -0,0 +1,29 @@
...
+import "runtime.proto";
+
+message CheckedCall {
+  string artifact_name = 1;
+  string artifact_version = 2;
+  exonum.runtime.AnyTx inner = 3;
+}
+
+message Batch {
+  repeated exonum.runtime.AnyTx inner = 1;
```
aleksuss: The `Batch` doesn't contain `artifact_name` and `artifact_version` because we know that it will be composed correct ?

slowli: `Batch` is responsible for batching. If the user wants so, he can wrap the `Batch` into `CheckedCall`, and/or wrap internal calls in the `Batch` in `CheckedCall`s.

services/middleware/Cargo.toml
skletsun: Yes, I faced probably the same issue in EJB - https://jira.bf.local/browse/ECR-4036

PR #1598, opened 2019-12-04
Add supervisor mode parameter to CLI. Refactor exonum-cli crate [ECR-3874]

vitvakatu:
## Overview

Adds new optional `supervisor-mode` to `generate-template` node configuration step.

List of changes:
- New `supervisor-mode` parameter for `generate-template` step.
- `exonum::node::NodeConfig` duplicated in `exonum-cli` and refactored to a new structure. Now it consist of public and private parts.
- Refactored intermediate configs for all configuration steps. Now it is only `NodePublicConfig` (created at `generate-template` step and modified on `generate-config`), `NodePrivateConfig` (create at `generate-config`) and `NodeConfig` (created at `finalize`, combined from `NodePublicConfig` and `NodePrivateConfig`).
- `ConfigManager` structure moved from `exonum::helpers::config` to `exonum-cli`, renamed to `DefaultConfigManager`. Introduced `ConfigManager` trait used by Exonum to update connect list stored on disk.
- Tests in `exonum-cli` are refactored and updated with respect to changes above.

This is the story about how to start adding a single CLI parameter and end up refactoring half of the whole framework.

Q: so much changes! Why?
A: I needed to store `supervisor_node` in the node configuration to be able to use it on the node startup. However, `NodeConfig` was a Core structure and I didn't want to make a dependency on `exonum_supervisor` crate just to bring one enum to the structure. Besides, `NodeConfig` feel being at the wrong place - the format of the node configuration on the disk does not need to match the one in the memory. So the `NodeConfig` was duplicated to the `exonum_cli` and also refactored a lot for better readability. (see next question).

Q: Why split `NodeConfig` to two parts and not using flat structure?
A: Because it was too hard for a user to decide which field in the node configuration file can be changed (without breaking the consensus), and which can't. Such split into `public` and `private` parts improves the situation a lot. Also it seems like I get a nice deduplication by using `NodePublicConfig` in place of old `SharedConfig` and `CommonConfigTemplate`. The latter is tricker, cause I used optional `validator_keys` to generate them on `generate-config` step, not on `generate-template`.

---

See: https://jira.bf.local/browse/ECR-3874

### Definition of Done


cli/src/command/generate_config.rs
```diff
@@ -145,21 +145,23 @@ impl ExonumCommand for GenerateConfig {
...
         };
         // Save public config separately.
-        save_config_file(&shared_config, &public_config_path)?;
+        save_config_file(&public_config, &public_config_path)?;
 
         let private_config = NodePrivateConfig {
             listen_address,
             external_address: self.peer_address.to_string(),
             master_key_path: master_key_path.clone(),
+            api: Default::default(),
```
popzxc: Can `..Default::default()` be used here?

cli/src/command/generate_template.rs
```diff
@@ -34,17 +35,25 @@ pub struct GenerateTemplate {
...
+    ///
+    /// Default: simple.
+    #[structopt(long)]
+    pub supervisor_mode: Option<SupervisorMode>,
 }
 
 impl ExonumCommand for GenerateTemplate {
     fn execute(self) -> Result<StandardResult, Error> {
-        let config_template = CommonConfigTemplate {
+        let supervisor_mode = self.supervisor_mode.unwrap_or(SupervisorMode::Simple);
```
popzxc: I don't like that default value is hidden deep in the sources. It's not as explicit as it was before.

exonum/src/helpers/config.rs
```diff
@@ -12,133 +12,13 @@
...
-    {
-        let mut current_config: NodeConfig = ConfigFile::load(path)?;
-        current_config.connect_list = connect_list;
-        ConfigFile::save(&current_config, path)?;
-
-        Ok(())
-    }
+/// Interface of the Config Manager usable for updating node configuration on
+/// the fly.
+pub trait ConfigManager {
```
popzxc: It's a bit weird that `exonum` has an interface to store config, but doesn't have one to load one. It seems to be an indirect coupling with `exonum-cli`.

cli/src/command/generate_config.rs
vitvakatu: It can't, because `NodePrivateConfig` does not implement `Default`. Besides, I don't want to use it here, as explicit declaration will catch refactoring issues in the future (you will be unable to forget one of the fields).

exonum/src/helpers/config.rs
vitvakatu: :thinking: Does it needs an interface to load config? It shouldn't be possible to do while the node is running anyway. The idea is that Exonum itself does not have facilities to operate with node configuration files. The external crate (exonum-cli) loads configs and converts them to the in-memory representation of Exonum, and also allows to update the connect list on the disk.

popzxc: I don't say it needs it :) I only mean that it looks strange, like an incomplete abstraction. Not sure what will be the best solution here.

vitvakatu: Yeah I kinda agree, just don't know how to solve this.

cli/src/config.rs
```diff
@@ -16,56 +16,118 @@
...
-pub struct SharedConfig {
-    /// Template for common configuration
-    pub common: CommonConfigTemplate,
-    /// Public node
-    pub node: NodePublicConfig,
+    /// Consensus configuration.
+    pub consensus: ConsensusConfig,
+    /// General configuration.
+    pub general: GeneralConfig,
+    /// Keys of the node.
```
popzxc: I think it worth mentioning when this field can be `None` and what will that mean for the node. And maybe explain that "Keys" doesn't mean private keys.

cli/src/config.rs
```diff
@@ -16,56 +16,118 @@
...
-/// `NodePrivateConfig` collects all public and secret keys.
-#[derive(Debug, Serialize, Deserialize)]
+/// Private configuration of the node.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct NodePrivateConfig {
-    /// Listen address.
+    /// Network listening address.
     pub listen_address: SocketAddr,
-    /// External address.
+    /// Remote Network address used by this node.
```
popzxc: Not sure about phrasing. What is "Remote Network address" and how is it exactly "used"?

cli/src/config.rs
```diff
@@ -16,56 +16,118 @@
...
     pub external_address: String,
     /// Path to the master key file.
     pub master_key_path: PathBuf,
+    /// Api configuration.
```
popzxc:
```suggestion
    /// API configuration.
```

cli/src/config.rs
```diff
@@ -16,56 +16,118 @@
...
+    pub api: NodeApiConfig,
+    /// Network configuration.
+    pub network: NetworkConfiguration,
+    /// Memory pool configuration.
+    pub mempool: MemoryPoolConfig,
+    /// Optional database configuration.
+    #[serde(default)]
+    pub database: DbOptions,
+    /// Transaction Verification Thread Pool size.
```
popzxc:
```suggestion
    /// Amount of threads used for transactions verification.
```

cli/src/config.rs
```diff
@@ -16,56 +16,118 @@
...
+    pub thread_pool_size: Option<u8>,
+    /// Node's ConnectList.
```
popzxc:
```suggestion
    /// Information about peers within network.
```

cli/src/lib.rs
```diff
@@ -142,28 +144,34 @@ impl NodeBuilder {
 
         if let StandardResult::Run(run_config) = command.execute()? {
             // Add builtin services to genesis config.
+            let supervisor = match run_config.node_config.public_config.general.supervisor_mode {
```
popzxc: I'd suggest to move `Supervisor` creation into a separate function.

cli/tests/cli.rs
```diff
@@ -20,16 +20,26 @@ extern crate pretty_assertions;
...
-    node::{ConnectInfo, ConnectListConfig, NodeConfig},
+    node::{ConnectInfo, ConnectListConfig},
 };
 use exonum_cli::command::{Command, ExonumCommand, StandardResult};
 use serde_derive::*;
 use structopt::StructOpt;
 
+use exonum::blockchain::ValidatorKeys;
+use exonum::crypto::gen_keypair;
+use exonum_cli::command::finalize::Finalize;
```
popzxc: We use nested imports usually.

services/supervisor/src/lib.rs.orig
```diff
@@ -0,0 +1,412 @@
+// Copyright 2019 The Exonum Team
```
popzxc: Merge artifact.

services/supervisor/src/mode.rs
```diff
@@ -103,3 +105,51 @@ impl Mode {
         }
     }
 }
+
+impl Default for Mode {
```
popzxc: I don't like it. `Simple` mode is not default by its nature, it's default within `exonum-cli` and nothing more than it. On the `Supervisor` level those modes do have equal rights.

services/supervisor/src/mode.rs
```diff
@@ -29,9 +29,11 @@ use exonum_merkledb::access::Access;
 use exonum_proto::ProtobufConvert;
 
 use super::{multisig::MultisigIndex, proto, DeployRequest};
+use failure::{self, format_err};
```
popzxc: We usually (try to) use the "external crates" -> "std" -> "local crate" imports ordering.

cli/tests/testdata/config/more_validators/cfg/0/node.toml
```diff
@@ -1,65 +1,71 @@
...
+[[public_config.consensus.validator_keys]]
+consensus_key = "531cd1323febe624a34cda4efa68d9bb78c9e59e0fcb9b5702a8f94dbf7080e1"
+service_key = "afdb7d47ea28bdb28cb8384d652bb241fe367b349f3158b7c2f89df8d1c31cc7"
+
+[[public_config.consensus.validator_keys]]
+consensus_key = "6f3f614692d678b664be7eaafbe280d14e2d4191299ae146874eec389c93ece7"
+service_key = "7662a94afdaac0b24dc239fd151ed5f9d9a384f5f240540c1e4f382d2b5a526d"
+
+[public_config.general]
+supervisor_mode = "Simple"
```
popzxc: Is the first letter capitalized on purpose?

cli/src/command/finalize.rs
```diff
@@ -63,33 +63,36 @@ pub struct Finalize {
...
-    fn validate_configs(public_configs: Vec<SharedConfig>) -> Result<ValidatedConfigs, Error> {
+    fn validate_configs(public_configs: Vec<NodePublicConfig>) -> Result<ValidatedConfigs, Error> {
         let mut map = BTreeMap::new();
         let mut config_iter = public_configs.into_iter();
         let first = config_iter
             .next()
             .ok_or_else(|| format_err!("Expected at least one config in PUBLIC_CONFIGS"))?;
-        let common = first.common;
-        map.insert(first.node.validator_keys.consensus_key, first.node);
+        map.insert(first.validator_keys.unwrap().consensus_key, first.clone());
```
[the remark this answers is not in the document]
vitvakatu: Not really, it could only mean manual configs editing. Added proper error message though

cli/src/config.rs
vitvakatu: 🤷‍♂ I just copied docstrings from NodeConfig in core. I have no idea how to describe this field in a proper way. @alekseysidorov ?

vitvakatu:
Thank you for the review, @popzxc

cli/src/command/generate_template.rs
```diff
@@ -22,10 +23,13 @@ use std::path::PathBuf;
 
 use crate::{
     command::{ExonumCommand, StandardResult},
-    config::{CommonConfigTemplate, GeneralConfig},
+    config::{GeneralConfig, NodePublicConfig},
     io::save_config_file,
 };
 
+/// Default supervisor service mode
+pub const DEFAULT_SUPERVISOR_MODE: SupervisorMode = SupervisorMode::Simple;
```
aleksuss: Why don't implement `Default` trait for `SupervisorMode` ?

CHANGELOG.md
```diff
@@ -47,6 +47,11 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Added hashed version of `Entry` called `ProofEntry`, which participates
   in the state aggregation. (#1553)
 
+#### exonum-cli
```
popzxc: Since some public interfaces were affected by this feature (namely, `Node::with_blockchain`), I think it should be mentioned in the `Breaking changes` section.

cli/src/command/generate_template.rs
```diff
@@ -34,17 +38,25 @@ pub struct GenerateTemplate {
     /// Number of validators in the network.
     #[structopt(long)]
     pub validators_count: u32,
+    /// Supervisor service mode. Possible options are "simple" and "decentralized".
+    ///
+    /// Default: simple.
+    #[structopt(long)]
+    pub supervisor_mode: Option<SupervisorMode>,
```
aleksuss: Is `Option` really needed here ?

cli/src/command/generate_template.rs
popzxc: It was implemented, and I requested it to be removed. I'll quote myself:

>I don't like it _(remark: it == `impl Default for Mode`)_. Simple mode is not default by its nature, it's default within exonum-cli and nothing more than it. On the Supervisor level those modes do have equal rights.

cli/src/command/generate_template.rs
vitvakatu: It is an optional parameter. We can make it required though, I do not have any arguments for either way

popzxc: I think making it required makes more sense. Overall, we can't assume what kind of supervisor user wants, except for the `run-dev` mode.

popzxc:
@slowli can you please give it a look?

cli/tests/cli.rs
```diff
@@ -531,3 +526,140 @@ fn test_clear_cache() {
...
+
+fn run_node_with_supervisor(supervisor_mode: SupervisorMode) -> Result<(), failure::Error> {
+    let testnet_dir = tempfile::tempdir()?;
+
+    let common_config_path = concat_path(testnet_dir.path(), "common.toml");
+
+    let generate_template = GenerateTemplate {
+        common_config: common_config_path.clone(),
+        validators_count: 1,
+        supervisor_mode: supervisor_mode,
```
aleksuss:
```suggestion
        supervisor_mode,
```

cli/src/command/finalize.rs
```diff
@@ -63,49 +64,73 @@ pub struct Finalize {
...
+
     fn create_connect_list_config(
         public_configs: &[NodePublicConfig],
         secret_config: &NodePrivateConfig,
     ) -> ConnectListConfig {
         let peers = public_configs
             .iter()
             .filter(|config| {
-                config.validator_keys.consensus_key != secret_config.keys.consensus_pk()
+                config.validator_keys.unwrap().consensus_key != secret_config.keys.consensus_pk()
```
aleksuss: Does it make sense to replace `unwrap()` with `except("Couldn't get validators keys")` ?

cli/src/command/finalize.rs
```diff
@@ -63,49 +64,73 @@ pub struct Finalize {
...
-        let common = first.common;
-        map.insert(first.node.validator_keys.consensus_key, first.node);
+            .ok_or_else(|| format_err!("Expected at least one config in <public-configs>"))?;
+        let consensus_key = Self::get_consensus_key(&first)?;
+        public_configs.insert(consensus_key, first.clone());
 
         for config in config_iter {
-            if common != config.common {
-                bail!("Found config with different common part.");
+            if first.consensus != config.consensus {
```
alekseysidorov: Why you do not use `ensure` macro here?

cli/src/config.rs
alekseysidorov: I also have no ideas for the better naming.

cli/src/config_manager.rs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        let mut current_config: NodeConfig = load_config_file(path)?;
+        current_config.private_config.connect_list = connect_list;
+        save_config_file(&current_config, path)?;
+
+        Ok(())
+    }
+}
+
+impl ConfigManager for DefaultConfigManager {
```
alekseysidorov: Are you sure that there are not ready solutions for this task in the crates.io?

cli/src/config.rs
```diff
@@ -16,56 +16,122 @@
...
     /// Validator keys.
     #[serde(skip)]
     pub keys: Keys,
 }
+
+/// Configuration for the `Node`.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct NodeConfig {
+    /// Private configuration of the node.
+    pub private_config: NodePrivateConfig,
```
slowli: If I understand correctly, private / public config correspond to what is called local / global config in the [Exonum docs](https://exonum.com/doc/version/latest/architecture/configuration/). If so, we should use single set of terms everywhere.

cli/src/config_manager.rs
```diff
@@ -0,0 +1,84 @@
...
+pub struct UpdateRequest(ConnectListConfig);
+
+impl DefaultConfigManager {
+    /// Creates a new `ConfigManager` instance for the given path.
+    pub fn new<P>(path: P) -> Self
+    where
+        P: AsRef<Path> + Send + 'static,
+    {
+        let (tx, rx) = mpsc::channel();
+        let handle = thread::spawn(move || {
```
slowli: Does `handle` do anything? In the old implementation, it was `join`ed on stop, but now this code is seemingly gone. Unlike some resources, threads are not "released" if the handle is dropped.

services/supervisor/src/mode.rs
```diff
@@ -27,11 +27,15 @@ use exonum::helpers::byzantine_quorum;
...
 use exonum_proto::ProtobufConvert;
+use failure::{self, format_err};
+
+use std::str::FromStr;
 
 use super::{multisig::MultisigIndex, proto, DeployRequest};
 
 /// Supervisor operating mode.
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
```
slowli: Nit / comment: Since `Copy` (non)implementation significantly influences the public interface of a type, we should be careful about implementing `Copy` for public types that may be extended in the future (of which `Mode` is a good example).

exonum/src/node/mod.rs
```diff
@@ -514,11 +482,6 @@ impl NodeHandler {
         let is_enabled = api_state.is_enabled();
         api_state.set_node_role(node_role);
 
-        let config_manager = match config_file_path {
```
slowli: Nit / slightly off-topic: Could you please fix the usage of `NodeHandler.config_manager` in `node/events.rs`?

```rust
if self.config_manager.is_some() {
    // ...
    self.config_manager.as_ref().unwrap().store_connect_list(/* ... */)
}
```

can be replaced with `if let` construction.

exonum/src/helpers/config.rs
```diff
@@ -12,133 +12,13 @@
...
-        current_config.connect_list = connect_list;
-        ConfigFile::save(&current_config, path)?;
-
-        Ok(())
-    }
+/// Interface of the Config Manager usable for updating node configuration on
+/// the fly.
+pub trait ConfigManager: Send {
+    /// Update connect list in the node configuration.
+    fn store_connect_list(&self, connect_list: ConnectListConfig);
```
slowli: Should the method take `&self` or `&mut self`? The second option is more permissive for implementors (e.g., the list could be stored directly in the manager without needing interior mutability), and, as far as I see, it is allowed by the `NodeHandler` code.

cli/src/config_manager.rs
```diff
@@ -0,0 +1,83 @@
...
+pub struct UpdateRequest(ConnectListConfig);
+
+impl DefaultConfigManager {
+    /// Creates a new `ConfigManager` instance for the given path.
+    pub fn new<P>(path: P) -> Self
+    where
+        P: AsRef<Path> + Send + 'static,
+    {
+        let (tx, rx) = mpsc::channel();
+        thread::spawn(move || {
```
popzxc: Is this correct that we don't store the `handle` and not join the thread?

cli/src/lib.rs
```diff
@@ -90,19 +90,21 @@
...
 
 use std::sync::Arc;
 
-use crate::command::{Command, ExonumCommand, StandardResult};
+use crate::command::{run::NodeRunConfig, Command, ExonumCommand, StandardResult};
+use crate::config_manager::DefaultConfigManager;
 
 pub mod command;
 pub mod config;
+pub mod config_manager;
```
popzxc: Does it have to be public?

cli/src/command/finalize.rs
```diff
@@ -116,7 +139,7 @@ impl Finalize {
 impl ExonumCommand for Finalize {
     fn execute(self) -> Result<StandardResult, Error> {
         let secret_config: NodePrivateConfig = load_config_file(&self.secret_config_path)?;
```
aleksuss: I'd rename it to `private_config`.

cli/src/config_manager.rs
popzxc: @vitvakatu ping

vitvakatu: Separate task https://jira.bf.local/browse/ECR-4047 is created to address this.

popzxc: Ok.

cli/src/command/run_dev.rs
```diff
@@ -82,7 +82,7 @@ impl ExonumCommand for RunDev {
         let node_config_file_name = "node.toml";
 
         let finalize = Finalize {
-            secret_config_path: self.artifact_path(SEC_CONFIG_FILE_NAME),
+            private_config_path: self.artifact_path(SEC_CONFIG_FILE_NAME),
```
slowli: Nit: The `SEC_CONFIG_FILE_NAME` name still sort of implies "secret" instead of "private."

cli/tests/cli.rs
```diff
@@ -555,17 +528,17 @@ fn different_supervisor_modes_in_public_configs() -> Result<(), failure::Error>
     };
 
     let testnet_dir = tempfile::tempdir()?;
-    let pub_config_1_path = concat_path(testnet_dir.path(), "pub1.toml");
-    let pub_config_2_path = concat_path(testnet_dir.path(), "pub2.toml");
-    let sec_config_path = concat_path(testnet_dir.path(), "sec.toml");
+    let pub_config_1_path = testnet_dir.path().join("pub1.toml");
+    let pub_config_2_path = testnet_dir.path().join("pub2.toml");
+    let sec_config_path = testnet_dir.path().join("sec.toml");
```
slowli: Nit: `sec_config_path` should be renamed to `private_config_path`.

cli/tests/cli.rs
```diff
@@ -647,7 +620,7 @@ fn run_node_with_supervisor(supervisor_mode: SupervisorMode) -> Result<(), failu
 
     if let StandardResult::Run(config) = run.execute()? {
         assert_eq!(
-            config.node_config.public_config.general.supervisor_mode,
+            &config.node_config.public_config.general.supervisor_mode,
```
slowli: Micro-nit: IMO, `assert_eq!(a, *b)` looks more idiomatic than `assert_eq!(&a, b)`.

PR #1605, opened 2019-12-06
Allow to stop service instances [ECR-3762]

alekseysidorov:
Definition of done.

- [x] Stopped services do not participate in the dispatcher work.
- [x] Unit tests that check that there is no possibility to interact with the stopped services and read their information schemas.
- [x] Support in the runtime and in the dispatcher to stop service instances with the corresponding tests.
- [x] Support of stop service workflow in the supervisor with the corresponding tests. 
- [x] Update changelog entry


exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -126,21 +126,30 @@ message ExecutionStatus {
...
+  ArtifactStatus status = 2;
 }
 
+// Current state of service instance in dispatcher.
 message InstanceState {
-    exonum.runtime.InstanceSpec spec = 1;
-    InstanceStatus status = 2;
+  // Status of a service instance.
+  enum Status {
+    // The service instance has no status.
```
popzxc: What does "has no status" mean?

exonum/src/runtime/types.rs
```diff
@@ -483,12 +519,39 @@ pub struct InstanceState {
...
+        self.status.is_active() || self.next_status.is_active()
+    }
+
+    /// Sets next status as current and changes next status to `None`
+    ///
+    /// # Panics
+    ///
+    /// - If next status is already `None`.
+    pub(crate) fn commit_next_status(&mut self) {
+        assert_ne!(
```
popzxc: To determine `None`-status (if we really need it), I'd suggest to use `Option`. And should we really panic here, isn't it safe to just return an error?

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -401,57 +452,64 @@ impl Dispatcher {
...
+                if status.is_active() {
+                    Some(descriptor)
+                } else {
+                    None
+                }
+            })
     }
 
-    /// Notify the runtimes that it has to shutdown.
+    /// Notifies the runtimes that it has to shutdown.
```
slowli: Nit / off-topic: Poor grammar; "...that they have to shut down" looks correct.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -401,57 +452,64 @@ impl Dispatcher {
...
         instance: &InstanceSpec,
+        status: InstanceStatus,
     ) -> Result<(), ExecutionError> {
         // Notify the runtime that the service has been committed.
         let runtime = self
             .runtimes
             .get_mut(&instance.artifact.runtime_id)
             .ok_or(Error::IncorrectRuntime)?;
-        runtime.commit_service(snapshot, instance)?;
+        // TODO Rewrite runtime API to pass status to runtime.
```
slowli: Will this TODO be solved within this PR? I imaging the runtime should be signaled when its service is shut down.

exonum/src/runtime/types.rs
```diff
@@ -411,42 +411,78 @@ impl ProtobufConvert for ArtifactStatus {
...
+        (*self as u32).to_bytes()
+    }
+
+    fn from_bytes(bytes: Cow<'_, [u8]>) -> Result<Self, failure::Error> {
+        let code = u32::from_bytes(bytes)?;
+        match code {
+            0 => Ok(InstanceStatus::None),
+            1 => Ok(InstanceStatus::Active),
+            2 => Ok(InstanceStatus::Stopped),
+            other => Err(format_err!("Instance status {} is unknown.", other)),
```
slowli: Nit: `Instance status with code {} ...` looks more interpretable, IMO.

exonum/src/runtime/types.rs
```diff
@@ -483,12 +519,39 @@ pub struct InstanceState {
     pub spec: InstanceSpec,
     /// Service instance activity status.
     pub status: InstanceStatus,
+    /// Next status of instance if the value is not `None`.
+    pub next_status: InstanceStatus,
```
slowli: Nit: Could make sense to name this field `pending_status`.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -44,13 +44,68 @@ mod tests;
...
     runtime_id: u32,
     name: String,
+    status: InstanceStatus,
+}
+
+/// Lookup table for the committed service instances.
+#[derive(Debug, Default)]
+struct CommittedServices {
+    instances: BTreeMap<InstanceId, ServiceInfo>,
+    instance_names: BTreeMap<String, InstanceId>,
```
slowli: Are keys of `instance_names` need to be ordered? Otherwise, I'd use `HashMap` (AFAIK, it's faster).

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -87,12 +142,10 @@ impl Dispatcher {
...
         for state in schema.instances().values() {
-            debug_assert_eq!(
-                state.status,
-                InstanceStatus::Active,
-                "BUG: Service instance should not be in pending state."
-            );
-            self.start_service(snapshot, &state.spec)?;
+            let status = state
+                .status
+                .expect("BUG: Service instance should has state.");
```
popzxc:
```suggestion
                .expect("BUG: Stored service instance should have a determined state.");
```

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -373,57 +421,61 @@ impl Dispatcher {
...
-            InstanceQuery::Name(name) => {
-                // TODO: This may be slow.
-                let id = Schema::new(fork).instances().get(name)?.spec.id;
-                Some(InstanceDescriptor { id, name })
-            }
-        }
+        self.service_infos
+            .get_instance(id)
+            .and_then(|(descriptor, status)| {
+                if status.is_active() {
```
popzxc: This pattern has several occurrences within code. Maybe it's possible to move it into some method, like `descriptor.from_status(status)`?

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -67,9 +67,9 @@ impl<T: Access> Schema<T> {
     }
 
     /// Returns a pending instances queue used to notify the runtime about service instances
-    /// to be committed.
-    fn pending_instances(&self) -> ListIndex<T::Base, InstanceSpec> {
-        self.access.clone().get_list(PENDING_INSTANCES)
+    /// to be committed with a new status.
```
popzxc: I'd rephrase it like "about service instances info to be updated".

exonum/src/runtime/mod.rs
```diff
@@ -315,7 +315,7 @@ pub trait Runtime: Send + fmt::Debug + 'static {
         parameters: Vec<u8>,
     ) -> Result<(), ExecutionError>;
 
-    /// Permanently adds a service to the runtime.
+    /// Permanently adds a service to the runtime. TODO [ECR-3762]
```
popzxc: This comment seems to be incorrect.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -149,20 +149,42 @@ impl Schema<&Fork> {
...
     }
 
+    /// Adds information about stopping service instance to the schema.
+    pub(crate) fn initiate_stopping_service(&mut self, instance_id: InstanceId) {
+        let mut instances = self.instances();
+
+        let instance_name = self
+            .instance_ids()
+            .get(&instance_id)
+            .expect("BUG: attempt to stop a nonexistent service instance.");
```
popzxc: I'd mention in the description that this method assumes the input data to be correct.

alekseysidorov: Do you think that `BUG` prefix is not enough?

popzxc: I think it's better to know that function panics before it actually panics :)

popzxc: For example, such a note will be useful for implementers of custom `Supervisor` service.

services/supervisor/tests/supervisor/service_lifecycle.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    let instance_id = start_inc_service(&mut testkit);
+    // Stop service instance.
+    execute_transaction(
+        &mut testkit,
+        ConfigPropose::immediate(1)
+            .stop_service(instance_id)
+            .sign_for_supervisor(keypair.0, &keypair.1),
+    )
+    .unwrap()
```
popzxc: Please change `unwrap`/`unwrap_err` with `expect`/`expect_err` for better error reporting in places that are actually being tested.

exonum/src/runtime/dispatcher/schema.rs
alekseysidorov: I'll write "panic" clause in the method documentation.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -44,13 +47,68 @@ mod tests;
...
+}
+
+impl CommittedServices {
+    fn insert(&mut self, id: InstanceId, info: ServiceInfo) {
+        let name = info.name.clone();
+        self.instances.insert(id, info);
+        self.instance_names.insert(name, id);
+    }
+
+    fn get_runtime_for_active_instance(&self, id: InstanceId) -> Option<u32> {
```
aleksuss: To be more clear, I'd propose to rename it to `get_runtime_id_for_active_instance` ?

exonum/src/runtime/mod.rs
```diff
@@ -307,32 +313,42 @@ pub trait Runtime: Send + fmt::Debug + 'static {
...
+    /// This method notifies runtime about service instance state changes in the dispatcher so
+    /// that the runtimes perform corresponding actions in according to changes in the service
+    /// instance state.
+    ///
+    /// Method is called for a specific service instance during the `Runtime` lifetime in the
+    /// following cases:
     ///
-    /// This method is called *once* for a specific service instance during `Runtime` lifetime:
+    /// - For newly added instances, or modified existing the method is called when the fork
+    /// with the corresponding changes is committed.
```
slowli: Nit: Lists with indented items look better, IMO:

```rust
/// - Lorem ipsum
///   foo bar
```

exonum/src/runtime/mod.rs
```diff
@@ -341,18 +357,25 @@ pub trait Runtime: Send + fmt::Debug + 'static {
...
+    /// (The latter kind of errors should be produced during the preceding
+    /// `initiate_adding_service` call.) The error should contain a description allowing the node
+    /// administrator to determine the root cause of the error and (ideally) recover the node by
+    /// eliminating it.
+    fn commit_service_status(
         &mut self,
         snapshot: &dyn Snapshot,
         spec: &InstanceSpec,
+        status: InstanceStatus,
     ) -> Result<(), ExecutionError>;
```
slowli: Should we continue to return an error here, or ~pull the plug~ just advise to panic? Tbh, I struggle to find a real use case for returning an error.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -149,20 +149,42 @@ impl Schema<&Fork> {
...
         );
-        instance_ids.put(&id, name);
-        self.pending_instances().push(spec);
+        self.pending_instances().put(&instance_name, pending_status);
+        instance_ids.put(&instance_id, instance_name);
         Ok(())
     }
 
+    /// Adds information about stopping service instance to the schema.
+    pub(crate) fn initiate_stopping_service(&mut self, instance_id: InstanceId) {
```
slowli: To mirror @popzxc's concerns, it looks sort of inconsistent that `initiate_adding_service` method above returns a `Result` while this one panics. Further, the 2 panics in this method are different: the first one can be triggered by the incorrect input, while the second one is something that should never-ever occur.

exonum/src/runtime/mod.rs
```diff
@@ -649,10 +672,19 @@ impl<'a> ExecutionContext<'a> {
...
+            .initiate_adding_service(spec)
             .map_err(From::from)
     }
 
+    /// Initiates stopping of an existing service instance in the blockchain. The stopping
+    /// service is active (i.e., does process transactions or `after_transactions` hook)
+    /// until the block built on top of the provided `fork` is committed.
+    ///
+    /// This method should be called for the exact context passed to the runtime.
+    pub(crate) fn initiate_stopping_service(&mut self, instance_id: InstanceId) {
```
slowli: Not sure I understand why this is placed in `ExecutionContext` or takes `self` by mutable ref. (Cf. `CallContext::start_artifact_registration`, which calls to a static method in `Dispatcher`.) For symmetry with `initiate_adding_service`? Since the method is private, I don't think future compatibility should apply.

exonum/src/runtime/rust/mod.rs
```diff
@@ -489,13 +493,22 @@ impl Runtime for RustRuntime {
...
+            InstanceStatus::Active => {
+                let instance = self.new_service(spec)?;
+                self.add_started_service(instance);
+            }
+
+            InstanceStatus::Stopped => {
+                self.remove_started_service(spec);
+            }
+        }
         self.new_services_since_last_block = true;
```
slowli: Nit: This variable should probably be renamed for clarity to something like `service_changed_since_last_block`. I initially thought that you've forgotten to switch off HTTP API of stopped services because of it 🙃

services/supervisor/src/transactions.rs
```diff
@@ -92,26 +92,43 @@ impl StartService {
...
+}
+
+impl StopService {
+    fn validate(&self, context: &CallContext<'_>) -> Result<(), ExecutionError> {
+        let instance = context
+            .data()
+            .for_dispatcher()
+            .get_instance(self.instance_id)
+            .ok_or_else(|| {
+                Error::MalformedConfigPropose
```
slowli: Nit: I wonder if it could make sense to split `MalformedConfigPropose` to more fine-grained variants. Could be done separately.

services/supervisor/tests/supervisor/service_lifecycle.rs
slowli: Nit: Missing semicolon. Also, it would be nice to check that the service HTTP API is switched off.

exonum/src/runtime/types.rs
```diff
@@ -447,13 +482,49 @@ pub struct InstanceState {
...
-        Self { spec, status }
+        Self {
+            spec,
+            status: Some(status),
+            pending_status: None,
+        }
+    }
+
+    /// Indicates whether the service instance current or pending status is active.
+    pub fn is_active(&self) -> bool {
```
slowli: I'm not sure that this is a good name; IMO, it sort of implies that the service is active now. I'd suggest to have a method returning status accounting for possible pending status, i.e.,

```rust
self.pending_status.or(self.status)
```

...and then act on this status depending on the caller.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -188,11 +244,15 @@ impl Dispatcher {
         // Currently the only check is that destination service exists, but later
         // functionality of this method can be extended.
         let call_info = &tx.as_ref().call_info;
-        Schema::new(snapshot)
+        let instance = Schema::new(snapshot)
             .get_instance(call_info.instance_id)
             .ok_or(Error::IncorrectInstanceId)?;
 
-        Ok(())
+        if instance.is_active() {
```
slowli: Are you sure this is the correct filter condition? If I understand correctly, this is *formally* correct (`snapshot` with which `check_tx` is called cannot have pending service statuses set), but logically processing transactions with a pending transition to active state is a mistake.

exonum/src/runtime/types.rs
```diff
@@ -367,59 +367,94 @@ impl Display for ArtifactStatus {
...
 
+impl BinaryValue for InstanceStatus {
+    fn to_bytes(&self) -> Vec<u8> {
+        (*self as u32).to_bytes()
+    }
+
+    fn from_bytes(bytes: Cow<'_, [u8]>) -> Result<Self, failure::Error> {
+        let code = u32::from_bytes(bytes)?;
+        match code {
+            1 => Ok(InstanceStatus::Active),
```
slowli: Nit: Isn't it possible to use `InstanceState_Status` constants here? Also, what's stopping from using `from_pb` / `to_pb` methods in `BinaryValue` implementation? enum variants seem to be encoded with the LEB128 encoding, so the current `BinaryValue` impl is not consistent with the Protobuf presentation.

services/supervisor/src/transactions.rs
alekseysidorov: ECR-3984

exonum/src/runtime/types.rs
alekseysidorov: It is possible but painful since `InstanceState_Status` has no `repr(i32)` attribute.
Also this type doesn't implement `rust_protobuf::Message` trait, so I can't just use `ProtobufConvert`.

I think that the best solution is extending `BinaryValue` macros to support enums with discriminant.
ECR-3941

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -149,17 +149,54 @@ impl Schema<&Fork> {
             return Err(Error::ServiceIdExists);
         }
 
-        let id = spec.id;
-        let name = spec.name.clone();
+        let instance_id = spec.id;
+        let instance_name = spec.name.clone();
+        let pending_status = InstanceStatus::Active;
```
popzxc: Micro-nit: You can write here `Some(InstanceStatus::Active)` to have a bit clearer constructor call below.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -149,17 +149,54 @@ impl Schema<&Fork> {
...
+    ) -> Result<(), Error> {
+        let mut instances = self.instances();
+        let mut pending_instances = self.pending_instances();
+
+        let instance_name = self
+            .instance_ids()
+            .get(&instance_id)
+            .ok_or(Error::IncorrectInstanceId)?;
+
+        let mut state = instances
```
popzxc: Micro-nit: If `state` will have methods like `into_active`/`into_stopped` which will return a new object with changed field, we will be able to get rid of the mutable variable here, which IMO will be more `rust`-ish.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -149,17 +149,54 @@ impl Schema<&Fork> {
...
+
+        if state.pending_status.is_some() {
+            return Err(Error::ServicePending);
+        }
+
+        // Modify instance status.
+        let pending_status = InstanceStatus::Stopped;
+        state.pending_status = Some(pending_status);
+        pending_instances.put(&instance_name, pending_status);
+        instances.put(&instance_name, state);
```
popzxc: Hmm... It looks a bit odd that we add `Stopped` service as pending (which means that this action is to be done yet), and update the instance state with the same value (which means that service is already stopped).
Wouldn't it cause the weird behaviour when service is not stopped yet but considered so by the core?

exonum/src/runtime/dispatcher/tests.rs
```diff
@@ -144,22 +144,25 @@ struct SampleRuntime {
     runtime_type: u32,
     instance_id: InstanceId,
     method_id: MethodId,
-    new_services: Vec<InstanceId>,
-    new_service_sender: Sender<(u32, Vec<InstanceId>)>,
+    services: BTreeMap<InstanceId, InstanceStatus>,
+    // Make services order predictable.
```
popzxc:
```suggestion
    // `BTreeMap` is used to make services order predictable.
```

exonum/src/runtime/mod.rs
```diff
@@ -203,10 +209,10 @@ impl From<RuntimeIdentifier> for u32 {
 ///
 /// ```text
 /// LIFE ::= initialize (GENESIS | RESUME) BLOCK* shutdown
-/// GENESIS ::= (deploy_artifact | start_adding_service commit_service)* after_commit
+/// GENESIS ::= (deploy_artifact | initiate_adding_service commit_service)* after_commit
 /// RESUME ::= (deploy_artifact | commit_service)* on_resume
 /// BLOCK* ::= PROPOSAL+ COMMIT
-/// PROPOSAL ::= before_transactions* (execute | start_adding_service)* after_transactions*
+/// PROPOSAL ::= before_transactions* (execute | initiate_adding_service)* after_transactions*
```
popzxc: Shouldn't `update_service_status` be mentioned here as well?

exonum/src/runtime/mod.rs
popzxc: In theory, looking far-far-far away to the future, I can imagine splitting "stopped" service state into two different states: "maintenance", used for e.g. migration to the new version, and "suspended", used e.g. for emulation of "trading hours" for a stock service. In that sense, there will be some "unsuspendable" services like `Supervisor`.

But currently I don't see a need in returning an error here as well.

services/supervisor/README.md
```diff
@@ -6,7 +6,7 @@
 
 Exonum-supervisor is a main service of the [Exonum blockchain framework](https://exonum.com/).
 
-It is capable of deploying and starting new services.
+It is capable of deploying and starting new services and stopping existing ones.
```
popzxc:
```suggestion
It is capable of deploying / starting new services and stopping existing ones.
```

services/supervisor/src/transactions.rs
```diff
@@ -92,32 +95,48 @@ impl StartService {
...
+                    .with_description("Instance with the specified ID is absent.")
+            })?;
+
+        match instance.status {
+            Some(InstanceStatus::Active) => Ok(()),
+            _ => Err(Error::MalformedConfigPropose.with_description(format!(
+                "Discarded stop of the already stopped instance: {}",
```
popzxc: Maybe "Discarded an attempt to stop the already stopped service instance"?

CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - State hash aggregation is now performed automatically by MerkleDB.
   The relevant methods in `Runtime` and `Service` in Rust runtime
   have been removed. (#1553)
+- `commit_service` was renamed to the `update_service_status` and now takes
```
```suggestion
- `commit_service` has been renamed to the `update_service_status` and now takes
```

services/supervisor/src/transactions.rs
```diff
@@ -386,13 +405,21 @@ impl Supervisor {
...
-                        return Err(Error::MalformedConfigPropose.into());
+                        return Err(Error::MalformedConfigPropose.with_description(
+                            "Discarded multiple instances with the same name in one request.",
+                        ));
                     }
                     start_service.validate(&context)?;
                 }
+
+                ConfigChange::StopService(stop_service) => {
+                    if !modified_instances.insert(stop_service.instance_id) {
```
popzxc: What if `StartService` and `StopService` for the same instance will appear in one config proposal?

CHANGELOG.md
```diff
@@ -37,6 +41,14 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `before_transactions` hook for services was introduced. (#1577)
 - `ErrorMatch` was introduced to test (e.g., using the testkit) that
   an `ExecutionError` has an expected type, error message and/or location. (#1585)
+- Service instances can now be stopped.
+
+  Active service instance can be stopped by the corresponding request by the
```
aleksuss: Maybe `to the Supervisor` rather then `by the...` ?

CHANGELOG.md
```diff
@@ -51,6 +63,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `Supervisor` service now can have initial configuration and implements
   `Configure` interface. (#1587)
+- Added `ConfigChange::StopService` to make requests to stop the service
```
```suggestion
- `ConfigChange::StopService` has been added to make requests to stop the service
```

examples/sample_runtime/src/main.rs
```diff
@@ -113,20 +113,31 @@ impl Runtime for SampleRuntime {
...
+            InstanceStatus::Stopped => {
+                let instance = self.started_services.remove(&spec.id);
+                println!("Stopping service {}: {:?}", spec, instance);
+            }
+        }
+
         Ok(())
     }
 
     /// Starts a new service instance and sets the counter value for this.
```
aleksuss: Is the description valid ? Maybe `Initiates adding a new service...` ?

exonum/src/runtime/dispatcher/schema.rs
aleksuss: Good catch, but `pending_status` is used in the method `put` on the line 164. `Some(..)` vs `unwrap()` ? 😄

popzxc: Ah, indeed. Those sneaky `Copy` types got me again :male_detective:

services/supervisor/src/transactions.rs
alekseysidorov: The `StopService` request will be invalid because the service instance is not running.

exonum/src/runtime/dispatcher/error.rs
```diff
@@ -51,6 +51,8 @@ pub enum Error {
     UnauthorizedCaller = 11,
     /// Malformed arguments for calling a service interface method.
     MalformedArguments = 12,
+    /// The service instance is already pending.
```
slowli: Nit: IMO, something like "service instance is already transitioning to a new status" would be more understandable.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -220,6 +220,20 @@ impl Dispatcher {
             .map_err(From::from)
     }
 
+    /// Initiates stopping of an existing service instance in the blockchain. The stopping
+    /// service is active (i.e., does process transactions or `after_transactions` hook)
```
slowli: Nit: I'd rephrase as "(i.e., processes transactions and the `after_transactions` hook)".

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -220,6 +220,20 @@ impl Dispatcher {
...
+    /// until the block built on top of the provided `fork` is committed.
+    ///
+    /// This method should be called for the exact context passed to the runtime.
```
slowli: This is a copy-paste artifact, isn't it? There is no context involved in the call.

exonum/src/runtime/mod.rs
```diff
@@ -209,11 +209,11 @@ impl From<RuntimeIdentifier> for u32 {
 ///
 /// ```text
 /// LIFE ::= initialize (GENESIS | RESUME) BLOCK* shutdown
-/// GENESIS ::= (deploy_artifact | initiate_adding_service commit_service)* after_commit
-/// RESUME ::= (deploy_artifact | commit_service)* on_resume
-/// BLOCK* ::= PROPOSAL+ COMMIT
+/// GENESIS ::= (deploy_artifact | initiate_adding_service update_service_status)* after_commit
+/// RESUME ::= (deploy_artifact | update_service_status)* on_resume
+/// BLOCK* ::= PROPOSAL + COMMIT
```
slowli: `+` is a quantity modifier ("one or more"). IMO, it makes little sense to separate it from the quantified action (i.e., `PROPOSAL`).

alekseysidorov: Oh, I wasn't be attentive, I'll correct this.

PR #1607, opened 2019-12-10
Api versioning [ECR-3897]

popzxc:
**Upd:** Now this PR not only adds a new functionality, but also refactors a bit `actix` code and `NamedWith`. Hope you'll like it.

This PR adds a possibility to create deprecated endpoints, and generate endpoints that will return `Gone` or `MovedPermanently` HTTP responses.

Unfortunately, this PR involves a decent amount of boilerplate code, but this is the fault of the current API approach: we implement (sort of) the same thing thrice: as generic interface in the `api` module, as concrete implementation for `actix`, and as an intermediate layer for the `rust` runtime.

Hopefully, extracting `api` into the separate crate later will allow us to reduce the code duplication.

[Here](https://github.com/exonum/exonum/blob/ba8489e804dc909d138a07f23deed6f4866440c2/test-suite/testkit/tests/api/api_service.rs#L56) you can see how new features can be used.


exonum/src/api/mod.rs
```diff
@@ -71,10 +82,62 @@ pub trait ApiBackend: Sized {
...
         self.raw_handler(Self::Handler::from(named_with))
     }
 
+    /// Creates an endpoint which will return "301 Moved Permanently" HTTP status code
+    /// to the incoming requests.
+    /// Response will include a "Location" header denoting a new location of the resource.
+    fn moved_permanently(
+        &mut self,
+        name: &'static str,
+        new_location: &'static str,
```
slowli: I think user agents such as browsers [redirect to the provided location](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301); i.e., it should be specific for this particular request. Hence, I think that service developers should be provided with a handler instead of a static replacement. Static replacement could make more sense with the `Link` header with `rel="latest-version"` / `rel="alternate"` as per the [`Deprecation` header draft](https://tools.ietf.org/html/draft-dalal-deprecation-header-01).

popzxc: From [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.4.2):

>The server SHOULD generate a Location header field in the response
   containing a preferred URI reference for the new permanent URI.  The
   user agent MAY use the Location field value for automatic
   redirection.

exonum/src/api/backends/actix.rs
```diff
@@ -131,12 +187,47 @@ impl ResponseError for api::Error {
...
+        };
+
+        let warning_text = format!(
+            "Deprecated API: This endpoint is deprecated, \
+             see the documentation to find an alternative. \
+             {}",
+            expiration_note
+        );
+
+        response.header(header::WARNING, warning_text);
```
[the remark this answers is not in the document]
popzxc: Well, for some reason I thought that `actix-web` transmutes the string to the required format. Now when I think about it, I can't understand what caused me for such a conviction :sweat_smile: 

Regarding the warn agent: it should be `( uri-host [ ":" port ] ) / pseudonym`, but it's optional (can be `-`).

Warn date, which is also a part of spec, is optional as well, so I think I'll just format the string to something like `"299 - \"Deprecated API: ...\""`.

exonum/src/api/mod.rs
popzxc: So, I guess, `MovedPermanently` is intended for basically just redirecting to a new location without any changes (even changing the request type is not recommended), and current signature suitable for such a convention. Or am I wrong here?

exonum/src/runtime/rust/api.rs
```diff
@@ -160,10 +162,136 @@ impl ServiceApiScope {
...
+    ///                 serde_urlencoded::to_string(query).expect("Unable to serialize query.");
+    ///     
+    ///             let new_location = format!("{}?{}", moved_immutable_new_location, query);
+    ///     
+    ///             Ok(new_location)
+    ///         },
+    ///         EndpointMutability::Immutable,
+    ///     );
+    /// ```
+    pub fn moved_permanently<Q, F>(
```
slowli: Why are separate methods required once we have dedicated error types for `Moved Permanently` / `Gone` statuses? This looks unnecessarily complex and awkward to me.  (In fact, I was going to propose create these error types and handle endpoints via errors, but the prerequisites for this seem to be already here!)

exonum/src/runtime/rust/api.rs
```diff
@@ -160,10 +162,136 @@ impl ServiceApiScope {
         self
     }
 
+    /// Same as `endpoint`, but also add a warning about this endpoint being deprecated to the response.
+    pub fn deprecated_endpoint<Q, I, F, R>(
```
slowli: Is it possible to use the following decorator-based approach? (Not sure about the accuracy of the "decorator" term.)

```rust
struct ServiceApi;

impl ServiceApi {
    fn some_handler(
        state: &api::ServiceApiState<'_>,
        arg: SomeRequest,
    ) -> api::Result<SomeResponse> {
        // ...
    }
}

let scope: &mut ServiceApiScope = ...;
scope.endpoint(Deprecated::new(ServiceApi::some_handler).with_date(Utc.ymd(2020, 06, 20))
```

Would minimize the amount of abstractions and could allow to introduce new "decorators" in the future.

exonum/src/runtime/rust/api.rs
popzxc: My original idea was to free users from creating any handlers by themselves for those cases, but seems like it turned out to be too confusing. I'll remove those methods and make returning error for those scenarios the intended use case.

exonum/src/runtime/rust/api.rs
popzxc: Well, it seems that currently implementing this feature on the `rust` runtime API level will require significant architecture changes, so I suggest to left things as-is *for now*, but return to this question once we will face the need in this functionality for at least one more thing.

exonum/src/runtime/rust/api.rs
```diff
@@ -160,10 +161,87 @@ impl ServiceApiScope {
...
+                    name: descriptor.1.as_ref(),
+                },
+            );
+            let result = handler(&state, query);
+            Box::new(result.into_future())
+        };
+        // Mark endpoint as deprecated.
+        let mut handler = Deprecated::from(handler);
+        if let Some(deprecates_on) = deprecates_on {
+            handler = handler.with_date(deprecates_on)
```
slowli: Nit: Missing semicolon.

test-suite/testkit/src/api.rs
```diff
@@ -261,9 +283,25 @@ where
...
+        let headers = response.headers();
+        for (header, expected_value) in expected_headers.iter() {
+            let header_value = headers.get(header).unwrap_or_else(|| {
+                panic!(
+                    "Response {:?} was expected to have header {}, but it isn't present",
+                    response, header
+                );
+            });
+
+            assert_eq!(header_value, expected_value);
```
slowli: Nit: It could make sense to add a message to `assert_eq` macro for additional context, something like `"Unexpected value of response header {}"`.

test-suite/testkit/Cargo.toml
```diff
@@ -21,6 +21,7 @@ travis-ci = { repository = "exonum/exonum" }
 [dependencies]
 actix = "0.7.9"
 actix-web = { version = "0.7.18", default-features = false }
+chrono = { version = "0.4.6" }
```
slowli: Nit: Why isn't a short form (`chrono = "0.4.6"`) used here?

exonum/src/runtime/rust/api.rs
```diff
@@ -160,10 +161,87 @@ impl ServiceApiScope {
...
+        self
+    }
+
     /// Return a mutable reference to the underlying web backend.
     pub fn web_backend(&mut self) -> &mut crate::api::backends::actix::ApiBuilder {
         self.inner.web_backend()
     }
+
+    /// Returns the base for service URLs, e.g. "/api/services/xnm-token/".
+    pub fn url_base(&self) -> String {
```
slowli: IMO, a better alternative to exposing this method is to prepend the URL base automatically to the URL returned by the user. Or do you think there is a valid case where the service would redirect to API of another service?

exonum/src/api/backends/actix.rs
```diff
@@ -474,28 +502,83 @@ impl From<AllowOrigin> for Cors {
...
+    fn json_responses() {
+        use chrono::TimeZone;
+
+        let actual_response = json_response(Actuality::Actual, 123);
+        assert_responses_eq(actual_response, HttpResponse::Ok().json(123));
+
+        let deprecated_response_no_deadline = json_response(Actuality::Deprecated(None), 123);
+        let expected_warning_text =
+            "Deprecated API: This endpoint is deprecated, \
+             see the documentation to find an alternative. \
```
slowli: Now, here's the drawback of the current approach: If a user wants a custom warning text, she would need to create PR in the core / modify core sources / petition the UN / ... Whereas with decorator mechanics, this would be significantly easier, and even adding new functionality to the `Deprecated` decorator (e.g., `description` / `doc_url` method) would likely be non-breaking and easy to migrate to.

exonum/src/api/backends/actix.rs
```diff
@@ -131,116 +132,143 @@ impl ResponseError for api::Error {
...
+/// <warn-code> in our case is 299, which means a miscellaneous persistent warning.
+/// <warn-agent> is optional, so we set it to "-".
+/// <warn-text> is a warning description, which is taken as an only argument.
+/// <warn-date> is not required.
+/// For details you can see RFC 7234, section 5.5: Warning.
+fn create_warning_header(warning_text: &str) -> String {
+    const WARNING_NUMBER: u16 = 299;
+    const WARNING_AGENT: &str = "-";
+
+    format!("{} {} \"{}\"", WARNING_NUMBER, WARNING_AGENT, warning_text)
```
slowli: Micro-nit: I wonder whether constants are inlined by the compiler here. Could give a (microscopic) performance boost.

exonum/src/api/backends/actix.rs
```diff
@@ -131,116 +132,143 @@ impl ResponseError for api::Error {
...
 }
 
-impl<Q, I, F> From<NamedWith<Q, I, FutureResult<I>, F, Immutable>> for RequestHandler
+/// Takes `HttpRequest` as a parameter and extracts query:
+/// - If request is immutable, the query is parsed from query string,
+/// - If request is mutable, the query is parsed from the request body as JSON.
+fn extract_query<Q>(
+    request: HttpRequest,
+    mutability: EndpointMutability,
+) -> Box<dyn Future<Item = Q, Error = actix_web::error::Error>>
```
slowli: Nit: Couldn't this be `impl Future`?

exonum/src/api/mod.rs
```diff
@@ -40,38 +39,46 @@ use crate::{api::node::SharedNodeState, blockchain::Blockchain};
...
 pub trait ApiBackend: Sized {
     /// Concrete endpoint handler in the backend.
     type Handler;
     /// Concrete backend API builder.
     type Backend;
 
-    /// Add the given endpoint handler to the backend.
-    fn endpoint<N, Q, I, R, F, E>(&mut self, name: N, endpoint: E) -> &mut Self
+    /// Adds the given endpoint handler to the backend.
+    fn endpoint<Q, I, R, F, E>(&mut self, name: &'static str, endpoint: E) -> &mut Self
```
slowli: Is `'static` lifetime necessary? Adding endpoints with dynamic naming is a valid use case, IMO.

exonum/src/api/with.rs
```diff
@@ -42,40 +42,135 @@ pub type FutureResult<I> = Box<dyn Future<Item = I, Error = error::Error>>;
...
-#[derive(Debug)]
-pub struct Immutable;
+/// Endpoint actuality.
+#[derive(Debug, Clone)]
+pub enum Actuality {
+    /// Endpoint is suitable for use.
+    Actual,
+    /// Endpoint is not recommended to use, the support of it will end soon.
+    /// Contains optional value denoting the endpoint expiration date.
+    Deprecated(Option<chrono::Date<chrono::Utc>>),
```
slowli: Nit: Fully qualified names look weird, IMO. Also, are you sure `Date` (rather than `DateTime`) will always be enough? The `Deprecation` HTTP header draft uses a `DateTime` equivalent, for example.

test-suite/testkit/Cargo.toml
popzxc: Because it's in this form for main `exonum` `Cargo.toml`. I think we should revise our `Cargo.toml`'s and fix all issues like this at once.

exonum/src/runtime/rust/api.rs
popzxc: I thought that maybe some service will implement a frontend for the network, and instead of direct communication with some node it will be recommended to interact with a server (which will proxy requests to the network). Or is it unrealistic?

exonum/src/api/backends/actix.rs
popzxc: But should he? Of course, deprecation warnings aren't standardized (yet), but IMO it will be better for end-side users to expect messages from the framework in some certain format, rather than look at the docs for every different service to check if it customizes responses.

exonum/src/runtime/rust/api.rs
slowli: Jeez, what happens if a node is behind a proxy in general? The location may just be incorrect in this case since the base URL may be incorrect because the proxy layer may add a prefix to node API URLs. Using a relative `Location` URL in this case may make more sense, but requires more work from the service developer.

exonum/src/api/backends/actix.rs
popzxc: Unfortunately no, match arms then will have incompatible types.

exonum/src/runtime/rust/api.rs
popzxc: I was just answering the question why I did it that way; no need for anger.

I've tried to implement suggested approach, but faced a problem: `MovedPermanently` endpoins are now the same as usual endpoints, and we can't distinguish the returned result until the handler is invoked (within `actix.rs`). As a result, I wasn't able to implement this easily. The only solution I found is to include the service name into the error, and do prepending when converting `api::Error` into `HttpResponse`, but it looks like a kludge :pensive:

exonum/src/api/backends/actix.rs
alekseysidorov: And what about Either?

popzxc: `Either` will make the return type more complex. As far as I know `Box` doesn't bring any overhead other than memory allocation, and since we aren't afraid of creating `Vec`s, this isn't a big problem, is it?

test-suite/testkit/tests/api/api_service.rs
```diff
@@ -0,0 +1,124 @@
...
+        // Moved endpoints.
+        let url_base = public_scope.url_base();
+        let moved_mutable_new_location = format!("{}/{}", &url_base, "ping-pong-deprecated-mut");
+        let moved_immutable_new_location = format!("{}/{}", &url_base, "ping-pong");
+        public_scope
+            .endpoint_mut(
+                "moved-mutable",
+                move |_state: &ServiceApiState<'_>, _query: PingQuery| -> api::Result<u64> {
+                    Err(api::Error::MovedPermanently(
+                        moved_mutable_new_location.clone(),
```
aleksuss: Needless `clone` ?

exonum/src/api/backends/actix.rs
slowli: > `Either` will make the return type more complex.

How though? The return type will be `impl Future<..>`.

> As far as I know Box doesn't bring any overhead other than memory allocation

Well, memory allocation in heap *is* an overhead, so we use arrays where it's possible (e.g., in crypto). Additionally, `Box<dyn _>` brings in dynamic dispatch, which is also slower. I doubt this overhead is significant, but if we *can* use stack-enabled / statically dispatched `impl Future`, why not?

exonum/src/api/backends/actix.rs
slowli: IMO, "see the documentation" bit sort of asks for a link 🙃 Otherwise, it's unclear which documentation to search for.

exonum/src/runtime/rust/api.rs
slowli: No anger implied, sorry. The expression I was going for was "oh my god, it's really difficult to account for the reverse proxy use case, isn't it?"

By using relative URL I meant literally using something like

```rust
api::Error::MovedPermanently("../v2/improved-endpoint")
```

since `Location` HTTP header allows relative URLs. This, of course, would shift the problem to service developers; they would need to compute the number of `../`s correctly.

popzxc: Ah, I get it now.

Hm, and what if we will provide a helper function like `redirect_link(from, to)`, which will produce a relative link from old URL to the new one? It seems that it's pretty simple for implementation, but it will free service developers from counting `..`s.

test-suite/testkit/tests/api/api_service.rs
popzxc: It's required, since it's a `move` closure.

exonum/src/api/backends/actix.rs
popzxc: It's now possible to add an additional description, which will be included to the deprecation warning (thus the beginning of the warning remains well-defined, but additional info can be provided).

exonum/src/runtime/rust/api.rs
popzxc: Added a `relative_to` method instead of `url_base`.

exonum/src/runtime/rust/api.rs
```diff
@@ -160,10 +162,120 @@ impl ServiceApiScope {
...
+    ///
+    /// ```
+    /// use exonum::runtime::rust::api::ServiceApiScope;
+    ///
+    /// fn main() {
+    ///    let relative_path = ServiceApiScope::relative_to("hello/world", "exonum/rust/app");
+    ///    assert_eq!(&relative_path, "../../exonum/rust/app");
+    /// }
+    /// ```
+    pub fn relative_to(old_endpoint: &str, new_endpoint: &str) -> String {
```
slowli: What if the endpoint path (relative to the service root) is a part of `ServiceApiState`? In this case, it would possible to encapsulate this logic in something like

```rust
scope.endpoint("v1/foo", |state, query| {
    let new_query = // based on `query`
    Err(state.moved_permanently("v2/foo").with_query(&new_query))
})
```

CHANGELOG.md
```diff
@@ -35,9 +35,14 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 #### exonum
 
 - `before_transactions` hook for services was introduced. (#1577)
+
 - `ErrorMatch` was introduced to test (e.g., using the testkit) that
   an `ExecutionError` has an expected type, error message and/or location. (#1585)
 
+- New types of error were added to `api::Error`: `Gone` and `MovedPermanently`. (#1607)
```
slowli: Nit: awkward wording; something like "new error variants were added" looks better.

test-suite/testkit/tests/api/api_service.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    fn wire(builder: &mut ServiceApiBuilder) {
+        let public_scope = builder.public_scope();
+
+        // Normal endpoint.
+        public_scope.endpoint("ping-pong", Self::ping_pong);
+
+        // Deprecated endpoints.
+        public_scope
+            .deprecated_endpoint("ping-pong-deprecated", None, None, Self::ping_pong)
```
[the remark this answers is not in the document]
popzxc: Done, now it's possible to use `Deprecated` builder from `exonum::api`.

exonum/src/runtime/rust/api.rs
popzxc: Done, now it's possible to create `MovedPermanently` response with builder.

exonum/src/runtime/rust/api.rs
```diff
@@ -40,18 +39,25 @@ pub struct ServiceApiState<'a> {
     broadcaster: Broadcaster<'a>,
     // TODO Think about avoiding of unnecessary snapshots creation. [ECR-3222]
     snapshot: Box<dyn Snapshot>,
+    /// Endpoint relative URL.
```
slowli: Nit: I'd specify "endpoint path relative to the service root".

exonum/src/api/error.rs
```diff
@@ -57,6 +58,44 @@ pub enum Error {
     Unauthorized,
 }
 
+/// A helper structure allowing to build `MovedPermanently` response from the
+/// composite parts.
+#[derive(Debug)]
+pub struct MovedPermanentlyErrorBuilder {
```
slowli: Nit: It could make sense to refactor this type as just `MovedPermantentlyError`, with a conversion into `api::Error`, since the error is ready once constructed.

PR #1609, opened 2019-12-10
Move incorrect tx check to the explorer [ECR-3907]

Kormick:
## Overview

Moved incorrect tx check to the explorer. Also added the same check in the websocket handle.

### DIFF
--- a/exonum/src/api/websocket.rs
+++ b/exonum/src/api/websocket.rs
@@ -346,8 +346,7 @@ impl Handler<Broadcast> for Server {
 impl Handler<Transaction> for Server {
     type Result = Result<TransactionResponse, failure::Error>;
 
-    /// Checks if incoming transaction is valid with Blockchain::check_tx.
-    /// Broadcasts valid transactions and ignores invalid.
+    /// Broadcasts transaction if the check was passed, and returns an error otherwise.
     fn handle(
         &mut self,
         Transaction { tx }: Transaction,
--- a/exonum/src/api/websocket.rs
+++ b/exonum/src/api/websocket.rs
@@ -346,6 +346,8 @@ impl Handler<Broadcast> for Server {
 impl Handler<Transaction> for Server {
     type Result = Result<TransactionResponse, failure::Error>;
 
+    /// Checks if incoming transaction is valid with Blockchain::check_tx.
+    /// Broadcasts valid transactions and ignores invalid.
     fn handle(
         &mut self,
         Transaction { tx }: Transaction,
@@ -353,11 +355,11 @@ impl Handler<Transaction> for Server {
     ) -> Self::Result {
         let msg = SignedMessage::from_hex(tx)?;
         let tx_hash = msg.object_hash();
+        let verified = msg.into_verified()?;
+        Blockchain::check_tx(&self.blockchain.snapshot(), &verified)?;
+
         // FIXME Don't ignore message error.
-        let _ = self
-            .blockchain
-            .sender()
-            .broadcast_transaction(msg.into_verified()?);
+        let _ = self.blockchain.sender().broadcast_transaction(verified);
         Ok(TransactionResponse { tx_hash })
     }
 }
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -168,6 +168,15 @@ impl Blockchain {
     pub fn service_keypair(&self) -> &(PublicKey, SecretKey) {
         &self.service_keypair
     }
+
+    /// Performs several shallow checks that transaction is correct.
+    ///
+    /// Returned `Ok(())` value doesn't necessarily mean that transaction is correct and will be
+    /// executed successfully, but returned `Err(..)` value means that this transaction is
+    /// **obviously** incorrect and should be declined as early as possible.
+    pub fn check_tx(&self, tx: &Verified<AnyTx>) -> Result<(), ExecutionError> {
+        Dispatcher::check_tx(&self.snapshot(), tx)
+    }
 }
 
 /// Mutable blockchain capable of processing transactions.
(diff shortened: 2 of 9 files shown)


test-suite/testkit/tests/counter/main.rs
```diff
@@ -1018,3 +1019,35 @@ fn test_explorer_with_before_transactions_error() {
...
+
+    // Send invalid transaction.
+    let (pubkey, key) = crypto::gen_keypair();
+    let tx = Reset.sign(SERVICE_ID + 1, pubkey, &key);
+    let data = hex::encode(tx.to_bytes());
+    let response = api
+        .public(ApiKind::Explorer)
+        .query(&json!({ "tx_body": data }))
+        .post::<TransactionResponse>("v1/transactions")
+        .unwrap_err();
```
popzxc: Nit: I'd suggest to use `expect_err` and `expect`, since the return value type is the topic of this test. With `expect_err`/`expect` we will receive a more meaningful messages if the test will fail.

exonum/src/node/consensus.rs
```diff
@@ -626,7 +626,7 @@ impl NodeHandler {
             bail!("Received already processed transaction, hash {:?}", hash)
         }
 
-        if let Err(e) = self.blockchain.check_tx(&msg) {
+        if let Err(e) = Blockchain::check_tx(&self.blockchain.snapshot(), &msg) {
```
popzxc: Snapshot is already available here, we don't have to create another one. Or am I wrong?

exonum/src/blockchain/tests.rs
```diff
@@ -765,11 +765,11 @@ fn test_check_tx() {
     );
 
     let correct_tx = TestAdd { value: 1 }.sign(TEST_SERVICE_ID, keypair.0, &keypair.1);
-    blockchain.check_tx(&correct_tx).unwrap();
+    Blockchain::check_tx(&blockchain.snapshot(), &correct_tx).unwrap();
```
popzxc: Nit: you can create a snapshot once to reuse it.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -181,11 +181,13 @@ impl Dispatcher {
     /// Returned `Ok(())` value doesn't necessarily mean that transaction is correct and will be
     /// executed successfully, but returned `Err(..)` value means that this transaction is
     /// **obviously** incorrect and should be declined as early as possible.
-    pub(crate) fn check_tx(&self, tx: &Verified<AnyTx>) -> Result<(), ExecutionError> {
-        // Currently the only check is that destination service exists, but later
```
popzxc: Why was the comment removed?

exonum/src/api/websocket.rs
```diff
@@ -346,18 +346,20 @@ impl Handler<Broadcast> for Server {
 impl Handler<Transaction> for Server {
     type Result = Result<TransactionResponse, failure::Error>;
 
+    /// Checks if incoming transaction is valid with Blockchain::check_tx.
+    /// Broadcasts valid transactions and ignores invalid.
```
popzxc: Nit: I'd rephrase it like "Broadcasts transaction if the check was passed, and returns an error otherwise.` (something similar can be applied to the `ExplorerApi::add_transaction` as well).

exonum/tests/websocket/main.rs
```diff
@@ -99,6 +100,24 @@ fn test_send_transaction() {
...
+    let tx = CreateWallet::new(pk, "Bob").sign(SERVICE_ID + 1, pk, &sk);
+    let tx_json =
+        serde_json::to_string(&json!({ "type": "transaction", "payload": { "tx_body": tx }}))
+            .unwrap();
+    client.send_message(&OwnedMessage::Text(tx_json)).unwrap();
+
+    // Check response on sent message.
+    let resp_text = recv_text_msg(&mut client).unwrap();
+    assert_eq!(
+        serde_json::from_str::<serde_json::Value>(&resp_text).unwrap(),
```
slowli: Nit: Could make sense to introduce a `let response: Value = ...` binding for this expression for the code to be more readable.

exonum/tests/websocket/main.rs
```diff
@@ -99,6 +100,24 @@ fn test_send_transaction() {
...
+        json!({
+            "result": "error",
+            "description": "Execution error `dispatcher:7` occurred: Suitable runtime for the given service instance ID is not found."
```
slowli: Nit: Overly long line; something like this should work:

```rust
json!({
    "result": "error",
    "description": "Execution error `dispatcher:7` occurred: Suitable runtime \
                    for the given service instance ID is not found.",
})
```

exonum/src/blockchain/mod.rs
```diff
@@ -168,6 +168,15 @@ impl Blockchain {
     pub fn service_keypair(&self) -> &(PublicKey, SecretKey) {
         &self.service_keypair
     }
+
+    /// Performs several shallow checks that transaction is correct.
+    ///
+    /// Returned `Ok(())` value doesn't necessarily mean that transaction is correct and will be
+    /// executed successfully, but returned `Err(..)` value means that this transaction is
+    /// **obviously** incorrect and should be declined as early as possible.
+    pub fn check_tx(snapshot: &dyn Snapshot, tx: &Verified<AnyTx>) -> Result<(), ExecutionError> {
```
slowli: (Slightly) off-topic: I'm not sure that the API endpoint should return `ExecutionError` as is, especially given the recent work on unifying error format. Could be solved later, I guess.

exonum/tests/websocket/main.rs
```diff
@@ -99,6 +100,24 @@ fn test_send_transaction() {
         })
     );
 
+    // Send invalid transaction.
+    let (pk, sk) = gen_keypair();
+    let tx = CreateWallet::new(pk, "Bob").sign(SERVICE_ID + 1, pk, &sk);
+    let tx_json =
+        serde_json::to_string(&json!({ "type": "transaction", "payload": { "tx_body": tx }}))
```
slowli: Nit: Ugly formatting; binding `json!(..)` to a variable could improve it.

exonum/src/blockchain/mod.rs
popzxc: Yeah, I think we'll handle it within the API errors unification topic.

PR #1611, opened 2019-12-11
Fix couple of bugs in MerkleDB

slowli:
## Overview

This PR fixes incorrect implementation of `Snapshot` for `Patch`. The previous implementation didn't take into account that indexes can be deleted within the `Patch`; this led to potential stale / inconsistent reads from `Patch`es or `Fork`s.

Independently, the byte iterator implementation for `ViewChanges` was incorrect; it entered an infinite loop in `peek()` once a `Delete` change was encountered. This bug was not previously exposed since the `peek` method was not used; but with the correction of `Patch` behavior, the situation has changed.

### DIFF
--- a/components/merkledb/src/views/mod.rs
+++ b/components/merkledb/src/views/mod.rs
@@ -381,10 +381,14 @@ where
     fn peek(&mut self) -> Option<(&[u8], &[u8])> {
         loop {
             match self.inner.peek() {
-                Some((key, &Change::Put(ref value))) => {
+                Some((key, Change::Put(ref value))) => {
                     return Some((key.as_slice(), value.as_slice()));
                 }
-                Some((_, &Change::Delete)) => {}
+                Some((_, Change::Delete)) => {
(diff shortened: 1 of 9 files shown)


components/merkledb/src/views/mod.rs
```diff
@@ -183,37 +183,19 @@ impl<T: RawAccess> View<T> {
...
-                }
-            }
-
-            if changes.is_cleared() {
-                return None;
-            }
-        }
-
-        self.snapshot().get(&self.address, key)
+        self.changes
```
popzxc: This pattern is encountered multiple times within the code. Can it somehow be encapsulated?

slowli: It's used 2 times x 2 different variations. I think encapsulation would only increase the amount of code (the necessary function to encapsulate changes would need to include at least 4 params: `snapshot`, `changes`, `name` and `key`), so I didn't do that.

popzxc: Makes sense.

PR #1612, opened 2019-12-12
Implement node API for retrieving call status [ECR-3887]

SergeiMal:
## Overview

---

See: https://jira.bf.local/browse/ECR-3887

### Definition of Done


exonum/src/api/node/public/explorer.rs
```diff
@@ -276,6 +308,72 @@ impl ExplorerApi {
...
+                CallInBlock::before_transactions(tx_info.content().payload().call_info.instance_id);
+            fill_call_status(call_in_block, CallKind::BeforeTransactions);
+
+            let call_in_block: CallInBlock =
+                CallInBlock::transaction(tx_info.location().position_in_block());
+            fill_call_status(call_in_block, CallKind::Transaction);
+
+            let call_in_block: CallInBlock =
+                CallInBlock::after_transactions(tx_info.content().payload().call_info.instance_id);
+            fill_call_status(call_in_block, CallKind::AfterTransactions);
```
alekseysidorov: This code looks unreadalbe, here we need to decomposite this method into the set of clear helpers.

popzxc: And boilerplate code should be generalized if possible.

exonum/src/api/node/public/explorer.rs
```diff
@@ -276,6 +316,80 @@ impl ExplorerApi {
...
+            TransactionInfo::InPool { .. } => {
+                let description =
+                    serde_json::to_string(&json!({ "type": "not committed" })).unwrap();
+                Err(ApiError::NotFound(description))
+            }
+        }?;
+
+        let block_height = tx_info.location().block_height();
+        let block_info = explorer.block(block_height).ok_or_else(|| {
+            ApiError::NotFound(format!(
```
popzxc: Why in some places error messages are encoded as JSON, and in some it's just a plain text?

exonum/src/api/node/public/explorer.rs
```diff
@@ -276,6 +316,80 @@ impl ExplorerApi {
...
+                "Requested block height ({}) exceeds the blockchain height ({})",
```
popzxc: What means "requested block height"? There is no block height in the query.

exonum/src/api/node/public/explorer.rs
```diff
@@ -171,6 +171,46 @@ impl AsRef<[u8]> for TransactionHex {
...
+    Transaction,
+    /// Call kind of `before_transactions` hook
+    BeforeTransactions,
+    /// Call kind of `after_transactions` hook
+    AfterTransactions,
+}
+
+/// Call status query parameters.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+pub struct CallStatusQuery {
```
popzxc: To be honest, it looks like an attempt to mix everything at once.
I'd expect the query to be something like:

```
pub enum CallKind {
    Transaction(hash),
    BeforeTransactions(height),
    AfterTransactions(height),
}
```

All three types of the request are different and I'm not sure that we should support requesting data for more than one at the time (meaning that if user is interested in the transaction result, before_transactions result and after_transactons result, he should do three different requests to the endpoint).

test-suite/testkit/tests/counter/main.rs
```diff
@@ -1052,3 +1053,175 @@ fn test_explorer_add_transaction_with_invalid_transaction() {
...
+    let block = testkit.create_block_with_transaction(tx.clone());
+
+    let response = api
+        .public(ApiKind::Explorer)
+        .query(&CallStatusQuery {
+            hash: tx.object_hash(),
+            call_kind: None,
+        })
+        .get::<CallStatusResponse>("v1/call_status")
+        .expect("Transaction send to finish with error.");
```
popzxc: I don't quite understand the meaning of "Transaction send to finish with error.".

test-suite/testkit/tests/counter/main.rs
```diff
@@ -1052,3 +1053,175 @@ fn test_explorer_add_transaction_with_invalid_transaction() {
...
+
+    assert!(response.before_transactions.is_some());
+    assert!(response.transaction.is_some());
+    assert!(response.after_transactions.is_some());
+
+    assert!(response
+        .before_transactions
+        .clone()
```
popzxc: Why are `clone`s are required?

exonum/src/api/node/public/explorer.rs
SergeiMal: Left as is because in case with height it need instance id to build CallInBlock

popzxc: And why can't we include `InstanceId` to the request?
```
pub enum CallKind {
    Transaction(hash),
    BeforeTransactions(height, instance_id),
    AfterTransactions(height, instance_id),
}
```

Requesting result of the service hook with hash of the transaction seems logically incorrect: what if there was no transactions for the service within block, and we still want to see the result of `AfterTransactions`?

And it's overall inconvenient and counterintuitive for users to request not a tx by tx hash.

exonum/src/api/node/public/explorer.rs
```diff
@@ -171,6 +171,35 @@ impl AsRef<[u8]> for TransactionHex {
...
```
slowli: I expect the query to serialize to something like

```text
height=10&type=transaction&index=3
```

or

```text
height=10&type=after_transactions&instance_id=5
```

I think the easiest way to accomplish this is by using

```rust
struct CallStatusQuery {
    height: Height, // or u64
    #[serde(flatten)]
    call: CallInBlock,
}
```

If we want to support searching by `tx_hash` too (fair enough), then it could be accomplished with

```rust
#[serde(untagged)]
enum CallStatusQuery {
    TransactionHash { tx_hash: Hash },
    CallWithHeight {
        height: Height, // or u64
        #[serde(flatten)]
        call: CallInBlock,
    },
}
```

exonum/src/explorer/mod.rs
```diff
@@ -553,6 +553,18 @@ impl<'a> BlockchainExplorer<'a> {
         Some(TransactionInfo::Committed(tx))
     }
 
+    /// Return status of call in block
+    pub fn call_status(
+        &self,
+        block_height: Height,
+        call_location: &CallInBlock,
```
slowli: Nit: `CallInBlock` is a `Copy` type (and quite small in size); it could make sense to pass it by value.

exonum/src/runtime/error.rs
```diff
@@ -428,7 +429,7 @@ where
 ///
 /// [`ErrorKind`]: enum.ErrorKind.html
 /// [`CallSite`]: struct.CallSite.html
-#[derive(Clone, Debug, Fail, BinaryValue)]
+#[derive(Clone, Debug, Deserialize, Serialize, Fail, BinaryValue)]
```
slowli: Implementing `(De)Serialize` is not necessary; see the comment about `ExecutionStatus`.

exonum/src/runtime/error.rs
```diff
@@ -88,7 +88,8 @@ use crate::{
 ///
 /// [`catch_unwind`]: https://doc.rust-lang.org/std/panic/fn.catch_unwind.html
 /// [`DispatcherError`]: ../enum.DispatcherError.html
-#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
```
slowli: Implementing `(De)Serialize` is not necessary; see the comment about `ExecutionStatus`.

exonum/src/api/node/public/explorer.rs
```diff
@@ -276,6 +351,44 @@ impl ExplorerApi {
...
+                let tx_info = explorer.transaction(&tx_hash).ok_or_else(|| {
+                    ApiError::NotFound(format!("Unknown transaction hash ({})", tx_hash))
+                })?;
+                let tx_info = match tx_info {
+                    TransactionInfo::Committed(info) => Ok(info),
+                    TransactionInfo::InPool { .. } => Err(ApiError::NotFound(format!(
+                        "Requested transaction ({}) is not executed yet",
+                        tx_hash
+                    ))),
+                }?;
```
slowli: Nit: Calling `?` here looks weird, IMO. Something like this could be more idiomatic:

```rust
let tx_info = match tx_info {
    TransactionInfo::Committed(info) => info,
    TransactionInfo::InPool { .. } => {
        let err = ApiError::NotFound(/* ... */);
        return Err(err);
    }
};
```

exonum/src/api/node/public/explorer.rs
```diff
@@ -171,6 +172,80 @@ impl AsRef<[u8]> for TransactionHex {
     }
 }
 
+/// Status of a call.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CallStatusResponse {
```
slowli: Couple of notes here:

- `status` is redundant because it is contained in the proof anyway
- It could make sense to return the `BlockProof` (block header with precommits), at least if the transaction is queried by hash (in this case, the caller may reasonably not have `BlockProof` locally)

exonum/src/api/node/public/explorer.rs
```diff
@@ -171,6 +172,80 @@ impl AsRef<[u8]> for TransactionHex {
...
+    /// Call status
+    pub status: ExecutionStatus,
+    /// Call execution proof
+    pub call_proof: MapProof<CallInBlock, ExecutionError>,
```
slowli: This is bad. We have a `serde` transform for `ExecutionError` (in `runtime::error::execution_error` module), but it cannot be applied here because `ExecutionError` is not a field. Then again, there was some talk about returning proofs in Protobuf. @aleksuss, can you comment on this?

aleksuss: Yeah. We have a plan to return proofs in protobuf format but in scope of other ticket.

popzxc: I think that *currently* mixing `json` and `protobuf` will cause only inconvenience.
IMO for the proof contents the explicit deserialization format derived from the `rust` structure is not a bad thing: it's easier to write a parser depending on the structure layout (if you'll want to check the proof).

exonum/src/api/node/public/explorer.rs
```diff
@@ -171,6 +172,80 @@ impl AsRef<[u8]> for TransactionHex {
...
+}
+
+/// Call status query parameters.
+#[derive(Debug, Clone)]
+#[derive(Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum CallStatusQuery {
```
popzxc: Since this query isn't obvious, I'd expect a doc-comment for it to contain an example of the expected `json`. It will be nice to have an example of output `json` in doc-comment for endpoint.

exonum/src/api/node/public/explorer.rs
aleksuss: Let's discuss this topic today on a grooming.

exonum/src/api/node/public/explorer.rs
```diff
@@ -171,6 +172,80 @@ impl AsRef<[u8]> for TransactionHex {
...
+        /// Location of an call within a block.
+        #[serde(flatten)]
+        call: CallKind,
+    },
+}
+
+/// The kind of requested call
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum CallKind {
```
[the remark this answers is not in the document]
popzxc: What if we will split the `call_status` endpoint into two separate endpoints? One for query with transaction hash, and one with `CallKind` + `height` as a query. It should simplify the overall structure.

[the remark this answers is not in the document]
vitvakatu: Involving this idea: what if we have separate endpoints for transactions, before_commits and after_commits?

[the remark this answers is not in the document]
aleksuss: Hm. Sounds not bad. Something like this: 
```
/api/explorer/v1/call_status/transaction?hash=....
/api/explorer/v1/call_status/after_transactions?height=5
/api/explorer/v1/call_status/before_transactions?height=5
``` 
?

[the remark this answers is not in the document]
vitvakatu: yep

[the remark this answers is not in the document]
aleksuss: @SergeiMal 👆

exonum/src/explorer/mod.rs
```diff
@@ -553,6 +553,18 @@ impl<'a> BlockchainExplorer<'a> {
         Some(TransactionInfo::Committed(tx))
     }
 
+    /// Return status of call in block
+    pub fn call_status(
```
slowli: I'm against exposing this method since it's ambiguous for `before_transactions` / `after_transactions` hooks whether the call was executed successfully or not executed at all. It could be possible to use more low-level methods in the core schema instead.

exonum/src/api/node/public/explorer.rs
```diff
@@ -171,6 +171,23 @@ impl AsRef<[u8]> for TransactionHex {
...
+}
+
+/// Call status query parameters.
+/// To check `before_transactions` or `after_transactions` call
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CallStatusQuery {
+    /// Height of a block
+    pub height: Height,
+    /// Numerical service identifier.
+    pub id: u32,
```
slowli: Nit: Using `InstanceId` instead of `u32` would be more idiomatic. Same for `service_id` instead of `id`.

test-suite/testkit/tests/counter/main.rs
```diff
@@ -918,3 +921,187 @@ fn test_explorer_add_transaction_with_invalid_transaction() {
...
+        .description()
+        .contains("Number 13"));
+
+    let response = api
+        .public(ApiKind::Explorer)
+        .query(&TransactionQuery {
+            hash: tx.object_hash(),
+        })
+        .get::<CallStatusResponse>("v1/call_status/transaction")
+        .expect("Explorer Api unexpectedly failed 3");
```
slowli: Nit: What's with the numbers? I think using backtrace is much more reliable to determine the error cause in tests rather than relying on messages.

exonum/src/api/node/public/explorer.rs
```diff
@@ -276,6 +293,59 @@ impl ExplorerApi {
             })
     }
 
+    /// Return call status of committed transaction.
```
popzxc: Nit: "Returns" instead of "return" here and in other places.

exonum/src/api/node/public/explorer.rs
```diff
@@ -276,6 +293,59 @@ impl ExplorerApi {
...
+
+        let call_in_block = CallInBlock::transaction(tx_info.location().position_in_block());
+        let block_height = tx_info.location().block_height();
+
+        let status = ExecutionStatus(explorer.call_status(block_height, call_in_block));
+        Ok(CallStatusResponse { status })
+    }
+
+    /// Return call status of `before_transactions` hook.
+    pub fn before_transactions_status(
```
popzxc: Nit: you can provide a type of the hook as an argument from `wire` method, and reduce `before_transactions_status` / `after_transactions_status` to one function, e.g. `hook_status`.

exonum/src/api/node/public/explorer.rs
SergeiMal: `fn block` and `fn blocks` annotated with "return" too

aleksuss: Feel free to change it there as well

exonum/src/runtime/error.rs
```diff
@@ -429,6 +430,7 @@ where
 /// [`ErrorKind`]: enum.ErrorKind.html
 /// [`CallSite`]: struct.CallSite.html
 #[derive(Clone, Debug, Fail, BinaryValue)]
+#[derive(Serialize, Deserialize)]
```
slowli: Is this `derive` still necessary? Same with the `derive(Serialize, Deserialize)` for `ErrorKind`.

exonum/src/api/node/public/explorer.rs
```diff
@@ -400,6 +470,18 @@ impl ExplorerApi {
                 let blockchain = self.blockchain.clone();
                 move |query| Self::block(blockchain.snapshot().as_ref(), query)
             })
+            .endpoint("v1/call_status/transaction", {
```
slowli: Without the authenticity proof, this endpoint is just a worse version of the `transaction` endpoint. I think we should get this PR merged already, so this can be solved separately.

exonum/src/api/node/public/explorer.rs
```diff
@@ -400,6 +470,18 @@ impl ExplorerApi {
...
+                let blockchain = self.blockchain.clone();
+                move |query| Self::transaction_status(blockchain.snapshot().as_ref(), query)
```
slowli: Nit: `&blockchain.snapshot()` instead of `.as_ref()` would be more concise. Again, could be solved separately.

test-suite/testkit/tests/counter/main.rs
```diff
@@ -906,3 +909,184 @@ fn test_explorer_add_invalid_transaction() {
...
+
+    let response = api
+        .public(ApiKind::Explorer)
+        .query(&CallStatusQuery {
+            height: Height(2),
+            service_id: SERVICE_ID,
+        })
+        .get::<CallStatusResponse>("v1/call_status/before_transactions")
+        .expect("Explorer Api unexpectedly failed");
+    assert!(response.status.0.is_err());
```
slowli: Nit: This is redundant given that `unwrap_err()` is called later. Also, a separate `let err = response.status.0.unwrap_err()` binding would probably decrease the number of lines 🙃

CHANGELOG.md
```diff
@@ -98,6 +98,12 @@ Indexes iterators names has been shortened to `Iter`, `Keys` and `Values`. (#162
 - `ErrorMatch` was introduced to test (e.g., using the testkit) that
   an `ExecutionError` has an expected type, error message and/or location. (#1585)
 
+- The set of endpoints were introduced.
```
slowli: Nit: I'd rephrase

> We introduced a set of public endpoints to retrieve the status of calls executed within a block:
> - `v1/call_status/transaction`: gets the status of a transaction
> - `v1/call_status/before_transactions`: gets the status of a `before_transactions` hook
> - `v1/call_status/after_transactions`: gets the status of an `after_transactions` hook

PR #1614, opened 2019-12-13
Enumerate group index [ECR-3904]

pavel-mukhanov:
Added method `keys` which return all the keys from group.


components/merkledb/src/group.rs
```diff
@@ -106,6 +108,69 @@ where
...
+
+        GroupIter::new(self.prefix.clone(), keys, self.access.clone())
+    }
+}
+
+///TODO: add doc
+#[derive(Debug)]
+pub struct GroupIter<K, T, I> {
+    base: IndexAddress,
+    keys: Vec<K>,
```
slowli: Storing all keys in the memory looks *wildly* inefficient in terms of memory use. OTOH, building right on top of a `IndexesPool` iterator is impossible - since `IndexesPool` would be borrowed, this would mean that operation on any other index is impossible during iteration! Honestly, I don't know what's the best solution here; using a random access on `IndexesPool` once the iterator is advanced seems viable, but is probably inefficient (in terms of performance). It could make sense to simply have a method in `Group` returning `Vec<K>` (a vector of all keys in the group). This would clearly notify the user that the method can use a large amount of RAM.

components/merkledb/src/views/metadata.rs
```diff
@@ -337,6 +339,33 @@ impl<T: RawAccess> IndexesPool<T> {
...
+            let key = get_suffix(entry);
+            keys.push(key);
+        }
+
+        keys
+    }
+}
+
+fn get_suffix<K: BinaryKey + Clone>(entry: (Vec<u8>, Vec<u8>)) -> K {
+    // TODO: simplify this
```
slowli: I'm not sure what this logic supposed to do. As I imagine `suffixes` method, it should work on "group" addresses only (i.e., addresses not containing the `id_in_group` part; the method signature may be changed to `prefix: &str` correspondingly). The method should iterate over keys that start with `${prefix}\0` and remove the corresponding number of bytes (`prefix.len() + 1`) at the beginning of the key.

components/merkledb/src/group.rs
```diff
@@ -142,4 +207,35 @@ mod tests {
         // The next line fails to compile because `Snapshot` cannot be written to:
         // group.get(&3).push("quux".to_owned());
     }
+
+    #[test]
+    fn group_iter() {
```
slowli: Please add a test that the iterator does not include longer index names, e.g., `"group_group_"` or `("group_group_", &0_u8)` and names situated after the group, e.g., `"unrelated"`.

components/merkledb/src/group.rs
alekseysidorov: But what if we change `iter` signature from `&self` to `self` to take onwership of the `IndexesPool`?

slowli: What will this change? For example, if the `IndexesPool` is built on top of a `&Fork`, it can be instantiated at most once at any given time. If we have one instance in the iterator, we cannot have another instance anywhere else, i.e., we cannot instantiate new indexes during iteration. (I suppose with some tinkering it may become possible to instantiate indexes we're iterating over, but it's still not enough IMO.)

components/merkledb/src/group.rs
```diff
@@ -91,7 +93,7 @@ where
 impl<T, K, I> Group<T, K, I>
 where
     T: Access,
-    K: BinaryKey + ?Sized,
+    K: BinaryKey + ?Sized + fmt::Debug,
```
alekseysidorov: Is this trait bound complete necessarily?

pavel-mukhanov: Forgot to remove

components/merkledb/src/group.rs
```diff
@@ -106,11 +107,31 @@ where
...
+where
+    T: RawAccess,
+    K: BinaryKey + Clone,
+    I: FromAccess<T>,
+{
+    /// Return all the keys from this group.
+    ///
+    /// Note: use this method carefully, because storing all keys in memory may
+    /// consume a large amount of RAM.
+    pub fn keys<V: BinaryValue>(&self) -> Vec<K> {
```
slowli: 1. What does the `V` type param represent?

2. The method should return `K::Owned`. I guess this is the reason you didn't use `Group<_, str, _>` in the tests.

3. Nit / could be solved separately: It *could* make sense to provide API like

```rust
fn keys_unchecked(&self) -> impl Iterator<Item = K>
```

provided that it's clearly documented that creating any new index while the iterator is in scope may lead to a panic. Or it could be worth providing the iterator for the readonly accesses only (e.g., for `T: AsReadonly<Readonly = T>`) 🤔

components/merkledb/src/views/metadata.rs
```diff
@@ -337,6 +339,19 @@ impl<T: RawAccess> IndexesPool<T> {
...
         (metadata, is_phantom)
     }
+
+    pub(super) fn suffixes<K: BinaryKey + Clone>(&self, prefix: &str) -> Vec<K> {
+        let prefix = concat_keys!(prefix, INDEX_NAME_SEPARATOR);
+        let mut keys = Vec::new();
+
+        for entry in self.0.iter::<Vec<u8>, Vec<u8>, Vec<u8>>(&prefix) {
+            let split = entry.0.split_at(prefix.len());
+            let key = K::read(&split.1).borrow().clone();
```
slowli: What's going on here? Shouldn't the return value be `Vec<K::Owned>`?

components/merkledb/src/views/address.rs
```diff
@@ -1,7 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 use std::{borrow::Cow, num::NonZeroU64};
 
 use crate::BinaryKey;
 
+/// Separator between the name and the additional bytes in family indexes.
+pub const INDEX_NAME_SEPARATOR: &[u8] = &[0];
```
slowli: There's `SEPARATOR_CHAR` several lines lower; please reuse it. Also, ideally the logic should be confined to this module (i.e., the visibility of the constant shouldn't change), but this is a nitpick.

components/merkledb/src/views/metadata.rs
```diff
@@ -337,6 +339,19 @@ impl<T: RawAccess> IndexesPool<T> {
         self.set_len(len + 1);
         (metadata, is_phantom)
     }
+
+    pub(super) fn suffixes<K: BinaryKey + Clone>(&self, prefix: &str) -> Vec<K> {
+        let prefix = concat_keys!(prefix, INDEX_NAME_SEPARATOR);
+        let mut keys = Vec::new();
+
+        for entry in self.0.iter::<Vec<u8>, Vec<u8>, Vec<u8>>(&prefix) {
```
slowli: Nit: IMO, `for (key, _) in ...` would read better.

components/merkledb/src/views/mod.rs
```diff
@@ -348,6 +348,30 @@ impl<T: RawAccessMut> View<T> {
     }
 }
 
+//TODO: Wrapper to prevent abstraction leak. Maybe there is better way.
+pub(crate) struct IndexesPoolWrapper<T: RawAccess> {
```
slowli: ~Abstractions shmabstractions~ `IndexesPool` is already made crate-private by the migration PR, so this looks redundant.

components/merkledb/src/group.rs
```diff
@@ -106,11 +107,31 @@ where
     }
 }
 
+impl<T, K, I> Group<T, K, I>
+where
+    T: RawAccess,
+    K: BinaryKey + Clone,
```
slowli: This is an incorrect bound; the correct one is `BinaryKey + ?Sized`.

components/merkledb/src/group.rs
```diff
@@ -106,11 +107,31 @@ where
     }
 }
 
+impl<T, K, I> Group<T, K, I>
+where
+    T: RawAccess,
+    K: BinaryKey + ?Sized,
+    I: FromAccess<T>,
+{
+    /// Return iterator over the keys from this group.
```
aleksuss:
```suggestion
    /// Returns iterator over the keys from this group.
```

components/merkledb/src/views/metadata.rs
```diff
@@ -163,7 +164,7 @@ impl Default for IndexType {
 /// See also `BinaryAttribute`.
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct IndexMetadata<V = Vec<u8>> {
-    identifier: u64,
+    pub identifier: u64,
```
aleksuss: Seems that visibility `pub(self)` is enough.

CHANGELOG.md
```diff
@@ -121,6 +121,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Added mechanism to extend block header. Block now contains
   key-value storage `additional_headers` which can contain binary data. (#1602)
+  
+- Added method `keys` to `Group` index which return all the keys from the group.
```
```suggestion
- Added method `keys_unchecked` to `Group` index which return all the keys from the group.
```

components/merkledb/src/group.rs
```diff
@@ -106,11 +107,31 @@ where
...
 
+impl<T, K, I> Group<T, K, I>
+where
+    T: RawAccess,
+    K: BinaryKey + ?Sized,
+    I: FromAccess<T>,
+{
+    /// Returns an iterator over the keys from this group.
+    ///
+    /// **Note:** use this method carefully, because storing all keys in memory may
```
popzxc: I'd say "avoid calling `collect` on the returned iterator, since ...", because obtaining iterator itself doesn't consume a large amount of RAM (or does it?).

components/merkledb/src/views/address.rs
```diff
@@ -1,7 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 use std::{borrow::Cow, num::NonZeroU64};
 
 use crate::BinaryKey;
 
+/// Separator between the name and the additional bytes in family indexes.
+pub const INDEX_NAME_SEPARATOR: &[u8] = &[SEPARATOR_CHAR];
```
popzxc: I suggest to have it `pub(super)`: despite the module being private, the strict visibility is easier to understand at first glance.

components/merkledb/src/views/metadata.rs
```diff
@@ -344,6 +345,19 @@ impl<T: RawAccess> IndexesPool<T> {
         self.set_len(len + 1);
         (metadata, is_phantom)
     }
+
+    pub(crate) fn suffixes<K: BinaryKey + ?Sized>(&self, prefix: &str) -> Vec<K::Owned> {
+        let prefix = concat_keys!(prefix, INDEX_NAME_SEPARATOR);
+        let mut keys = Vec::new();
+
+        for (key, _) in self.0.iter::<Vec<u8>, Vec<u8>, Vec<u8>>(&prefix) {
```
popzxc: Bikeshedding/discussion: `Vec<u8>, Vec<u8>, Vec<u8>` looks dirty, maybe provide type aliases for them within the function body?

components/merkledb/src/group.rs
pavel-mukhanov: We consume RAM before obtaining iterator when we call `suffixes` method.

popzxc: Then I think that the note should be expanded with explanation why exactly RAM is consumed, since it's not really obvious -- usually, creating an iterator is a cheap operation.

components/merkledb/src/views/metadata.rs
pavel-mukhanov: It's possible to remove first Vec<u8>, maybe it won't be so dirty.

CHANGELOG.md
```diff
@@ -121,6 +121,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Added mechanism to extend block header. Block now contains
   key-value storage `additional_headers` which can contain binary data. (#1602)
+  
+- Added method `keys_unchecked` to `Group` index which return all the keys
+from the group. (#1614)
```
```suggestion
  from the group. (#1614)
```

components/merkledb/src/indexes/group.rs
```diff
@@ -106,11 +107,32 @@ where
...
+    K: BinaryKey + ?Sized,
+    I: FromAccess<T>,
+{
+    /// Returns an iterator over the keys from this group.
+    ///
+    /// **Note:** use this method carefully, because it obtains all available keys for this
+    /// group in `suffixes` method and stores them in `Vec`. This operation can consume a large
+    /// amount of RAM.
+    ///
+    /// **Note:** creating any new indexes while this iterator is in scope may lead to panic.
```
slowli: Sorry for obtuse remarks: *This* implementation cannot panic in the described scenario since it does not borrow the metadata view (unlike the implementation in #1637, which I admittedly specified in a very vague manner).

Because this implementation still has memory overhead, I'd suggest to use the name `buffered_keys` (or `keys_buffered`) and return a `Vec<K::Owned>` directly.

components/merkledb/src/views/metadata.rs
```diff
@@ -344,6 +345,19 @@ impl<T: RawAccess> IndexesPool<T> {
         self.set_len(len + 1);
         (metadata, is_phantom)
     }
+
+    pub(crate) fn suffixes<K: BinaryKey + ?Sized>(&self, prefix: &str) -> Vec<K::Owned> {
+        let prefix = concat_keys!(prefix, INDEX_NAME_SEPARATOR);
```
slowli: Nit: I'd rather have the prefix logic encapsulated in the `views::address` module, as in #1637.

components/merkledb/src/views/metadata.rs
```diff
@@ -344,6 +345,19 @@ impl<T: RawAccess> IndexesPool<T> {
...
+        let mut keys = Vec::new();
+
+        for (key, _) in self.0.iter::<_, Vec<u8>, Vec<u8>>(&prefix) {
+            let split = key.split_at(prefix.len());
```
slowli: Why not use `&key[prefix.len()..]` here?

components/merkledb/src/views/metadata.rs
```diff
@@ -344,6 +345,19 @@ impl<T: RawAccess> IndexesPool<T> {
         self.set_len(len + 1);
         (metadata, is_phantom)
     }
+
+    pub(crate) fn suffixes<K: BinaryKey + ?Sized>(&self, prefix: &str) -> Vec<K::Owned> {
+        let prefix = concat_keys!(prefix, INDEX_NAME_SEPARATOR);
+        let mut keys = Vec::new();
+
+        for (key, _) in self.0.iter::<_, Vec<u8>, Vec<u8>>(&prefix) {
```
slowli: Using `bytes_iter` instead of `iter` here would save `Vec` allocations. (Don't know whether the compiler is smart enough to remove them, but I wouldn't rely on this fact.)

components/merkledb/src/indexes/group.rs
```diff
@@ -106,11 +107,32 @@ where
...
+    pub fn keys_unchecked(&self) -> impl Iterator<Item = K::Owned> {
+        let indexes_pool = IndexesPool::new(self.access.clone());
+        indexes_pool.suffixes::<K>(self.prefix.name()).into_iter()
```
slowli: This is incorrect for groups in which `prefix` has an `id_in_group` component. Such groups are impossible to create via `AccessExt`, but are possible to create via `FromAccess` (including via derive macro). You should either provide a generic implementation (as in #1637), or check if the group's prefix has `id_in_group` and panic in this case (or otherwise handle it).

components/merkledb/src/views/metadata.rs
```diff
@@ -346,13 +345,13 @@ impl<T: RawAccess> IndexesPool<T> {
...
-        let prefix = concat_keys!(prefix, INDEX_NAME_SEPARATOR);
+    pub(crate) fn suffixes<K: BinaryKey + ?Sized>(&self, prefix: &IndexAddress) -> Vec<K::Owned> {
+        let prefix = prefix.qualified_prefix();
         let mut keys = Vec::new();
+        let mut bytes_iter = self.0.iter_bytes(&prefix);
 
-        for (key, _) in self.0.iter::<_, Vec<u8>, Vec<u8>>(&prefix) {
-            let split = key.split_at(prefix.len());
-            let suffix = K::read(&split.1);
+        while let Some(key) = bytes_iter.next() {
```
slowli: `bytes_iter` doesn't stop when the generated keys no longer start with `prefix`, so this must be checked manually:

```rust
while let Some((key, _)) = bytes_iter.next() {
    if !key.starts_with(&prefix) {
        break;
    }
    // ...
}
```

Note that this leads to incorrect values returned from `buffered_keys`, possible panics (if one of the following keys is shorter than `prefix`) and undefined behavior (if the key type is `str` and a successive key is not valid UTF-8). Unfortunately, the tests as written don't capture this; IMO, they should capture all these situations.

components/merkledb/src/indexes/group.rs
```diff
@@ -119,10 +119,18 @@ where
     /// group in `suffixes` method and stores them in `Vec`. This operation can consume a large
     /// amount of RAM.
     ///
-    /// **Note:** creating any new indexes while this iterator is in scope may lead to panic.
-    pub fn keys_unchecked(&self) -> impl Iterator<Item = K::Owned> {
+    /// # Panics
+    ///
+    /// Panics if group prefix contains `id_in_group` component. Such group can be created via
+    /// `FromAccess` trait.
+    pub fn buffered_keys(&self) -> Vec<K::Owned> {
```
slowli: Another thing: I think this method works incorrectly for `Prefixed` / `Migration` accesses. Unlike with `id_in_group`, we cannot afford to not support them, since the main use case for group enumeration is migration. Please add the corresponding tests.

aleksuss:
Closed in favour #1662

PR #1615, opened 2019-12-13
Fix websocket issue during different kind of WebSocket connect

sfffaaa:
## Overview

WebSocket could not receive any information if different kinds of WebSockets connected.
For example:
1. The block WebSocket is connected.
2. The tx WebSocket is connected.
3. The client of the block WebSocket cannot receive any block data.

The root cause is the ApiNodeState only has one address for WebSocket server's address, so I make the broadcast_server_address as a vector.
BTW, I didn't clean the address vector because the behavior of "shutdown_broadcast_server" only occurs during the node stops.

### DIFF
--- a/exonum/src/api/node/mod.rs
+++ b/exonum/src/api/node/mod.rs
@@ -45,7 +45,7 @@ struct ApiNodeState {
     node_role: NodeRole,
     majority_count: usize,
     validators: Vec<ValidatorKeys>,
-    broadcast_server_address: Option<Addr<websocket::Server>>,
+    broadcast_server_addresses: Vec<Option<Addr<websocket::Server>>>,
     tx_cache_len: usize,
 }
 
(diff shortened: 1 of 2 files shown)


popzxc:
Thanks for the contribution!
Implementation looks good. :+1: 

Can you please add a test for several parallel clients receiving messages?

exonum/src/api/node/mod.rs
```diff
@@ -45,7 +45,7 @@ struct ApiNodeState {
     node_role: NodeRole,
     majority_count: usize,
     validators: Vec<ValidatorKeys>,
-    broadcast_server_address: Option<Addr<websocket::Server>>,
+    broadcast_server_addresses: Vec<Option<Addr<websocket::Server>>>,
```
alekseysidorov: Hmm, It seems that `Option` became redundant in this context.

sfffaaa: Sure, Thanks for your reviewing. I'll fix it on Tuesday or Wednesday.

exonum/src/api/node/mod.rs
```diff
@@ -247,8 +247,8 @@ impl SharedNodeState {
 
     pub(crate) fn shutdown_broadcast_server(&self) {
         let state = self.node.read().expect("Expected read lock");
-        for address in state.broadcast_server_addresses.iter() {
-            address.do_send(websocket::Terminate);
+        for server in state.broadcast_server_addresses.iter() {
+            server.do_send(websocket::Terminate);
```
aleksuss: Maybe `connection` ? What do you thinks? IMO, connection is more appropriate name here.

sfffaaa: I don't have a preference for the naming. I follow the original name, server, before the refinement. However, I also can change it to connection. These two names, server or connection, are fine to me both.

PR #1618, opened 2019-12-14
Proofread runtime docs

Maria-Nosyk:
## Overview

---

### Definition of Done

### DIFF
--- a/exonum/src/runtime/mod.rs
+++ b/exonum/src/runtime/mod.rs
@@ -12,85 +12,87 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Common building blocks for creating runtimes for the Exonum blockchain.
+//! Common building blocks that compose runtimes for the Exonum blockchain.
 //!
-//! Each runtime contains specific services to execute transactions, process events,
+//! Each runtime contains specific services that execute transactions, process events,
 //! provide user APIs, etc. A unified dispatcher redirects all the calls
-//! and requests to the appropriate runtime environment. Thus, a blockchain interacts with the
+//! and requests to an appropriate runtime environment. Thus, a blockchain interacts with the
 //! dispatcher, and not with specific runtime instances.
 //!
 //! # Artifacts
 //!
-//! Each runtime has its own [artifacts] registry from which users can create a service.
-//! The artifact identifier is required by the runtime to construct service instances.
+//! Each runtime has its own [artifacts] registry. Users can create services from the stored artifacts.
+//! An artifact identifier is required by the runtime to construct service instances.
 //! In other words, an artifact identifier is similar to a class name, and a specific
 //! service instance - to a class instance. A single artifact may be used to instantiate
 //! zero or more services.
 //!
-//! The format of an artifact ID is uniform across runtimes (it is essentially a string),
-//! but the runtime may customize artifact deployment via runtime-specific deployment arguments.
+//! The format of the artifact ID is uniform across runtimes - it is essentially a string.
+//! But the runtime may customize artifact deployment via runtime-specific deployment arguments.
 //!
 //! # Artifact Lifecycle
 //!
-//! 1. An artifact is assembled in a way specific to the runtime. For example, the artifact may
+//! 1. An artifact is assembled in a way specific to the runtime. For example, an artifact may
 //!   be compiled from sources and packaged using an automated build system.
 //!
 //! 2. The artifact with the service is deployed on the blockchain. The decision to deploy the
 //!   artifact and the deployment spec are usually performed by the blockchain administrators.
-//!   The corresponding logic is customizable via a [supervisor service](#supervisor-service).
+//!   The corresponding logic is customizable via the [supervisor service](#supervisor-service).
 //!   What deployment entails depends on the runtime; e.g., the artifact may be downloaded
 //!   by each Exonum node, verified for integrity and then added into the execution environment.
 //!
-//! 3. For each node, an artifact may be deployed either asynchronously or synchronously /
+//! 3. For each node, an artifact may be deployed either asynchronously or synchronously, that is
 //!   in a blocking manner. The supervisor usually first commands a node to deploy the artifact
-//!   asynchronously via [`Mailbox`] once the decision to start deployment is reached
-//!   by the blockchain administrators. Async deployment speed and outcome may differ among nodes.
+//!   asynchronously via [`Mailbox`], once the decision to start deployment is reached
+//!   by the blockchain administrators. Asynchronous deployment speed and outcome may differ among nodes.
 //!
-//! 4. The supervisor translates local deployment outcomes into a consensus-agreed result.
+//! 4. The supervisor translates the local deployment outcomes into a consensus-agreed result.
 //!   For example, the supervisor may collect confirmations from the validator nodes that have
-//!   successfully deployed the artifact, and once all the validator nodes have sent
-//!   their confirmations, the artifact is *committed*. Being part of the service logic,
+//!   successfully deployed the artifact. Once all the validator nodes have sent
+//!   their confirmations, the artifact is *committed*. As a part of the service logic,
 //!   artifact commitment is completely deterministic, agreed via consensus,
 //!   and occurs at the same blockchain height for all nodes in the network.
 //!
-//! 5. Once the artifact is committed, every node in the network is required to have it deployed
+//! 5. Once the artifact is committed, every node in the network must have it deployed
 //!   in order to continue functioning.
-//!   If a node has not deployed the artifact previously, deployment becomes blocking; the node
+//!   If a node has not deployed the artifact previously, deployment becomes blocking. The node
 //!   does not participate in consensus or block processing until the deployment is completed
 //!   successfully. If the deployment is unsuccessful, the node stops indefinitely.
-//!   Due to deployment confirmation mechanics built into the supervisor, it is reasonable
+//!   The deployment confirmation mechanics is built into the supervisor. Thus, it is reasonable
 //!   to assume that a deployment failure at this stage is local to the node and
 //!   could be fixed by the node admin.
 //!
 //! # Service Lifecycle
 //!
-//! 1. Once the artifact is committed, it is possible to instantiate a corresponding service.
-//!   Each instantiation request contains an ID of a previously deployed artifact,
+//! 1. Once the artifact is committed, it is possible to instantiate the corresponding service.
+//!   Each instantiation request contains an ID of the previously deployed artifact,
 //!   a string instance ID, and instantiation arguments in a binary encoding
-//!   (by convention, Protobuf). As with the artifacts, the logic controlling instantiation
+//!   (by convention, Protobuf). As with the artifacts, the logic that controls instantiation
 //!   is encapsulated in the supervisor service.
 //!
-//! 2. During instantiation, the service is assigned a numeric ID, which is used to reference
+//! 2. During instantiation the service gets a numeric ID, which is used to reference
 //!   the service in transactions. The runtime can execute initialization logic defined
 //!   in the service artifact; e.g., the service may store some initial data in the storage,
-//!   check service dependencies, etc. The service (or the enclosing runtime) may signal that
-//!   the initialization failed, in which case the service is considered not instantiated.
+//!   check service dependencies, etc. If the service (or the enclosing runtime) signals that
+//!   the initialization failed, the service is considered not instantiated.
 //!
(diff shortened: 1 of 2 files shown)


exonum/src/runtime/mod.rs
```diff
@@ -210,13 +211,13 @@ impl From<RuntimeIdentifier> for u32 {
...
 ///
-/// The ordering for the "readonly" method `is_artifact_deployed` in relation
+/// The ordering for the "read-only" method `is_artifact_deployed` in relation
 /// to the lifecycle above is not specified.
 ///
 /// # Consensus and Local Methods
 ///
-/// The following methods should return the same result given the same arguments for all nodes
-/// in the blockchain network:
+/// The following methods should return the same result if they contain the same arguments for all
```
popzxc: This phrasing looks a bit weird: methods do not contain arguments, but arguments are provided into methods (to obtain some result).

exonum/src/runtime/mod.rs
```diff
@@ -227,85 +228,93 @@ impl From<RuntimeIdentifier> for u32 {
...
-    /// node restart, including restoring the deployed artifacts / started service instances
-    /// for all runtimes.
+    /// Notifies the runtime that the dispatcher has completed re-initialization after the
+    /// node restart. Re-initialization includes restoring the deployed artifacts / started service instances
+    /// for all the runtimes.
     ///
-    /// This method is called *no more than once* during `Runtime` lifetime. It is called iff
-    /// the blockchain had genesis block when the node was started. The blockchain state
-    /// is guaranteed to not change between `initialize` and `on_resume` calls.
+    /// This method is called *maximum once* during the `Runtime` lifetime. To call the method,
```
popzxc: IMO the new variant slightly shifts the main point: original line has a strong accent on the conditions, under which this method **will** be invoked, but the new line describes conditions in which this method **may** be invoked. Maybe something like "Blockchain will invoke the method iff it has a genesis block when the node is started"?

exonum/src/runtime/mod.rs
```diff
@@ -227,85 +228,93 @@ impl From<RuntimeIdentifier> for u32 {
...
-    /// long time).
-    /// Alternatively, "garbage" services may be removed from `Runtime` in `after_commit`
+    /// The constructor runs *exactly once* during the blockchain lifetime for each successfully
+    /// initialized service instance. That is to say, the constructor is *not* called on a node restart.
+    ///
+    /// At the same time, when `start_adding_service` is called,
+    /// there is no guarantee that the service will eventually get to the blockchain via `commit_service`.
+    /// If the consensus accepts an alternative block proposal without the service inside,
+    /// commit of the service will fail.
+
```
popzxc: Empty like here seems to be incorrect.

exonum/src/runtime/mod.rs
```diff
@@ -317,38 +326,41 @@ pub trait Runtime: Send + fmt::Debug + 'static {
...
+    /// This method is called *once* for a specific service instance during the `Runtime` lifetime:
     ///
     /// - For newly added instances, the method is called when the fork with the corresponding
     ///   `start_adding_service()` call is committed.
     /// - After a node restart, the method is called for all existing service instances.
     ///
-    /// It is guaranteed that `start_adding_service()` was called with the same `spec` earlier
-    /// and returned `Ok(())`. The results of the call (i.e., changes to the blockchain state)
-    /// are guaranteed to be persisted from the call.
+    /// Calling this method guarantees that `start_adding_service()` has been called with the same `spec`
```
popzxc: Maybe replace "calling" with "invocation" to avoid weird "Calling guarantees that ... has been called"?

exonum/src/runtime/mod.rs
```diff
@@ -12,109 +12,110 @@
...
 //! provide user APIs, etc. A unified dispatcher redirects all the calls
-//! and requests to the appropriate runtime environment. Thus, a blockchain interacts with the
+//! and requests to an appropriate runtime environment. Thus, a blockchain interacts with the
 //! dispatcher, and not with specific runtime instances.
 //!
 //! # Artifacts
 //!
-//! Each runtime has its own [artifacts] registry from which users can create a service.
-//! The artifact identifier is required by the runtime to construct service instances.
+//! Each runtime has its own [artifacts] registry. Users can create services from the stored artifacts.
```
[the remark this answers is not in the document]
aleksuss: Even less. **80**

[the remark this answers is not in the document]
alekseysidorov: I think that 80 is too short for the wide screens era.

exonum/src/runtime/mod.rs
```diff
@@ -12,109 +12,110 @@
...
 //!
 //! 3. Once the service is instantiated, it can process transactions and interact with the
 //!   external users in other ways. Different services instantiated from the same artifact
 //!   are independent and have separate blockchain storages. Users can distinguish services
 //!   by their IDs; both numeric and string IDs are unique within a blockchain.
 //!
-//! Warning: state hash of the block is not affected by services, instantiated within it. The
-//! only exception is the genesis block where object hashes of builtin services are included
-//! to the resulting block state hash. This behavior difference will be unified in next release.
+//! Warning: the services, instantiated within a block, do not affect the state hash of that block. The
```
slowli: This doesn't seem true any more with the automated state hash aggregation. @alekseysidorov, could you confirm? If so, the paragraph should be removed.

exonum/src/runtime/mod.rs
```diff
@@ -12,109 +12,110 @@
...
 //! # Supervisor Service
 //!
 //! A supervisor service is a service that has additional privileges. This service
-//! allows to deploy artifacts and instantiate new services after the blockchain is launched
+//! allows deploying artifacts and instantiating new services after the blockchain is launched
 //! and running. Other than that, it looks like an ordinary service.
-//! The supervisor should be present during the blockchain start, otherwise no new artifacts / services
-//! could ever be added to the blockchain.
+//! To enable adding new artifacts / services to the blockchain after its start, the supervisor
+//! must be present in the system binary file during the blockchain start.
```
slowli: I suppose this is technically correct with the current supervisor implementation, but AFAIK there are no obstacles preventing a non-Rust supervisor implementation. Thus, I'd suggest to reform to

> [...] the supervisor must be one of the builtin service instances.

Possibly, it's worth it specifying above what a builtin service instance is.

exonum/src/runtime/mod.rs
```diff
@@ -227,85 +228,93 @@ impl From<RuntimeIdentifier> for u32 {
...
-    /// The runtime should discard the instantiated service instance after completing this method,
-    /// unless there are compelling reasons to retain it (e.g., creating an instance takes very
-    /// long time).
-    /// Alternatively, "garbage" services may be removed from `Runtime` in `after_commit`
+    /// The constructor runs *exactly once* during the blockchain lifetime for each successfully
+    /// initialized service instance. That is to say, the constructor is *not* called on a node restart.
+    ///
+    /// At the same time, when `start_adding_service` is called,
+    /// there is no guarantee that the service will eventually get to the blockchain via `commit_service`.
+    /// If the consensus accepts an alternative block proposal without the service inside,
```
slowli: Not sure that service being "inside a proposal" is the best wording. IMO, something like this would be better:

> The consensus may accept an alternative block proposal, in which the service is not instantiated or instantiated with different parameters.

exonum/src/runtime/mod.rs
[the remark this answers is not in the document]
aleksuss: Markdown linter has a limit in 80 characters

exonum/src/runtime/rust/mod.rs
```diff
@@ -450,7 +450,7 @@ impl Runtime for RustRuntime {
         self.blockchain = Some(blockchain.clone());
     }
 
-    // We need to propagate changes in the services immediately after initialization.
+    // Propagate changes in the services immediately after initialization.
```
aleksuss:
```suggestion
    // Propagates changes in the services immediately after initialization.
```
?

exonum/src/runtime/mod.rs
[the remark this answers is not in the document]
popzxc: But it's not a markdown, it's a doc-comment.

exonum/src/runtime/mod.rs
```diff
@@ -227,85 +228,93 @@ impl From<RuntimeIdentifier> for u32 {
...
     /// because of the time dependence between `commit_service` and `after_commit` described above.
-    /// The runtime should commit long-term resources for the service after a `commit_service()` call.
-    /// Since discarded instances persist their state in a discarded fork, no further action
-    /// is required to remove this state.
     ///
-    /// # Return value
+    /// According to the principle of the consensus algorithm, the process of forming a new block
+    /// takes place in RAM. A patch with changes is then applied to the database. If a block proposal
+    /// is not accepted, a fork with the patch containing this proposal is discarded from RAM.
+    /// The service instances, already discarded by the runtime, persist their state only in the
```
slowli: It does not matter whether the service instance is discarded by the runtime - the service state is placed within the patch anyway (this is not something a runtime can influence). Thus, I'd remove "already discarded by the runtime". I'd also remove the preceding 3 sentences because they are too generic / don't really concern the documented method.

exonum/src/runtime/mod.rs
```diff
@@ -388,63 +400,61 @@ pub trait Runtime: Send + fmt::Debug + 'static {
...
+    /// This method is called *after* all `commit_service` calls related
     /// to the same block. The method is called exactly once for each block in the blockchain,
     /// including the genesis block.
     ///
-    /// A block is not yet persisted when this method is called; the up to date block information
-    /// is provided in the `snapshot`. It corresponds exactly to the information
-    /// eventually persisted; i.e., no modifying operations are performed on the block.
+    /// A block is not yet persisted when this method is called. The `snapshot` provides an up-to-date
+    /// block information. It corresponds exactly to the information
+    /// eventually persisted. This means that no modifying operations are performed on the block.
```
slowli: Nit: Since the last sentence rewords the penultimate one, it could be safely removed.

exonum/src/runtime/rust/mod.rs
```diff
@@ -14,11 +14,11 @@
 
 //! The current runtime is for running native services written in Rust.
 //!
-//! A set of artifacts available to deploy in the Rust runtime is static. The set is defined at the time
-//! of compilation. Once created, the set can be changed only by the node binary recompilation.
+//! In the Rust runtime a set of service artifacts that you want to deploy is static. The set is defined at the time
```
slowli: Nit: IMO, "you may want to deploy" would look more accurate. During blockchain initialization, a user may not instantiate services from all available artifacts; the corresponding services may be instantiated later.

exonum/src/runtime/rust/mod.rs
```diff
@@ -61,10 +61,10 @@
 //!     WalletAlreadyExists = 1,
 //! }
 //!
-//! // Define the transaction interface for your service by creating a trait with
+//! // Define a transaction interface for your service by creating a `Transactions` trait with
 //! // the following attribute and method signatures.
-//! // This attribute implements `Interface` trait for this trait and `Transaction`
-//! // trait for each argument.
+//! // This attribute implements the `Interface` trait for the `Transactions` trait and a `Transaction`
```
slowli: Nit: This was already in the text, but "implementing a trait for trait" looks super-weird from the Rust perspective. (In reality, `Interface` is implemented for a trait *object* based on the `Transactions` trait - `impl Interface for dyn Transactions`.)

exonum/src/runtime/mod.rs
alekseysidorov: Yes this line is outdated and even more incorrect at the moment.

exonum/src/runtime/mod.rs
```diff
@@ -210,13 +207,13 @@ impl From<RuntimeIdentifier> for u32 {
...
 ///
-/// The ordering for the "readonly" method `is_artifact_deployed` in relation
+/// The ordering for the "read-only" method `is_artifact_deployed` in relation
 /// to the lifecycle above is not specified.
 ///
 /// # Consensus and Local Methods
 ///
-/// The following methods should return the same result given the same arguments for all nodes
-/// in the blockchain network:
+/// The following methods should return the same result if provided arguments the same for all
```
popzxc:
```suggestion
/// The following methods should return the same result if provided arguments are the same for all
```

exonum/src/runtime/mod.rs
popzxc: @aleksuss this suggestion still holds

PR #1619, opened 2019-12-16
Fix services behaviour on the genesis block [ECR-3945]

popzxc:
This PR fixes the following bugs:

- Builtin services weren't actually activated immediately in the genesis block;
- Consequence: `after_transactions` hook for builtin services was not invoked;
- Errors in the `after_transactions` hook didn't lead to the blockchain creation failure;
- `Supervisor` service was using `blockchain::Schema::height` in `after_transactions` hook despite being always builtin.


exonum/src/runtime/rust/macros.rs
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Macros useful for developing `rust` services.
+
+/// Skips the invocation of the service hook for the genesis block.
+/// This macro is intended to be used within `rust` service hooks.
+#[macro_export]
+macro_rules! skip_for_genesis {
```
slowli: Bikeshedding: I don't think that encapsulating this check into a macro is warranted. It could make more sense to introduce a checking method for `BlockchainData` or `CallContext`, like this:

```rust
fn after_transactions(&self, ctx: CallContext<'_>) -> Result<(), ExecutionError> {
    if ctx.in_genesis_block() {
        return Ok(());
    }
}
```

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -117,9 +117,20 @@ impl Dispatcher {
...
+        // Mark service as active.
+        self.activate_pending(fork);
+        // Remove service from pending.
+        Schema::new(fork as &Fork).take_pending_instances();
+        // Flush changes, so the snapshot will be created with all the recently added changes.
+        // We can do it safely, since any failure on the genesis block level will result in the
+        // overall blockchain creation failure.
+        fork.flush();
+        // Start service within runtime.
+        self.start_service((fork as &Fork).snapshot(), &spec)
```
slowli: This is very much a hack relying on the peculiarities of `impl RawAccess for Fork`. I'd much prefer if all services were activated at once and based on a `Patch` - like it was before, with the exception that `after_transactions` would be called after that.

exonum/src/blockchain/tests.rs
```diff
@@ -362,7 +390,117 @@ fn create_blockchain(
...
+#[test]
+fn after_transactions_failure_causes_genesis_failure() {
+    let blockchain_result = maybe_create_blockchain(
+        vec![ServiceGenesisPanicImpl.into()],
+        vec![ServiceGenesisPanicImpl
+            .artifact_id()
+            .into_default_instance(TEST_SERVICE_ID, TEST_SERVICE_NAME)],
+    );
+
+    const EXPECTED_ERR_TEXT: &str =
```
slowli: Ouch. Would it make sense to check that the error contains `"after_commit failed for at least one service"` instead?

exonum/src/blockchain/tests.rs
```diff
@@ -362,7 +390,117 @@ fn create_blockchain(
...
+    );
+
+    // After creation of the genesis block, check that value was set.
+    let snapshot = blockchain.snapshot();
+    let index = snapshot.get_list("service_good.val");
+    assert_eq!(index.len(), 1);
+    assert_eq!(index.get(0), Some(1));
+}
+
+/// Checks that if `after_transactions` fails on the genesis level,
```
slowli: Nit: I'd reword "on the genesis level" to "on the genesis block" here and elsewhere in order to unify the used terms.

exonum/src/runtime/rust/service.rs
```diff
@@ -92,7 +92,20 @@ pub trait Service: ServiceDispatcher + Debug + 'static {
     /// act similarly on different nodes. In other words, the service should only use data available
     /// in the provided `CallContext`.
     ///
+    /// Note that if service was added within genesis block, it will be activated **immediately** and
+    /// thus `after_transactions` will be invoked for such a service after the genesis block creation.
+    /// If you aren't interested in the invocation for the genesis block, consider using
+    /// [`skip_for_genesis`] macro.
+    ///
+    /// Also note that invocation of [`blockchain::Schema::height`] will **panic** if invoked within
+    /// `after_transactions` of the genesis block. If you aren't going to skip the invocation for genesis
```
slowli: Nit: The *invocation* on the genesis block is out of the control of the service; it can influence whether it's *processed* the same way on the genesis block.

exonum/src/runtime/rust/service.rs
```diff
@@ -92,7 +92,20 @@ pub trait Service: ServiceDispatcher + Debug + 'static {
...
+    /// block and need to know current height, use [`blockchain::Schema::next_height`] instead.
```
slowli: Nit: Awkward wording; `next_height` does not return the *current* height. I'd replace with

> If you aren't going to skip the handler for the genesis block, use `blockchain::Schema::next_height` to infer the current blockchain height.

exonum/src/runtime/rust/tests.rs
```diff
@@ -198,7 +198,13 @@ impl<T: Runtime> Runtime for Inspected<T> {
         context: ExecutionContext<'_>,
         instance_id: u32,
     ) -> Result<(), ExecutionError> {
-        let height = CoreSchema::new(&*context.fork).height();
+        let schema = CoreSchema::new(&*context.fork);
+        // Skip genesis block execution.
```
slowli: Nit: This can be more accurately worded as "Skip adding events during genesis block execution". The `inner` runtime is notified about the call in any case.

services/supervisor/src/lib.rs
```diff
@@ -299,11 +300,13 @@ impl Service for Supervisor {
     }
 
     fn after_transactions(&self, mut context: CallContext<'_>) -> Result<(), ExecutionError> {
+        skip_for_genesis!(context);
+
         let mut schema = Schema::new(context.service_data());
         let configuration = schema.supervisor_config();
         let core_schema = context.data().for_core();
-        let validator_count = core_schema.consensus_config().validator_keys.len();
         let height = core_schema.height();
```
slowli: It seems that `height` is used only as `height.next()` further down in the code. Thus, could it make sense to assign `let height = core_schema.next_height()` and *not* skip the genesis block?

exonum/src/blockchain/tests.rs
```diff
@@ -491,16 +492,18 @@ fn after_transactions_failure_causes_genesis_failure() {
...
-         [(AfterTransactions { id: 0 }, ExecutionError { kind: Unexpected, \
-         description: \"42\", runtime_id: Some(0), call_site: Some(CallSite \
-         { instance_id: 0, call_type: AfterTransactions }) })]";
+    const EXPECTED_ERR_TEXT: &str = "`after_transactions` failed for at least one service";
     let actual_err = blockchain_result
         .expect_err("Blockchain shouldn't be created after failure within genesis block");
 
     // Unfortunately, `failure::Error` doesn't implement `PartialEq`, so we have to string-compare them.
-    assert_eq!(&format!("{}", actual_err), EXPECTED_ERR_TEXT);
+    let error_string = format!("{}", actual_err);
```
slowli: Nit: It's possible to use `ToString` here:

```rust
assert!(error.to_string().contains(EXPECTED_ERR_TEXT));
```

exonum/src/blockchain/tests.rs
```diff
@@ -491,16 +492,18 @@ fn after_transactions_failure_causes_genesis_failure() {
...
+
+    assert_eq!(
```
slowli: Why is `assert_eq!` used instead of `assert!` here?

exonum/src/runtime/rust/service.rs
```diff
@@ -94,18 +94,19 @@ pub trait Service: ServiceDispatcher + Debug + 'static {
     ///
     /// Note that if service was added within genesis block, it will be activated **immediately** and
     /// thus `after_transactions` will be invoked for such a service after the genesis block creation.
-    /// If you aren't interested in the invocation for the genesis block, consider using
-    /// [`skip_for_genesis`] macro.
+    /// If you aren't interested in the processing of for the genesis block, you can use
+    /// [`CallContext::in_genesis_block`] method and exit early for `true` return value.
```
slowli: > for `true` return value

Nit: I'd rephrase as: "exit early if `true` is returned".

exonum/src/runtime/rust/service.rs
```diff
@@ -94,18 +94,19 @@ pub trait Service: ServiceDispatcher + Debug + 'static {
...
+    /// `after_transactions` of the genesis block. If you are going to process the genesis
+    /// block and need to know current height, use [`blockchain::Schema::next_height`] to infer the
+    /// current blockchain height.
     ///
     /// Services should not rely on a particular ordering of `Service::after_transactions` invocations.
     ///
     /// [`blockchain::Schema::height`]: ../../blockchain/schema/struct.Schema.html#method.height
     /// [`blockchain::Schema::height`]: ../../blockchain/schema/struct.Schema.html#method.next_height
-    /// [`skip_for_genesis`]: ../../macro.skip_for_genesis.html
+    /// [`CallContext::in_genesis_block`]: call_context/struct.CallContext.html#method.in_genesis_block
```
slowli: Isn't `call_context` module private?

exonum/src/runtime/rust/tests.rs
```diff
@@ -199,7 +199,7 @@ impl<T: Runtime> Runtime for Inspected<T> {
         instance_id: u32,
     ) -> Result<(), ExecutionError> {
         let schema = CoreSchema::new(&*context.fork);
-        // Skip genesis block execution.
+        // Skip adding events during genesis block execution.
```
slowli: Sorry that didn't point out this before, but checking whether `after_transactions` is run twice on the genesis block could be useful. Would it be difficult to refactor this method to record `schema.next_height()` instead of `schema.height()`?

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - State hash aggregation is now performed automatically by MerkleDB.
   The relevant methods in `Runtime` and `Service` in Rust runtime
   have been removed. (#1553)
+- `after_transactions` hook is now invoked on the genesis block for the builtin
+  services. Note that calling `blockchain::Schema::height` within `after_transactions`
+  hook may cause a panic for a builtin service. (#1619)
```
slowli: Nit: "~may~ will cause a panic"; this isn't probabilistic.

CHANGELOG.md
```diff
@@ -37,6 +40,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `before_transactions` hook for services was introduced. (#1577)
 - `ErrorMatch` was introduced to test (e.g., using the testkit) that
   an `ExecutionError` has an expected type, error message and/or location. (#1585)
+- New `blockchain::Schema` method `next_height` was added as a non-panicking
+  alternative to `height`. (#1619)
+- New macro `skip_for_genesis` was added for `rust` services to skip service hook
```
slowli: This is outdated, isn't it?

exonum/src/runtime/mod.rs
```diff
@@ -412,6 +412,8 @@ pub trait Runtime: Send + fmt::Debug + 'static {
     /// `after_transactions` is called for every service active at the beginning of the block
     /// (i.e., services instantiated within the block do **not** receive a call) exactly
     /// once for each block.
+    /// Note that services instantiated within genesis block are activated **immediately** and
+    /// thus `after_transactions` is invoked for them.
```
slowli: Nit: I'd remove "Note that" and specify "[...] invoked for them in the genesis block".

exonum/src/runtime/rust/tests.rs
popzxc: Yeah, that looks like a good idea. As an addition, I made **all** the hooks use `next_height`, because
- It's better to have the same timescale everywhere;
- It allowed us to replace `Option<Height>` with `Height` for `CommitService` event.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -117,12 +117,28 @@ impl Dispatcher {
         spec: InstanceSpec,
         constructor: Vec<u8>,
     ) -> Result<(), ExecutionError> {
-        // Start the built-in service instance.
+        // Add the service instance to the service.
```
aleksuss: Looks a bit weird

PR #1625, opened 2019-12-17
Improve error message for failed deploy of Rust service [ECR-3973]

Kormick:
## Overview

Improved error message for failed service deploy:
- Renamed code to local_error_code in ErrorKind
- Added more detailed description for UnableToDeploy error


exonum/benches/criterion/block.rs
```diff
@@ -298,7 +298,9 @@ mod cryptocurrency {
             if arg.seed % 2 == 0 {
                 Ok(())
             } else {
-                let error_kind = ErrorKind::Service { code: 15 };
+                let error_kind = ErrorKind::Service {
+                    local_error_code: 15,
```
popzxc: To avoid misunderstanding: in [ECR-3973] it is written that `ErrorKind::Runtime { code: 1 }` can be confused with `runtime_id`, so renaming is going to make errors more friendly to users.

exonum/src/runtime/error/execution_result.rs
```diff
@@ -31,17 +31,23 @@ pub(super) struct ExecutionStatus {
...
+                ErrorKind::Service { local_error_code } => {
+                    (ExecutionType::ServiceError, Some(local_error_code))
+                }
             };
 
             ExecutionStatus {
                 typ,
                 description: err.description.clone(),
-                code,
+                code: local_error_code,
```
popzxc: Hm... I think it's better to unify everything to have the same terminology.

exonum/src/runtime/error.rs
```diff
@@ -456,6 +456,16 @@ impl ExecutionError {
         Self::new(ErrorKind::Service { code }, description)
     }
 
+    /// Creates a runtime error for use in runtime code.
+    pub fn runtime(runtime_id: u32, code: u8, description: impl Into<String>) -> Self {
```
vitvakatu: Funny, I had the same issue. You don't need this method, you can create runtime error using `ExecutionFail::with_description` method: 
`rust::Error::UnableToDeploy.with_description(description).into()`

@slowli perhaps the docs aren't clear enough, as even two developers have already stuck with it?

popzxc: I think asymmetry with `ExecutionError::service` does the thing here.

vitvakatu: Yep, I agree, I think that was the main reason for me. Sorry for requesting changes @Kormick , I pressed wrong button actually. I'm pretty okay with adding this method (and I've actually created a task to do so, but closed it when found another solution).

vitvakatu: Though using `ExecutionFail` trait is a bit better because the interface is much safer in this case - you're not forced to use numeric IDs (even if you convert them from enum)

Kormick: Can we provide runtime id with it?

vitvakatu: Good catch, actually not. 🤔 With `runtime` method, I don't see much sense in using `ExecutionFail` for runtime errors, because having runtime ID is useful. Which means we need either to add runtime ID to `ExecutionFail` somehow, or simplify `ExecutionFail` derive macros and use it for service errors only.
Please create a task for whatever decision. 

cc @slowli @aleksuss

slowli: > Can we provide runtime id with it?

The runtime ID should be added automatically during error handling within the dispatcher; see, e.g. [here](https://github.com/exonum/exonum/blob/1a23a5c6bcb9c349ea096818e5cf952dc6669261/exonum/src/runtime/dispatcher/mod.rs#L321). If this information is not set, this should be fixed within the dispatcher.

vitvakatu: set_runtime_id method is module-private, we can't use it in EJB.

exonum/src/runtime/mod.rs
```diff
@@ -191,6 +191,27 @@ impl From<RuntimeIdentifier> for u32 {
...
+            1 => Ok(RuntimeIdentifier::Java),
+            _ => Err(()),
+        }
+    }
+}
+
+impl std::fmt::Display for RuntimeIdentifier {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            RuntimeIdentifier::Rust => write!(f, "Rust"),
```
slowli: Nit: you can use [`f.write_str`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.write_str).

exonum/src/runtime/mod.rs
```diff
@@ -191,6 +191,27 @@ impl From<RuntimeIdentifier> for u32 {
...
+    fn try_from(id: u32) -> Result<Self, Self::Error> {
+        match id {
+            0 => Ok(RuntimeIdentifier::Rust),
+            1 => Ok(RuntimeIdentifier::Java),
+            _ => Err(()),
+        }
+    }
+}
+
+impl std::fmt::Display for RuntimeIdentifier {
```
slowli: Nit: Since `fmt` is in scope, you could use `fmt::Display`.

exonum/src/runtime/mod.rs
```diff
@@ -191,6 +191,27 @@ impl From<RuntimeIdentifier> for u32 {
     }
 }
 
+impl std::convert::TryFrom<u32> for RuntimeIdentifier {
```
slowli: `TryFrom` is a part of the public interface, so having an adequate error type here would be useful. I think it makes sense to rather define the transform `u32 -> Result<RuntimeIdentifier, ()>` as a private function for `RuntimeIdentifier`.

exonum/src/runtime/error.rs
```diff
@@ -526,21 +536,24 @@ impl Display for ExecutionError {
...
             )?;
         } else if let Some(runtime_id) = self.runtime_id {
             write!(
                 formatter,
-                "Execution error `{kind}` occurred in runtime {id}",
+                "Execution error with code `{kind}` occurred in {runtime}",
                 kind = self.kind,
-                id = runtime_id
+                runtime = match RuntimeIdentifier::try_from(runtime_id) {
+                    Ok(runtime) => format!("{} runtime", runtime),
```
slowli: Nit: IMO, it would make more sense to include "runtime" qualifier into `impl Display for RuntimeIdentifier` and simply use `runtime.to_string()` here.

exonum/src/runtime/error.rs
Kormick: I've added set of runtime_id in the Dispatcher. Should we leave ExecutionError::runtime method?

exonum/src/runtime/mod.rs
```diff
@@ -194,6 +194,25 @@ impl From<RuntimeIdentifier> for u32 {
     }
 }
 
+impl RuntimeIdentifier {
```
alekseysidorov: What do you think about this crate?
https://crates.io/crates/num-derive
It could be helpful to avoid of handwritten implementations.

Kormick: Looks good, but I'm not sure if we want to add new dependencies for this case.

popzxc: Agree, adding a new dependency just to save several lines is an overkill.

alekseysidorov: The similar code is not only present here in this crate. But this is just a little remark.

popzxc: You can create an issue and mark it for the next grooming, so we'll discuss whether it worth it :)

exonum/src/runtime/error.rs
slowli: @vitvakatu That's the idea behind automatic assignment of information - the runtimes should not care about this stuff knowing that it happens automatically behind the scenes.

@Kormick I don't see a reason to retain this method. I'd rather think about removing `service` constructor 🙃

exonum/src/runtime/rust/mod.rs
```diff
@@ -435,6 +441,18 @@ impl RustRuntime {
...
     }
+
+    fn artifacts_to_pretty_string(&self) -> String {
+        if self.available_artifacts.is_empty() {
+            return "None".to_string();
+        }
+
+        self.available_artifacts
+            .keys()
+            .map(|id| format!("{}", id))
```
slowli: Nit: Using `ToString::to_string` here would look more idiomatic here, IMO.

PR #1627, opened 2019-12-17
Move sandbox tests into the separate crate [ECR-3931]

alekseysidorov:
This action should be done before I will move the Rust runtime into separate crate.


test-suite/consensus-tests/src/lib.rs
```diff
@@ -1074,8 +1052,12 @@ impl SandboxBuilder {
     }
 }
 
-impl Schema<&Fork> {
+pub trait SchemaExt {
```
slowli: IMO, this extension trait is unwarranted: the method is used in a single place, so it could be defined as `reject_transaction(&Schema, &Hash)`.

exonum/src/blockchain/block.rs
```diff
@@ -125,8 +125,8 @@ pub struct Block {
 }
 
 impl Block {
-    /// Insert new additional header to the block.
-    pub(crate) fn add_header<K: BlockHeaderKey>(&mut self, value: K::Value) {
+    /// Inserts new additional header to the block.
+    pub fn add_header<K: BlockHeaderKey>(&mut self, value: K::Value) {
```
slowli: This method should have `#[doc(hidden)]` attribute.

exonum/src/blockchain/mod.rs
```diff
@@ -498,9 +498,9 @@ impl BlockchainMut {
     }
 }
 
-/// Return transaction from persistent pool. If transaction is not present in pool, try
-/// to return it from transactions cache.
-pub(crate) fn get_transaction<T: RawAccess>(
+/// Returns transaction from the persistent pool. If transaction is not present in the pool, tries
+/// to return it from the transactions cache.
+pub fn get_transaction<T: RawAccess>(
```
slowli: This function and the one below should have `#[doc(hidden)]` attribute.

exonum/src/runtime/blockchain_data.rs
```diff
@@ -132,3 +132,18 @@ impl SnapshotExt for dyn Snapshot {
         mount_point_for_service(self, id)
     }
 }
+
+/// Extension trait for `Fork` allowing to writeable access blockchain data in a more structured way.
+///
+/// Be very careful with blockchain manipulations, they can destroy the data consistency!
+#[doc(hidden)]
+pub trait ForkExt {
```
slowli: Nit: IMO, exposing a constructor for `blockchain::Schema` (with `#[doc(hidden)]` and maybe feature gate) could be more versatile.

test-suite/consensus-tests/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
```
slowli: A README for the crate would be appreciated. Despite the crate not being published, README could still help understand the purpose and contents of the crate.

PR #1628, opened 2019-12-18
Move indexes to separate module.

pavel-mukhanov:
All indexes has been moved to separate module.


CHANGELOG.md
```diff
@@ -52,6 +52,14 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `Supervisor` service now can have initial configuration and implements
   `Configure` interface. (#1587)
 
+### Internal Improvements
```
popzxc: I believe this is a breaking change, since import paths for *proof indexes were changed.

pavel-mukhanov: Not sure if it really breaking, because external APIs remain unchanged.

pavel-mukhanov: Import paths for *proof indexes only changed in MerkleDB.

popzxc: Ah, sure.

components/merkledb/src/lib.rs
```diff
@@ -159,28 +159,23 @@
...
     views::{AsReadonly, IndexAddress, IndexType, ResolvedAddress, SystemSchema},
 };
 // Workaround for 'Linked file at path {exonum_merkledb_path}/struct.ProofMapIndex.html
 // does not exist!'
 #[doc(no_inline)]
 pub use self::{
-    proof_list_index::{ListProof, ProofListIndex},
-    proof_map_index::{MapProof, ProofMapIndex, RawProofMapIndex},
+    indexes::proof_list_index::{ListProof, ProofListIndex},
+    indexes::proof_map_index::{self, MapProof, ProofMapIndex, RawProofMapIndex},
```
popzxc: Why is `self` re-exported for `proof_map_index`, but it not for `proof_list_index`?

pavel-mukhanov: For clarity and backward compatibility. From `proof_list_index` module we need to export only `ProofListIndex` and `ListProof`.  From `proof_map_index` we need to export `Raw`, `Hashed`, `ProofPath` and so on.  And I think it's more clear to use `proof_map_index::Raw` than `merkledb::Raw`.

popzxc: Hm... I'm a bit concerned about asymmetry here, but the explanation seems to be reasonable. OK then.

slowli:
AFAIK, Java binding uses exact iterator types in the native part of the binding, so they need to remain exportable from the crate. Is this correct, @vitvakatu? If so, I'd suggest to keep modules for separate indexes public and shorten iterator names akin to the naming used in `std` (`Iter` / `Keys` / `Values`).

vitvakatu:
@slowli yes, thank you for pinging.

components/merkledb/src/indexes/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+
+//! All available `MerkleDB` indexes.
+
+pub use self::{entry::Entry, group::Group, proof_entry::ProofEntry};
+
+mod entry;
+mod group;
+mod proof_entry;
+
+pub mod key_set_index;
```
slowli: Nit: I'd remove `_index` suffix from the index modules.

CHANGELOG.md
```diff
@@ -72,6 +72,11 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - `Supervisor` structure isn't generic anymore. (#1587)
 
+#### exonum-merkledb
+
+- All indexes has been moved to separate module `indexes`. Indexes iterators
```
slowli: > All indexes has been moved to separate module `indexes`.

Not from the user's perspective they weren't.

.travis.yml
```diff
@@ -103,6 +103,11 @@ jobs:
     - touch target/doc/std/option/enum.Option.html
     - touch target/doc/std/primitive.usize.html
     - touch target/doc/std/primitive.char.html
+    - touch target/doc/trait.BinaryKey.html
+    - touch target/doc/trait.BinaryValue.html
+    - touch target/doc/struct.ListIndex.html
```
slowli: Why are the new fake files necessary? As far as I can see, the relevant links indeed don't work; e.g., the link to `../struct.ListIndex.html` does not work for `SparseListIndex` if it is accessed from the crate root. Same for the link to `BinaryKey` from `KeySetIndex`.

PR #1629, opened 2019-12-18
Privatization [ECR-3946]

popzxc:
This PR removes unused public APIs, and reduces the visibility of several public items (which shouldn't be public by their nature).

Main changes:

- The following public APIs were removed/made private in `exonum`:
  - `helpers::path_relative_from` function;
  - `messages::BinaryValue` public re-export;
  - `Blockchain::pool_size`, `Blockchain::get_saved_peers` and
    `Blockchain::remove_peer_with_pubkey` methods;
  - `node::state` module (constants from `node::state` are now accessible in
    `node::constants` module).
- `SparseListIndex::indices` method was renamed to `SparseListIndex::indexes` (and `indexes` is used everywhere instead of `indices`).
- The following public APIs were removed/made private `testkit`:
  - `compare` module;
  - `txvec` macro;
  - `TestKit::probe_all` and `TestKit::probe` methods.
- Some methods that were declared `pub` were made `pub(crate)` (most of them were in fact private, since the modules containing them are private, but IMO it's better to have the lowest possible visibility just in case).


slowli:
Here's some types in the main crate I find kind of redundant. (`api`, `explorer`, `node` and `runtime::rust` modules will be moved out soon, so I've skipped them.)
 
- `impl_serde_hex_for_binary_value` (may belong to the MerkleDB crate?)
- `blockchain::{error reexports}`
- `blockchain::FatalError` (looks semantically incorrect; in reality, it’s a DB error)
- `blockchain::InstanceCollection` (haven’t we agreed to remove it?)
- `helpers::ZeroizeOnDrop` (???)
- `helpers::Milliseconds` (???)
- `helpers::generate_testnet_config` (needs docs or be removed / hidden)
- `helpers::create_rust_runtime_and_genesis_config` (???)
- `helpers::config` (affected by #1598)
- `helpers::user_agent` (why is this exported as a separate module? Probably needs to be hidden.)
- `messages`: most of the module. Most of messages pertain to consensus and thus don’t need to be a part of public API.
- `proto`: Should this module be `#[doc(hidden)]`? Since message types generated by Protobuf consist entirely of public fields, there is at least technical possibility to make otherwise forward-compatible changes breaking when evolving Protobuf declarations (e.g., adding a field).
- `runtime::error`: Since most of its contents are re-exported in `runtime`, it could make sense to make the module private.
- `runtime::Mailbox`: The opposite case; `Mailbox::push` accepts an `Action` argument, but `Action` is not exported.
- `runtime::RuntimeInstance` (the type looks redundant and is a result of a sub-optimal `Node` constructor)

The methodology to obtain this list was as follows:

1. Compile the crate docs locally.
2. Open the crate docs.
3. Look around and note the stuff looking sort of cringe.

test-suite/testkit/src/lib.rs
```diff
@@ -421,50 +417,12 @@ impl TestKit {
...
-        let snapshot = self.snapshot();
-        self.rollback();
-        snapshot
-    }
-
-    /// Executes a transaction given the current state of the blockchain but does not
-    /// commit execution results to the blockchain. The execution result is the same
-    /// as if a transaction was included into a new block; for example,
-    /// a transaction included into one of previous blocks does not lead to any state changes.
     ///
```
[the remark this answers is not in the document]
popzxc: Ah, it's a merge artifact =_=

popzxc:
@slowli thanks for suggestion, I've implemented (almost) all and it seems that the provided list is (close to be) complete.

What was done also:
- `clear_consensus_messages_cache` was put under `doc(hidden)`.

What was left untouched:
- `messages` module -- I attempted to reduce visibility, but that cause a lot of pain, since in fact most of structures leak outside (through `events` module, which is `doc(hidden)` but still public and `node` module); making items less public there deserves a whole new PR;
- `RuntimeInstance` structure -- we use `Into<RuntimeInstance>` in the blockchain builder;
- `runtime::Mailbox` structure -- it's used as an argument in the `after_commit` hook and we can hide it easily.

alekseysidorov:
Keep in mind that some of the methods are used at the Rust runtime. But Rust runtime will be moved to a separate crate very soon.

popzxc:
> Keep in mind that some of the methods are used at the Rust runtime. But Rust runtime will be moved to a separate crate very soon.

Better to close interfaces now and then discuss which ones we need to open again, IMHO.

exonum/src/node/consensus.rs
```diff
@@ -206,7 +206,10 @@ impl NodeHandler {
 
     /// Handles the `Block` message. For details see the message documentation.
     // TODO: Write helper function which returns Result. (ECR-123)
```
aleksuss: Is this TODO actual ?

exonum/src/node/mod.rs
```diff
@@ -23,7 +23,15 @@ pub use self::{
 };
 
 // TODO: Temporary solution to get access to WAIT constants. (ECR-167)
```
aleksuss: I suppose this TODO is outdated as well and can be removed.

popzxc: Don't think so, we still have to think whether it's applicable to have those constants in public.

exonum/src/node/consensus.rs
popzxc: Nope, I'll remove it.

CHANGELOG.md
```diff
@@ -35,10 +41,41 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+  - `blockchain::{error reexports}` (available from `runtime::`);
+  - `blockchain::FatalError` public re-export;
+  - `blockchain::InstanceCollection` structure;
+  - `helpers::ZeroizeOnDrop` trait;
+  - `helpers::Milliseconds` type;
+  - `helpers::generate_testnet_config`, `helpers::create_rust_runtime_and_genesis_config`
+    and `helpers::clear_consensus_messages_cache` functions;
+  - `helpers::config` and `helpers::user_agent` modules;
+  - `proto` module;
+  - `runtime::error` module (`catch_panic` was added to the list of public
```
alekseysidorov: May you sort lines by the method names?

exonum/src/blockchain/schema.rs
```diff
@@ -108,7 +108,8 @@ pub struct Schema<T> {
 
 impl<T: Access> Schema<T> {
     /// Constructs information schema based on the given `access`.
-    pub(crate) fn new(access: T) -> Self {
+    #[doc(hidden)]
```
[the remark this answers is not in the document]
popzxc: Good idea, but I think it can be a fully separate crate, not even a projects in `components` folder.

PR #1632, opened 2019-12-19
Fix documentation in runtime.proto

vitvakatu:
As exonum protofiles are not Rust-specific, we cannot use Rust idioms in the documentation.

## Overview

---

### Definition of Done


exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -146,9 +146,8 @@ message ArtifactState {
 message InstanceState {
   // Status of a service instance.
   enum Status {
-    // The service instance has no status, i.e. this value corresponds to
-    // `Option::None` on the Rust code level and other corresponds to
-    // `Some(...)`.
+    // No value, e.g. there are no pending changes to the instance
```
alekseysidorov: But with this phrase, I also suggest using `Option` in Rust.

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -157,8 +156,8 @@ message InstanceState {
   }
   // Service instance specification.
   exonum.runtime.InstanceSpec spec = 1;
-  // Service instance activity status.
+  // Service instance status.
   Status status = 2;
-  // Pending status of instance if the value is not `NONE`.
+  // Pending status of the instance.
```
alekseysidorov: You removed assertion that the `Status::NONE` cannot be pending.

exonum/src/proto/schema/exonum/runtime.proto
vitvakatu: Who is the recipient of such suggestion? Does the user of protofiles need to know what Rust type to use?

exonum/src/proto/schema/exonum/runtime.proto
vitvakatu: 🤔 Waaait. Does it mean that `pending_status` must never be equal to Status::NONE? If so, what is the purpose of Status::NONE?

aleksuss: There is a comment `// Reserved for the further usage.` 🙂

alekseysidorov: This status exists only between the transaction execution that initiates the instance status change and the block commit, thus it can only visible if you read dispatcher schema. But this status is still important to distinguish instance states during the block execution.
Purpose of Status::NONE is to avoid protobuf design flaws.

exonum/src/proto/schema/exonum/runtime.proto
alekseysidorov: Not only Rust language has optional values. Moreover, this remark is a hint for how to interpret `Status::NONE`.

vitvakatu:
The PR is closed, separate task to fix the documentation will be created

PR #1635, opened 2019-12-20
Allow to return `IndexProof` by full index name [ECR-4013]

Kormick:
## Overview

Implemented proof_for_index in SnapshotExt. BlockchainData::proof_for_service_index uses this method now.


vitvakatu:
I suppose the method should be `#[doc(hidden)]`

exonum/src/runtime/blockchain_data.rs
```diff
@@ -119,6 +109,18 @@ pub trait SnapshotExt {
...
+    /// Returns a proof for a Merkelized index with the specified name.
+    ///
+    /// # Return value
+    ///
+    /// If the index does not exist or is not Merkelized, returns `None`. Note that this may
+    /// occur before the index is accessed the first time, which, depending on the service logic,
+    /// may happen indefinitely after the service has been initialized. Thus, `unwrap`ping the
+    /// returned value may unexpectedly lead to a panic unless the index is initialized early
+    /// (e.g., during service initialization).
+    #[doc(hidden)]
```
alekseysidorov: I am not sure that this method should be hidden.

popzxc: [Find an agreement](https://github.com/exonum/exonum/pull/1635#issuecomment-567901022) :)

IMO it's ok for this method to be hidden - it's unlikely that service authors will want to get a proof for an index by its full name, so I tend to the `doc(hidden)` variant.

vitvakatu: https://wiki.bf.local/display/EJB/Java+Proofs+P2%3A+Creating+IndexProofs?focusedCommentId=40828999#comment-40828999

Kormick: I don't have permission :(

Kormick: Now I do

slowli: My initial remark that the method should be hidden was not very well thought out. I'm not against the method being public; it doesn't seem possible to misuse or abuse it.

vitvakatu: I'm afraid making this method public will force breaking changes later, as this piece of functionality is needed for EJB only, and might conflict with later changes.

PR #1636, opened 2019-12-20
Fix panic on multiple proposes on the same height [ECR-4030]

popzxc:
This PR fixes node behaviour if several proposes correspond to the same round, and one of them is applied: now, if the height is bumped during processing proposals become complete, no more proposals are processed.

Corresponding test in the `consensus-tests` was added as well.


exonum/src/node/consensus.rs
```diff
@@ -309,13 +337,19 @@ impl NodeHandler {
             if our_block_hash != block_hash {
                 panic!(
                     "Full propose: wrong state hash. Either a node's implementation is \
```
slowli: Nit: It's "wrong block hash", isn't it?

exonum/src/node/consensus.rs
```diff
@@ -309,13 +337,19 @@ impl NodeHandler {
...
-                     incorrect or validators majority works incorrectly"
+                     incorrect or validators majority works incorrectly. Calculated hash \
+                     is {:?}, but hash in the propose is {:?}",
```
slowli: AFAIK, `block_hash` is not included in proposals (and is not computed at the time a proposal is created). Instead, `block_hash` is mentioned in precommits.

exonum/src/node/consensus.rs
```diff
@@ -309,13 +337,19 @@ impl NodeHandler {
             if our_block_hash != block_hash {
```
slowli: Sorry for off-topic, but I can't help but notice that the code above is immensely incomprehensible to me. (I think I've written about it ~a year ago, but the code hasn't disappeared somehow 🙃)

- Why does `take_unknown_propose_with_precommits` return a vector if the cycle returns after the first iteration? Is this an error with the cycle or an error with `take_unknown_propose_with_precommits`?
- Can there be multiple values in the returned vector? When?

exonum/src/node/consensus.rs
```diff
@@ -496,14 +542,15 @@ impl NodeHandler {
                     self.check_propose_and_broadcast_precommit(round, propose_hash, block_hash);
                     // Commit if has consensus
```
slowli: Nit / off-topic: This comment looks grammatically incorrect; could you fix it?

exonum/src/node/consensus.rs
```diff
@@ -660,9 +707,14 @@ impl NodeHandler {
 
         let full_proposes = self.state.check_incomplete_proposes(hash);
         // Go to handle full propose if we get last transaction.
+        let mut height_bumped = false;
         for (hash, round) in full_proposes {
             self.remove_request(&RequestData::ProposeTransactions(hash));
-            self.handle_full_propose(hash, round);
+            // If a new height was achieved, no more proposals for this height
+            // should be processed. However, we still have to remove requests.
```
slowli: Can the logic further down the method (i.e., handling a full block) work incorrectly if the new height is achieved here?

exonum/src/node/state.rs
```diff
@@ -870,6 +870,19 @@ impl State {
             }
         }
 
+        // Sort proposes, so the order of processing is predictable.
+        full_proposes.sort_unstable_by(|a, b| {
```
slowli: I'm not sure I understand what sorting achieves here. Is it useful for testing purposes only?

test-suite/consensus-tests/src/lib.rs
```diff
@@ -585,11 +587,13 @@ impl Sandbox {
 
         for _ in 0..expected_set.len() {
             if let Some((real_addr, real_msg)) = self.pop_sent_message() {
-                assert_eq!(
-                    expected_msg,
-                    real_msg.as_raw(),
-                    "Expected to broadcast other message"
-                );
+                if expected_msg != real_msg.as_raw() {
```
slowli: Why not use `assert_eq!` with a custom message?

test-suite/consensus-tests/src/lib.rs
```diff
@@ -548,33 +548,35 @@ impl Sandbox {
         }
     }
 
-    pub fn broadcast<T: TryFrom<SignedMessage>>(&self, msg: &Verified<T>) {
+    pub fn broadcast<T>(&self, msg: &Verified<T>)
+    where
+        T: TryFrom<SignedMessage> + std::fmt::Debug,
```
slowli: Nit / style: Not a fan of fully qualified names. Since `Debug` is used multiple times in the module, it could make sense to import it.

exonum/src/node/consensus.rs
popzxc: Yeah, I forgot to fix the note.

test-suite/consensus-tests/src/lib.rs
popzxc: Because we compare raw message, but display not raw.

exonum/src/node/state.rs
popzxc: Not sure. It guarantees that proposes are processed according to the round number (primary key), and only then according to the hash.
The second one is useful for testing, yeah, but the first one is useful overall IMO - why would we want to process rounds in order 3-0-2-1?

exonum/src/node/consensus.rs
popzxc: Don't think so: if we're involved in the processing of the new height approval, we shouldn't be lagging behind, and thus shouldn't request blocks.

test-suite/consensus-tests/src/lib.rs
```diff
@@ -585,11 +587,13 @@ impl Sandbox {
...
-                assert_eq!(
-                    expected_msg,
-                    real_msg.as_raw(),
-                    "Expected to broadcast other message"
-                );
+                if expected_msg != real_msg.as_raw() {
+                    // Use full messages instead of raw for reporting an assertion failure.
+                    panic!(
+                        "Expected to broadcast other message: expected {:?}, but got {:?}",
+                        &msg, &real_msg
```
aleksuss: References are redundant here, I suppose.

test-suite/consensus-tests/src/lib.rs
aleksuss: Just out of curiosity. Is there some benefit of displaying not raw message here ? In below's panics we display raw messages.

popzxc: Yes, there is: raw message contains payload only as a sequence of bytes, and wrapped message provides not serialized data (e.g. `Propose { ... }`) as well. While I was fixing the bug, I had plenty of panics here and believe me -- raw message in log is really useless :(
I think we should avoid displaying raw messages in panics/assertions everywhere, but it's a topic for a different PR. I'll create an issue for that.

exonum/src/node/consensus.rs
popzxc: Indeed it seems that there can't be more than one element in the vector.
I renamed methods to `add_propose_confirmed_by_majority` / `take_confirmed_propose`, made them operate `Option` instead of `Vec` and added comments explaining the usage conditions.

exonum/src/node/consensus.rs
popzxc: I added a comment explaining this, hopefully helpful. However, this is a place for a future refactoring.

exonum/src/node/state.rs
```diff
@@ -870,9 +870,53 @@ impl State {
...
+        // descending order, since it's more likely to commit a propose with
+        // greater round, and we don't want to process a lot of proposes from
+        // lower rounds.
+        // However, if the amount of proposes isn't big, overhead from sorting
+        // is not desired.
+
+        // TODO: Clarify the value for this constant (ECR-4050).
+        const MIN_PROPOSES_AMOUNT_FOR_SORTING: usize = 10;
+
+        if full_proposes.len() >= MIN_PROPOSES_AMOUNT_FOR_SORTING {
```
slowli: Sorry for being obtuse in a personal conversation: I've meant the exact opposite logic for this threshold. If the number of proposes is small, we can sort them with a relatively small overhead. If the amount of proposes is large (which should almost never happen), the overhead may become significant, so we don't sort them.

exonum/src/node/state.rs
```diff
@@ -870,9 +870,53 @@ impl State {
...
+            full_proposes.sort_unstable_by(|a, b| {
```
slowli: Nit: Does tuple destructuring here makes the formatter go crazy? Otherwise, it could make things clearer.

exonum/src/node/consensus.rs
```diff
@@ -658,13 +708,28 @@ impl NodeHandler {
...
+            if !height_bumped {
+                height_bumped = self.handle_full_propose(hash, round) == RoundAction::NewHeight;
+            }
         }
 
+        // Scenario 2&3: We're processing an already committed block.
```
slowli: Nit: "Scenarios 2 and 3:" would look more readable, IMO. (`2 & 3 == 2` 🙃)

exonum/src/node/state.rs
popzxc: The same applies for my inattentive reading; sorry as well :)

exonum/src/node/consensus.rs
```diff
@@ -409,8 +445,9 @@ impl NodeHandler {
                 );
             }
 
-            self.lock(prevote_round, propose_hash);
+            return self.lock(prevote_round, propose_hash);
```
alekseysidorov: "if with else" is more ideomatic way in my opinion.

PR #1637, opened 2019-12-21
Owning iterators & group key enumeration

slowli:
## Overview

This is another approach to key enumeration within a group. It introduces *the owning iterator*. The owning iterator is implemented via a self-referential struct via pinning (I've tried to use [`rental`](https://docs.rs/rental/0.5.4/rental/index.html), but didn't get it to work; the crux of the issue is that the tail part of the struct - i.e., the iterator - needs to be taken by a *mutable* reference and return a *shared* reference to its data). The iteration over group keys is then implemented as a particular use case of the owning iterator over the metadata index.

The resulting API is still clunky; to slightly soften restrictions on it, I've changed the logic of writing metadata for indexes. Previously, the metadata was written eagerly on each change; now, it's written on index drop. There's still a plenty of situations where using an iterator will lead to a panic, so a `buffered_keys()` alternative is still implemented.

Owning iterators may be of separate interest; they make it possible to convert indexes into iterators, which may be useful, e.g., to return such an iterator from a function. The issue with owning iterators is that I'm not 100% sure they are memory-safe (there's some `unsafe` code involved).

See also #1614

### DIFF
--- a/components/merkledb/src/access/mod.rs
+++ b/components/merkledb/src/access/mod.rs
@@ -8,14 +8,16 @@ pub use self::extensions::AccessExt;
 pub use crate::views::{AsReadonly, RawAccess, RawAccessMut};
 
 use crate::validation::assert_valid_name_component;
-use crate::views::{IndexAddress, IndexMetadata, IndexType, ViewWithMetadata};
+use crate::views::{
+    GroupKeys, IndexAddress, IndexMetadata, IndexType, IndexesPool, ViewWithMetadata,
+};
 
 mod extensions;
 
 /// High-level access to database data.
 ///
 /// This trait is not intended to be implemented by the types outside the crate; indeed,
-/// it instantiates `ViewWithMetadata`, which is crate-private. Correspondingly, `Access` methods
+/// it instantiates several crate-private types. Correspondingly, `Access` methods
 /// rarely need to be used directly; use [its extension trait][`AccessExt`] instead.
 ///
 /// [`AccessExt`]: trait.AccessExt.html
@@ -52,6 +54,15 @@ pub trait Access: Clone {
         addr: IndexAddress,
         index_type: IndexType,
     ) -> Result<ViewWithMetadata<Self::Base>, AccessError>;
+
+    /// Returns an iterator over keys in a group with the specified address.
+    ///
+    /// # Panics
+    ///
+    /// If the access is built on top a `Fork`, an attempt to create an index from the fork
+    /// while iterating over keys will result in a panic. This is because such an operation
+    /// may invalidate the iterator.
+    fn group_keys(self, base_addr: IndexAddress) -> GroupKeys<Self::Base>;
 }
 
 impl<T: RawAccess> Access for T {
@@ -67,6 +78,10 @@ impl<T: RawAccess> Access for T {
         index_type: IndexType,
     ) -> Result<ViewWithMetadata<Self::Base>, AccessError> {
         ViewWithMetadata::get_or_create(self, &addr, index_type)
+    }
+
+    fn group_keys(self, base_addr: IndexAddress) -> GroupKeys<Self::Base> {
+        IndexesPool::new(self).group_keys(&base_addr)
     }
 }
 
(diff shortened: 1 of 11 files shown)


components/merkledb/src/views/owning_iter.rs
```diff
@@ -0,0 +1,337 @@
...
+        });
+
+        unsafe {
+            // Elongate the iterator lifetime to `'static`.
+            //
+            // SAFETY:
+            // `T` may not have static lifetime (indeed, `T == &Fork` or `T == &dyn Snapshot`
+            // are two frequent use cases), but since `Inner` contains `T` as the type param,
+            // `iter` is always valid during the lifetime of the `Inner` instance.
+            let iter: Box<dyn Iterator> = mem::transmute(boxed.view.iter_bytes(from));
```
slowli: It's possible to get rid of `transmute` here in favor of reference lifetime elongation:

```rust
struct Inner<'a, T: RawAccess + 'a> {
    view: View<T>,
    iter: Option<Box<dyn Iterator + 'a>>,
    _pin: PhantomPinned,
}

let view_ptr = &boxed.view as *const View<T>;
let view_ref: &'a View<T> = unsafe { &*view_ptr };
let iter: Box<dyn Iterator + 'a> = View::iter_bytes(view_ref, from);
```

In code using the iterator, the lifetime `'a` would be automatically set by the compiler. For example, the iterator in `Group` would be redefined as

```rust
pub fn keys<'a>(&self) -> Keys<'a, T::Base, K>
``` 

Not sure whether this complication is warranted though.

components/merkledb/src/access/mod.rs
```diff
@@ -52,6 +54,15 @@ pub trait Access: Clone {
...
     ) -> Result<ViewWithMetadata<Self::Base>, AccessError>;
+
+    /// Returns an iterator over keys in a group with the specified address.
+    ///
+    /// # Panics
+    ///
+    /// If the access is built on top a `Fork`, an attempt to create an index from the fork
+    /// while iterating over keys will result in a panic. This is because such an operation
+    /// may invalidate the iterator.
+    fn group_keys(self, base_addr: IndexAddress) -> GroupKeys<Self::Base>;
```
alekseysidorov: So can we fix this issue by introducing an additional lifetime?

components/merkledb/src/views/owning_iter.rs
```diff
@@ -0,0 +1,337 @@
...
+    db::Iterator,
+    views::{RawAccess, View},
+};
+
+#[derive(Debug)]
+struct Inner<T: RawAccess> {
+    // Safety: Not accessed during iteration.
+    view: View<T>,
+    iter: Option<Box<dyn Iterator>>,
+    _pin: PhantomPinned,
```
alekseysidorov: Did you see the `pin-project` crate? It may be usable in this case.

components/merkledb/src/group.rs
```diff
@@ -104,12 +105,98 @@ where
...
+    /// the [`buffered_keys`] method or otherwise delay the DB modification until after
+    /// the iterator is dropped.
+    ///
+    /// [`Fork`]: struct.Fork.html
+    /// [`ReadonlyFork`]: struct.ReadonlyFork.html
+    /// [`buffered_keys`]: #method.buffered_keys
+    ///
+    /// # Examples
+    ///
+    /// This example shows incorrect use of the iterator:
```
popzxc:
```suggestion
    /// This example shows **incorrect** use of the iterator:
```

components/merkledb/src/views/owning_iter.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+impl<T: RawAccess> Inner<T> {
```
popzxc: To be honest I like neither of two existing solutions. On the one hand, loading all the group keys into RAM is dangerous, and on the other the solution which I don't find very elegant.

Taking into account that in most cases you will not iterate through group keys often, maybe it's possible to create an intermediate approach:
- Define some fixed amount of entries that we can safely store in RAM;
- Using `View::iter` load the first bunch of entries and allow iterating through them;
- Once all the loaded entries are consumed, load the next portion using `View::iter_from`;
- Finally, stop iterating, when all the entries with the given prefix are consumed.

Will this work or is it even more inappropriate?

slowli: I've thought about this approach briefly. It is incorrect in the general case (the keys can change between `iter_from` calls), but I *think* it could work for readonly access, which is the use case for migrations. I'll try this approach, most probably in a separate branch (since this one has owning iterators as the defining feature).

slowli:
I've moved the relevant code parts to #1662. I think owning iterators can make sense, but they could be implemented later.

PR #1639, opened 2019-12-23
Service schemas as public interface [ECR-4007]

slowli:
## Overview

This PR introduces tools for schemas as a part of the public service interface. More widely, the `runtime::versioning` module in the main crate contains versioning tools and documentation on how service developers should version services.

---

See: https://jira.bf.local/browse/ECR-4007


components/derive/src/lib.rs
```diff
@@ -130,64 +131,82 @@ pub fn from_access(input: TokenStream) -> TokenStream {
...
-/// * `#[service_factory(implements("path_1", "path_2"))]`
+/// Path to the module that was generated by the build script, which
+/// contains the original Protobuf source files of the service. If omitted, no Protobuf sources
+/// will be included with the service artifact.
 ///
-/// Path list to the interfaces which have been implemented by the service.
+/// ## `crate`
 ///
-/// ## Optional
+/// ```text
```
popzxc: Nit: Maybe `ignore` to get syntax highlighting?

examples/cryptocurrency-advanced/backend/src/schema.rs
```diff
@@ -24,14 +24,24 @@ use exonum::crypto::{Hash, PublicKey};
...
+    /// Public part of the schema.
+    #[from_access(flatten)]
+    pub public: SchemaInterface<T>,
     /// History for specific wallets.
     pub wallet_history: Group<T, PublicKey, ProofListIndex<T::Base, Hash>>,
 }
 
+/// Public part of the cryptocurrency schema.
+#[derive(Debug, FromAccess, RequireArtifact)]
+pub struct SchemaInterface<T: Access> {
```
popzxc: Bikeshedding: IMO `InterfaceSchema` / `PublicSchema` is better, since `XXXInterface` is usually associated with some description abstracted over a hidden implementation.

aleksuss: What about `SchemaAccess` ??? 🤔

popzxc: We already have `Access` as an abstraction, and it will cause even more ambiguity, as for me.

slowli: One of reasons for using `*Interface` is that `*Schema` is processed specially by `#[derive(FromAccess)]`. I'm thinking about removing special processing and using `Schema` for public part of schema and `SchemaImpl` for the full / private part. What do you think about it?

popzxc: Sounds good.

PR #1638, opened 2019-12-23
Use rust-rocksdb 0.13 [ECR-3848]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-3848

### Definition of Done

- [ ] There are no TODOs left in the merged code
- [ ] Change is covered by automated tests
- [ ] Benchmark results are attached (if applicable)
- [ ] The [coding guidelines] are followed
- [ ] Public API has proper documentation
- [ ] Changelog is updated if needed (in case of notable or breaking changes)
- [ ] The continuous integration build passes

[coding guidelines]: https://github.com/exonum/exonum/blob/master/CONTRIBUTING.md#conventions

PR #1641, opened 2019-12-24
Rust runtime crate [ECR-3931]

alekseysidorov:
Complete split off Rust runtime from the Exonum crate.


exonum/src/node/mod.rs
```diff
@@ -948,8 +945,7 @@ impl Node {
     /// Creates node for the given services and node configuration.
     pub fn new(
         database: impl Into<Arc<dyn Database>>,
-        external_runtimes: impl IntoIterator<Item = impl Into<RuntimeInstance>>,
-        services: impl IntoIterator<Item = Box<dyn ServiceFactory>>,
+        with_runtimes: impl FnOnce(mpsc::Sender<UpdateEndpoints>) -> Vec<RuntimeInstance>,
```
popzxc: It makes sense to add a verbose explanation what does this argument mean and what should if contain.

runtimes/rust/src/api.rs
```diff
@@ -241,7 +241,7 @@ impl ServiceApiScope {
     }
 
     /// Return a mutable reference to the underlying web backend.
-    pub fn web_backend(&mut self) -> &mut crate::api::backends::actix::ApiBuilder {
+    pub fn web_backend(&mut self) -> &mut exonum::api::backends::actix::ApiBuilder {
```
popzxc: Seems that this method is only used in `testkit` counter service for an example authorization. Maybe remove it at all?

runtimes/rust/src/call_context.rs
```diff
@@ -88,16 +114,14 @@ impl<'a> CallContext<'a> {
     ///
     /// This method can only be called by the supervisor; the call will panic otherwise.
     #[doc(hidden)]
+    #[deprecated]
```
popzxc: It isn't deprecated right now, is it? Currently it's the only API for service registration that we have, AFAIK.

popzxc: Same applies to other methods marked as `deprecated`.

runtimes/rust/src/lib.rs
```diff
@@ -262,31 +262,30 @@ pub use self::{
...
 use self::api::ServiceApiBuilder;
 
 mod call_context;
 mod runtime_api;
 mod service;
 mod stubs;
-#[cfg(test)]
-mod tests;
+// #[cfg(test)]
+// mod tests;
```
popzxc: I suggest to add some warning-spawning code here, because it's really easy to forget to uncomment this later, e.g.:

```rust
// TODO: This is placed here so warning will be spawned
// until code below is uncommented.
const FIXME: u8 = 42;
// #[cfg(test)]
// mod tests;
```

runtimes/rust/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+categories = ["cryptography"]
+description = "The runtime is for running native services written in Rust."
+
+[badges]
+travis-ci = { repository = "exonum/exonum" }
+
+[dependencies]
+exonum = { version = "0.13.0-rc.2", path = "../../exonum" }
+exonum-crypto = { version = "0.13.0-rc.2", path = "../../components/crypto" }
+exonum-derive = { version = "0.13.0-rc.2", path = "../../components/derive" }
```
popzxc: Since derive macros are part of a public interface for a rust runtime, shouldn't we provide public re-exports for required macros?

runtimes/rust/src/call_context.rs
alekseysidorov: I use deprecated marker to mark methods which I want to remove by this PR.

popzxc: Got it. With a comment it would have been more clear :)

runtimes/rust/src/api.rs
slowli: It was / is used in some projects with the same purpose it's used in testkit, i.e., to implement endpoint authorization. Since there is no other ways to provide these capabilities, I'd be wary to remove this method. It may be marked as hidden / unstable, though.

runtimes/rust/src/lib.rs
```diff
@@ -351,16 +350,31 @@ impl RustRuntime {
...
         let service_factory = service_factory.into();
         let artifact = service_factory.artifact_id();
         trace!("Added available artifact {}", artifact);
         self.available_artifacts.insert(artifact, service_factory);
         self
     }
 
+    /// Adds a new available to deploy service factories to the runtime and returns
+    /// a modified `RustRuntime` object for further chaining.
+    pub fn with_available_services(
```
slowli: Nit / subjective: This method looks redundant.

runtimes/rust/src/lib.rs
```diff
@@ -351,16 +350,31 @@ impl RustRuntime {
             .expect("Method called before Rust runtime is initialized")
     }
 
-    /// Adds a new service factory to the runtime and returns
+    /// Adds a new available to deploy service factory to the runtime and returns
     /// a modified `RustRuntime` object for further chaining.
-    pub fn with_factory(mut self, service_factory: impl Into<Box<dyn ServiceFactory>>) -> Self {
+    pub fn with_available_service(
```
slowli: Why did you change the name? I liked the old one better; for one, it's shorter 🙃 It also reflects the naming of the argument type (`ServiceFactory`).

runtimes/rust/src/service.rs
```diff
@@ -97,15 +96,14 @@ pub trait Service: ServiceDispatcher + Debug + 'static {
     /// If you aren't interested in the processing of for the genesis block, you can use
     /// [`CallContext::in_genesis_block`] method and exit early if `true` is returned.
     ///
-    /// Also note that invocation of [`blockchain::Schema::height`] will **panic** if invoked within
-    /// `after_transactions` of the genesis block. If you are going to process the genesis
-    /// block and need to know current height, use [`blockchain::Schema::next_height`] to infer the
-    /// current blockchain height.
+    /// Also note that invocation of `exonum::blockchain::Schema::height` will **panic**
```
slowli: Nit: "Also not that" looks redundant, and the `exonum::blockchain::Schema::height` is way too long. Are you against putting a link like `https://docs.rs/exonum/latest/exonum/blockchain/struct.Schema.html#method.height`?

runtimes/rust/src/call_context.rs
slowli: Do you plan to remove these methods in this PR or separately?

runtimes/rust/src/api.rs
```diff
@@ -241,7 +239,7 @@ impl ServiceApiScope {
...
```
slowli: Nit: Fully qualified names in the middle of the code look ugly.

runtimes/rust/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Another good point to use Rust runtime are services with high performance
+requirements because the runtime has lowest overhead.
+
+In the Rust runtime a set of service artifacts that you may want to deploy is
+static. The set is defined at the time of compilation. Once the set is created,
+you can change it only by the node binary recompilation.
+
+The Rust runtime does not provide any level of service isolation from the
+operation system. Therefore, the security audit of the artifacts that should be
```
slowli: Nit: "audit of the deployed artifacts" is shorter and more readable, IMO.

exonum/src/runtime/mod.rs
```diff
@@ -730,3 +727,96 @@ impl<'a> From<(InstanceId, &'a str)> for InstanceDescriptor<'a> {
...
+    pub fn initiate_stopping_service(&self, instance_id: InstanceId) -> Result<(), ExecutionError> {
+        Dispatcher::initiate_stopping_service(self.0.fork, instance_id)
+    }
+}
+
+/// Collection of unstable execution context features.
+#[doc(hidden)]
+pub trait ExecutionContextUnstable {
+    /// Re-borrows an execution context with the same interface name.
+    fn reborrow(&mut self) -> ExecutionContext<'_>;
```
slowli: Why is this method unstable?

exonum/src/runtime/mod.rs
```diff
@@ -730,3 +727,96 @@ impl<'a> From<(InstanceId, &'a str)> for InstanceDescriptor<'a> {
...
+    /// Re-borrows an execution context with the specified interface name.
+    fn reborrow_with_interface<'s>(&'s mut self, interface_name: &'s str) -> ExecutionContext<'s>;
+    /// Returns the service matching the specified query.
+    fn get_service<'q>(&self, id: impl Into<InstanceQuery<'q>>) -> Option<InstanceDescriptor<'_>>;
```
slowli: Why is this unstable?

exonum/src/runtime/mod.rs
```diff
@@ -609,6 +609,17 @@ impl<'a> ExecutionContext<'a> {
         }
     }
 
+    /// Returns extensions required for the Supervisor service implementation.
+    ///
+    /// Make sure that this method invokes by the instance with the [`SUPERVISOR_INSTANCE_ID`]
```
slowli: Grammar: ~invokes~ is invoked by.

Also, I don't quite like this kind of guarantees. Would it make sense to leave parts (or all) of `CallContext` in the main crate? IMO, there's nothing inherently Rustic about `CallContext`, and `InstanceDescriptor` necessary for its function can be obtained from `Dispatcher`, if I'm not mistaken.

exonum/src/node/mod.rs
slowli: Will this constructor signature be improved after `api` is moved to a separate crate? IMO, it looks quite terrible. (I don't have a better solution in mind yet, but I'll try to come up with one.) Additionally, `UpdateEndpoints` in the current form cannot work with multiple runtimes (any update will override the entire set of endpoints).

examples/timestamping/backend/src/lib.rs
```diff
@@ -37,11 +37,9 @@ pub mod transactions;
 
 use exonum::{
     merkledb::BinaryValue,
-    runtime::{
-        rust::{api::ServiceApiBuilder, CallContext, Service},
-        DispatcherError, ExecutionError,
-    },
+    runtime::{DispatcherError, ExecutionError},
```
slowli: The Rust runtime should probably re-export frequently used types from the `runtime` module. Alternatively, there should be a facade crate reexporting the Rust runtime along with other components. (The main crate currently reexports `crypto` and `merkledb`, but it can't reexport the Rust runtime for obvious reasons.)

components/derive/src/exonum_interface.rs
```diff
@@ -306,21 +306,21 @@ impl ExonumService {
 
             let method = quote! {
                 fn #name(&self, context: Ctx, arg: #arg_type) -> Self::Output {
-                    #cr::runtime::rust::GenericCall::generic_call(
+                    #cr::GenericCall::generic_call(
                         self,
                         context,
                         #descriptor,
-                        #cr::merkledb::BinaryValue::into_bytes(arg),
+                        exonum_merkledb::BinaryValue::into_bytes(arg),
```
slowli: This forces users to explicitly export `exonum_merkledb` crate, although it is already guaranteed to be reexported by `exonum`. For this reason, `exonum::merkledb` could be better.

components/derive/src/execution_fail.rs
```diff
@@ -20,20 +20,20 @@ use syn::{Attribute, Data, DeriveInput, Expr, Lit, Meta, MetaNameValue, Variant}
...
 
-use super::{find_meta_attrs, CratePath};
+use super::{find_meta_attrs, ExonumCratePath};
 
 #[derive(Debug, FromMeta)]
 #[darling(default)]
 struct ExecutionFailAttrs {
     #[darling(rename = "crate")]
-    cr: CratePath,
+    cr: ExonumCratePath,
```
slowli: Nit: I'd use `MainCratePath` / `RustRuntimeCratePath` naming.

runtimes/rust/src/lib.rs
slowli: Will the tests be restored in this PR?

services/supervisor/src/proto_structures.rs
```diff
@@ -20,14 +20,15 @@ use exonum::{
     exonum_merkledb::ObjectHash,
```
slowli: Off-topic: Huh, does the main crate expose both `merkledb` and `exonum_merkledb`? Seems weird.

popzxc: I already grumbled about that. If I remember correctly, this is due to some issues with derive macros. And I totally agree that it's way too weird =_=

runtimes/rust/src/api.rs
alekseysidorov: I am not sure. Without web backend you cannot write custom http handlers.

runtimes/rust/src/service.rs
alekseysidorov: Yes, because links like this is too fragile.

exonum/src/runtime/mod.rs
alekseysidorov: I just move all `pub(crate)` methods to this trait. If you think that I should make some of them public, I will do it.

exonum/src/node/mod.rs
alekseysidorov: At the moment I don't know how I can make this API better. 
I may write todo comment with explanation why this API so terrible and what we should do to fix it.

components/derive/src/lib.rs
```diff
@@ -248,15 +248,31 @@ pub(crate) fn find_meta_attrs(name: &str, args: &[Attribute]) -> Option<NestedMe
...
 }
 
-impl ToTokens for CratePath {
+impl ToTokens for MainCratePath {
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        self.0.to_tokens(tokens)
+    }
+}
+
+#[derive(Debug, FromMeta, PartialEq, Eq)]
```
aleksuss: Are `PartialEq` and `Eq` traits really needed here ?

exonum/src/node/mod.rs
aleksuss: Does it make sense to create type alias for `FnOnce(mpsc::Sender<UpdateEndpoints>) -> Vec<RuntimeInstance>` ?

alekseysidorov: I think that it is too verbose.

runtimes/rust/Cargo.toml
alekseysidorov: How I can do this? If I add `pub use exonum_derive::*` I get error.

```
the name `ServiceFactory` is defined multiple times
`ServiceFactory` must be defined only once in the macro namespace of this module
```

exonum/src/runtime/mod.rs
alekseysidorov: https://jira.bf.local/browse/ECR-4075

CHANGELOG.md
```diff
@@ -159,6 +159,11 @@ Indexes iterators names has been shortened to `Iter`, `Keys` and `Values`. (#162
 
 - `impl_serde_hex_for_binary_value` macro was moved from core to `merkledb`. (#1629)
 
+#### exonum-rust-runtime
+
+- Rust runtime module was moved from the `exonum` crate into the separate
```
popzxc: I think it should be as well noted in `exonum` section as breaking change and in `exonum-cli` as new feature.

components/derive/src/service_factory.rs
```diff
@@ -120,20 +120,20 @@ impl ToTokens for ServiceFactory {
         let (impl_generics, ty_generics, where_clause) = self.generics.split_for_impl();
 
         let expanded = quote! {
-            impl #impl_generics #cr::runtime::rust::ServiceFactory for #name #ty_generics #where_clause {
-                fn artifact_id(&self) -> #cr::runtime::ArtifactId {
-                    #cr::runtime::ArtifactId {
-                        runtime_id: #cr::runtime::RuntimeIdentifier::Rust as _,
+            impl #impl_generics #cr::ServiceFactory for #name #ty_generics #where_clause {
+                fn artifact_id(&self) -> exonum::runtime::ArtifactId {
```
slowli: You've changed the types to local for the `ServiceDispatcher` derive macro, but not here. Is this intentional?

runtimes/rust/src/lib.rs
```diff
@@ -248,6 +243,12 @@
 //! # }
 //! ```
 
+pub use exonum::runtime::{
```
slowli: Nit / subjective: I wonder whether gathering these types in a module would make more sense:

```rust
pub use exonum::runtime as generic; // name TBD
```

This would be more structured, but more verbose when specifying imports.

runtimes/rust/tests/runtime.rs
```diff
@@ -0,0 +1,1249 @@
...
+};
+use futures::{sync::mpsc, Future};
+use serde_derive::*;
+
+use std::{
+    collections::BTreeMap,
+    sync::{Arc, Mutex},
+};
+
+macro_rules! impl_binary_value_for_bincode {
```
slowli: I wonder whether it makes sense to put this macro in the MerkleDB crate. `bincode` serialization is similarly used in several other tests (e.g., in #1657).

runtimes/rust/tests/runtime.rs
```diff
@@ -0,0 +1,1249 @@
...
+
+    fn default_instance(&self) -> InstanceInitParams {
+        self.artifact_id()
+            .into_default_instance(Self::INSTANCE_ID, Self::INSTANCE_NAME)
+            .with_constructor(Init::default())
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Init {
```
slowli: Nit: Since we don't care about future / backward compatibility in tests, it could make sense to use `String` directly.

runtimes/rust/tests/runtime.rs
alekseysidorov: I think the best way is to add `bincode` support to the `BinaryValue` derive.

slowli: Yeah, makes sense.

CHANGELOG.md
```diff
@@ -77,6 +77,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   To get the sources of an artifact, use query `type=artifact&name=$name&version=$version`,
   where `$name` and `$version` are replaced with appropriate values. (#1590)
 
+- Rust runtime module was moved from the `exonum` crate into the separate
+  `exonum-rust-runtime` crate. (#1641)  
 - `update_service_status` now does not return a value. (#1659)
```
aleksuss: Split entities with an empty line, pls.

runtimes/rust/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2019 Exonum Team
```
slowli: Nit: We've reverted to the original text of the license in #1638.

Cargo.toml
```diff
@@ -18,6 +18,8 @@ members = [
     "test-suite/testkit",
     "test-suite/testkit/server",
 
+    "runtimes/rust/",
```
slowli: Nit: What's with the ending slash? Other `members` don't have it.

PR #1648, opened 2019-12-25
Process deployment failures properly [ECR-3926]

popzxc:
This PR fixes the bug with the node trying to deploy artifact on every height after a failure until the deadline height is achieved.
Also, this PR introduces a new private API endpoint, allowing us to get the deployment status for any request.

I decided to go with the simplest possible way to just fix the bug and focus on other important tasks for 1.0 release.
I hope that we can rewrite the whole deployment approach in a more proper manner after the release, and for now it's enough.

Regarding changelog: I've put the note about changing the `DeployConfirmation` transaction to the `breaking change` section (since it's a part of a public API), but we clearly state in docs that users should not work with `Supervisor` transactions directly, so this change is quasi-breaking.


services/supervisor/src/api.rs
```diff
@@ -100,6 +185,39 @@ impl PrivateApi for ApiImpl<'_> {
...
+            // Request in the list of failed deployments.
+            return Ok(DeployState::Failed(height).into());
+        }
+
+        let dispatcher_schema = self.0.data().for_dispatcher();
+        let core_schema = self.0.data().for_core();
+        let validator_count = core_schema.consensus_config().validator_keys.len();
+        // Check that artifact is deployed and was confirmed *for the same deploy request*.
+        if dispatcher_schema.get_artifact(&request.artifact).is_some()
+            && schema.deploy_confirmations.confirmations(&request) == validator_count
```
slowli: This looks awkward, TBH. Would it make sense to record the deploy status for each request explicitly in the supervisor schema, rather than just recording failures? If so, this would be a better solution, IMO; it would be more transparent and clean.

exonum/src/helpers/types.rs
```diff
@@ -113,6 +115,21 @@ impl Height {
...
 }
 
+impl BinaryValue for Height {
+    fn to_bytes(&self) -> Vec<u8> {
+        self.0.into_bytes()
+    }
+
+    fn from_bytes(value: Cow<'_, [u8]>) -> Result<Self, failure::Error> {
+        Ok(Self(
+            <u64 as BinaryValue>::from_bytes(value).expect("Error while deserializing value"),
```
slowli: Why does this method panic? `BinaryValue` deals with untrusted input, so this looks inappropriate.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -540,7 +540,7 @@ pub enum Action {
     StartDeploy {
         artifact: ArtifactId,
         spec: Vec<u8>,
-        and_then: Box<dyn FnOnce() -> ExecutionFuture + Send>,
+        and_then: Box<dyn FnOnce(Result<(), ()>) -> ExecutionFuture + Send>,
```
slowli: The field should be renamed `then` to be consistent with `future` combinators.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -566,7 +566,7 @@ impl Action {
             } => {
                 dispatcher
                     .deploy_artifact(artifact.clone(), spec)
-                    .and_then(|()| and_then())
+                    .then(|result| and_then(result.map_err(drop)))
```
slowli: Why is the error dropped? It could provide useful information to the handler.

services/supervisor/src/api.rs
```diff
@@ -15,18 +15,100 @@
...
+            spec,
+            deadline_height,
+        }
+    }
+}
+
+/// State of the deployment performed by `Supervisor`.
+#[derive(Debug, Clone, PartialEq)]
+#[derive(Serialize, Deserialize)]
+pub enum DeployState {
```
slowli: Nit: AFAIK, we try to uniformly use `#[serde(rename_all = "snake_case")]` for enums.

services/supervisor/src/proto/service.proto
```diff
@@ -33,9 +33,11 @@ message DeployRequest {
 }
 
 // Confirmation that the artifact was successfully deployed by the validator.
-message DeployConfirmation {
+message DeployResult {
   // Artifact identifier.
-  exonum.runtime.ArtifactId artifact = 1;
+  DeployRequest request = 1;
+  // Result of deployment.
+  bool success = 2;
```
slowli: I'm not sure whether the supervisor should make nodes agree on the error, but at least recording the error on the blockchain would be useful for auditing purposes.

services/supervisor/src/proto_structures.rs
```diff
@@ -226,24 +228,22 @@ pub struct ConfigProposalWithHash {
...
 impl_serde_hex_for_binary_value! { StopService }
 impl_serde_hex_for_binary_value! { ConfigPropose }
 impl_serde_hex_for_binary_value! { ConfigVote }
 
-impl From<DeployRequest> for DeployConfirmation {
-    fn from(v: DeployRequest) -> Self {
-        Self {
-            artifact: v.artifact,
-        }
+impl From<(DeployRequest, bool)> for DeployResult {
```
slowli: Nit: This falls in the category of obscure `From` conversions, IMO.

services/supervisor/tests/supervisor/deploy_failures.rs
```diff
@@ -0,0 +1,517 @@
+// Copyright 2019 The Exonum Team
```
slowli: Great job with these tests 👍

exonum/src/helpers/types.rs
popzxc: I just copy-pasted an implementation from `Round` several lines below :slightly_smiling_face:.
I'll fix it there as well then.

popzxc:
I'm dumb and have ran test only for added file :man_facepalming: 
Gonna fix it ASAP.

popzxc:
Should be fine now.

As a result of suggested improvements, the deployment state is stored within blockchain and can be obtained through `Supervisor` schema.

services/supervisor/src/transactions.rs
```diff
@@ -340,40 +342,51 @@ impl SupervisorInterface<CallContext<'_>> for Supervisor {
...
-            if confirmations == validator_count {
-                log::trace!(
-                    "Registering deployed artifact in dispatcher {:?}",
-                    deploy_request.artifact
-                );
-
-                // Remove artifact from pending deployments.
+        match deploy_result.result {
+            Ok(()) => {
+                let confirmations = schema.deploy_confirmations.confirm(&deploy_request, author);
```
slowli: Nit: I'd try to extract this block into a separate method.

services/supervisor/src/proto_structures.rs
```diff
@@ -241,9 +272,10 @@ impl_serde_hex_for_binary_value! { StopService }
 impl_serde_hex_for_binary_value! { ConfigPropose }
 impl_serde_hex_for_binary_value! { ConfigVote }
 
-impl From<(DeployRequest, bool)> for DeployResult {
-    fn from((request, success): (DeployRequest, bool)) -> Self {
-        Self { request, success }
+impl DeployResult {
+    /// Creates a new `DeployRequest` object.
+    pub fn new(request: DeployRequest, result: Result<(), ExecutionError>) -> Self {
```
slowli: Nit: Could make sense to have two constructors for OK and erroneous `result`s.

exonum/src/runtime/mod.rs
```diff
@@ -170,7 +170,8 @@ use crate::{
 
 mod blockchain_data;
 mod dispatcher;
-pub(crate) mod error;
+#[doc(hidden)]
+pub mod error;
```
slowli: I think it's possible to export only `execution_error` module in `error` without exposing the `error` module itself.

services/supervisor/src/deploy_state.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn to_pb(&self) -> Self::ProtoStruct {
+        let mut pb = Self::ProtoStruct::new();
+        match self {
+            DeployFailCause::Deadline => pb.set_deadline(Default::default()),
+            DeployFailCause::DeployError { height, error } => {
+                let mut error_info_pb = pb_supervisor::ErrorInfo::new();
+                let pb_error = ProtobufConvert::to_pb(error);
+                error_info_pb.set_error(pb_error);
+                error_info_pb.set_height(height.0);
+                pb.set_error(error_info_pb)
```
slowli: Nit: Missing semicolon.

services/supervisor/src/proto/schema.proto
```diff
@@ -18,8 +18,45 @@ package exonum.supervisor;
...
+    }
+}
+
+// Deployment state.
+message DeployState {
+    enum Type {
+        // Deploy is in process.
+        PENDING = 0;
+        // Deployment failed.
+        FAIL = 1;
```
slowli: What do you think about adding `TIMEOUT` variant for the deadline and getting rid of `DeployFailCause` altogether?

services/supervisor/src/api.rs
```diff
@@ -15,18 +15,87 @@
...
 };
 
+/// Query for retrieving information about deploy state.
+/// This is flattened version of `DeployRequest` which can be
+/// encoded via URL query parameters.
+#[derive(Debug, Clone, PartialEq)]
+#[derive(Serialize, Deserialize)]
+pub struct DeployInfoQuery {
+    /// Artifact identifier as string, e.g. `0:exonum-supervisor:0.13.0-rc.2"
+    pub artifact: String,
```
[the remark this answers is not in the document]
popzxc: Do you mind if these dependencies will be added in the next (follow-up) PR?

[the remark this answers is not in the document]
slowli: Sure; the follow-up PR is good.

services/supervisor/src/deploy_state.rs
```diff
@@ -135,12 +102,15 @@ impl ProtobufConvert for DeployState {
         let state = match pb.get_state() {
             PENDING => DeployState::Pending,
             SUCCESS => DeployState::Succeed,
+            TIMEOUT => DeployState::Timeout,
             FAIL => {
-                if !pb.has_cause() {
+                if !pb.has_error() {
                     return Err(format_err!("Protobuf representation of `DeployState` has type `FAIL` but has no cause set"));
```
slowli: Micro-nit: overly long string.

PR #1650, opened 2019-12-25
Migration tools [ECR-4056]

slowli:
## Overview

This PR introduces basic support of migrations from the core side / Rust runtime. Note that the migration workflow is **not** integrated into the runtime workflow (this is outside the scope of this PR).

Some notes:

- ~#1645 is incorporated into this PR (will probably have to rebase this PR once it's merged to exclude the corresponding commit)~ rebased
- API of the main crate needs decoupling with #1639
- There are several potential conflicts with #1641 (e.g., I've reworked `Node` constructor to be somewhat bearable) 

Follow-up tasks:

- [ECR-4078, ECR-4079] Improve migration script UX
- [ECR-4080] Test fault tolerance of migration scripts
- [ECR-4081] More thorough docs & doc examples

---

See: https://jira.bf.local/browse/ECR-4056


exonum/src/runtime/rust/mod.rs
```diff
@@ -322,45 +359,93 @@ impl Instance {
...
+            "Added available artifact {} with migration support",
+            artifact
+        );
+        self.available_artifacts
+            .insert(artifact, Box::new(service_factory));
+        self
+    }
+
+    /// Completes the build process.
+    #[doc(hidden)]
```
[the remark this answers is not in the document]
slowli: Probably shouldn't be 🙃 The idea is that the argument of the method (`mpsc::Sender`) exposes implementation details that we would like to remain hidden, but hiding the method may not be the best approach.

exonum/src/runtime/migrations.rs
```diff
@@ -0,0 +1,485 @@
...
+
+use std::{collections::BTreeMap, fmt};
+
+use crate::runtime::InstanceSpec;
+
+/// Atomic migration script.
+pub struct MigrationScript {
+    end_version: Version,
+    name: String,
+    logic: Box<dyn FnOnce(&mut MigrationContext) + Send>,
```
popzxc: Shouldn't it be `-> Result<(), ExecutionError>` (or `MigrationError`, dunno).

exonum/src/runtime/migrations.rs
```diff
@@ -0,0 +1,485 @@
...
+
+    /// Signals to return an error if the starting version is less than the specified version.
+    pub fn forget_before(mut self, version: Version) -> Self {
+        self.check_prerelease(&version);
+        self.min_start_version = Some(version);
+        self
+    }
+
+    /// Adds a migration script at the specified version.
+    pub fn migrate<F>(mut self, version: Version, script: F) -> Self
```
popzxc: Hm, IMHO this name is confusing, I'd name it `add_migration_script` or `add_migration`.

exonum/src/runtime/migrations.rs
```diff
@@ -0,0 +1,485 @@
...
+        self.scripts.insert(version, script);
+        self
+    }
+
+    /// Selects a list of migration scripts based on the provided start version of the artifact.
+    pub fn select(
+        self,
+        start_version: &Version,
+    ) -> Result<Vec<MigrationScript>, DataMigrationError> {
+        if !self.support_prereleases && start_version.is_prerelease() {
```
popzxc: In `check_prerelease` similar check is an assertion, and here we just return an error. I think it'll be better to have a consistent reactions.

exonum/src/runtime/migrations.rs
```diff
@@ -0,0 +1,485 @@
...
+        if !self.support_prereleases {
+            assert!(
+                !version.is_prerelease(),
+                "Prerelease versions require using `with_prereleases` constructor"
+            );
+        }
+    }
+
+    /// Signals to return an error if the starting version is less than the specified version.
+    pub fn forget_before(mut self, version: Version) -> Self {
```
popzxc: Maybe just `set_min_version`?

exonum/src/runtime/rust/mod.rs
```diff
@@ -288,20 +289,56 @@ mod stubs;
...
+        self.0.create_instance()
+    }
+}
+
+impl<T> MigrateData for WithoutMigrations<T> {
+    fn migration_scripts(
+        &self,
+        _start_version: &Version,
+    ) -> Result<Vec<MigrationScript>, DataMigrationError> {
+        Err(DataMigrationError::UnsupportedStart)
```
popzxc: I think it's better to have a separate type of error for such a situation.

exonum/src/runtime/rust/mod.rs
```diff
@@ -288,20 +289,56 @@ mod stubs;
 #[cfg(test)]
 mod tests;
 
+trait FactoryWithMigrations: ServiceFactory + MigrateData {}
+impl<T: ServiceFactory + MigrateData> FactoryWithMigrations for T {}
+
+/// Wrapper around a service factory that does not support migrations.
+#[derive(Debug)]
+struct WithoutMigrations<T>(T);
```
popzxc: Can user one day change their mind and switch to `WithMigrations`? If not, I'd suggest to add a big warning like "This cannot be undone! Think twice, think thrice and consult your psychotherapist before doing that." just to avoid tears and blames later :)

exonum/src/runtime/migrations.rs
slowli: This is intentional. Using a prerelease version in in `migrate` is the decision made by the service developer, so we want to inform that he is wrong in his ways as quickly and as prominently as possible. (Ideally, during compilation time, but I didn't come up with the way to do that just yet, although it seems possible via with proc macros.)

In contrast, here `start_version` is *not* chosen by the service developer, so panicking would be inappropriate.

exonum/src/runtime/rust/mod.rs
slowli: What do you mean by switching to `WithMigrations`? If we consider recompiling the node binary acceptable, then sure, the user should just switch `with_factory` to `with_migrating_factory` in the Rust runtime builder and replace the node binary.

popzxc: Yeah, that exactly what I meant.

Then I think it will be nice to have a guide-level explanation of how to make your non-migration (at the moment) service support migration, e.g. "Add a new version of artifact, add a factory for it to the rust runtime, create migration script, change invoking of `with_factory` for artifact from which you want to migrate to `with_migrating_factory`, etc".

slowli: This makes sense, but IMO this should be documented once we have workflow for the dispatcher / supervisor.

exonum/src/runtime/migrations.rs
slowli: Would you mind if I change it in a follow-up PR corresponding to ECR-4080?

popzxc: Sure!

PR #1643, opened 2019-12-24
Implement Protobuf serialization for IndexProof

Kormick:
## Overview

Implemented Protobuf serialization for Verified, BlockProof and IndexProof.

WIP: Getting compilation error after adding merkledb proto files to the build and proto description for IndexProof that uses MapProof.

PR #1651, opened 2019-12-25
Document AdditionalHeaders and hide OrderedMap [ECR-4061]

popzxc:


exonum/src/blockchain/block.rs
```diff
@@ -100,6 +112,10 @@ impl AdditionalHeaders {
 ///
 /// The header only contains the amount of transactions and the transactions root hash as well as
 /// other information, but not the transactions themselves.
+///
+/// Note that this structure is export-only, meaning that one can rely on the serialization format
+/// provided by corresponding `protobuf` definitions, but cannot expect `exonum` to accept
+/// and process `Block` structure created outside of the `exonum` core.
```
dmitry-timofeev: That basically says "_Block_ protobuf message shall not be used to create messages outside of the core" because the present Block _message_ definition does not say in which order the entries are ordered.

This restriction is buried inside Exonum Rust code, and not prominently specified in the Block _message_ — which is part of the API. It also has an important implication for client code — "Always keep the block in its serialized form [if you need to verify anything in it]" and "Do not serialize block from JSON/object form".

Finally, not being able to create block proofs is likely to make testing of BlockProof verification in light clients harder or even hardly possible.

Whilst I find 'Do not serialize blocks' mostly reasonable, I am not sure I sense the benefits of unspecified ordering with its implications. Does the core need to under-specify this? @slowli ?

popzxc: No, this says not exactly that. This says that `Block` structures provided by core can be deserialized and processed as user wants, all the operations with `Block` are performed as per `protobuf` file structure definitions; but at the same time if user will create a `Block` structure based on protobuf definition and pass it to the core, there is no guarantee that this structure will be considered correct.

Rules are the same as for `PublicKey` from `exonum-crypto`: we guarantee, that `PublicKey` structures provided and serialized by core can be deserialized into actual `PublicKey`, but we do not guarantee that every object serialized to the `protobuf` `PublicKey` will be processed by `crypto`, since by it's nature it's just an array of bytes, and not every array of bytes is `PublicKey`.

Thus, as a conclusion: user is free to do anything he wants with the obtained serialized message (if it was obtained from core). `exonum` still follows the rules of the `protobuf` specification, it is complete enough to work with an obtained object; but this specification may be not complete enough to create a valid `Block` message from scratch.

popzxc: Also other note: user can try to create a `Block` object based on the `protobuf` specification -- there is no such a restriction. The only restriction here is introduced once user wants to provide this constructed object to the core, it is a restriction of `Block` as a part of `core` API and concrete `rust` code, not as a part of `protobuf` specification.

dmitry-timofeev: Block in Exonum imposes a restriction on the order of entries in additional headers which is not specified in the message, making it impossible to create a valid Block message without digging into Rust code.

popzxc: Can you please describe the scenario when one will want to create a `Block` message from scratch?
This structure is highly tied to the main blockchain, which is managed by core only, and in my opinion there should be no need to "create a valid Block message" outside of the core.

dmitry-timofeev: For instance, unit testing of block proof verification — to create a correct block message:

popzxc: For unit testing, one can obtain several `Block` objects from `exonum` and store them as a test vector. Maybe, `exonum` core should provide such a test vector for those purposes.

Despite of that, IMHO making some detail of implementation public just for easier testing approach is a bad idea.

dmitry-timofeev: "Getting them from exonum" is far from unit testing — you will need a testkit for the target language (so no unit tests for JS or C# — or do you suggest firing an exonum node for unit tests?).

>
IMHO making some detail of implementation public just for easier testing approach is a bad idea.

It is not exactly an implementation detail when *the block binary representation is a public API* — as the clients getting a Block are expected to serialize it correctly: https://github.com/exonum/exonum/pull/1643/files#diff-4bc3b98fd5517a9bb8cc14a0a18c89aeR34 (here they probably _would_ given an original message)

popzxc: I suggest to run exonum once, collect the data required in the file once, and then use the file for unit test, plain and simple. You can find this approach, e.g., in [ethereum project](https://github.com/ethereum/tests/tree/develop/BlockchainTests).

popzxc: More than that, this file will be language-independent and can be created once for all the target platforms (that's what I meant when I told that maybe `exonum` should provide such a test vector).

dmitry-timofeev: If we consider additional headers in the Block-message as merely a list (as this PR achieves), then I think it might work. byte[] <-> Block-message conversion will preserve that order; Block-message -> Block _shall_ preserve that order; {Block properties} -> Block shall use the specification in core Block-struct regarding their ordering. My mistake was that I still considered them a map. If the ordering is specified in Block-struct, I think it is enough.

popzxc: Yes, that's exactly the idea :) Glad that you find this acceptable.

exonum/src/proto/schema/exonum/blockchain.proto
```diff
@@ -29,7 +35,7 @@ message Block {
   exonum.crypto.Hash tx_hash = 5;
   exonum.crypto.Hash state_hash = 6;
   exonum.crypto.Hash error_hash = 7;
-  OrderedMap additional_headers = 8;
+  AdditionalHeaders additional_headers = 8;
```
bullet-tooth: Changing the type without changing the field number leads to breaking backward compatibility. Is it fine?
See: https://stackoverflow.com/questions/27362742/changing-data-type-for-a-field-in-protocol-buffer

popzxc: This field was introduced after the `0.13.0-rc.2` version (see [blockchain.proto](https://github.com/exonum/exonum/blob/v0.13.0-rc.2/exonum/src/proto/schema/exonum/blockchain.proto#L23) on release tag), thus changing the type between releases is perfectly fine.

slowli: @bullet-tooth I think it's fine; we've never released a version with `OrderedMap`; it was added after the 0.13 RC.

exonum/src/proto/schema/exonum/key_value_sequence.proto
```diff
@@ -26,6 +26,7 @@ message KeyValue {
   bytes value = 2;
 }
 
-message OrderedMap {
+// A sequence of repeated key-value pairs.
```
slowli: Nit: "repeated" looks redundant; "sequence" implies that the field is repeated, and it's seen from the type declaration.

exonum/src/helpers/ordered_map.rs
```diff
@@ -21,22 +21,26 @@ use protobuf::Message;
 
 use std::{borrow::Cow, collections::BTreeMap, iter::FromIterator};
 
-/// Protobuf wrapper type to store small maps of non-scalar keys and values.
-/// Stored keys are ordered and duplicate keys are forbidden.
+/// Protobuf-encodable type to store small maps of non-scalar keys and values.
+///
+/// This structure uses on `KeyValueSequence` from `key_value_sequence.proto` as
+/// a backend, but adds the verification logic to it:
+/// - Keys are sorted in a lexicographical order;
```
slowli: Nit: a blank line before the list could improve readability.

PR #1653, opened 2019-12-26
Remove extern crate: components [ECR-4004]

Kormick:
## Overview

Removed extern crate in components. 
Also, fixed order of use/mod.


aleksuss:
Don't forget about doctests. There could be hidden `extern crate ...`s as well 😄

PR #1659, opened 2019-12-26
Make infallible runtime/dispatcher methods panicking [ECR-4018]

popzxc:
Now methods which logically should never fail do not return an error, but do panic instead.


exonum/src/runtime/mod.rs
```diff
@@ -389,21 +389,17 @@ pub trait Runtime: Send + fmt::Debug + 'static {
...
-    /// description allowing the node administrator to determine
-    /// the root cause of the error and recover the node(s) by eliminating the cause.
-    ///
-    /// This error / panic must not be common to all nodes in the network.
-    /// The errors common for the whole network should be produced during the preceding
-    /// `initiate_adding_service` call.
+    /// This method does not return a value, meaning that any error occurred during this method execution
+    /// is considered critical and should lead to the node stopping.
+    ///
+    /// It is assumed that if `initiate_adding_service` didn't return an error previously, it is able
```
slowli: Grammar: "it is able" -> "the runtime is able".

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -228,16 +237,18 @@ impl Dispatcher {
...
-    ) -> Result<(), ExecutionError> {
-        // TODO: revise dispatcher integrity checks [ECR-3743]
+    ///
+    /// # Panics
+    ///
+    /// This method assumes that `deploy_artifact` was previously called for the corresponding
+    /// `ArtifactId` and deployment was completed successfully.
+    /// If any error happens within `commit_artifact`, it is considered either a bug in the
+    /// `Supervisor` service or `Dispatcher` itself, and as a result, this method will panic.
+    pub(crate) fn commit_artifact(fork: &Fork, artifact: ArtifactId, deploy_spec: Vec<u8>) {
```
slowli: I'm slightly hesitant about removing an error from this method. If I understand correctly, this places certain additional requirements on the supervisor (sort of; since panics and errors are processed in the same way by the dispatcher, breaking these requirements will lead to the rollback rather than to unexpected behavior).

exonum/src/runtime/rust/mod.rs
```diff
@@ -504,10 +504,14 @@ impl Runtime for RustRuntime {
...
         status: InstanceStatus,
-    ) -> Result<(), ExecutionError> {
+    ) {
         match status {
             InstanceStatus::Active => {
-                let instance = self.new_service(spec)?;
+                let instance = self
+                    .new_service(spec)
+                    .expect("BUG: Attempt to create a new service instance failed; \
+                             within `instantiate_adding_service` we were able to create a new instance,
```
slowli: Nit: no `\` at the end of the line. Also, is this "hanging" style of formatting allowed by the formatter? I thought it would fix it to something like:

```rust
let instance = self
    .new_service(spec)
    .expect(
        "BUG: Attempt to create a new service instance failed; \
         within `instantiate_adding_service` we were able to create a new instance, \
         but now we are not.",
     );
```

exonum/src/blockchain/builder.rs
```diff
@@ -71,7 +71,7 @@ impl BlockchainBuilder {
         let has_genesis_block = !Schema::new(&snapshot).block_hashes_by_height().is_empty();
 
         if has_genesis_block {
-            blockchain.dispatcher.restore_state(&snapshot)?;
+            blockchain.dispatcher.restore_state(&snapshot);
         } else {
             blockchain.create_genesis_block(self.genesis_config)?;
```
slowli: Should this method still return a `Result` though? I don't think there are many (if any) scenarios, in which the error here can be meaningfully recovered from.

examples/sample_runtime/src/main.rs
```diff
@@ -119,10 +119,10 @@ impl Runtime for SampleRuntime {
         _snapshot: &dyn Snapshot,
         spec: &InstanceSpec,
         status: InstanceStatus,
-    ) -> Result<(), ExecutionError> {
+    ) {
         match status {
             InstanceStatus::Active => {
-                let instance = self.start_service(spec)?;
+                let instance = self.start_service(spec).unwrap();
```
slowli: Nit: It could be helpful to add a comment why `unwrap()` is appropriate here, but not when `start_service` is called in `Runtime::initiate_adding_service`.

exonum/src/runtime/dispatcher/mod.rs
popzxc: IMHO, `Supervisor` is kind of entity on which we can surely trust -- if it behaves incorrectly, we are doomed anyway.
Knowing the context of `commit_artifact` -- artifact should be previously *successfully* deployed, and the deploy should happen not only for this node, but for the whole network, and all of those conditions are checked by the `Supervisor` only -- it makes sense to make `Supervisor` responsible for this method not panicking.
Overall, if this method will panic, then `Supervisor` somehow did let it panic, and this is already a big problem.

exonum/src/runtime/error.rs
```diff
@@ -273,6 +273,18 @@ impl ErrorMatch {
         Self::new(fail.kind(), fail.description())
     }
 
+    /// Creates a matcher for `Unexpected` kind of errors.
+    /// By default it will match any description.
+    pub fn unexpected() -> Self {
```
slowli: Nit: Maybe `any_unexpected` would be a better name?

exonum/src/runtime/dispatcher/tests.rs
```diff
@@ -228,9 +227,8 @@ impl Runtime for SampleRuntime {
                 self.services.insert(spec.id, new_status);
                 self.new_services.insert(spec.id, new_status);
             }
-            Ok(())
         } else {
-            Err(DispatcherError::IncorrectRuntime.into())
+            panic!(DispatcherError::IncorrectRuntime)
```
slowli: Nit: Does panicking with an error (instead of a string) make sense here? Is the error extracted via `catch_unwind`? Otherwise, it could make sense to panic with a string, IMO.

PR #1668, opened 2019-12-27
Use uint32 for tx position in block: [ECR-4083]

dmitry-timofeev:
## Overview

2^32 - 1 is plenty to index tx position in block.

The number of tx is already limited to a way smaller number.

See ECR-4083

[skip ci]

---

### Definition of Done

### DIFF
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -307,7 +307,7 @@ impl BlockchainMut {
         }
 
         // Save & execute transactions.
-        for (index, hash) in (0..).zip(tx_hashes.iter()) {
+        for (index, hash) in (0..).zip(tx_hashes) {
             self.execute_transaction(*hash, height, index, &mut fork, tx_cache);
         }
 
--- a/exonum/src/blockchain/mod.rs
+++ b/exonum/src/blockchain/mod.rs
@@ -35,6 +35,7 @@ use futures::Future;
 
 use std::{
     collections::{BTreeMap, HashMap},
+    convert::TryInto,
     iter,
     sync::Arc,
 };
@@ -308,7 +309,13 @@ impl BlockchainMut {
 
         // Save & execute transactions.
         for (index, hash) in tx_hashes.iter().enumerate() {
-            self.execute_transaction(*hash, height, index, &mut fork, tx_cache);
+            self.execute_transaction(
+                *hash,
+                height,
+                index.try_into().unwrap(),
+                &mut fork,
+                tx_cache,
+            );
         }
 
         // During processing of the genesis block, this hook is already called in another method.
@@ -368,7 +375,7 @@ impl BlockchainMut {
         &self,
         tx_hash: Hash,
         height: Height,
-        index: usize,
+        index: u32,
         fork: &mut Fork,
         tx_cache: &mut BTreeMap<Hash, Verified<AnyTx>>,
     ) {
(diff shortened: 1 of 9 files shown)


test-suite/consensus-tests/tests/basic.rs
```diff
@@ -278,7 +278,10 @@ fn test_store_txs_positions() {
     let locations = schema.transactions_locations();
     for (expected_idx, hash) in hashes.iter().enumerate() {
         let location = locations.get(hash).unwrap();
-        assert_eq!(expected_idx as u64, location.position_in_block());
+        assert_eq!(
+            u32::try_from(expected_idx).unwrap(),
```
popzxc: You can use `(0..).zip(hashes.iter())` to get the `u32` values generated instead of `usize` produced by `enumerate`.

With that, there will be no need in `unwrap`.

aleksuss: Does it worth that ? I don't think so. In the end it's just a test.

popzxc: IMHO it doesn't worsen the readability, and lets us get rid of bogus `try_from` (since we know that the values can't exceed the `u32` limit and `unwrap`ping it anyway).
Despite of this being a test part, I think we should treat all our code with the same love and care :)

However, if you insist, you can leave this as is, I don't mind.

popzxc: Other option (if you don't like the `(0..)` going first) is to zip them in the other order: `hashes.iter().zip(0..)`. Then you'll just have to swap `expected_idx` and `hash`.

exonum/src/blockchain/mod.rs
```diff
@@ -298,7 +299,13 @@ impl BlockchainMut {
 
         // Save & execute transactions.
         for (index, hash) in tx_hashes.iter().enumerate() {
-            self.execute_transaction(*hash, height, index, &mut fork, tx_cache);
+            self.execute_transaction(
+                *hash,
+                height,
+                index.try_into().unwrap(),
```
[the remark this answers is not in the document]
aleksuss: Index is `usize` because its an index of `Vec` which has a `usize` type. And we don't make any conversions inside of `execute_transaction` method, cause signature of `execute_transaction` has been changed as well.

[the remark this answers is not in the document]
popzxc: Ah, sorry, I've accidentally opened a file from `master` instead of the actual one; nevermind.

But nevertheless, we can avoid `try_into().unwrap()` here as well, by using `zip`.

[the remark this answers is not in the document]
aleksuss: Ok. Here it obviously makes sense 👍🏻

exonum/src/blockchain/mod.rs
```diff
@@ -307,7 +307,7 @@ impl BlockchainMut {
         }
 
         // Save & execute transactions.
-        for (index, hash) in tx_hashes.iter().enumerate() {
+        for (index, hash) in (0..).zip(tx_hashes.iter()) {
```
slowli: Nit: Isn't `iter()` redundant here? That is, `(0..).zip(tx_hashes)` would work as well.

exonum/src/proto/schema/exonum/blockchain.proto
```diff
@@ -34,15 +34,15 @@ message Block {
...
+  uint32 position_in_block = 2;
 }
 
 // Location of an isolated call within a block.
 message CallInBlock {
   oneof call {
     // Call of a transaction within the block. The value is the zero-based
     // transaction index.
-    uint64 transaction = 1;
+    uint32 transaction = 1;
```
slowli: Will we solve possible issues with field naming in a separate PR? AFAIK, @dmitry-timofeev is not quite satisfied with the present naming; I sort of agree. Would something like `transaction_index` / `service_before_transactions` / `service_after_transactions` be better?

test-suite/consensus-tests/tests/basic.rs
```diff
@@ -276,9 +276,9 @@ fn test_store_txs_positions() {
     let snapshot = sandbox.blockchain().snapshot();
     let schema = snapshot.for_core();
     let locations = schema.transactions_locations();
-    for (expected_idx, hash) in hashes.iter().enumerate() {
+    for (expected_idx, hash) in (0u32..).zip(hashes.iter()) {
```
slowli: Nit: Would `(0_u32..).zip(&hashes)` work here? IMO, this would be more idiomatic.

exonum/src/proto/schema/exonum/blockchain.proto
aleksuss: These changes was made by @dmitry-timofeev 's commit 😄 And he's changed only type of the field.

aleksuss: Anyway, I think it would be better to change naming in another PR.

PR #1671, opened 2020-01-02
Improve `node` module in main crate [ECR-3947]

slowli:
## Overview

This PR reduces the visibility of items in the `node` module. The overall idea is that any types used by `NodeHandler` are considered an implementation detail, while configs, `Node` and `ApiSender` are public (more details on public types are provided in the module docs).

Wishlist:

- [ ] Builder for configs (there is one in the `helpers` module; should it be generalized?)
- [ ] Examples of using `Node` in the docs

---

See: https://jira.bf.local/browse/ECR-3947


cli/src/config.rs
```diff
@@ -98,11 +100,8 @@ impl Into<CoreNodeConfig> for NodeConfig {
             network: self.private_config.network,
             api: self.private_config.api,
             mempool: self.private_config.mempool,
-            services_configs: Default::default(),
-            database: self.private_config.database,
```
slowli: `database` and `master_key_path` fields of `NodeConfig` were not used by the core code. Is it safe to remove them? Unlike `services_configs`, which was not set by the CLI crate (therefore, I assume it *is* safe to remove), these two fields *are* set by the CLI crate, so the may have a purpose somewhere.

examples/sample_runtime/src/main.rs
```diff
@@ -301,6 +298,7 @@ fn main() {
 
     let blockchain_ref = blockchain.as_ref().to_owned();
     let node = Node::with_blockchain(blockchain, channel, node_cfg, None);
+    let shutdown_handle = node.shutdown_handle();
```
slowli: `ShutdownHandle` is now separated from `ApiSender`, although it uses `ApiSender` internally. This is to provide separation of concerns and prevent misuse.

examples/sample_runtime/src/main.rs
```diff
@@ -313,7 +311,7 @@ fn main() {
             spec: Vec::default(),
         };
         let tx = service_keypair.request_artifact_deploy(SUPERVISOR_INSTANCE_ID, request);
-        api_sender.broadcast_transaction(tx).unwrap();
+        api_sender.broadcast_transaction(tx).wait().unwrap();
```
slowli: `ApiSender` now provides async broadcast. This is because it's easy to go from async to sync (e.g., by using `wait()` as here), but not the other way around.

exonum/src/events/mod.rs
```diff
@@ -45,7 +45,39 @@ pub type SyncSender<T> = Wait<Sender<T>>;
 /// This kind of events is used to schedule execution in next event-loop ticks
 /// Usable to make flat logic and remove recursions.
 #[derive(Debug, PartialEq)]
-pub enum InternalEvent {
+pub struct InternalEvent(pub(crate) InternalEventInner);
+
+impl InternalEvent {
+    // FIXME: review visibility
```
popzxc: Is this going to be done within the current PR?

slowli: Yeah. Since `InternalEvent` is hidden from the docs + these methods are used in consensus tests, I plan to leave the methods public and remove `#[doc(hidden)]` (since it's obsolete).

cli/src/config.rs
vitvakatu: `service_configs` is the artifact from pre-DS past. `database`, however, allows (or should allow) tuning RocksDB options and was supported in the past. If it is not used, I'd consider it a bug.

slowli: As far as I see, all `Node` constructors have DB supplied to them in the assembled form. The `cli` crate itself [does use `database` options from `NodePrivateConfig`](https://github.com/exonum/exonum/blob/3c5bf23c090c6a6cc8ae184ce18f1cab34ac2fc0/cli/src/lib.rs#L173); this code is not changed by this PR. Is this OK?

vitvakatu: OK, looks like storing the database options in NodeConfig is not needed.

PR #1673, opened 2020-01-02
Remove extern crate log in exonum [ECR-4004]

Kormick:
Removed `extern crate log` in exonum.


slowli:
I feel that using fully qualified macros, such as `log::info!(...)`, would be more readable, but this is the matter of opinion, obviously.

popzxc:
I agree with @slowli that for `log` the fully qualified names will be better, but IMHO it's a topic for a different task -- this PR only removes `extern crate` and nothing more than it.

I'd also propose to later use `failure::` as a prefix for `ensure`/`bail`. Once all the 2018-edition routine is complete, I think I'll create a PR to perform such a transformation everywhere :)

PR #1674, opened 2020-01-02
Remove extern crate serde_json in exonum

Kormick:
Removed `extern crate serde_json` in exonum.


exonum/src/runtime/types.rs
```diff
@@ -23,6 +23,8 @@ use exonum_proto::ProtobufConvert;
 use failure::{bail, ensure, format_err};
 use semver::Version;
 use serde_derive::{Deserialize, Serialize};
+#[cfg(test)]
+use serde_json::json;
```
popzxc: Can you please wrap the tests existing in this file in 
```rust
#[cfg(test)]
mod tests {
..
}
```

and move this import there?

PR #1679, opened 2020-01-03
Allow to derive `BinaryValue` from `bincode` serialization [ECR-4086]

slowli:
## Overview

This PR allows to derive `BinaryValue` from `bincode` serialization. The new serialization format is then used in most tests. There are also other minor changes; e.g., all types necessary for deriving `ObjectHash` / `BinaryValue` are now reexported from the MerkleDB crate.

---

See: https://jira.bf.local/browse/ECR-4086


components/merkledb/benches/schema_patterns.rs
```diff
@@ -13,17 +13,15 @@
 // limitations under the License.
 
 use criterion::{black_box, Bencher, Benchmark, Criterion, Throughput};
-use exonum_derive::FromAccess;
+use exonum_derive::*;
```
popzxc: Nit/style: I don't really like wildcard imports, IMHO specifying concrete imports really helps when you try to get what's going on in the module seeing it first time (or after a long time).

examples/timestamping/backend/src/main.rs
```diff
@@ -14,10 +14,11 @@
...
 use exonum_cli::NodeBuilder;
 
-fn main() -> Result<(), failure::Error> {
+fn main() {
     exonum::helpers::init_logger().unwrap();
     NodeBuilder::new()
         .with_service(exonum_time::TimeServiceFactory::default())
         .with_service(exonum_timestamping::TimestampingService)
         .run()
+        .unwrap();
```
popzxc: Nit: `expect` will be a bit more friendly :)

PR #1680, opened 2020-01-06
Revise common runtime errors [ECR-3984]

popzxc:
This PR introduces the following changes:

- `runtime/error.rs` was splitted into smaller sub-modules, since it became pretty big and hard to understand;
- `Dispatcher::Error` enum was removed, instead it was replaced by two enums: `CoreError` (meant to be used by core most of time) and `CommonError` (which represents the errors that can occur in different contexts);
- `Common` variant was added to `ErrorKind` enum;
- error codes for `Supervisor` were changed and split into sub-groups, allowing to add relevant errors without mixing their codes.


services/supervisor/src/errors.rs
```diff
@@ -15,39 +15,60 @@
...
+/// - 0 - 15: Common `Supervisor` errors.
+/// - 16 - 31: Errors related to artifacts.
+/// - 32 - 47: Errors related to service instances.
+/// - 48 - 64: Errors related to configuration changes.
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[derive(ExecutionFail)]
 pub enum Error {
-    /// Artifact has been already deployed.
-    AlreadyDeployed = 0,
+    // # General errors that can occur within supervisor interaction.
```
slowli: What do you think about splitting the error enum itself into several enums for different error categories?

vitvakatu:
Is runtime supposed to return `CommonError`s, or will it work if we return custom `Runtime` errors instead?

Adding new ErrorKind affects current EJB's policy on errors, that was just established.
cc @dmitry-timofeev

popzxc:
From docs:

```rust
/// ## `Common` errors
///
/// `Common` errors set provides various error codes that can occur within `Runtime`
/// and `Service` lifecycle. They are intended to be reused in the service and runtime code instead
/// of defining new error codes with the same effect.
```

Thus, nothing will break if you won't use `Common` errors type, but it is recommended to use this type of errors rather that define the same error as runtime/service-specific.

Note that this kind of errors was not invented out of nowhere, it's just error codes from old `DispatcherError` which occurred to be used not only by core, but also by runtimes/services.

popzxc:
> The present kinds try to tell the source of the error.

That's not quite correct, as for now. Even if we ignore the errors that now called `common` (which are used in core, Rust runtime and services), we have e.g. `Middleware` service which [spawns](https://github.com/exonum/exonum/blob/9cd688bf349b668d4f89c81aaec7b72dd82843eb/services/middleware/src/transactions.rs#L186) a dispatcher error to mimic a dispatcher behavior.

To obtain the source of error, `ExecutionError` [provides](https://github.com/exonum/exonum/blob/master/exonum/src/runtime/error.rs#L519) separate methods (`call_site` and `runtime_id`), which are set by the core and guaranteed to be correct.

Regarding the usefulness to the client: as clients of `exonum` (as a framework) are service authors, this is useful for them, since some basic errors are already present, and thus author is able to write less code and less documentation. The `common` errors are not just called common, these errors can occur in **any** service, and IMHO it makes sense to provide them out of the box.

dmitry-timofeev:
By clients I considered the service clients that submit transactions and _need_ to distinguish between different types of errors (otherwise a single "unexpected" error with optional description is enough). I see that by using common errors you don't need to create your own error variant. But a transaction still must document its errors so that these clients can build reliable handling of them. If a service now can produce errors of different kinds: service + common — then it becomes more difficult to handle because the clients need to consider different _kinds_ of errors on top of kind 'service' and error codes.

If error kind does not communicate the error source (it is a known issue in the present code — but one I suggested to fix in ECR-4038) then what does it communicate? 

Also, why Exonum services are allowed to create errors of any kind? In Java they are allowed to create 'service' only, and may cause 'unexpected', but never — runtime or dispatcher.

popzxc:
>then what does it communicate

It communicates the kind of error, which means the defined error set. There are several sets:
- Core (fixed set of error codes, that *usually* correspond to the core code);
- Common (fixed set of error codes that are well-known and can occur anywhere);
- Runtime (non-fixed set of errors, specific to the runtime in which it occurred);
- Service (non-fixed set of errors, specific to the service in which it occurred);
- Unexpected (error that happened unpredictably, there is no other documentation other than provided description).

What it *essentially* means -- it's a hint for a place where you can find docs for this error code.

For finding the source of error, as I pointed above, we have a special interface designed exactly for this purpose.

> Also, why Exonum services are allowed to create errors of any kind?

One of the reasons you can see above as well: for example, we have a `Middleware` service which mimics core behavior and can spawn `DispatcherError` (`CoreError` in terms of this PR).

dmitry-timofeev:
The kinds/categories make sense, thanks for clarification, but, given they must be handled by the clients (and understood), I'd prefer a simpler and stricter definition that is currently present.

Call site, OTOH, tells which method was invoked, not where the error occurred. Therefore, 'core' from the core during transaction execution is indistinguishable from 'core' in a service transaction.

popzxc:
>Call site, OTOH, tells which method was invoked, not where the error occurred.

Hmm. If you will look at the current dispatcher code ([1](https://github.com/exonum/exonum/blob/master/exonum/src/runtime/dispatcher/mod.rs#L227), [2](https://github.com/exonum/exonum/blob/master/exonum/src/runtime/dispatcher/mod.rs#L337), [3](https://github.com/exonum/exonum/blob/master/exonum/src/runtime/dispatcher/mod.rs#L330)), you'll see that the location can be set partially.

Together with the guarantees that core provide (e.g. that it will not request runtime to execute transaction of non-existent service), it gives us the following:
- If `runtime_id` and `call_site` are both set, then error occurred in the service code (since the request was designated to the service, and core was sure that target service exists in a concrete runtime).
- If `runtime_id` is set and `call_site` is not set, then error occurred in the runtime code (e.g. within deployment, as in the [1] link above).
- If none of `runtime_id` and `call_site` is set, then the error occurred in the core code (for example, core can't find suitable runtime to process the request, as in [3] link above).

Or am I missing something?

popzxc:
Note that in [3] error occurs in `execute` method, and still have to destination information set, exactly because error originates in core code.

services/explorer/tests/api.rs
```diff
@@ -522,11 +522,10 @@ fn test_explorer_add_invalid_transaction() {
         .query(&json!({ "tx_body": data }))
         .post::<TransactionResponse>("v1/transactions")
         .expect_err("Expected transaction send to finish with error.");
-    let error_body = "Execution error with code `core:3` occurred: Suitable runtime \
-                      for the given service instance ID is not found.";
+    let error_body = ExecutionError::from(CoreError::IncorrectInstanceId).to_string();
     assert_matches!(
         response,
-        ApiError::BadRequest(ref body) if body == error_body
+        ApiError::BadRequest(ref body) if body == &error_body
```
slowli: Micro-nit: `*body == error_body` looks more idiomatic.

services/supervisor/src/errors.rs
```diff
@@ -14,29 +14,24 @@
...
+/// Error codes 0-15.
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[derive(ExecutionFail)]
-pub enum Error {
-    // # General errors that can occur within supervisor interaction.
-    // Error codes 0-31.
-    // ------------------
+pub enum CommonError {
     /// Transaction author is not a validator.
     UnknownAuthor = 0,
```
slowli: Nit: Is this error variant substantially different from `UnauthorizedCaller` in the core `CommonError`?

exonum/src/runtime/error/error_match.rs
```diff
@@ -128,6 +94,38 @@ impl ErrorMatch {
...
+    }
+}
+
+impl PartialEq<ExecutionError> for ErrorMatch {
+    fn eq(&self, other: &ExecutionError) -> bool {
+        other.eq(self)
+    }
+}
+
 pub(super) enum StringMatch {
```
slowli: Nit: It seems possible to move `ErrorMatch` to this module and make `StringMatch` private, if a single test covering matching functionality is moved in this module, too.

dmitry-timofeev:
CallSite interpretation makes sense, thanks. I think the docs for it could be improved (in the messages) to say in which cases it might not be set.

Also, if you go with common error kind, the bit on what [error kind communicates](https://github.com/exonum/exonum/pull/1680/files#diff-869b87af3e145bb0196793b8fc7f8116R79) might need to be updated.

popzxc:
Indeed, I'll update the docs in a follow-up PR.

exonum/src/runtime/error/execution_error.rs
```diff
@@ -12,30 +12,238 @@
...
+        &self.description
+    }
+
+    /// Returns the ID of a runtime in which this error has occurred. If the runtime is not known
+    /// (e.g. error originates in the core code), returns `None`.
+    pub fn runtime_id(&self) -> Option<u32> {
+        self.runtime_id
+    }
+
+    #[inline]
```
alekseysidorov: What is reason to use `inline` directive explicitly?

popzxc: Dunno, it was there before, I left it as-is.

PR #1681, opened 2020-01-06
Allow to use slash in artifact names [ECR-4077]

popzxc:


components/merkledb/src/validation.rs
```diff
@@ -40,10 +40,10 @@ pub fn is_valid_index_name_component(prefix: &str) -> bool {
 /// Checks that a character is allowed in an index name.
 ///
 /// Only these combination of symbols are allowed:
-/// `[0-9]`, `[a-z]`, `[A-Z]`, `_`, `-`.
+/// `[0-9]`, `[a-z]`, `[A-Z]`, `/`, `_`, `-`.
 pub fn is_allowed_index_name_char(c: u8) -> bool {
     match c {
-        b'0'..=b'9' | b'A'..=b'Z' | b'a'..=b'z' | b'-' | b'_' => true,
+        b'0'..=b'9' | b'A'..=b'Z' | b'a'..=b'z' | b'-' | b'_' | b'/' => true,
```
slowli: Will the index names containing `/` really work? I thought the names are mapped one-to-one to RocksDB column families, which are mapped to file names. Correspondingly, I'd want to see that index names like `test/foo`, `test/../../bar` or `/etc/shadow` don't lead to unexpected results. I'd suggest to allow slashes *only* for artifact names instead.

popzxc: Ah, I didn't think about it.
My original idea was to keep things as simple as possible, so there will be less extra-cases.
But no problem, I'll make slash available only for artifact names.

PR #1685, opened 2020-01-08
Brush up explorer [ECR-4084] [ECR-4092]

slowli:
## Overview

This PR improves various aspects of the blockchain explorer, such as API documentation and tests.

---

See: https://jira.bf.local/browse/ECR-4084, https://jira.bf.local/browse/ECR-4092


components/explorer/src/api/websocket.rs
```diff
@@ -0,0 +1,182 @@
...
+/// Response to a WebSocket client.
+#[serde(tag = "result", rename_all = "snake_case")]
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub enum Response<T> {
+    /// Successful response.
+    Success {
+        /// Payload attached to the response.
+        response: T,
+    },
+    /// Erroneous response.
```
popzxc: Nit: I'm not sure that this response can be called "erroneous", the response *itself* is successful (if it was sent, received and parsed), but it rather contains a report about an occurred error.

components/explorer/Cargo.toml
```diff
@@ -19,6 +19,7 @@ travis-ci = { repository = "exonum/exonum" }
 exonum = { version = "0.13.0-rc.2", path = "../../exonum" }
 
 chrono = { version = "0.4.6", features = ["serde"] }
+hex = '0.4.0'
```
aleksuss:
```suggestion
hex = "0.4.0"
```

PR #1687, opened 2020-01-08
Add wait for network starting instead of sleep()

ivan-ochc:


test-suite/exonum-py-tests/suite/common.py
```diff
@@ -87,3 +90,37 @@ def launcher_networks(network: ExonumNetwork) -> List[Dict[str, Any]]:
...
+    wait_api_to_start(network)
+    wait_for_block(network, 1)
+
+
+def wait_for_block(network: ExonumNetwork, height: int = 1) -> None:
+    """Wait for block at specific height"""
+    for validator_id in range(network.validators_count()):
+        host, public_port, private_port = network.api_address(validator_id)
+        client = ExonumClient(host, public_port, private_port)
+        end = time.time() + 20
```
popzxc: Nit: You can just add a constant e.g. `RETRIES_AMOUNT`, and use `for` loop instead of `while`, like:

```python
for _ in range(RETRIES_AMOUNT):
    if client.public_api.get_block(height) == 200:
        break
    time.sleep(0.5)
```

same applies to the `wait_api_to_start` function.

PR #1688, opened 2020-01-09
Update tests to the latest version of exonum-launcher

ivan-ochc:


test-suite/exonum-py-tests/exonum_tests/deploy.py
```diff
@@ -108,7 +107,11 @@ def test_deploy_regular_exceed_deadline_height(self):
             explorer = launcher.explorer()
 
             launcher.deploy_all()
-            launcher.wait_for_deploy()
+            try:
```
popzxc: Can be rewritten as:

```python
with self.assertRaises(ExecutionFailError):
    launcher.wait_for_deploy()
```

PR #1690, opened 2020-01-10
Move `api` module out of main crate [ECR-3949]

slowli:
## Overview

This PR moves `api` module out of the main crate, introducing two new crate:

- `exonum-api` is an *upstream* crate, which defines API building primitives. It is reexported by `exonum`. (Not sure if this is the best idea, but it provides low-effort backward compatibility for existing code.)
- `exonum-system-api` is a *downstream* crate with a *node plugin* defining system API. Plugins are sort of like services, but more low-level.

Unresolved questions:

- Should we communicate that `exonum::api` module is not restricted by semver restrictions on the main crate? *Is* it not restricted? What about other reexports (`crypto`, `keys`, `merkledb`)? Maybe, remove the reexport altogether?
- Is plugin a good abstraction? Explorer could theoretically be refactored into a plugin. However, we may want to introduce helper indexes in the explorer for more efficient data retrieval, which is not possible with the current plugin interface (plugins cannot write to DB).

---

See: https://jira.bf.local/browse/ECR-3949


components/api/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+edition = "2018"
+authors = ["The Exonum Team <exonum@bitfury.com>"]
+homepage = "https://exonum.com/"
+repository = "https://github.com/exonum/exonum"
+documentation = "https://docs.rs/exonum-api"
+readme = "README.md"
+license = "Apache-2.0"
+keywords = ["exonum", "http"]
+categories = ["cryptography"]
+description = "HTTP API engine the Exonum framework"
```
aleksuss: `.. of the ...` ?

components/system-api/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+edition = "2018"
+authors = ["The Exonum Team <exonum@bitfury.com>"]
+homepage = "https://exonum.com/"
+repository = "https://github.com/exonum/exonum"
+documentation = "https://docs.rs/exonum-system-api"
+readme = "README.md"
+license = "Apache-2.0"
+keywords = ["exonum", "http"]
+categories = ["cryptography"]
+description = "System API plugin the Exonum framework"
```
aleksuss:
```suggestion
description = "System API plugin of the Exonum framework"
```

components/system-api/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "exonum-system-api"
+version = "0.13.0-rc.2"
+edition = "2018"
+authors = ["The Exonum Team <exonum@bitfury.com>"]
```
aleksuss:
```suggestion
authors = ["The Exonum Team <contact@exonum.com>"]
```

components/api/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "exonum-api"
+version = "0.13.0-rc.2"
+edition = "2018"
+authors = ["The Exonum Team <exonum@bitfury.com>"]
```
aleksuss: An email address has been changed
```suggestion
authors = ["The Exonum Team <contact@exonum.com>"]
```

exonum/src/node/consensus.rs
```diff
@@ -634,6 +634,10 @@ impl NodeHandler {
         };
 
         let snapshot = self.blockchain.snapshot();
+        for plugin in &self.plugins {
```
popzxc: Nit: Not sure if this should be called from the consensus code.

services/supervisor/tests/supervisor/main.rs
```diff
@@ -63,21 +64,24 @@ fn assert_count_is_not_set(api: &TestKitApi, service_name: &'static str) {
     assert!(response.is_err());
 }
 
-fn artifact_exists(api: &TestKitApi, name: &str) -> bool {
-    let artifacts = &api.dispatcher_info().artifacts;
-    artifacts.iter().any(|a| a.name == name)
+#[allow(clippy::let_and_return)] // doesn't work otherwise
```
popzxc: Why exactly?

slowli: Borrow checker. Here's the error messages I'm getting locally with the following code

```rust
let snapshot = testkit.snapshot();
let artifacts = snapshot.for_dispatcher().service_artifacts();
artifacts.keys().any(|artifact| artifact.name == name)
```

```text
error[E0597]: `*snapshot` does not live long enough
  --> services\supervisor\tests\supervisor\main.rs:70:21
   |
70 |     let artifacts = snapshot.for_dispatcher().service_artifacts();
   |                     ^^^^^^^^ borrowed value does not live long enough
71 |     artifacts.keys().any(|artifact| artifact.name == name)
   |     ---------------- a temporary with access to the borrow is created here ...
72 | }
   | -
   | |
   | `*snapshot` dropped here while still borrowed
   | ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `exonum_merkledb::indexes::proof_map::Keys<'_, exonum::runtime::types::ArtifactId>`
   |
   = note: The temporary is part of an expression at the end of a block. Consider forcing this temporary to be dropped sooner, before the block's local variables are dropped. For example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block.

error[E0597]: `artifacts` does not live long enough
  --> services\supervisor\tests\supervisor\main.rs:71:5
   |
71 |     artifacts.keys().any(|artifact| artifact.name == name)
   |     ^^^^^^^^^-------
   |     |
   |     borrowed value does not live long enough
   |     a temporary with access to the borrow is created here ...
72 | }
   | -
   | |
   | `artifacts` dropped here while still borrowed
   | ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `exonum_merkledb::indexes::proof_map::Keys<'_, exonum::runtime::types::ArtifactId>`
   |
   = note: The temporary is part of an expression at the end of a block. Consider forcing this temporary to be dropped sooner, before the block's local variables are dropped. For example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block.
```

exonum/src/node/consensus.rs
slowli: This is exactly [where the equivalent call was](https://github.com/exonum/exonum/blob/0f622c7b0787f7c8947421d8f75bccb2334c548e/exonum/src/node/consensus.rs#L575) for the explorer before it was moved out.

services/supervisor/tests/supervisor/main.rs
popzxc: That's weird. In theory, `any` for iterator is not lazy, and all the variables should be dropped only after result of `any` is collected.

Here's somewhat similar working example: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c96cc80a558cf5f54812ee758a2e116d).

exonum/src/node/consensus.rs
popzxc: I see. Probably we should later move all the code that is not strictly tied to the consensus to somewhere else. For now it's fine to left things as-is, I guess.

PR #1692, opened 2020-01-13
Improve ExecutionError docs in runtime.proto

popzxc:
Follow-up for #1680.

Improves the docstrings in the `runtime.proto`.

### DIFF
--- a/exonum/src/proto/schema/exonum/runtime.proto
+++ b/exonum/src/proto/schema/exonum/runtime.proto
@@ -92,7 +92,7 @@ enum ErrorKind {
 // Option with set call site and unset runtime ID is not valid, receiving a message
 // with such a combination means receiving a malformed message.
 message ExecutionError {
-  // The kind of error that indicates the type of occurred error.
+  // The kind of error that indicates its type.
   ErrorKind kind = 1;
   // User defined error code that can have different meanings for the different
   // error kinds.
@@ -104,7 +104,7 @@ message ExecutionError {
   oneof runtime {
     // Identifier of runtime in which error originates.
     uint32 runtime_id = 4;
-    // The error originates not in the runtime code.
+    // The error does not originate in the runtime code.
     google.protobuf.Empty no_runtime_id = 5;
   }
 
@@ -112,7 +112,7 @@ message ExecutionError {
   oneof call_info {
     // Information about service call from which error originates.
     CallSite call_site = 6;
-    // The error originates not in the service code.
+    // The error does not originate in the service code.
     google.protobuf.Empty no_call_site = 7;
   }
 }
--- a/exonum/src/proto/schema/exonum/runtime.proto
+++ b/exonum/src/proto/schema/exonum/runtime.proto
@@ -67,8 +67,9 @@ message GenesisConfig {
 }
 
 // The type of ExecutionError.
-// Note that ErrorKind doesn't communicate the source of the error, for this
-// see the 'runtime_id' and 'call_site' fields of ExecutionError.
+// Note that ErrorKind isn't a primary source for determining the origin of the error.
+// First, see the 'runtime_id' and 'call_site' fields of ExecutionError, and only then
+// you can rely on `ErrorKind` to resolve ambiguity.
 enum ErrorKind {
   // An unexpected error which does not have a specific cause.
   // The description of the error is the only source of information about this kind of errors.
@@ -84,6 +85,7 @@ enum ErrorKind {
 }
 
 // Result of unsuccessful runtime execution.
+//
 // ExecutionError message provides the information about the source of the error.
 // The source of error is determined as following:
 // - If both runtime ID and call site are set, then error originates in the service code.
@@ -91,6 +93,9 @@ enum ErrorKind {
 // - If none of runtime ID and call site is set, then error originates in the core code.
 // Option with set call site and unset runtime ID is not valid, receiving a message
 // with such a combination means receiving a malformed message.
+//
+// Though in most cases 'runtime_id' and 'call_site' are enough to deduce the source of error,
+// 'ErrorKind' type can be used to resolve ambiguity.
 message ExecutionError {
   // The kind of error that indicates its type.
   ErrorKind kind = 1;
@@ -104,7 +109,7 @@ message ExecutionError {
   oneof runtime {
     // Identifier of runtime in which error originates.
     uint32 runtime_id = 4;
-    // The error does not originate in the runtime code.
+    // There was no runtime to process an erroneous call.
     google.protobuf.Empty no_runtime_id = 5;
   }
 
@@ -112,7 +117,7 @@ message ExecutionError {
   oneof call_info {
     // Information about service call from which error originates.
     CallSite call_site = 6;
-    // The error does not originate in the service code.
+    // There was no service to process an erroneous call.
     google.protobuf.Empty no_call_site = 7;
   }
 }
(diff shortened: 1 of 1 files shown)


exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -66,34 +66,53 @@ message GenesisConfig {
...
   SERVICE = 3;
+  // A common error which can occur in different contexts. See common error codes for details.
   COMMON = 4;
 }
 
 // Result of unsuccessful runtime execution.
+// ExecutionError message provides the information about the source of the error.
+// The source of error is determined as following:
+// - If both runtime ID and call site are set, then error originates in the service code.
+// - If runtime ID is set, and call site is not set, then error originates in the runtime code.
```
dmitry-timofeev: The question not to this code, but Runtime spec. As only the runtime can know if an error occurred in service or its code, is it its responsibility to communicate that to the core?

@vitvakatu shall we modify anything?

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -66,34 +66,53 @@ message GenesisConfig {
...
-    // The runtime associated with the error is unknown.
+    // The error originates not in the runtime code.
     google.protobuf.Empty no_runtime_id = 5;
   }
 
+  // Call site will be set if the error originates in code related to the certain service.
   oneof call_info {
+    // Information about service call from which error originates.
     CallSite call_site = 6;
+    // The error originates not in the service code.
```
dmitry-timofeev: ... does not originate ...?

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -66,34 +66,53 @@ message GenesisConfig {
...
+// ExecutionError message provides the information about the source of the error.
+// The source of error is determined as following:
+// - If both runtime ID and call site are set, then error originates in the service code.
+// - If runtime ID is set, and call site is not set, then error originates in the runtime code.
+// - If none of runtime ID and call site is set, then error originates in the core code.
+// Option with set call site and unset runtime ID is not valid, receiving a message
+// with such a combination means receiving a malformed message.
 message ExecutionError {
-  // The kind of error that indicates in which module the error occurred.
+  // The kind of error that indicates the type of occurred error.
```
dmitry-timofeev: that indicates its type?

exonum/src/proto/schema/exonum/runtime.proto
vitvakatu: Setting call site is possible only in Core (as the method `set_call_site` is package-private). Though I don't understand how the Core will understand if an error occurred in service code.

popzxc: Core provides some guarantees on the requests that it makes to runtime:

If core requests a runtime to invoke a `BeforeTransactions` hook for service, then it's sure that this service exists in the runtime and runtime should be able to dispatch the call.
Thus, the only possible variant of managed error here is the error in the service code.

Error in runtime code there may happen only because of the bug in either runtime or core code, and for this purpose we have `Unexpected` error type, and even in this condition call site friendly will tell us that error happened when core requested runtime X to perform something for service Y, and the error description will contain something like "BUG: Core unexpectedly requested to invoke hook for non-existent service".

OTOH, if core performs a request to the runtime *itself* (rather than to a service), e.g. artifact deployment, there is obviously no service call site, and the source of (expected and valid) error will surely be in the runtime code.

vitvakatu: I see, thank you, @popzxc 

@dmitry-timofeev no changes are needed on our side. The only thing we can alter is to utilize `CommonError`, but if I understand it correctly, it is not entirely necessary.

dmitry-timofeev: > Error in runtime code there may happen only because of the bug in either runtime or core code

Not necessarily. "Transaction with the given id does not exist", "Malformed arguments" — those errors happen in _runtime_ code during tx processing, which performs routing and decoding correspondingly. They may be caused by a client bug (invalid arguments are passed), a runtime bug (we can't route/decode correctly), or a service bug (it uses incorrect tx id).

popzxc: Hm.  As I see it:

If the destination service exists, then both of non-existent tx id and malformed arguments errors belong to the service code.

Runtime was requested just to pass the certain arguments to the service. Service is present, but it cannot process the request. Thus, the error originates in the service.

Even if the runtime is capable of deserializing arguments for the service call, it's still an additional feature that runtime authors did introduce to make service authors life easier. Service could attempt to deserialize data by itself and face the same issue.

One more point here is that `ExecutionError` in most cases denotes some sort of *expected* error, which can happen in a normal workflow. For bugs we have `Unexpected` error type, and IMHO we should not try to mix bugs with expected errors (caused by incorrect input provided by user), since hopefully `Unexpected` errors will never occur in production code.

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -125,7 +144,9 @@ message CallSite {
 // Result of runtime execution.
 message ExecutionStatus {
   oneof result {
+    // Successful execution.
     google.protobuf.Empty ok = 1;
+    // Execution happened with an error.
```
slowli: Nit / grammar: "Execution ended with an error" or "An error occurred during execution" look more grammatically correct, IMO.

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -66,34 +66,53 @@ message GenesisConfig {
...
   // Optional description which doesn't affect `object_hash`.
   string description = 3;
 
+  // Runtime ID will be set if the error originates in code related to the certain runtime.
   oneof runtime {
-    // Runtime identifier associated with the error.
+    // Identifier of runtime in which error originates.
     uint32 runtime_id = 4;
-    // The runtime associated with the error is unknown.
+    // The error does not originate in the runtime code.
```
slowli: This seems incorrect; at least this was not my intent when I designed these data structures in #1585. If `runtime_id` is set, this means there is a runtime responsible for processing the erroneous call. The error may originate either directly in the runtime, or in the service, depending on the *error kind*.

Similarly, if `call_site` is set, the error does not necessarily originate in the service code. For example, if there is no interface, a corresponding `CommonError` will be raised (at least in the Rust runtime). In other words, both error kind and error location are needed to fully interpret the error (e.g., service error codes don't tell much if we don't know which service they belong to - which information `call_site` provides).

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -66,34 +66,53 @@ message GenesisConfig {
   repeated InstanceInitParams builtin_instances = 3;
 }
 
-// The kind of ExecutionError.
+// The type of ExecutionError.
+// Note that ErrorKind doesn't communicate the source of the error, for this
+// see the 'runtime_id' and 'call_site' fields of ExecutionError.
```
slowli: I think both `ErrorKind` and `runtime_id` / `call_site` may be necessary to interpret the error; see the other comment.

exonum/src/proto/schema/exonum/runtime.proto
dmitry-timofeev: I don't agree that such errors belong to or originate in the service _code_ because they happen outside of it, they are raised by the runtime code. They are certainly related to a particular service, but do not originate in its code. 

The preserved id of the related service is definitely useful for debuggability. But the current doc says "originates in the service code". I suggest to relax it to "relates to a particular service", because the current phrasing says it is thrown from the service code.

popzxc: Sure, makes sense.

PR #1695, opened 2020-01-13
Re-use #is_aborted method:

dmitry-timofeev:
## Overview

Re-use #is_aborted method in #finish impl.

---

### Definition of Done


dmitry-timofeev:
#finish is similar to #merge, so I request Rust-guys to review bigger deduplication opportunities there (e.g., if we don't care if finish creates a new fork (as merge does).

popzxc:
Yeah, that's a good point, but I think we can defer all the refactoring to post-1.0 time.
Now, let's do a (hopefully not last) big dash and reach this milestone :)

slowli:
The code was written this way because of the borrow checker (one thing distinguishing Rust from Java 🙃). As you may see, the changed version has failed to compile:

```text
error[E0382]: borrow of moved value: `self`
   --> components/merkledb/src/migration.rs:530:12
    |
528 |         let patch = self.fork.unwrap().into_patch();
    |                     --------- value moved here
529 |         let hash = Migration::new(&self.namespace, &patch).state_hash();
530 |         if self.is_aborted() {
    |            ^^^^ value borrowed here after partial move
    |
    = note: move occurs because `self.fork` has type `std::option::Option<db::Fork>`, which does not implement the `Copy` trait
```

The gist is that the compiler doesn't (want to?) see that `self.is_aborted()` doesn't use `self.fork`.

dmitry-timofeev:
I see, thanks, so it doesn't work because a part of the object state was already "moved" and after that you can't take a reference to the original object, does it?

This could work if we don't care if it checks the abort signal exactly before the merge. This version does compile:

```rust
        if self.is_aborted() {
            Err(MigrationError::Aborted)
        } else {
            let patch = self.fork.unwrap().into_patch();
            let hash = Migration::new(&self.namespace, &patch).state_hash();
            self.db.merge(patch).map_err(MigrationError::Merge)?;
            Ok(hash)
        }
    }
```

slowli:
@dmitry-timofeev The following works too:

```rust
pub fn finish(mut self) -> Result<Hash, MigrationError> {
    let patch = self.fork.take().unwrap().into_patch();
    let hash = Migration::new(&self.namespace, &patch).state_hash();
    if self.is_aborted() {
        Err(MigrationError::Aborted)
    } else {
        self.db.merge(patch).map_err(MigrationError::Merge)?;
        Ok(hash)
    }
}
```

Note the use of [`Option::take`](https://doc.rust-lang.org/std/option/enum.Option.html#method.take) on `self.fork`. This allows to take out `fork` from `self` without the `MigrationHelper` instance considered partially moved.

dmitry-timofeev:
Thanks! Shall I update the PR with one of the suggested solutions, or close the PR and keep the code as is?

slowli:
@dmitry-timofeev I'd suggest to update this PR with my suggestion since it does not move the abortion check further from the merge application (creating a patch from a fork can take a surprising amount of time).

PR #1696, opened 2020-01-13
Implement refined HTTP errors [ECR-4101]

Kormick:
Initial version for refined HTTP errors.
Added `test_endpoint` in the explorer API and `ServiceApiScope::endpoint_new` handler for testing.
If the current implementation is ok, then we can replace all endpoints with it.

WIP:
- Do we need to set a version of the artifact to the `source` field?
- Should we set a deprecated warning header if we return an error from the deprecated endpoint?
- How can we handle errors not from endpoint level (i.e. if we failed to deserialize query)? I've failed to find the place where such errors are converted to the response.


exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+    /// Short description of the error.
+    pub title: String,
+    /// Detailed description of the error.
+    pub detail: String,
+    /// HashMap with additional parameters.
+    pub params: HashMap<String, String>,
+    /// Source of the error.
+    pub source: String,
+    /// Internal error code.
+    pub error_code: i8,
```
popzxc: Why `i8`?

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+    }
+
+    /// Sets `detail` of an error.
+    pub fn detail(mut self, detail: String) -> Self {
+        self.detail = detail;
+        self
+    }
+
+    /// Inserts new value into `params`.
+    pub fn param(mut self, key: String, value: String) -> Self {
```
popzxc: I'd suggest to take `Into<String>` everywhere instead of `String` to get more flexibility.

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+
+/// API HTTP error struct.
+#[derive(Fail, Debug, Serialize)]
+pub struct ApiError {
+    /// HTTP error code.
+    #[serde(skip)]
+    pub http_code: HttpCode,
+    /// A URI reference to the documentation or possible solutions for the problem.
+    #[serde(rename = "type")]
+    pub error_type: String,
```
popzxc: Since the type already contains `*Error` in its name, I think that we can name this field just `type` without any ambiguity.

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+}
+
+impl std::fmt::Display for ApiError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&serde_json::to_string(self).unwrap())
+    }
+}
+
+impl ApiError {
+    fn default() -> Self {
```
popzxc: Probably it's better to rename this method to `new` and make it take `http_code` as an argument.

popzxc: ...and then it could make sense to make this method public, so we don't have to bother ourselves with creating a constructor method for every possible HTTP response code.

services/explorer/src/api.rs
```diff
@@ -529,6 +529,22 @@ impl ExplorerApi {
         Ok(CallStatusResponse { status })
     }
 
+    fn test_endpoint(
+        _schema: Schema<&dyn Snapshot>,
+        query: TransactionQuery,
+    ) -> Result<CallStatusResponse, HttpApiError> {
+        let response = HttpApiError::BadRequest()
```
popzxc: Looks cool! :+1:

exonum/src/api/error.rs
slowli: `type` is a Rust keyword 🙃

slowli: I think something like `docs_uri` could be better; IMO, `type` (or `error_type`) is not very descriptive in this context.

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+    pub http_code: HttpCode,
+    /// A URI reference to the documentation or possible solutions for the problem.
+    #[serde(rename = "type")]
+    pub error_type: String,
+    /// Short description of the error.
+    pub title: String,
+    /// Detailed description of the error.
+    pub detail: String,
+    /// HashMap with additional parameters.
+    pub params: HashMap<String, String>,
```
slowli: Nit: I'd suggest to add extensible fields later and start with a simpler version.

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+    /// Source of the error.
+    pub source: String,
+    /// Internal error code.
+    pub error_code: i8,
+}
+
+impl std::fmt::Display for ApiError {
```
slowli: Shouldn't `Display` be human-readable? IMO, something like displaying `title` + `details` could be more appropriate.

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
 use serde::Serialize;
+
+use std::collections::HashMap;
 use std::io;
 
 use crate::node::SendError;
 
+/// List of possible HTTP response codes.
+#[derive(Debug, Serialize)]
+pub enum HttpCode {
```
slowli: Huh, why not reuse the status from `actix`?

exonum/src/api/backends/actix.rs
```diff
@@ -143,6 +143,20 @@ impl ResponseError for api::Error {
...
+    fn error_response(&self) -> HttpResponse {
+        let mut response = match self.http_code {
+            HttpCode::BadRequest => HttpResponse::BadRequest(),
+            HttpCode::NotFound => HttpResponse::NotFound(),
+            HttpCode::NotImplemented => HttpResponse::NotImplemented(),
+        };
+
+        response
+            .header(header::CONTENT_TYPE, "application/problem+json")
+            .body(self.to_string())
```
slowli: I think it's inappropriate use of `to_string`; it could make more sense to use `serde_json::to_string` directly.

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+    /// HTTP error code.
+    #[serde(skip)]
+    pub http_code: HttpCode,
+    /// A URI reference to the documentation or possible solutions for the problem.
+    #[serde(rename = "type")]
+    pub error_type: String,
+    /// Short description of the error.
+    pub title: String,
+    /// Detailed description of the error.
+    pub detail: String,
```
slowli: Nit: It could make sense to skip serializing empty fields:

```rust
#[serde(default, skip_serializing_if = "String::is_empty")]
pub detail: String,
```

services/explorer/src/api.rs
```diff
@@ -529,6 +529,22 @@ impl ExplorerApi {
         Ok(CallStatusResponse { status })
     }
 
+    fn test_endpoint(
```
slowli: Nit: It could make sense to test stuff in the `api` test in testkit; it's specifically designed for this purpose 🙃

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+    /// HashMap with additional parameters.
+    pub params: HashMap<String, String>,
+    /// Source of the error.
+    pub source: String,
```
slowli: Do you plan to make `source` more structured in this PR? If not, it could make sense to not define it initially.

exonum/src/api/error.rs
popzxc: Ah, always these limitations get me :upside_down_face:

exonum/src/api/error.rs
Kormick: Should be `u8`?

popzxc: I think so, unless there is a good reason to allow negative error codes.

exonum/src/api/error.rs
```diff
@@ -16,10 +16,118 @@
...
+    #[allow(non_snake_case)]
+    pub fn NotFound() -> Self {
+        Self {
+            http_code: HttpCode::NotFound,
+            ..Self::default()
+        }
+    }
+
+    /// Sets `error_type` of an error.
+    pub fn error_type(mut self, error_type: String) -> Self {
```
aleksuss: Maybe `&str` would be better from the UX point of view?
```rust
.error_type("Some type of error")
```
vs
```rust
.error_type("Some type of error".to_string()) 
``` 
What do you think ?

popzxc: I already suggested to use `Into<String>`: https://github.com/exonum/exonum/pull/1696#discussion_r365887349

Kormick: With Into<String> will be almost the same.

aleksuss: @popzxc, even better 👍🏻

aleksuss: @Kormick `Into` allow you to use `&str` and `String` at the same time. Take a look: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=236b56b94c702cea1f93ee64e4443525

test-suite/testkit/src/api.rs
```diff
@@ -360,6 +423,51 @@ where
 
         Err(error)
     }
+
+    fn response_to_api_result_new<R>(mut response: Response) -> api::ApiResult<R>
+    where
+        R: DeserializeOwned + 'static,
+    {
+        fn extract_error(response: &mut Response) -> Option<api::ApiError> {
```
popzxc: Probably it would make sense to implement `ApiError::parse(u16, AsRef<str>)` method and here just use 

```rust
ApiError::parse(code, body).expect("Failed to decode api error")
```

test-suite/testkit/src/api.rs
```diff
@@ -360,6 +423,51 @@ where
...
+            let code = response.status();
+            let body = response.text().expect("Unable to get response text");
```
popzxc: However, if this will move into `ApiError::parse` you should be careful with `expect` / `unwrap` usage.

components/api/src/error.rs
```diff
@@ -14,10 +14,85 @@
...
+#[derive(Fail, Debug, Serialize)]
+pub struct ApiError {
+    /// HTTP error code.
+    #[serde(skip)]
+    pub http_code: StatusCode,
+    /// A URI reference to the documentation or possible solutions for the problem.
+    #[serde(rename = "type", default, skip_serializing_if = "String::is_empty")]
+    pub docs_uri: String,
+    /// Short description of the error.
+    #[serde(default, skip_serializing_if = "String::is_empty")]
```
popzxc: And one more nit: since `ApiError` doesn't derive `Deserialize`, `default` attribute is redundant.

test-suite/testkit/src/api.rs
```diff
@@ -360,6 +423,51 @@ where
...
+
+            let mut error = api::ApiError::new(code);
+            if let Some(json_obj) = serde_json::from_str::<serde_json::Value>(&body)
+                .unwrap()
+                .as_object()
+            {
+                if json_obj.contains_key("type") {
```
slowli: Is there anything preventing implementing error parsing via `impl Deserialize for ApiError`? It would be a lot less verbose.

test-suite/testkit/tests/api/main.rs
```diff
@@ -132,3 +134,35 @@ fn moved() {
...
+#[test]
+fn endpoint_with_new_error_type() {
+    let (_testkit, api) = init_testkit();
+
+    // Check OK response.
+    let ok_query = PingQuery { value: 64 };
+    let response: u64 = api
+        .public(ApiKind::Service("api-service"))
+        .query(&ok_query)
+        .get_new::<u64>("new-error-type")
```
slowli: Nit: Turbofish is redundant here since you've specified `response` type earlier.

runtimes/rust/src/api.rs
```diff
@@ -164,6 +166,32 @@ impl ServiceApiScope {
...
+        R: IntoFuture<Item = I, Error = crate::api::ApiError> + 'static,
+    {
+        let blockchain = self.blockchain.clone();
+        let descriptor = self.descriptor.clone();
+        self.inner
+            .endpoint(name, move |query: Q| -> crate::api::ApiFutureResult<I> {
+                let desc = descriptor.clone();
+                let state = ServiceApiState::from_api_context(
+                    &blockchain,
+                    (desc.0, desc.1.as_ref()).into(),
```
slowli: Nit / subjective: IMO, destructuring `desc` above could be more readable:

```rust
let (instance_id, instance_name) = descriptor.clone();
```

runtimes/rust/src/api.rs
```diff
@@ -164,6 +166,32 @@ impl ServiceApiScope {
...
+                    name,
+                );
+                let result = handler(&state, query);
+                let future = result
+                    .into_future()
+                    .map_err(move |err| err.source(format!("{}:{}", desc.1, desc.0)));
```
slowli: Does this mean that the error source will be overridden even if the user sets it? If this is intended, I'd suggest to make `ApiError::source()` hidden from docs to not trip the user; if not, then the user-provided source should not be overridden.

Also, it could make sense exploring more structured source presentation, e.g.

```rust
pub struct ApiError {
    // ...
    source: serde_json::Value,
}
```

or even

```rust
pub struct ApiError<S> {
    // ...
    source: S,
}
```

...although this can be done after this PR is merged.

test-suite/testkit/src/api.rs
popzxc: It won't have an error code set. Not sure if this can be considered correct deserializing. Usually when you deserialize something, you expect to obtain a complete object.

slowli: I'd think we can agree to go with an almost correct serialization in this case. Another way around is to split `ApiError` into body and code and deserialize the body:

```rust
struct ApiError {
    pub http_code: StatusCode,
    pub body: ApiErrorBody,
}

#[derive(Serialize, Deserialize)]
struct ApiErrorBody {
    // all fields except for `http_code`...
}
```

popzxc: Yeah, option with split body/code sounds good.

components/api/src/error.rs
```diff
@@ -14,10 +14,99 @@
...
+        write!(f, "{}: {}", self.body.title, self.body.detail)
+    }
+}
+
+impl ApiError {
+    /// Builds a ApiError with the given `http_code`.
+    pub fn new(http_code: StatusCode) -> Self {
+        Self {
+            http_code,
+            body: ApiErrorBody {
```
popzxc: Nit: You can just derive `Default` for `ApiErrorBody`.

components/api/src/error.rs
```diff
@@ -14,10 +14,99 @@
...
+    }
+
+    /// Sets `error_code` of an error.
+    pub fn error_code(mut self, error_code: u8) -> Self {
+        self.body.error_code = Some(error_code);
+        self
+    }
+
+    /// Tries to create `ApiError` from JSON.
+    pub fn parse(http_code: StatusCode, body: &str) -> std::result::Result<Self, String> {
```
popzxc: Nit: It's better to use `failure::Error` for error type. You can create `failure::Error` for string with `failure::format_err!`.

components/api/src/error.rs
```diff
@@ -14,10 +14,99 @@
 
 //! The set of errors for the Exonum API module.
 
+use actix_web::http::StatusCode;
```
popzxc: Nit: Since this type is required to parse `ApiError`, it's better to re-export this type via `pub use`, so clients won't have to import `actix_web` just to get this type available.

components/api/src/error.rs
slowli: I'd suggest to retain at least the error message here. Perhaps the easiest way is to use `Result<Self, serde_json::Error>`.

components/api/src/error.rs
```diff
@@ -14,49 +14,116 @@
...
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}: {}", self.body.title, self.body.detail)
+    }
+}
+
+impl Error {
+    /// Builds a ApiError with the given `http_code`.
+    pub fn new(http_code: HttpStatusCode) -> Self {
```
slowli: Could you add constructors for common erroneous status codes, e.g. `bad_request` (400) / `not_found` (404) / `forbidden` (403) / `internal` (500)? Also, the internal constructor may accept the error to immediately set the `detail`:

```rust
pub fn internal(cause: impl failure::Fail) { /* ... */ }
```

components/api/src/error.rs
```diff
@@ -14,49 +14,116 @@
...
+
+    /// Sets `error_code` of an error.
+    pub fn error_code(mut self, error_code: u8) -> Self {
+        self.body.error_code = Some(error_code);
+        self
+    }
+
+    /// Adds HTTP header, which will be added in `HttpResponse`
+    #[doc(hidden)]
+    pub fn header(mut self, key: HeaderName, value: &str) -> Self {
```
slowli: Does this method need to be public? As far as I see, it's only used in this file. I'd prefer to include more methods later as they become necessary.

components/system-api/src/public.rs
```diff
@@ -97,7 +97,7 @@ impl SystemApi {
 
     fn handle_stats_info(self, name: &'static str, api_scope: &mut ApiScope) -> Self {
         let self_ = self.clone();
-        api_scope.endpoint(name, move |_query: ()| {
+        api_scope.endpoint(name, move |_query: ()| -> api::Result<_> {
```
slowli: Nit: Specifying the return type is redundant here and in other places in this file.

runtimes/rust/src/api.rs
```diff
@@ -14,7 +14,7 @@
 
 //! Building blocks for creating HTTP API of Rust services.
 
-pub use exonum_api::{Deprecated, EndpointMutability, Error, FutureResult, Result};
+pub use exonum_api::{Deprecated, EndpointMutability, Error, FutureResult};
```
slowli: Why is `Result` no longer re-exported? As far as I see, this is the only thing forcing `cryptocurrency-advanced` and some other crates to have `exonum-api` as the direct dependency.

examples/cryptocurrency/src/lib.rs
```diff
@@ -283,7 +283,8 @@ pub mod contracts {
 /// Cryptocurrency API implementation.
 pub mod api {
     use exonum::crypto::PublicKey;
-    use exonum_rust_runtime::api::{self, ServiceApiBuilder, ServiceApiState};
+    use exonum_api as api;
```
slowli: This direct dependency could be avoided if `Error` had `not_found` constructor and if `Result` as continued to be re-exported.

components/api/src/error.rs
```diff
@@ -14,49 +14,116 @@
...
-    /// Unauthorized error. This error occurs when the request lacks valid
-    /// authentication credentials.
-    #[fail(display = "Unauthorized")]
-    Unauthorized,
+use serde::{Deserialize, Serialize};
+
+use std::fmt;
+
+/// API HTTP error struct.
+#[derive(Fail, Debug, PartialEq)]
```
slowli: I'd be wary of using `PartialEq` in tests: since error headers are not restored, this may lead to unexpected assertion fails. Also, from the point of view of future compatibility having `PartialEq` forces us to implement it for any possible extensions of the type; IMO, it's not worth it. Correspondingly, I'd suggest to compare the necessary fields separately in tests.

services/explorer/src/api.rs
```diff
@@ -417,18 +418,22 @@ impl ExplorerApi {
...
+                )));
         }
 
         let (upper, upper_bound) = if let Some(upper) = query.latest {
             if upper > explorer.height() {
-                return Err(ApiError::NotFound(format!(
+                return Err(api::Error::new(api::HttpStatusCode::NOT_FOUND)
+                    .title("Block not found")
+                    .detail(format!(
                     "Requested latest height {} is greater than the current blockchain height {}",
```
slowli: Nit: Weird formatting; the arguments to `format!` are not indented compared to the `format!` itself.

services/explorer/src/api.rs
```diff
@@ -571,13 +584,21 @@ impl ExplorerApi {
...
+                })
         };
 
         Box::new(
             verify_message(snapshot, query.tx_body)
                 .into_future()
-                .map_err(|e| ApiError::BadRequest(e.to_string()))
+                .map_err(|e| {
+                    api::Error::new(api::HttpStatusCode::BAD_REQUEST)
+                        .title("Failed to add transaction")
```
slowli: Nit: Slightly more verbose `title` (e.g., "Failed to add transaction to memory pool") would be more helpful to clients.

test-suite/testkit/src/server.rs
```diff
@@ -149,9 +153,9 @@ impl Handler<RollBack> for TestKitActor {
 
     fn handle(&mut self, RollBack(height): RollBack, _ctx: &mut Self::Context) -> Self::Result {
         if height == Height(0) {
-            return Err(api::Error::BadRequest(
-                "Cannot rollback past genesis block".into(),
-            ));
+            return Err(api::Error::new(api::HttpStatusCode::BAD_REQUEST)
+                .title("RollBack handler failed.")
```
slowli: Nit: I'd move "Cannot rollback past genesis block" to `title` and leave `detail` empty. IMO, this would make the error easier to understand: the client may not be aware about `RollBack` type, and it's not its handler fault that the request has failed.

test-suite/testkit/src/server.rs
```diff
@@ -120,10 +122,12 @@ impl Handler<CreateBlock> for TestKitActor {
         let block_info = if let Some(tx_hashes) = msg.tx_hashes {
             let maybe_missing_tx = tx_hashes.iter().find(|h| !self.0.is_tx_in_pool(h));
             if let Some(missing_tx) = maybe_missing_tx {
-                return Err(api::Error::BadRequest(format!(
-                    "Transaction not in mempool: {}",
-                    missing_tx.to_string()
-                )));
+                return Err(api::Error::new(api::HttpStatusCode::BAD_REQUEST)
+                    .title("CreateBlock handler failed.")
```
slowli: Nit: I'd suggest to change `title` to "Creating block failed" in order to be more understandable.

runtimes/rust/src/api.rs
```diff
@@ -149,16 +149,23 @@ impl ServiceApiScope {
...
         let descriptor = self.descriptor.clone();
         self.inner
-            .endpoint(name, move |query: Q| -> crate::api::FutureResult<I> {
-                let descriptor = (descriptor.0, descriptor.1.as_ref());
-                let state = ServiceApiState::from_api_context(&blockchain, descriptor.into(), name);
+            .endpoint(name, move |query: Q| -> FutureResult<I> {
+                let (instance_id, instance_name) = descriptor.clone();
+                let state = ServiceApiState::from_api_context(
+                    &blockchain,
+                    (instance_id, instance_name.as_ref()).into(),
```
popzxc: I think it's better to bind this conversion to the variable. Maybe you'll be able to use `Display` representation of `InstanceDescriptor` for error source as well (not sure, since lifetimes can cause problems).

runtimes/rust/src/api.rs
```diff
@@ -149,16 +149,23 @@ impl ServiceApiScope {
...
+                    name,
+                );
                 let result = handler(&state, query);
-                Box::new(result.into_future())
+                let future = result
+                    .into_future()
+                    .map_err(move |err| err.source(format!("{}:{}", instance_name, instance_id)));
```
popzxc: Nit: `Display` for `InstanceDescriptor` puts ID first, but here the ordering is reversed.

services/explorer/src/api.rs
```diff
@@ -480,46 +485,54 @@ impl ExplorerApi {
...
+    fn block(schema: Schema<&dyn Snapshot>, query: BlockQuery) -> api::Result<BlockInfo> {
         let explorer = BlockchainExplorer::from_schema(schema);
         explorer.block(query.height).map(From::from).ok_or_else(|| {
-            ApiError::NotFound(format!(
-                "Requested block height ({}) exceeds the blockchain height ({})",
-                query.height,
-                explorer.height()
-            ))
+            api::Error::new(api::HttpStatusCode::NOT_FOUND)
+                .title("Failed to get block info.")
```
popzxc: Nit: Title contains a dot, while other titles do not.

components/api/src/backends/actix.rs
```diff
@@ -116,20 +116,11 @@ impl ExtendApiBackend for actix_web::Scope<()> {
 
 impl ResponseError for ApiError {
```
popzxc: I think we need some unit tests which check that `ApiError` is converted to the actix `HttpResponse` correctly.

The logic of this type is not as simple anymore.

components/api/Cargo.toml
```diff
@@ -23,11 +23,10 @@ chrono = { version = "0.4.6", features = ["serde"] }
 failure = "0.1.5"
 futures = "0.1.25"
 log = "0.4.6"
-serde = "1.0.101"
+serde = { version = "1.0.101", features = ["derive"] }
```
alekseysidorov: Is it necessary in our code base or just more convenient than before?

Kormick: I had errors without it before, but now it works. I'll remove it.

PR #1697, opened 2020-01-13
Improve runtime docs

slowli:
## Overview

This PR improves runtime docs (non-controversial part, IMO) and removes reexports of `runtime` types from the Rust runtime crate (slightly more controversial part).


runtimes/rust/src/error.rs
```diff
@@ -21,7 +21,7 @@ use exonum_derive::ExecutionFail;
 #[derive(ExecutionFail)]
 #[execution_fail(kind = "runtime")]
 pub enum Error {
-    /// Unable to parse artifact identifier or specified artifact has non-empty spec.
+    /// Cannot deploy artifact because it has non-empty specification.
```
aleksuss:
```suggestion
    /// Cannot deploy an artifact because it has non-empty specification.
```

exonum/src/runtime/mod.rs
```diff
@@ -452,6 +471,9 @@ pub trait Runtime: Send + fmt::Debug + 'static {
 
     /// Dispatches payload to the method of a specific service instance.
     ///
+    /// The call is dispatched iff the service is considered active at the moment.
```
```suggestion
    /// The call is dispatched if the service is considered active at the moment.
```

aleksuss: Or it's not a typo and means *if and only if* ?

runtimes/rust/src/error.rs
slowli: The error concerns a specific artifact though. IMO, dropping an article in an error message is fine.

PR #1698, opened 2020-01-13
Separate Exonum node from core [ECR-4114]

slowli:
## Overview

The final boss of the main crate deconstruction, this PR removes entire node logic (i.e., `events` and `node` modules, most of `messages` and some other stuff) out of the main crate into a separate crate.

- The testkit optionally depends on the node via conditional compilation; Rust services do not depend on node code at all. This means compilation speed-up and lesser amount of dependencies. (As a pleasant side effect, service code now has no access to `ExternalMessage`s, meaning that it cannot shut down the node, for example.)
- `exonum-api` is no longer re-exported from the main crate (since it does not use it).
- Consensus and node tests are moved to the node module; most of previously hidden public types are made crate-private. This means that Rust runtime is now a dev dependency of the node crate, but this seems safe since I don't see a scenario in which the Rust runtime would depend on the node crate.

Discussion:

- Is there a better way to define messages? As of 6221057, the core defines `CoreMessage` (which contains transactions and precommits) and the node crate defines `ExonumMessage` (which contains all P2P messages *including* transactions and precommits), and the tags for transactions and precommits in both must agree.
- What are the implications for Java binding? I *think* in the worst case it can depend on `exonum-node` crate (but should it?). 

~⚠️ This PR includes commits from #1690 and will be rebased once that PR is merged.~ Done

---

See: https://jira.bf.local/browse/ECR-4114


exonum-node/Cargo.toml
```diff
@@ -0,0 +1,64 @@
+[package]
+name = "exonum-node"
+version = "0.13.0-rc.2"
+edition = "2018"
+authors = ["The Exonum Team <contact@exonum.com>"]
+homepage = "https://exonum.com/"
+repository = "https://github.com/exonum/exonum"
+documentation = "https://docs.rs/exonum"
```
popzxc: Nit: I think the link should be different. Not sure if we need to specify it now though.

exonum-node/src/consensus.rs
```diff
@@ -537,7 +541,12 @@ impl NodeHandler {
                     .prevotes(prevote_round, propose_hash)
                     .iter()
                     .map(|p| p.clone().into());
-                self.blockchain.save_messages(round, raw_messages);
+
+                let fork = self.blockchain.fork();
```
popzxc: Pattern "fork -> do one thing -> merge" occurs several time in the code. Could it make sense to create a HOF for that, e.g.
```rust
fn persist_changes<F: FnOnce(&Fork)>(&self, changes: F) {
    let fork = self.blockchain.fork();
    changes(&fork);
    self.blockchain.merge(fork.into_patch()).expect("..");
}
```

?

exonum-node/src/sandbox/mod.rs
```diff
@@ -12,13 +12,11 @@
 // See the License for the specific language governing permissions and
```
popzxc: Sandbox is now a veteran of traveling 🙃

exonum-node/src/schema.rs
```diff
@@ -0,0 +1,99 @@
...
+        ListIndex, MapIndex,
+    },
+    messages::Verified,
+};
+
+use std::iter;
+
+use crate::messages::{Connect, Message};
+
+const CONSENSUS_MESSAGES_CACHE: &str = "core.consensus_messages_cache";
```
popzxc: Would it make sense to prepend node indexes with `node.` instead of `core.`?

exonum-node/src/schema.rs
```diff
@@ -0,0 +1,99 @@
...
+const CONSENSUS_ROUND: &str = "core.consensus_round";
+const PEERS_CACHE: &str = "core.peers_cache";
+
+/// Schema for an Exonum node.
+#[derive(Debug)]
+pub(crate) struct NodeSchema<T> {
+    access: T,
```
popzxc: There are only 3 indexes in this schema, so I think using a declarative schema layout with eager initialization won't harm the performance.

exonum/Cargo.toml
```diff
@@ -19,33 +19,20 @@ travis-ci = { repository = "exonum/exonum" }
 
 [dependencies]
 log = "0.4.6"
-byteorder = { version = "1.2.7", features = [ "i128" ] }
 hex = "0.4"
-bit-vec = "0.6.0"
-rand = "0.7"
+bit-vec = "0.6.0" # ???
```
popzxc: ?

exonum/src/messages/signed.rs
```diff
@@ -76,8 +79,10 @@ impl_serde_hex_for_binary_value! { SignedMessage }
 /// See module [documentation](index.html#examples) for examples.
 #[derive(Clone, Debug)]
 pub struct Verified<T> {
-    pub(super) raw: SignedMessage,
-    pub(super) inner: T,
+    #[doc(hidden)] // FIXME: how to keep fields private?
```
popzxc: Just providing getters won't work?

exonum-node/src/schema.rs
slowli: I went for backward compatibility, which sort of makes sense for peers cache (not so much for other two indexes).

exonum-node/src/consensus.rs
```diff
@@ -642,7 +663,7 @@ impl NodeHandler {
             "COMMIT ====== height={}, proposer={}, round={}, committed={}, pool={}, hash={}",
             height,
             proposer,
-            round.map_or_else(|| "?".into(), |x| format!("{}", x)),
+            round.map_or_else(|| "?".to_owned(), |x| x.to_string()),
```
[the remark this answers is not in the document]
slowli: Doesn't work: `x` is `Round`, and `ToString::to_string` takes `self` by reference.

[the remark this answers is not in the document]
popzxc: Ah. Sadly, there is no `ToString::into_string`.

exonum-node/src/messages/types.rs
```diff
@@ -0,0 +1,693 @@
...
+
+use bit_vec::BitVec;
+use chrono::{DateTime, Utc};
+use exonum::{
+    blockchain::Block,
+    crypto::{Hash, PublicKey},
+    helpers::{Height, Round, ValidatorId},
+    impl_exonum_msg_try_from_signed,
+    merkledb::{BinaryValue, HashTag},
+    messages::{AnyTx, Precommit, SignedMessage},
```
popzxc: A bit of reflections: In theory, `Precommit` is the biggest thing that currently ties us to the BFT consensus (other one is storing `ConsensusConfig` in core schema, but it's much easier to make abstract).
If we'll be able to replace it with some abstract `ConsensusDecision`, it may be easier to implement nodes with different type of consensus (Casper, GRANDPA, or even plain simple PoW).
However, is this interesting to us, is a big question though.

exonum-node/src/proto/consensus.proto
```diff
@@ -18,31 +18,16 @@ syntax = "proto3";
 
 package exonum.consensus;
 
+// TODO: is it needed?
```
popzxc: I think since we put it in every file, there is no harm to have this as part of coding standard.

test-suite/testkit/src/builder.rs
```diff
@@ -265,14 +268,28 @@ impl TestKitBuilder {
...
 
-        TestKit::assemble(
-            TemporaryDB::new(),
-            network,
-            genesis_config,
-            self.additional_runtimes,
-            self.plugins,
-            self.api_notifier_channel,
-        )
+        #[cfg(feature = "exonum-node")]
```
popzxc: Nit: I'd prefer to have `cfg` on a function instead of code blocks, something like:

```rust
#[cfg(feature = "exonum-node")]
fn complete_initalization(self, teskit: mut Testkit) -> Teskit {
    testkit.set_plugins(self.plugins);
    testkit
}

#[cfg(not(feature = "exonum-node"))]
fn complete_initalization(self, teskit: Testkit) -> Teskit {
    testkit
}
```

and here just:


```rust
let testkit = TestKit::assemble(
    TemporaryDB::new(),
    network,
    genesis_config,
    self.additional_runtimes,
    self.api_notifier_channel,
);
self.complete_initalization(testkit)
```

test-suite/testkit/src/lib.rs
```diff
@@ -251,11 +248,51 @@ impl TestKit {
...
+    /// Needs to be called immediately after node creation.
+    #[cfg(feature = "exonum-node")]
+    pub(crate) fn set_plugins(&mut self, plugins: Vec<Box<dyn NodePlugin>>) {
+        debug_assert!(self.plugins.is_empty());
+        self.plugins = plugins;
+        self.api_aggregator = self.create_api_aggregator();
+    }
+
+    fn create_api_aggregator(&self) -> ApiAggregator {
+        #[cfg(feature = "exonum-node")]
```
popzxc: Nit: Same here.

test-suite/testkit/src/lib.rs
```diff
@@ -797,15 +827,29 @@ impl StoppedTestKit {
...
-            self.db,
-            self.network,
-            // TODO make consensus config optional [ECR-3222]
-            GenesisConfigBuilder::with_consensus_config(ConsensusConfig::default()).build(),
-            runtimes,
-            self.plugins,
-            self.api_notifier_channel,
-        )
+
+        #[cfg(feature = "exonum-node")]
```
popzxc: Nit: Same here.

test-suite/testkit/src/builder.rs
slowli: `self` is partially moved by this point, and the `plugins` field only exists if the `exonum-node` feature is on. Thus, conditionals on the call site seem inevitable.

popzxc: Sad.

exonum-node/src/connect_list.rs
```diff
@@ -14,14 +14,17 @@
...
-use crate::{
-    blockchain::ValidatorKeys,
-    crypto::PublicKey,
-    messages::{Connect, Verified},
-};
+
+#[cfg(test)]
+use crate::messages::Connect;
+#[cfg(test)]
+use exonum::messages::Verified;
```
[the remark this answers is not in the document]
slowli: Could you suggest how? This doesn't work

```rust
#[cfg(test)]
use ::{exonum::messages::Verified, crate::messages::Connect};
```

```text
error[E0433]: failed to resolve: global paths cannot start with `crate`
  --> exonum-node\src\connect_list.rs:25:36
   |
25 | use ::{exonum::messages::Verified, crate::messages::Connect};
   |                                    ^^^^^ global paths cannot start with `crate`
```

components/api/src/cors.rs
```diff
@@ -130,3 +130,30 @@ fn allow_origin_from_str() {
         AllowOrigin::Whitelist(vec!["http://a.org".to_string(), "http://b.org".to_string()]),
     );
 }
+
+#[test]
+fn test_allow_origin_toml() {
```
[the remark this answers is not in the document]
slowli: Well, the test was in the `exonum-cli` crate; I've just moved it close to `AllowOrigin`. I think the test sort of makes sense; it tests that `AllowOrigin` can be (de)serialized with TOML.

exonum-node/src/connect_list.rs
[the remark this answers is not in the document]
popzxc: Doesn't work like that?

```rust
#[cfg(test)]
use {
    super::messages::Connect,
    exonum::messages::Verified,
}
```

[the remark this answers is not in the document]
popzxc:
```rust
#[cfg(test)]
use {
    crate::messages::Connect,
    exonum::messages::Verified,
}
```

may also work (without `::` prefix).

components/api/src/cors.rs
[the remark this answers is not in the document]
popzxc: OK then.

exonum/src/proto/schema/exonum/messages.proto
```diff
@@ -0,0 +1,56 @@
...
+
+// Container for a verified message.
+message Verified {
+  SignedMessage raw = 1;
+}
+
+message CoreMessage {
+  oneof kind {
+    exonum.runtime.AnyTx any_tx = 1;
+    Precommit precommit = 4;
```
alekseysidorov: It may be better to change the numbering so that the `Precommit` message will have id = 2 to make numbering continuous.

slowli: ...But this will break compatibility with previous Exonum versions. Are we OK with this?

popzxc: I think so. Since there are no migrations in the previous release, and there are a lot of breaking changes already, this won't cause any serious harm IMHO.

PR #1699, opened 2020-01-14
Interfaces info [Draft] [ECR-3747]

popzxc:
Well, this PR brings a bunch of changes related to interfaces:

- In Rust runtime method IDs are not anymore assigned automatically, user have to specify them (this is also will be useful for upcoming transaction versioning feature).
- `CallInfo` now contains interface name (it should be empty for default interface call).
- `Runtime` trait now has `interfaces` method, which should return all the interfaces implemented by instance.
- Interfaces implemented by service are now stored in the `Dispatcher` schema.
- Both `Dispatcher` and `ExecutionContext` now check that target service implements certain interface.
- `Dispatcher` schema now provides method `get_instances_by_interface`, which can be used as a base for building `anycast`/`multicast` calls.

---

What has not been done:

- Determining `Supervisor` by its interface instead of ID. There are already a lot of discussion points in this PR, so I think it's better to do it in a separate PR.

---

A bit info about design decisions:

> Why interfaces for instance, and not for artifact?

Two reasons here:

1. It's easier to implement.
2. In theory, service can implement or not implement some interface depending on initialization params (e.g. it can have parameter `configurable` which enables `Configure` interface).

> Why do we need to specify method IDs manually?

Two reasons as well:

1. Explicit is better than implicit.
2. Later, with the transaction versioning, we will have to enforce that logic of transaction with certain ID does not change through all the life cycle of service, and determining ID by order of transaction in the trait is too fragile for such a strong invariant.


components/derive/src/exonum_interface.rs
```diff
@@ -164,6 +165,27 @@ impl TryFrom<&[Attribute]> for ExonumServiceAttrs {
...
+
+    fn try_from(args: &[Attribute]) -> Result<Self, Self::Error> {
+        find_meta_attrs("interface_method", args)
+            .map(|meta| Self::from_nested_meta(&meta))
+            .unwrap_or_else(|| {
+                let msg = format!(
+                    "Unable to find method ID mapping for method. \
+                     It should be specified, e.g. `#[interface_method(id = 0)]`",
+                );
+                return Err(darling::Error::custom(msg));
```
aleksuss: Is the `return` redundant ?

popzxc: Yep.

components/derive/src/lib.rs
```diff
@@ -249,11 +249,43 @@ pub fn service_factory(input: TokenStream) -> TokenStream {
...
+/// ## `id` (required)
+///
+/// ```text
+/// #[interface_method(id = 0)]
+/// ```
+///
+/// Numeric identifier of the method. Should be unique for every method in the trait.
+#[proc_macro_attribute]
+pub fn interface_method(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    // We don't modify the input stream, since `interface_method` attribute only
```
slowli: It could be possible to add a doc comment with the method ID mention, like

```rust
/// # Method identifier
///
/// This method has [method identifier](link to docs?) 0.
```

exonum/src/blockchain/tests/mod.rs
```diff
@@ -165,7 +167,7 @@ impl Transaction {
     ) -> Verified<AnyTx> {
         let tx = AnyTx {
             arguments: self.into_bytes(),
-            call_info: CallInfo::new(instance_id, 0),
+            call_info: CallInfo::new(instance_id, 0, ""),
```
slowli: It could make sense to retain the old constructor and add a new constructor for an interface call (say, `CallInfo::with_interface`).

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -40,6 +43,19 @@ const INSTANCES: &str = "dispatcher_instances";
 const PENDING_INSTANCES: &str = "dispatcher_pending_instances";
 const LOCAL_MIGRATION_RESULTS: &str = "dispatcher_local_migration_results";
 const INSTANCE_IDS: &str = "dispatcher_instance_ids";
+const SERVICE_INTERFACES: &str = "dispatcher_interfaces";
+
+#[derive(Debug, Serialize, Deserialize, BinaryValue)]
+#[binary_value(codec = "bincode")]
```
slowli: IMO, the use of `bincode` is unwarranted here. It can be gotten rid of by storing services in a `Group` of `KeySetIndex`es keyed by the instance ID, for example.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -118,6 +133,13 @@ impl<T: Access> Schema<T> {
         self.access.clone().get_proof_map(INSTANCES)
     }
 
+    /// Returns a mapping between service instance IDs and interfaces that they implement.
+    pub(crate) fn service_interfaces(
+        &self,
+    ) -> ProofMapIndex<T::Base, InstanceId, ServiceInterfaces> {
```
slowli: I'm having doubts that using `bincode` serialization with `HashSet` and then merkelizing the values is sound.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -165,6 +187,20 @@ impl<T: Access> Schema<T> {
     pub fn local_migration_result(&self, instance_name: &str) -> Option<MigrationStatus> {
         self.local_migration_results().get(instance_name)
     }
+
+    /// Returns all the services which implement the provided interface.
+    pub fn get_instances_by_interface(&self, interface_name: &'static str) -> Vec<InstanceId> {
```
slowli: Nit: Is `'static` bound necessary?

popzxc:
>But how is the runtime (or service instance) is supposed to retain this information? Should the service be made stateful for this purpose?

It can store the flag in schema. Since dynamic routing is a syntax sugar provided by `Rust` runtime, and actually service routs the request by itself, in theory in can first read some value from storage, and then "decide" whether it has (or not) some interface.

---

Regarding the motivation of the overall design of PR:

I considered service interfaces more like class interfaces, rather than permissions of an application.
In this sense, ability of service to declare interfaces is close to dynamic dispatching pattern.
Since in current scheme the service itself is capable of dispatching the call, it can adapt its dispatching behaviour based on the input parameters, which is an interesting feature, as for me.

Analogy with software permissions is good, but IMHO since we're creating a framework for *private* blockchains, network administrators first of all should know *what exactly* they are deploying. If they do deploy a service which is malware and tries to mimic the `Supervisor` interface, it's their fault.

And going further, if this approach would have been considered OK, I planned to introduce something like `interfaceLayout` entity, which declared mapping between method IDs and input parameters which should exist in every service that claim to implement that interface.

In this scenario, a `Supervisor` service would have been a service that can accept all the transactions that current `Supervisor` implementation can accept.

---

Regarding the adequateness of decisions -- I don't have a strong opinion on this topic, and of course it will be nice to discuss the best approach here.

Since the original task could had many possible solutions, I decided to implement a sort of PoC to have a field for a discussion.

Probably, I should've marked this PR as a draft.

---

tl;dr: I see your point and it totally makes sense, but my point was that network administrators are responsible for not deploying anything that can harm them (since it's a private network). I don't mind changing the approach.

popzxc:
Closing since it was decided to use a different approach and it'll be easier to implement it from scratch.

PR #1700, opened 2020-01-14
Update sub-project readmes [ECR-3993]

popzxc:
Don't be scared by the size of PR -- most of lines are `LICENSE` files added to sub-projects where they've been missing.

This PR adds `README.md` to sub-projects that didn't have it, and in some cases extends existing `README.md` files with more details.


cli/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+| Command            | Description
+| ------------------ | -----------
+| generate-template  | Generate common part of the nodes configuration
+| generate-config    | Generate public and private configs of the node
+| finalize           | Generate final node configuration
+| run                | Run the node with provided node config
+| run-dev            | Run the node with auto-generated config
+| maintenance        | Perform different maintenance actions
+| help               | Prints this message or the help of the given subcommand(s)
```
aleksuss: `Prints` of `Print` ? I'd prefer third form of verb here.

popzxc: This was copy-pasted from the actual `exonum-executable --help` output. I think it will be the best to leave things here as in output.

cli/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+Consult [the crate docs](https://docs.rs/exonum-cli) for more details.
+
+## How To Run the Network
+
+1. Generate common (template) part of the nodes configuration using
+  `generate-template` command. Generated `.toml` file must be spread
+  among all the nodes and must be used in the following configuration step.
+2. Generate public and secret (private) parts of the node configuration using
+ `generate-config` command. At this step, Exonum will generate master key, from
```
slowli: Nit: Inconsistent indentation (1 vs 2 in the surrounding list items).

components/api/README.md
```diff
@@ -4,10 +4,44 @@
...
-which uses  [`actix`] internally.
-The wrapper is used in [Rust services][rust-runtime] and in plugins
-for the Exonum node.
+`exonum-api` crate provides an extensible interface for building backend-agnostic
+HTTP APIs.
+
+Within Exonum, this crate is used by [Rust services][rust-runtime] and in
+plugins for the Exonum node.
+
+Under the hood `exonum-api` uses `actix`.
```
slowli: Nit: It could make sense to link to `actix` or remove the link at the end of the readme.

components/build/README.md
```diff
@@ -0,0 +1,63 @@
+# `build.rs` utility for Exonum
+
+[![Travis Build Status](https://img.shields.io/travis/exonum/exonum/master.svg?label=Linux%20Build)](https://travis-ci.com/exonum/exonum)
+[![License: Apache-2.0](https://img.shields.io/github/license/exonum/exonum.svg)](https://github.com/exonum/exonum/blob/master/LICENSE)
+![rust 1.36.0+ required](https://img.shields.io/badge/rust-1.36.0+-blue.svg?label=Required%20Rust)
+
+This crate simplifies writing `build.rs` for Exonum and Exonum services.
```
slowli: Nit: I'd use "build script" instead of `build.rs`; this [seems to be the common term used in the Rust documentation](https://doc.rust-lang.org/cargo/reference/build-scripts.html).

components/build/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Since `protobuf` is the Exonum default serialization format, `build.rs` is
+mostly used to compile `protobuf` files and generate corresponding code.
+Generated code is used later by the Exonum core and services.
+
+There are three predefined sets of protobuf sources available for use:
```
slowli: Nit: I'd not mention the number of sets, since it's easy to get wrong (like now, there are four sets mentioned, not three).

components/build/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Consult [the crate docs](https://docs.rs/exonum-build) for more details.
+
+## Examples
+
+Sample `build.rs` using `exonum-build`:
+
+```rust
+fn main() {
+    #[cfg(feature = "with-protobuf")]
```
slowli: Nit: Does it make sense to use conditional compilation in the example? Looks like an over-complication.

components/build/README.md
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[cfg(feature = "with-protobuf")]
+fn gen_proto_files() {
+    use exonum_build::ProtobufGenerator;
+    use std::env;
+
+    let current_dir = env::current_dir().expect("Failed to get current dir.");
+    let protos = current_dir.join("src/proto");
+    println!("cargo:protos={}", protos.to_str().unwrap());
```
slowli: Nit: This "export" part is unnecessary unless we want to get Protobuf declarations from the crate. IMO, it could make sense to either remove this part or document it properly (including the necessity to have `links` entry in the crate manifest, for example).

components/crypto/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Consult [the crate docs](https://docs.rs/exonum-crypto) for more details.
+
+## Examples
+
+Signing data and verifying the signature:
+
+```rust
+fn main() {
+  exonum_crypto::init();
```
slowli: Nit: Why did you use 2-space indentation here, but 4-space elsewhere? 🙃

components/keys/README.md
```diff
@@ -0,0 +1,26 @@
+# Exonum keys management
+
+[![Travis Build Status](https://img.shields.io/travis/exonum/exonum/master.svg?label=Linux%20Build)](https://travis-ci.com/exonum/exonum)
+[![License: Apache-2.0](https://img.shields.io/github/license/exonum/exonum.svg)](https://github.com/exonum/exonum/blob/master/LICENSE)
+![rust 1.36.0+ required](https://img.shields.io/badge/rust-1.36.0+-blue.svg?label=Required%20Rust)
+
+`exonum-keys` provides tools for storing and loading encrypted keys for
+Exonum core.
+
+This is a simply utility crate specialized for the Exonum core needs.
```
slowli: "simpl**e** utility crate"?

components/build/README.md
```diff
@@ -0,0 +1,63 @@
+# `build.rs` utility for Exonum
+
+[![Travis Build Status](https://img.shields.io/travis/exonum/exonum/master.svg?label=Linux%20Build)](https://travis-ci.com/exonum/exonum)
+[![License: Apache-2.0](https://img.shields.io/github/license/exonum/exonum.svg)](https://github.com/exonum/exonum/blob/master/LICENSE)
+![rust 1.36.0+ required](https://img.shields.io/badge/rust-1.36.0+-blue.svg?label=Required%20Rust)
+
+This crate simplifies writing `build.rs` for Exonum and Exonum services.
+
+Since `protobuf` is the Exonum default serialization format, `build.rs` is
```
slowli: Nit: Code formatting looks redundant; simply using "Protobuf" could suffice.

components/crypto/README.md
popzxc: Sublime Text for some reason decided that tab there should equal to two spaces ^^"

popzxc:
@slowli Yeah, at first I was concerned about including code snippets, but since we're going to test them automatically, I think it's not a big deal.

PR #1701, opened 2020-01-15
Add possibility to specify service method IDs [ECR-3896]

popzxc:
Since this piece of work is not directly related to ECR-3747, I decided to move it in a separate PR.

Now it's a first step for transaction versioning support in Rust runtime.

As an addition for `#[interface_method(id = 0)]` I added a possibility to increment IDs automatically:

```rust
    #[exonum_interface(auto_ids)]
    trait SampleInterface<Ctx> {
        type Output;
        fn timestamp(&self, ctx: Ctx, arg: String) -> Self::Output;
    }
```


components/derive/src/exonum_interface.rs
```diff
@@ -164,17 +167,39 @@ impl TryFrom<&[Attribute]> for ExonumServiceAttrs {
     }
 }
 
+#[derive(Debug, FromMeta)]
+struct MethodIdAttr {
```
slowli: Nit: Shouldn't this type be named `InterfaceMethodAttrs` or something like that?

components/derive/src/exonum_interface.rs
```diff
@@ -139,22 +140,24 @@ impl ServiceMethodDescriptor {
 
 #[derive(Debug, FromMeta)]
 #[darling(default)]
-struct ExonumServiceAttrs {
+struct ExonumInterfaceAttrs {
     #[darling(rename = "crate")]
     cr: RustRuntimeCratePath,
+    id_auto_increment: bool,
```
slowli: Nit / bikeshedding: Is the name intentionally long? 🙃 I'd prefer something shorter, like `auto_ids`.

components/derive/src/lib.rs
```diff
@@ -249,11 +249,52 @@ pub fn service_factory(input: TokenStream) -> TokenStream {
 ///
 /// Prefix of the `exonum` crate has two main values - `crate` or `exonum`. The default value
 /// is `exonum`.
+///
+/// ## `id_auto_increment`
+///
+/// ```text
+/// #[exonum_interface(id_auto_increment = true)]
```
slowli: Nit: Doesn't `#[exonum_interface(id_auto_increment)]` work as well? I'd prefer this form here and in actual code.

runtimes/rust/src/lib.rs
```diff
@@ -65,6 +65,7 @@
 //!     type Output;
 //!     // Each method of the trait should have a signature of the following format.
 //!     // The argument should implement the `BinaryValue` trait.
+//!     #[interface_method(id = 0)]
```
slowli: Nit: It could make sense to explain `interface_method` attribute here.

components/derive/src/exonum_interface.rs
popzxc: Yeah, the idea was to make it less pretty. However, I don't mind `auto_ids`.

PR #1702, opened 2020-01-15
Make the TemporaryDB more safety

aleksuss:
## Overview

---

### Definition of Done

### DIFF
--- a/components/merkledb/src/backends/temporarydb.rs
+++ b/components/merkledb/src/backends/temporarydb.rs
@@ -19,10 +19,13 @@ use tempfile::TempDir;
 
 use std::sync::Arc;
 
-use super::rocksdb::RocksDB;
+use crate::backends::rocksdb::{RocksDB, RocksDBSnapshot};
 use crate::{db::DB_METADATA, Database, DbOptions, Iter, Patch, ResolvedAddress, Result, Snapshot};
 
-/// Wrapper over the `RocksDB` backend which stores data in the temporary directory
+#[cfg(test)]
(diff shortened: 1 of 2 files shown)


components/merkledb/src/backends/temporarydb.rs
```diff
@@ -30,15 +30,20 @@ use crate::{db::DB_METADATA, Database, DbOptions, Patch, Result, Snapshot};
 #[derive(Debug)]
 pub struct TemporaryDB {
     inner: RocksDB,
-    dir: TempDir,
+    dir: Arc<TempDir>,
+}
+
+struct TemporarySnapshot {
```
popzxc: Nit: I'd add a docstring with some description like "`RocksDB` snapshot wrapper with the `TempDir` handle to prevent it from removing until all the snapshots and database itself are dropped"

components/merkledb/src/backends/temporarydb.rs
```diff
@@ -87,11 +97,23 @@ impl TemporaryDB {
...
+
+    fn temporary_snapshot(&self) -> TemporarySnapshot {
+        TemporarySnapshot {
+            snapshot: self.inner.rocksdb_snapshot(),
+            _dir: Arc::clone(&self.dir),
+        }
+    }
+
+    #[cfg(test)]
+    fn get_db_path(&self) -> &Path {
```
slowli: Nit: IMO, this method could be inlined; adding a separate one-line method just for tests looks like an overkill.

PR #1704, opened 2020-01-15
Forward-compatible auth [ECR-3932]

slowli:
## Overview

This PR makes `Caller` type (i.e., information about call authorization) forward-compatible and introduces its uniform representation (`CallerAddress`). The `cryptocurrency-advanced` example is changed to support generic `Caller`s, and a toy multisig service is implemented to test this functionality.

Discussion:

- Should `CallerAddress` be a type alias or a separate type (maybe even with another serialization format; cf. Base58Check and bech32 in Bitcoin)?

---

See: https://jira.bf.local/browse/ECR-3932


exonum/src/runtime/types.rs
```diff
@@ -46,6 +46,9 @@ pub type InstanceId = u32;
 /// Identifier of the method in the service interface required for the call.
 pub type MethodId = u32;
 
+/// Uniform presentation of a `Caller`.
+pub type CallerAddress = Hash;
```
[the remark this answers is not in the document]
slowli: Yeah, probably. Although it doesn't on its own answer the discussion question whether we need some error-detecting encoding for addresses (I *think* it could be added later.)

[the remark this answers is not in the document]
aleksuss: Agree, that we'll be able to change it in the future. If we will need it.

examples/cryptocurrency-advanced/backend/src/proto/service.proto
```diff
@@ -45,7 +45,7 @@ message CreateWallet {
 // Wallet information stored in the database.
 message Wallet {
   // `PublicKey` of the wallet.
```
aleksuss:
```suggestion
  // Address of the wallet's owner.
```

examples/cryptocurrency-advanced/backend/src/proto/service.proto
```diff
@@ -44,8 +44,8 @@ message CreateWallet {
 
 // Wallet information stored in the database.
 message Wallet {
-  // `PublicKey` of the wallet.
-  exonum.crypto.PublicKey pub_key = 1;
+   // Address of the wallet's owner.
```
popzxc:
```suggestion
  // Address of the wallet's owner.
```

examples/cryptocurrency-advanced/backend/src/transactions.rs
```diff
@@ -55,7 +55,7 @@ pub enum Error {
 #[protobuf_convert(source = "proto::Transfer", serde_pb_convert)]
 pub struct Transfer {
     /// `PublicKey` of receiver's wallet.
```
popzxc:
```suggestion
    /// `Address` of receiver's wallet.
```

exonum/src/runtime/types.rs
```diff
@@ -729,12 +730,176 @@ impl ProtobufConvert for MigrationStatus {
...
+            }
+        } else if pb.has_blockchain() {
+            Caller::Blockchain
+        } else {
+            bail!("No variant specified for `Caller`");
+        })
+    }
+}
+
+/// Uniform presentation of a `Caller`.
```
popzxc: It will be nice to have an explanation how clients (e.g. light clients) can obtain `CallerAddress, having the `PublicKey` of recipient.

PR #1705, opened 2020-01-16
Clarify docs in Blockchain schema [ECR-4063] [ECR-4098]

popzxc:
This PR adds a bit more explanations for `Blockchain` schema.

### DIFF
--- a/exonum/src/blockchain/schema.rs
+++ b/exonum/src/blockchain/schema.rs
@@ -126,7 +126,7 @@ impl<T: Access> Schema<T> {
     /// ended up with a particular status.
     /// For an execution that resulted in an error, this will be an usual proof of existence.
     /// If transaction was executed successfully, such a proof will be a proof of absence.
-    /// Since the number of transaction in a block is mentioned in the block header, user
+    /// Since the number of transactions in a block is mentioned in the block header, the user
     /// will be able to distinguish absence of error (meaning successful execution) from
     /// absence of transaction with such an index: if index is less than amount of transactions
     /// in block, proof denotes successful execution, otherwise transaction with such an index
--- a/exonum/src/blockchain/schema.rs
+++ b/exonum/src/blockchain/schema.rs
@@ -122,10 +122,15 @@ impl<T: Access> Schema<T> {
 
     /// Returns a record of errors that occurred during execution of a particular block.
     ///
-    /// This method can be used to retrieve a proof that execution of a certain transaction
-    /// ended up with a particular status. Since the number of transaction in a block is
-    /// mentioned in the block header, a proof of absence of an error for a transaction
-    /// with a particular index means that it was executed successfully.
+    /// This method can be used to build a proof that execution of a certain transaction
+    /// ended up with a particular status.
+    /// For an execution that resulted in an error, this will be an usual proof of existence.
+    /// If transaction was executed successfully, such a proof will be a proof of absence.
+    /// Since the number of transaction in a block is mentioned in the block header, user
+    /// will be able to distinguish absence of error (meaning successful execution) from
+    /// absence of transaction with such an index: if index is less than amount of transactions
+    /// in block, proof denotes successful execution, otherwise transaction with such an index
+    /// does not exist in the block.
     ///
     /// Similarly, execution errors of the `before_transactions` / `after_transactions` hooks can be proven
     /// to external clients. Discerning successful execution from a non-existing service requires prior knowledge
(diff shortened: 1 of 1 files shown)


exonum/src/blockchain/schema.rs
```diff
@@ -122,10 +122,15 @@ impl<T: Access> Schema<T> {
...
     ///
-    /// This method can be used to retrieve a proof that execution of a certain transaction
-    /// ended up with a particular status. Since the number of transaction in a block is
-    /// mentioned in the block header, a proof of absence of an error for a transaction
-    /// with a particular index means that it was executed successfully.
+    /// This method can be used to build a proof that execution of a certain transaction
+    /// ended up with a particular status.
+    /// For an execution that resulted in an error, this will be an usual proof of existence.
+    /// If transaction was executed successfully, such a proof will be a proof of absence.
+    /// Since the number of transaction in a block is mentioned in the block header, user
```
slowli:
```suggestion
    /// Since the number of transactions in a block is mentioned in the block header, the user
```

exonum/src/blockchain/schema.rs
```diff
@@ -378,7 +383,12 @@ where
 /// an error may occur. Since Exonum services may call each other's methods, `CallSite` is
 /// richer than `CallInBlock`.
 ///
+/// One example of difference between these types is [`CallType::Constructor`]: since services
+/// are constructed outside of the block processing routine, this kind of errors cannot be
```
slowli: I don't quite understand what the "block processing routine" is supposed to mean. Services *are* constructed as a result of usual block calls to the supervisor service, aren't they? (Well, builtin services being an exception.) The difference is that service constructors are not topmost calls.

popzxc: What I meant here is that service creation happens outside of "interacting points" within block (`before/after_transactions` hook, transaction execution). 

From the point of users view, services are constructed after `after_transactions` hooks, and there is no such a point of interaction for them, it's totally under the hood.

Do you have a suggestion how to phrase it with less ambiguity?

slowli: > services are constructed after `after_transactions` hooks

Ah, I understand now; we've got a terminology issue. I've understood "service creation" to be equivalent to `Runtime::initiate_starting_service` calls, and these calls happen *within* "interacting points" of a block. Instantiating service objects indeed happens both in these calls and during block commitment.

popzxc: Yeah, but invocation of `Runtime::initiate_starting_service` is hidden from users, so I decided that it's better to not disclose it.

exonum/src/blockchain/schema.rs
```diff
@@ -122,10 +122,15 @@ impl<T: Access> Schema<T> {
...
-    /// ended up with a particular status. Since the number of transaction in a block is
-    /// mentioned in the block header, a proof of absence of an error for a transaction
-    /// with a particular index means that it was executed successfully.
+    /// This method can be used to build a proof that execution of a certain transaction
+    /// ended up with a particular status.
+    /// For an execution that resulted in an error, this will be an usual proof of existence.
+    /// If transaction was executed successfully, such a proof will be a proof of absence.
+    /// Since the number of transactions in a block is mentioned in the block header, the user
+    /// will be able to distinguish absence of error (meaning successful execution) from
+    /// absence of transaction with such an index: if index is less than amount of transactions
```
dmitry-timofeev: I'd note that by the key itself it is not possible to tell if it belongs to the execution of a transaction with a certain message hash. My understanding is that a proper proof must first prove the tx inclusion in a block at a certain location, then — the call result (using an already proved location).

I had briefly described it here:
* https://github.com/exonum/exonum-java-binding/blob/d47524e388340b64d5e12366a658ac21d83ce931/exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/Blockchain.java#L302-L309
* https://github.com/exonum/exonum-java-binding/blob/d47524e388340b64d5e12366a658ac21d83ce931/exonum-java-binding/core/src/main/java/com/exonum/binding/core/blockchain/Blockchain.java#L81-L86

dmitry-timofeev: I.e., one shall not check the index against the number of transaction in a block, but _construct_ a key by using an already proved location.

popzxc: I added a note that this kind of proof should not be considered a proof of inclusion. Are you OK with such a statement?

popzxc: Oh snap! It seems like I forgot to push this change. Redirected the request (https://github.com/exonum/exonum/pull/1724#discussion_r369362951).

PR #1706, opened 2020-01-16
Add resume service support to supervisor [ECR-3773]

alekseysidorov:


services/supervisor/src/transactions.rs
```diff
@@ -131,6 +128,47 @@ impl StopService {
     }
 }
 
+impl ResumeService {
+    fn validate(&self, context: &CallContext<'_>) -> Result<(), ExecutionError> {
+        let instance = get_instance(context, self.instance_id)?;
+
+        match instance.status {
```
popzxc: Nit: you can just write `if instance.status != InstanceStatue::Stopped`

aleksuss: Really ? `Option` derives `PartialOrd` and `Eq`. Take a look [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b3ff083e7147f89c341020d7392a1f31)

services/supervisor/src/transactions.rs
```diff
@@ -131,6 +128,47 @@ impl StopService {
...
+                ConfigurationError::MalformedConfigPropose.with_description(format!(
+                    "Discarded an attempt to resume service with incorrect artifact version: \
+                     got {}, expected {}",
+                    self.artifact.version,
+                    instance.data_version()
+                )),
+            );
+        }
+
+        if instance.spec.artifact.name != self.artifact.name {
```
slowli: Nit: It could make sense to use [`ArtifactId::is_upgrade_of`](https://github.com/exonum/exonum/blob/616622f3b7e7bf7858590bae6b11732704fcfd7c/exonum/src/runtime/types.rs#L201) to be more forward-compatible. OTOH, the error message will be less specific in this case.

services/supervisor/src/proto_structures.rs
```diff
@@ -124,6 +137,8 @@ pub enum ConfigChange {
     StartService(StartService),
     /// Existing service instance stop request.
     StopService(StopService),
+    /// Previously stopped service instance resume request.
```
slowli: Nit: IMO, "... request" descriptions are weirdly phrased. I'd suggest something like "Request to resume a previously stopped service instance."

services/supervisor/src/proto/service.proto
```diff
@@ -65,6 +65,16 @@ message ServiceConfig {
   bytes params = 2;
 }
 
+// Request for the resume previously stopped service instance.
```
slowli:
```suggestion
// Request to resume a previously stopped service instance.
```

components/crypto/src/lib.rs
```diff
@@ -469,7 +469,7 @@ pub struct KeyPair {
 
 impl KeyPair {
     pub fn from_keys(public_key: PublicKey, secret_key: SecretKey) -> Self {
-        debug_assert!(
+        assert!(
```
slowli: Why was this changed?

services/supervisor/src/transactions.rs
alekseysidorov: You are right, today I am a little bit stupid.

CHANGELOG.md
```diff
@@ -118,6 +118,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Error code values were changed and split into several enum
   representing sub-groups. (#1680)
 
+- Added support for resuming previously stopped services. (#1706)
```
slowli: This isn't a breaking change, is it?

services/supervisor/tests/supervisor/service_lifecycle.rs
```diff
@@ -178,3 +177,128 @@ fn stop_already_stopped_service() {
             )
     )
 }
+
+#[test]
+fn resume_stopped_service() {
```
slowli: Nit: Could you add a test to check that it's impossible to request resume and some other operation regarding the service (e.g., changing its config) in the same update?

components/crypto/src/lib.rs
alekseysidorov: Because this code breaks tests in release mode and @aleksuss asked me to fix this micro issue.

services/supervisor/src/transactions.rs
alekseysidorov: `is_upgrade_of` has a little bit different behaviour.

PR #1707, opened 2020-01-16
Transaction versioning [ECR-3896]

popzxc:
This PR brings the following goods:

- Rules on transaction persistence are now documented in the `runtime` module docs (tl;dr: you can update the tx logic, but you should not put some transaction on the place from where other transaction was removed);
- `MethodRemoved` variant was added to the `CommonError` enum, it is recommended to always return it from removed methods;
- `interface_method` attribute now supports `removed` parameter, which removes method from the trait (so users don't have to implement it) and makes attempt to invoke this method return `CommonError::MethodRemoved`.

---

Which goods aren't brought:

- Handy interface for making transactions deprecated.

It seems that there is no easy way to do it right now, and since there is an alternative (just write in the docs that method is deprecated and will be removed soon), I decided that we can live without this feature in release.

Anyway, if such a functionality will become must-have, it shouldn't be a breaking change to introduce it later.

---

With this PR, the following workflow may be used for updating transactions:

1. Create some interface with transactions.
2. Within next release, fix a bug in method A.
3. Within next release, write in the docs that method A is deprecated and will be removed in the next release.
4. Within next release, mark method A as removed in interface, and create method B to be used instead.
5. ?????
6. PROFIT


components/derive/src/exonum_interface.rs
```diff
@@ -234,7 +286,7 @@ impl ExonumInterface {
                             let msg = format!("Method ID {} is already used", method_id);
```
slowli: It could make sense to perform a similar check for removed method IDs. As far as I see, presently it's possible to define a method with the same ID as the removed one, and at best the compiler will complain about unreachable code in the method ID matcher (although I'm not sure such warnings are not suppressed for macro-generated code).

popzxc: My bad, I forgot to add it.

components/derive/src/exonum_interface.rs
```diff
@@ -138,13 +139,57 @@ impl ServiceMethodDescriptor {
...
+                        Err(darling::Error::custom(msg).with_span(&lit))
+                    }
+                } else {
+                    let msg = "Incorrect method ID, must be an unsigned integer";
+                    Err(darling::Error::custom(msg).with_span(&item))
+                }
+            })
+            .collect();
+
+        match ids {
```
slowli: Nit: Can't it be shortened as `ids.map(|ids| Self { ids })`?

components/derive/src/exonum_interface.rs
```diff
@@ -219,9 +271,12 @@ impl ExonumInterface {
         // Process trait methods.
         let mut methods = Vec::with_capacity(item_trait.items.len());
         let mut has_output = false;
-        let mut used_method_ids = HashSet::new();
         let mut next_method_id = 0;
 
+        // Initialize used methods with removed method IDs.
+        let mut used_method_ids: HashSet<u32> =
```
slowli: Nit: It could make sense to have more specific error message if the existing method ID collides with one of removed IDs.

exonum/src/runtime/versioning.rs
```diff
@@ -69,6 +69,29 @@
...
+//!
+//! **Important:** Transaction numeric identifier is considered a constant during all the time of
+//! service existence. It means that if transaction was declared with certain ID, its logic can
+//! be updated (e.g., to fix a bug) or be removed, but it **never** should be replaced with other
+//! transaction.
+//!
+//! If transaction was removed from service, attempt to invoke it should always
+//! result in returning an `ExecutionError`.
+//!
+//! You should use [`CommonError::MethodRemoved`] to report the error in case a method was removed.
```
slowli: This isn't actually a link, is it? You should use the full name (`CommonError::MethodRemoved`) in the link below, and probably link to the variant (via `#variant.MethodRemoved`).

runtimes/rust/src/lib.rs
```diff
@@ -153,6 +153,56 @@
 //! # impl Service for StatefulService {}
 //! ```
 //!
+//! ## Transactions removing
```
slowli: Nit: "Removing Transactions" looks more appropriate.

PR #1708, opened 2020-01-16
Change type for precommits in BlockProof [ECR-4108]

aleksuss:
## Overview

Use `repeated SignedMessage` instead of  `repeated Verified` in `BlockProof` in the proto description.

---

See: https://jira.bf.local/browse/ECR-4108

### Definition of Done

### DIFF
--- a/exonum/src/messages/signed.rs
+++ b/exonum/src/messages/signed.rs
@@ -26,7 +26,7 @@ use std::{
 };
 
 use crate::{
-    crypto::{self, Hash, PublicKey, SecretKey, Signature},
+    crypto::{self, Hash, PublicKey, SecretKey},
     messages::types::SignedMessage,
     proto,
 };
@@ -231,22 +231,12 @@ where
     type ProtoStruct = proto::SignedMessage;
 
     fn to_pb(&self) -> Self::ProtoStruct {
-        let mut message = Self::ProtoStruct::new();
         let signed_message = self.as_raw();
-
-        message.set_payload(signed_message.payload.to_pb());
-        message.set_author(signed_message.author.to_pb());
-        message.set_signature(signed_message.signature.to_pb());
-        message
+        signed_message.to_pb()
     }
 
-    fn from_pb(mut pb: Self::ProtoStruct) -> Result<Self, Error> {
-        let signed_message = SignedMessage {
-            payload: pb.take_payload(),
-            author: PublicKey::from_pb(pb.take_author())?,
-            signature: Signature::from_pb(pb.take_signature())?,
-        };
-
+    fn from_pb(pb: Self::ProtoStruct) -> Result<Self, Error> {
+        let signed_message = SignedMessage::from_pb(pb)?;
         signed_message.into_verified()
     }
 }
--- a/exonum/src/messages/signed.rs
+++ b/exonum/src/messages/signed.rs
@@ -26,7 +26,7 @@ use std::{
 };
 
 use crate::{
-    crypto::{self, Hash, PublicKey, SecretKey},
+    crypto::{self, Hash, PublicKey, SecretKey, Signature},
     messages::types::SignedMessage,
     proto,
 };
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -106,6 +106,10 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
   This method is used to resume a previously stopped services.
   (#1693)
 
+- `BlockProof` structure has been changed. The field `precommits` has a type
```
dmitry-timofeev: Given BlockProof was added after the latest release, we could probably leave this change out.

exonum-node/src/sandbox/tests/basic.rs
```diff
@@ -234,12 +234,12 @@ fn test_retrieve_block_and_precommits() {
...
     let block_proof = bl_proof_option.unwrap();
     let block = block_proof.block;
-    let precommits: Vec<Verified<Precommit>> = block_proof.precommits;
     let expected_height = target_height.previous();
     let expected_block_hash = block.object_hash();
 
     assert_eq!(expected_height, block.height);
-    for precommit in precommits {
+    for precommit in block_proof.precommits {
+        let precommit: Verified<Precommit> = precommit.into_verified().unwrap();
```
popzxc: If we want to test here that signatures are correct, I think it worth to use `expect`.
Otherwise, `try_from` will be enough.

Same for other similar places in tests.

exonum/src/blockchain/block.rs
```diff
@@ -216,7 +216,7 @@ pub struct BlockProof {
     /// in the block, etc.
     pub block: Block,
     /// List of `Precommit` messages for the block.
-    pub precommits: Vec<Verified<Precommit>>,
+    pub precommits: Vec<SignedMessage>,
```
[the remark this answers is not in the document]
popzxc: In theory, I don't think that it's a bad idea to generalize approach for work with consensus decisions (see https://github.com/exonum/exonum/pull/1698#discussion_r366707308), but the current implementation raises some questions indeed.

components/explorer/src/lib.rs
```diff
@@ -629,10 +629,11 @@ impl<'a> BlockchainExplorer<'a> {
         self.schema.transactions().get(tx_hash)
     }
 
-    fn precommits(&self, block: &Block) -> Vec<Verified<Precommit>> {
+    fn precommits(&self, block: &Block) -> Vec<SignedMessage> {
         self.schema
             .precommits(&block.object_hash())
             .iter()
+            .map(Verified::into_raw)
```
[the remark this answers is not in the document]
popzxc: However, as @slowli correctly noted above, it going much further than what we talked about.

Thus it'll be better to create a separate task, discuss it and implement later.

exonum/src/messages/signed.rs
```diff
@@ -228,17 +228,27 @@ impl<T> ProtobufConvert for Verified<T>
...
+        message.set_signature(signed_message.signature.to_pb());
+        message
     }
 
     fn from_pb(mut pb: Self::ProtoStruct) -> Result<Self, Error> {
-        let raw = SignedMessage::from_pb(pb.take_raw())?;
-        raw.into_verified()
+        let signed_message = SignedMessage {
+            payload: pb.take_payload(),
+            author: PublicKey::from_pb(pb.take_author()).expect("Failed to convert from protobuf."),
```
popzxc: Shouldn't we return an error instead of `expect`?

aleksuss: Good catch. I suppose it makes sense 👍

exonum/src/messages/signed.rs
```diff
@@ -228,17 +228,26 @@ impl<T> ProtobufConvert for Verified<T>
...
+    type ProtoStruct = proto::SignedMessage;
 
     fn to_pb(&self) -> Self::ProtoStruct {
-        let mut verified = Self::ProtoStruct::new();
-        verified.set_raw(self.as_raw().to_pb());
-        verified
+        let mut message = Self::ProtoStruct::new();
+        let signed_message = self.as_raw();
+
+        message.set_payload(signed_message.payload.to_pb());
```
popzxc: Yet another nit: `SignedMessage` already implements `ProtobufConvert`. Can we reuse it here?

Like:

```rust
fn to_pb(&self) -> Self::ProtoStruct {
    self.as_raw().to_pb()
}
```

same for `from_pb`.

aleksuss: "An hour in the morning is worth two in the evening" 🙂

PR #1709, opened 2020-01-17
Add workaround for segfault in TestkitApi

alekseysidorov:
This PR adds ManualDrop wrapper to take control of `actix_web::TestServer` destruction.
This is necessary because the server does not wait for the correct stop of its threads, which leads to runtime errors.


alekseysidorov:
https://github.com/exonum/exonum/pull/1713
Closed in favor of more correct implementation.

PR #1710, opened 2020-01-17
Make public items non-exhaustive [ECR-3944]

popzxc:
This PR makes public items in core and its upstream dependencies (actually, `merkledb` only) non-exhaustive.

Not all items were modified, only these which may be extended in the future (but anyway, that was A LOT).

For details on non-exhaustive pattern, see the [corresponding RFC](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md).

Later, when we won't be tied to 1.36, we'll be able to replace phantom fields/variants with `#[non_exhaustive]` attribute.


components/merkledb/src/indexes/proof_map/node.rs
```diff
@@ -27,6 +27,9 @@ use crate::{BinaryKey, BinaryValue, HashTag, ObjectHash};
 const BRANCH_NODE_SIZE: usize = 2 * (HASH_SIZE + PROOF_PATH_SIZE);
 
 /// Enum that represents the type of the `ProofMap` node.
+///
+/// This type is not intended to be exhaustively matched. It can be extended in the future
```
slowli: This type is not public, is it? If it is, I'd view publicity as a problem, not exhaustive matching.

components/merkledb/src/migration.rs
```diff
@@ -547,6 +550,11 @@ pub enum MigrationError {
     /// Migration has been aborted.
     #[fail(display = "Migration was aborted")]
     Aborted,
+
+    /// Never actually generated.
```
slowli: I'd argue that `MigrationError` here should not admit new variants (at least in MerkleDB 1.x).

components/merkledb/src/views/address.rs
```diff
@@ -289,17 +289,25 @@ pub struct ResolvedAddress {
     /// for different views in the same column family. In other words, key spaces for two addresses
     /// with equal `name` and `id`s `Some(x)` and `Some(y)`, `x != y`, must not intersect.
     pub id: Option<NonZeroU64>,
+
+    /// No-op field for forward compatibility.
+    non_exhaustive: (),
```
slowli: I'd argue that breaking `ResolvedAddress` structure will lead to major breakage in MerkleDB, so this seems overly defensive. Also, the structure is only public because of its use in `Snapshot`, which - let's face it - we don't intend for third parties to implement.

components/merkledb/src/views/mod.rs
```diff
@@ -38,6 +38,9 @@ mod tests;
 /// Represents current view of the database by specified `address` and
 /// changes that took place after that view had been created. `View`
 /// implementation provides an interface to work with related `changes`.
+///
+/// This type is not intended to be exhaustively matched. It can be extended in the future
```
slowli: `View` is non-public, is it?

exonum-node/src/sandbox/mod.rs
```diff
@@ -1342,10 +1334,7 @@ mod unit_tests {
         // See comments to `test_sandbox_recv_and_send`.
         let (public, secret) = gen_keypair();
```
slowli: Nit: Could you rename `public` to `consensus_key` for better readability?

exonum/src/blockchain/block.rs
```diff
@@ -138,9 +138,37 @@ pub struct Block {
     pub error_hash: Hash,
     /// Additional information that can be added into the block.
     pub additional_headers: AdditionalHeaders,
+
+    /// No-op field for forward compatibility.
+    #[protobuf_convert(skip)]
```
slowli: I'd argue that `additional_headers` take care of possible extensions, so this looks redundant.

exonum/src/blockchain/config.rs
```diff
@@ -144,6 +166,39 @@ impl ConsensusConfig {
     /// Time that will be added to round timeout for each next round in terms of percent of first_round_timeout.
     pub const TIMEOUT_LINEAR_INCREASE_PERCENT: u64 = 10; // 10%
 
+    /// Creates a new `ConsensusConfig` object.
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
slowli: Could it make sense to introduce a builder here? I don't think having a method with 10 args is user-friendly 🙃

exonum/src/blockchain/config.rs
```diff
@@ -50,6 +50,22 @@ pub struct ValidatorKeys {
...
+
+    /// No-op field for forward compatibility.
+    #[protobuf_convert(skip)]
+    #[serde(default, skip)]
+    non_exhaustive: (),
+}
+
+impl ValidatorKeys {
+    /// Creates a new `ValidatorKeys` object.
+    pub fn new(consensus_key: PublicKey, service_key: PublicKey) -> Self {
```
slowli: This method will break if new fields are added to `ValidatorKeys`, right? Should it be documented as unstable then?

exonum/src/blockchain/mod.rs
```diff
@@ -220,7 +220,10 @@ impl BlockchainMut {
             .consensus_config_entry()
             .set(genesis_config.consensus_config);
 
-        for ArtifactSpec { artifact, payload } in genesis_config.artifacts {
+        for ArtifactSpec {
+            artifact, payload, ..
```
slowli: Meh, I'd prefer getting rid of destructuring here.

exonum/src/messages/types.rs
```diff
@@ -40,6 +40,24 @@ pub struct SignedMessage {
     pub author: PublicKey,
     /// Digital signature over `payload` created with `SecretKey` of the author of the message.
     pub signature: Signature,
+
+    /// No-op field for forward compatibility.
+    #[protobuf_convert(skip)]
```
slowli: I don't think extending `SignedMessage` is feasible in 1.x; it's a simple type encompassing data signed with a Ed25519 key.

exonum/src/runtime/versioning.rs
```diff
@@ -157,32 +157,35 @@ use crate::runtime::{ArtifactId, CoreError, ExecutionError, ExecutionFail};
...
+///     "1.5.7".parse().unwrap(),
+/// ).unwrap();
 /// req.try_match(&valid_artifact).unwrap();
 ///
 /// // This artifact is outdated.
-/// let outdated_artifact = ArtifactId {
-///     version: "1.2.0".parse().unwrap(),
-///     ..valid_artifact.clone()
-/// };
+/// let outdated_artifact = ArtifactId::new(
```
slowli: Nit: Could make sense to clone the `outdated_artifact` from `valid_artifact` and change its version.

exonum/src/runtime/versioning.rs
```diff
@@ -157,32 +157,35 @@ use crate::runtime::{ArtifactId, CoreError, ExecutionError, ExecutionFail};
...
-/// let valid_artifact = ArtifactId {
-///     runtime_id: RuntimeIdentifier::Rust as _,
-///     name: "some.Service".to_owned(),
-///     version: "1.5.7".parse().unwrap(),
-/// };
+/// let valid_artifact = ArtifactId::new(
+///     RuntimeIdentifier::Rust as u32,
+///     "some.Service".to_owned(),
+///     "1.5.7".parse().unwrap(),
+/// ).unwrap();
```
slowli: Nit: Maybe, it could make sense to introduce the `unchecked` constructor? Alternatively, you could use a `fn main() -> Result<(), failure::Error>` wrapper for the test.

runtimes/rust/src/lib.rs
```diff
@@ -640,6 +634,12 @@ impl Runtime for RustRuntime {
                 self.remove_started_service(spec);
             }
             InstanceStatus::Migrating(_) => { /* Do nothing. */ }
+            other => {
+                log::warn!(
```
slowli: I think the runtime should panic here, since this is a sign of the misconfiguration of the node binary.

services/supervisor/tests/supervisor/supervisor_config.rs
```diff
@@ -66,15 +66,13 @@ fn initial_configuration() {
...
-    let instance_spec = InstanceSpec {
-        id: SUPERVISOR_INSTANCE_ID,
-        name: Supervisor::NAME.into(),
-        artifact: Supervisor.artifact_id(),
-    };
-    let incorrect_instance = InstanceInitParams {
-        instance_spec,
-        constructor: incorrect_config,
-    };
+    let instance_spec = InstanceSpec::from_raw_parts(
```
slowli: Nit: `ArtifactId` has API to create `InstanceInitParams`:

```rust
let incorrect_instance = Supervisor
    .artifact_id()
    .into_default_instance(SUPERVISOR_INSTANCE_ID, Supervisor::NAME)
    .with_constructor(incorrect_config);
```

components/merkledb/src/indexes/proof_map/node.rs
popzxc: Ah, I automated the process of adding non-exhaustive guards, but when I was sifting them, forgot to remove comments in several places.

components/merkledb/src/migration.rs
popzxc: Then it'll be just inconsistent with the majority of other error types. Are we OK with that?

slowli: I think yes; users of the main crate are not exposed to this error directly, but rather via `MigrationError` from the `runtime::migrations` module.

exonum/src/runtime/versioning.rs
popzxc: Yeah, got idea, I've changed `unwrap`s with `?` as well.

components/merkledb/src/views/mod.rs
popzxc: It was (indirectly). Made it `pub(crate)`.

exonum/src/runtime/types.rs
```diff
@@ -214,16 +214,26 @@ impl ArtifactId {
...
-            non_exhaustive: (),
-        };
+        let artifact = Self::from_raw_parts(runtime_id.into(), name.into(), version);
         artifact.validate()?;
         Ok(artifact)
     }
 
+    /// Creates a new artifact identifier from prepared parts without any checks.
+    ///
+    /// Since the internal structure of `InstanceSpec` can change, this method is considered
```
slowli: `InstanceSpec` most probably needs to be changed to `ArtifactId`.

examples/sample_runtime/src/main.rs
```diff
@@ -151,6 +151,10 @@ impl Runtime for SampleRuntime {
             InstanceStatus::Migrating(_) => {
                 // We don't migrate service data in this demo.
             }
+
+            _ => {
+                // We aren't interesting in other possible statuses.
```
Kormick:
```suggestion
                // We aren't interested in other possible statuses.
```

runtimes/rust/src/lib.rs
```diff
@@ -635,8 +635,10 @@ impl Runtime for RustRuntime {
             }
             InstanceStatus::Migrating(_) => { /* Do nothing. */ }
             other => {
-                log::warn!(
-                    "Received non-expected service status: {}; ignoring the request",
+                panic!(
+                    "Received non-expected service status: {}; \
+                     Rust runtime isn't prepared to process this action, \
+                     probably node binary is configured incorrectly",
```
slowli: Nit: I'd specify that the rust runtime is probably outdated relative to the core library in this case; it could help pinpoint the problem. Also, I wonder if it could make sense to have a method `InstanceStatus::can_be_ignored()` for runtimes, which would allow to not panic here if a status change is not "significant" (i.e., can be ignored by a typical runtime).

PR #1711, opened 2020-01-17
Remove CallContext in favour of the ExecutionContext [ECR-4075]

alekseysidorov:


runtimes/rust/src/lib.rs
```diff
@@ -296,7 +294,7 @@ pub mod _reexports {
     //! Types necessary for `ServiceDispatcher` and `ServiceFactory` derive macros to work.
 
     pub use exonum::runtime::{
-        ArtifactId, CommonError, ExecutionError, MethodId, RuntimeIdentifier,
+        ArtifactId, CallContext, CommonError, ExecutionError, MethodId, RuntimeIdentifier,
```
alekseysidorov: I think that is more convenient to export `CallContext` as `pub use exonum::runtime::CallContext` because `Service` documentation refers on the `CallContext` structure, therefore if I don't write this reexport I will lose this link.

popzxc: Why can't you just paste a link as `https://docs.rs/exonum/latest/exonum/runtime/struct.CallContext.html`?

alekseysidorov: It will be more fragile in corner cases.

exonum/src/runtime/execution_context.rs
```diff
@@ -0,0 +1,409 @@
...
+/// Provides the current state of the blockchain and the caller information for the call
+/// which is being executed.
+///
+/// The call can mean a transaction call, `before_transactions` / `after_transactions` hook,
+/// or the service constructor invocation.
+#[derive(Debug)]
+pub struct ExecutionContext<'a> {
+    /// The current state of the blockchain. It includes the new, not-yet-committed, changes to
+    /// the database made by the previous transactions already executed in this block.
+    pub fork: &'a mut Fork,
```
slowli: We have a major problem here: having `&mut Fork` publicly available to services negates all the benefits from having data separation via `BlockchainData`! I know it may be frustrating, but I think the best way to deal with this is to leave a separate `CallContext` type in the Rust runtime crate as a thin wrapper around `ExecutionContext`, copying methods that we want to expose to the Rust services.

popzxc: Ouch. In theory, we can create type-based restriction for the `Fork` access (see [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b3b27efcc2e5bf65cc82133e9d2cd38a) ), expecting the runtimes to convert context into unprivileged, but not sure if such a complication worth it.

slowli: Generic approach with a separate type is more flexible, so IMO we should go with it.

popzxc: Sure.

exonum/src/runtime/blockchain_data.rs
```diff
@@ -40,14 +40,25 @@ impl<'a, T: RawAccess + AsReadonly> BlockchainData<'a, T> {
         }
     }
 
+    /// Returns unstructured readonly access to blockchain data.
+    ///
+    /// # Safety
+    ///
+    /// Trying to get access the executing service data can lead to panic.
```
slowli:
```suggestion
    /// Trying to access data of the executing service through the returned access
    /// can lead to a panic because of borrowing checks performed by the database.
```

exonum/src/runtime/execution_context.rs
```diff
@@ -0,0 +1,416 @@
...
+        DispatcherSchema, ExecutionError, InstanceDescriptor, InstanceId, InstanceQuery,
+        InstanceSpec, InstanceStatus, MethodId, SUPERVISOR_INSTANCE_ID,
+    },
+};
+
+/// Provides the current state of the blockchain and the caller information for the call
+/// which is being executed.
+///
+/// The call can mean a transaction call, `before_transactions` / `after_transactions` hook,
+/// or the service constructor invocation.
```
slowli: Nit: Resume routine should also be mentioned here.

exonum/src/runtime/execution_context.rs
```diff
@@ -0,0 +1,416 @@
...
+            })?;
+
+        // Add a service instance to the dispatcher schema.
+        DispatcherSchema::new(&*self.fork)
+            .initiate_adding_service(spec)
+            .map_err(From::from)
+    }
+
+    /// Re-borrows an execution context with the given instance descriptor.
+    fn reborrow<'d>(&'d mut self, instance: InstanceDescriptor<'d>) -> ExecutionContext<'d> {
```
slowli: Micro-nit: Why `'d`? 🙃 The method below uses `'s` as short for `self`.

exonum/src/runtime/execution_context.rs
```diff
@@ -0,0 +1,416 @@
...
+        &mut self,
+        called_instance: impl Into<InstanceQuery<'q>>,
+        interface_name: &str,
+        method_id: MethodId,
+        arguments: &[u8],
+        fallthrough_auth: bool,
+    ) -> Result<(), ExecutionError>;
+}
+
+impl<'a> ExecutionContextUnstable for ExecutionContext<'a> {
```
slowli: Nit: Can't `'a` be elided into `ExecutionContext<'_>` here?

package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
       "dev": true,
       "requires": {
-        "color-convert": "1.9.3"
+        "color-convert": "^1.9.0"
```
slowli: Nit: Is committing changes here intentional?

runtimes/rust/src/stubs.rs
```diff
@@ -55,7 +56,7 @@ pub trait Interface<'a> {
     /// Invokes the specified method handler of the service instance.
     fn dispatch(
         &self,
-        cx: CallContext<'a>,
+        cx: ExecutionContext<'a>,
```
slowli: Nit / off-topic: `cx` can be changed to `ctx` or `context`, since it's what's used throughout the codebase.

exonum/src/runtime/execution_context.rs
```diff
@@ -0,0 +1,416 @@
...
+        let core_schema = self.data().for_core();
+        core_schema.next_height() == Height(0)
+    }
+
+    /// Returns identifier of the service interface required for the call.
+    ///
+    /// Keep in mind that this getter, in fact, is a part of an unfinished "interfaces" feature.
+    /// It will be replaced in future releases. At the moment this field is always empty for
+    /// the primary service interface.
+    pub fn interface_name(&self) -> &str {
```
popzxc: Nit: shouldn't it be `doc(hidden)` then?

alekseysidorov: Interface name is used in the sample runtime example. So it is a little bit strange to use hidden features in the sample example.

package-lock.json
alekseysidorov: Oh, `npm install` script modifies this lock file automatically without my participation. I don't know how to disable this "feature".

PR #1713, opened 2020-01-17
Fix segfault after TestKitApi dropping

alekseysidorov:
This implementation is a complete fix not a workaround

### DIFF
--- a/test-suite/testkit/src/api.rs
+++ b/test-suite/testkit/src/api.rs
@@ -16,7 +16,12 @@
 
 pub use exonum_api::ApiAccess;
 
-use actix_web::{test::TestServer, App};
+use actix::{actors::signal, Addr, System};
+use actix_net::server::Server;
+use actix_web::{
+    server::{HttpServer, IntoHttpHandler},
+    App,
+};
 use exonum::{
     blockchain::ApiSender,
     messages::{AnyTx, Verified},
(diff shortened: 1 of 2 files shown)


test-suite/testkit/src/api.rs
```diff
@@ -378,3 +386,75 @@ fn create_test_server(aggregator: ApiAggregator) -> TestServer {
...
+                .keep_alive(5)
+                .workers(1)
+                .start();
+
+            tx.send((System::current(), local_addr, srv)).unwrap();
+            sys.run();
+        });
+
+        let (system, addr, backend) = rx.recv().unwrap();
+        System::set_current(system);
```
slowli: Couldn't this call be avoided if `System` is retained within `TestServer`? IMO, setting a `System` for the test thread may have unintended side effects.

test-suite/testkit/src/api.rs
```diff
@@ -378,3 +386,75 @@ fn create_test_server(aggregator: ApiAggregator) -> TestServer {
...
+        }
+    }
+
+    /// Construct test server url.
+    fn addr(&self) -> net::SocketAddr {
+        self.addr
+    }
+}
+
+impl Drop for TestServer {
```
slowli: Nit: Is it possible to add a test that the socket address becomes free after `TestServer` is dropped?

test-suite/testkit/src/api.rs
alekseysidorov: I just copypasted this code from the `actix_web::TestServer`.

test-suite/testkit/src/api.rs
alekseysidorov: I think that it doesn't make much sense. This code is necessary only to avoid hidden issues under the `actix-web` hood.

PR #1715, opened 2020-01-17
Improve migrations [ECR-4080] [ECR-4081]

slowli:
## Overview

This PR documents migrations and add tooling to test fault-tolerance of migration scripts.

Discussion:

- Is the approach for testing fault tolerance good enough? It resulted in some changes in MerkleDB, which are quite hacky, IMO.

---

See: https://jira.bf.local/browse/ECR-4080, https://jira.bf.local/browse/ECR-4081


components/merkledb/src/migration.rs
```diff
@@ -429,40 +429,49 @@ impl fmt::Debug for MigrationHelper {
 impl MigrationHelper {
     /// Creates a new helper.
     pub fn new(db: impl Into<Arc<dyn Database>>, namespace: &str) -> Self {
-        let (this, handle) = Self::with_handle(db, namespace);
-        handle.forget();
-        this
+        assert_valid_name_component(namespace);
```
popzxc: Since this structure is public, wouldn't it be better to return `Result<Self, failure::Error>` here?

slowli: The current implementation assumes that `namespace` comes from the trusted source and was checked previously. For example, in the Exonum context it is a service name, which is validated in advance (when the service is created).

PR #1716, opened 2020-01-18
Brush up time service [ECR-4085]

popzxc:
This PR encapsulates Time service internals, and documents the service and its API.


services/time/src/lib.rs
```diff
@@ -14,10 +14,75 @@
...
+//! Time service is designed to be a trusted source of actual time for the Exonum blockchain services.
+//! With this service, user services are able to access the calendar time supplied by validator nodes
+//! to the blockchain.
+//!
+//! Basics of the approach:
+//!
+//! Each validator at a specific time sends a transaction indicating its local time
+//! (usually immediately after the commit of each block). The time service maintains an index with the
+//! most current time values indicated separately by each validator. This index is updated after each
+//! transaction from any of the validators. A 1/3 percentile of these values (ordered by decreasing time)
```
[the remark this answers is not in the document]
popzxc: Indeed. It's a copy-paste from [docs at exonum.com](https://exonum.com/doc/version/latest/advanced/time/) though.

services/time/examples/exonum_time.rs
```diff
@@ -12,6 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! This example shows how to run the Exonum node with the Time service available.
+//!
+//! Though the Time service artifact will be available within node, it should be
```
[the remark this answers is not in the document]
popzxc: Unfortunately, `cli` is unable yet to do so. Default instances for `Supervisor` and `Explorer` services are added manually in `NodeBuilder::run` and it seems that there is no public interface to do it externally.

services/time/src/lib.rs
```diff
@@ -14,10 +14,75 @@
...
+//!
+//! # HTTP API
+//!
+//! REST API of the service is documented in the [`api` module](api/index.html).
+//!
+//! # Examples
+//!
+//! ## Use with TestKit
+//!
+//! ```
```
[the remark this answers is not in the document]
popzxc: I think it better to keep this example, since it focuses on the time service and explains details. It will also look good on the main docs page.

services/time/src/lib.rs
```diff
@@ -14,10 +14,91 @@
...
+//! testkit.create_blocks_until(Height(2));
+//!
+//! // Obtain time service schema.
+//! let snapshot = testkit.snapshot();
+//! let time_schema: TimeSchema<_> = snapshot.service_schema(TIME_SERVICE_NAME).unwrap();
+//!
+//! // Obtain time from the schema. Service can base its logic on this time.
+//! let time = time_schema.time.get();
+//!
+//! // With `MockServiceProvider` we can ensure that time is based on data provided by `TimeProvider`.
```
slowli: Micro-nit: This line seems overly long (doesn't fit into layout on my laptop). Some other lines (e.g., `time_schema` instantiation) could be shortened as well.

services/time/src/api.rs
```diff
@@ -12,7 +12,60 @@
...
+//! let response: Option<DateTime<Utc>> = api
+//!     .public(ApiKind::Service(TIME_SERVICE_NAME))
+//!     .get("v1/current_time")?;
+//!
+//! // Since no blocks were created yet, time is not available.
+//! assert!(response.is_none());
+//!
+//! // Create some blocks and try again.
+//! testkit.create_blocks_until(Height(5));
+//! let response: Option<DateTime<Utc>> = api.public(ApiKind::Service(TIME_SERVICE_NAME))
```
slowli: Reminder that this may be formatted more idiomatically.

services/time/src/api.rs
```diff
@@ -12,7 +12,60 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Exonum-time API.
+//! HTTP API for the time service. All APIs are accessible from the public HTTP server
+//! of the node.
```
slowli: What about the private endpoint getting times for current validators? Without its description, the purpose of `ValidatorTime` type is not quite clear. I don't think private APIs are necessarily private *interfaces*; the split is about access / privilege level, not stability.

popzxc:
Regarding your questions:

> Would it make sense to change the transaction argument type to Timestamp, or would it hurt forward compatibility?

I think that if we'll decide that we need a different type of transaction input, we'll be able to safely add a new transaction, marking the existing one as deprecated. For now IMHO we can leave things as-is.

> Does an empty Config message in the Protobuf sources make sense? What's its purpose, if any?

Yes, sort of (at least, it was done on purpose earlier). Earlier we had an implicit agreement that empty `Config` message means "no configuration parameters".

As a result of such a convention, e.g. `exonum-launcher` [relies](https://github.com/exonum/exonum-launcher/blob/master/exonum_launcher/instances/default_spec_loader.py#L39) on it.

If we'll decide to cancel it, the changes should be done at least there.

However, I see no harm in having such a message.

services/time/src/lib.rs
popzxc: I've split comment lines that exceed 80 symbols limit into several lines, but left source lines as-is, since they don't violate 100 symbols limit (default for source code lines as per [`rustfmt` guideline](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md)) -- splitting them will make code in docs look unnaturally formatted.

PR #1717, opened 2020-01-18
Add Resume CallType [ECR-4115]

popzxc:


exonum/src/runtime/error/mod.rs
```diff
@@ -240,6 +244,7 @@ impl fmt::Display for CallType {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             CallType::Constructor => formatter.write_str("constructor"),
+            CallType::Resume => formatter.write_str("resuming routine"),
```
alekseysidorov: I think that space here can be confusing for end users.

popzxc: It's also included in string for `Method` and for hooks. I did it the similar way as there.

PR #1720, opened 2020-01-20
Implement BinaryValue and ObjectHash traits for 1..12-ary tuples

integrer:
## Overview

Created implementations for BinaryValue and ObjectHash traits for 1..12-ary tuples using macro
implementations like [here](https://github.com/rust-lang/rust/blob/66b0c97070f422cb82baaaafc79ee94cab4396c5/src/libcore/tuple.rs#L6). 

---

### Definition of Done


components/merkledb/src/values.rs
```diff
@@ -109,6 +109,45 @@ macro_rules! impl_binary_value_scalar {
             }
         }
 
+        impl_object_hash_for_binary_value! { $type }
+    };
+    ($type:tt, $write_32:ident, $read_32:ident, $len_32:expr, $write_64:ident, $read_64:ident, $type_64:tt, $len_64:expr) => {
+        #[cfg(target_pointer_width = "32")]
```
popzxc: I don't think that platform-specific encoding is a good idea for `MerkleDB` traits, since the data layout should be the same for every node in the network. With such an approach, it can cause the hash divergence on different nodes (e.g. if one node runs on a 32-bit PC, and one on a 64-bit PC, and some merkelized index calculates hash from the `BinaryValue`-encoded data).

components/merkledb/src/values.rs
```diff
@@ -262,6 +399,109 @@ impl BinaryValue for [u8; HASH_SIZE] {
     }
 }
 
+// Implementations for Tuples up to 12 except Tuple1 because it deserializes as is
+tuple_impls!{
```
popzxc: IMHO it would look better as following:

```rust
impl_binary_value_for_tuple!{1; T1};
impl_binary_value_for_tuple!{2; T1, T2};
impl_binary_value_for_tuple!{3; T1, T2, T3};
impl_binary_value_for_tuple!{4; T1, T2, T3, T4};
...

```

components/merkledb/src/values.rs
```diff
@@ -137,6 +179,101 @@ impl BinaryValue for () {
...
+{
+    fn object_hash(&self) -> Hash {
+        exonum_crypto::hash(&self.to_bytes())
+    }
+}
+
+/// Converts vector of byte vectors into flat byte vector. Also keeps their
+/// sizes as converted usize values and writes them before the vectors
+/// themselves to the output vector to be able to convert it back
+fn chunks_to_bytes(chunks: &Vec<Vec<u8>>) -> Vec<u8> {
```
slowli: 1. `chunks: &[Vec<u8>]` would be more idiomatic.
2. I think the implementation using [`Vec::extend_from_slice`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend_from_slice) (and computing the needed capacity of the output vector) would be significantly faster since it would use memory copying primitives.

components/merkledb/src/values.rs
integrer: Thank you for your point of view.

I agree with you. I think it might be better to store platform-specific values as a 32-bit integer. But in this case, we need to solve the 32-bit variable overflow problem. This can happen if the buffer contains a value greater than 4 GB. But I think this is such an unlikely case to store huge unit values in the blockchain so that we can do this.

In addition, buffers > 2^32 bytes are not supported by 32-bit systems.

popzxc: We can store platform-specific values in `u64`, which will fit both 32-bit and 64-bit pointers :)

components/merkledb/src/values.rs
integrer: Thank you!

I think I will do something like.

components/merkledb/src/values.rs
integrer: But in this case we have to store useless zeroes in rest 4 bytes for each size value.

popzxc: I think that it's not a big problem: 
1. There are not so many 32-bit machines nowadays.
2. It's not a pretty common scenario to use `usize` / `isize` as a value to store. If one will be totally sure that value will never exceed 32-bit limit, they will be able to use `u32` / `i32` instead.

integrer:
I created implementation according yours comments in master branch.

aleksuss:
@integrer can i close this PR or you plan to commit some changes ?

integrer:
Yes, I no longer plan changes in this PR.

PR #1723, opened 2020-01-21
Move rust runtime API tests outside of testkit [ECR-4091]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-4091

### Definition of Done


runtimes/rust/Cargo.toml
```diff
@@ -25,11 +25,13 @@ exonum-proto = { version = "0.13.0-rc.2", path = "../../components/proto" }
 failure = "0.1.5"
 futures = "0.1.25"
 log = "0.4.8"
-protobuf = "2.8.0"
+protobuf = "2.10.1"
 serde = "1.0.101"
 serde_derive = "1.0.101"
 
 [dev-dependencies]
+exonum-testkit = { version = "0.13.0-rc.2", path = "../../test-suite/testkit" }
```
popzxc: `exonum-testkit` depends on `exonum-rust-time`, and `exonum-rust-runtime` has `exonum-testkit` as a dev dependency.

Won't it cause problems with uploading to `crates.io`?

aleksuss: Good point. I ran into a similar case some time ago and I used this [workaround](https://github.com/rust-lang/cargo/issues/4242#issuecomment-413203081).

popzxc: I see.

slowli: Wouldn't a better solution be creating a separate non-publishable crate for testing (like `consensus-tests` and `node-tests` before)? As far as I see, the tested functionality is public, so it is technically possible.

aleksuss: Hopefully it would.

test-suite/rust-runtime-api/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+    use assert_matches::assert_matches;
+    use exonum_rust_runtime::{api::Error as ApiError, ProtoSourceFile, ProtoSourcesQuery};
+    use pretty_assertions::assert_eq;
+
+    use std::collections::HashSet;
+
+    use crate::service::TestRuntimeApiService;
+    use exonum_testkit::{ApiKind, TestKit, TestKitApi, TestKitBuilder};
+
+    fn testkit_with_rust_service() -> (TestKit, TestKitApi) {
```
slowli: This function should probably be moved outside the test code and made public so that the compiler doesn't complain about dead code.

test-suite/rust-runtime-api/build.rs
```diff
@@ -14,9 +14,10 @@
 
 use exonum_build::ProtobufGenerator;
 
+// API Tests.
```
popzxc: Nit: This comment is misleading, it's not "API tests", it's a main function of a build script :slightly_smiling_face:

test-suite/rust-runtime-api/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod proto;
+mod service;
+
+#[cfg(test)]
```
popzxc: Do we need `cfg(test)` in a pure test crate?

test-suite/rust-runtime-api/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    use std::collections::HashSet;
+
+    use crate::service::TestRuntimeApiService;
+    use exonum_testkit::{ApiKind, TestKit, TestKitApi, TestKitBuilder};
+
+    fn testkit_with_rust_service() -> (TestKit, TestKitApi) {
+        let mut testkit = TestKitBuilder::validator()
+            .with_logger()
+            .with_validators(1)
```
popzxc: Isn't `with_validators(1)` default?

test-suite/rust-runtime-api/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
```
popzxc: Shouldn't we specify [`publish = false`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish--field-optional)?

test-suite/rust-runtime-api/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+name = "exonum-rust-runtime-api-tests"
+version = "0.0.0"
+edition = "2018"
+authors = ["The Exonum Team <exonum@bitfury.com>"]
+homepage = "https://exonum.com/"
+repository = "https://github.com/exonum/exonum"
+readme = "README.md"
+license = "Apache-2.0"
+keywords = ["exonum", "runtime", "api", "tests"]
```
popzxc: I think `crates.io`-related fields (and `badges` section) should be removed.

test-suite/rust-runtime-api/src/lib.rs
aleksuss: It just copy-pasted.

popzxc: My beloved quote:

>The Boy Scouts of America have a simple rule that we can apply to our profession.
> _**Leave the campground cleaner than you found it.**_

aleksuss: The words of wisdom ! 👍🏻

test-suite/rust-runtime-api/src/lib.rs
```diff
@@ -100,21 +99,34 @@ fn service_protos_with_service() {
 /// Rust-runtime API should return error in case of an incorrect artifact.
 #[test]
 fn service_protos_with_incorrect_service() {
+    use exonum_rust_runtime::_reexports::{ArtifactId, RuntimeIdentifier};
```
slowli: Since `_reexports` module is private, I'd suggest to use exports from the `exonum` crate instead. I don't think adding `exonum` as a direct dependency is a big deal in this case.

components/api/src/error.rs
```diff
@@ -24,7 +24,7 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 
 /// API HTTP error struct.
-#[derive(Fail, Debug, Default)]
+#[derive(Fail, Debug, Default, Eq, PartialEq)]
```
slowli: I'd suggest to **not** implement `PartialEq` / `Eq` for types in which equality may become a problem in the future (i.e., it's not obvious that type instances should be comparable). I've described the reasoning in #1696 - the implementation may be difficult to maintain if new fields are added to the error.

popzxc: I'm wondering when will we be able to use negative traits on custom traits...

```rust
impl !PartifalEq for OurType {}
```

PR #1726, opened 2020-01-22
Clarify a description of the time field of the CommittedTransaction [ECR-3911]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-3911

### Definition of Done

### DIFF
--- a/components/explorer/src/lib.rs
+++ b/components/explorer/src/lib.rs
@@ -373,6 +373,7 @@ impl<'a> IntoIterator for &'a BlockWithTransactions {
 /// | `location_proof` | [`ListProof`]`<`[`Hash`]`>` | Proof of transaction inclusion into a block |
 /// | `status` | (custom; see below) | Execution status |
 /// | `time` | [`DateTime`]`<`[`Utc`]`>` | Commitment time* |
+///
 /// \* By commitment time we mean an approximate commitment time of the block
 /// which includes the transaction. This time is a median time of the precommit local times
 /// of each validator.
--- a/components/explorer/src/lib.rs
+++ b/components/explorer/src/lib.rs
@@ -372,6 +372,9 @@ impl<'a> IntoIterator for &'a BlockWithTransactions {
 /// | `location` | [`TxLocation`] | Location of the transaction in the block |
 /// | `location_proof` | [`ListProof`]`<`[`Hash`]`>` | Proof of transaction inclusion into a block |
 /// | `status` | (custom; see below) | Execution status |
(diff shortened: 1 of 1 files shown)


components/explorer/src/lib.rs
```diff
@@ -372,6 +372,9 @@ impl<'a> IntoIterator for &'a BlockWithTransactions {
 /// | `location` | [`TxLocation`] | Location of the transaction in the block |
 /// | `location_proof` | [`ListProof`]`<`[`Hash`]`>` | Proof of transaction inclusion into a block |
 /// | `status` | (custom; see below) | Execution status |
+/// | `time` | [`DateTime`]`<`[`Utc`]`>` | An approximate commit time of the block which includes
```
popzxc: I'd suggest to use somewhat short description in the cell with a reference, which is explained below, like:

```markdown
| `time` | [`DateTime`]`<`[`Utc`]`>` | Block commitment time*

\* By commitment time we mean approximate commitment time for the block
which includes the transaction. This time is calculated as the arithmetic average
of the precommit local times of each validator.
```

IMHO with such an approach, the table will look nicer.

aleksuss: Yeah. Looks nice. But question is How will it be looking on the docs.rs ? 🤔 Have you seen such approach before ?

popzxc: What do you mean? It will look just like in the markdown, e.g. 

image

components/explorer/src/lib.rs
```diff
@@ -372,6 +372,10 @@ impl<'a> IntoIterator for &'a BlockWithTransactions {
 /// | `location` | [`TxLocation`] | Location of the transaction in the block |
 /// | `location_proof` | [`ListProof`]`<`[`Hash`]`>` | Proof of transaction inclusion into a block |
 /// | `status` | (custom; see below) | Execution status |
+/// | `time` | [`DateTime`]`<`[`Utc`]`>` | Commitment time* |
```
popzxc: I think it's better to have an empty line between table and text.

```suggestion
/// | `time` | [`DateTime`]`<`[`Utc`]`>` | Commitment time* |
///
```

components/explorer/src/lib.rs
```diff
@@ -372,6 +372,10 @@ impl<'a> IntoIterator for &'a BlockWithTransactions {
...
+/// \* By commitment time we mean an approximate commitment time of the block
+/// which includes the transaction. This time is calculated as the arithmetic average
```
dmitry-timofeev: It used to be the median time — has that changed?

aleksuss: Yeah. Good catch. It's still median time 😄

PR #1727, opened 2020-01-22
Support migrations in supervisor [ECR-4123]

popzxc:
This PR introduces the support of the migration workflow to the supervisor.

Overall, new interfaces are pretty similar to the ones for deployment process.

The main difference is that for deploy we provide a callback to report a deployment result, and for migrations we have to poll the migration state manually instead.


services/supervisor/src/lib.rs
```diff
@@ -55,6 +55,41 @@
 //! The operation of starting or resuming a service is treated similarly to a configuration change
 //! and follows the same rules.
 //!
+//! ## Migrations management
+//!
+//! Supervisor service provides a functionality to perform data migrations for services.
+//! Request for migration is sent through private REST API and contains the name of instance
+//! to migrate, end artifact version to achieve after migration, and deadline height until which
```
alekseysidorov: I think that it is a quite difficult to predict the number of blocks needed for successful migration.
In my opinion network administrators should revert broken migration by hand.

services/supervisor/src/migration_state.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub fn is_failed(&self) -> bool {
+        self.inner.is_failed()
+    }
+
+    pub fn update(&mut self, new_state: AsyncEventState) {
+        self.inner = new_state;
+    }
+}
+
+mod pb_expected_state_hash {
```
alekseysidorov: Thoughts: This approach may be usable e.g. for `prevoius_hash` of `BlockHeader`.

services/supervisor/src/lib.rs
popzxc: I thought about it, but OTOH network administrators won't be able to distinguish a long-running migration from a migration that went into an infinite loop, they both will look as `Pending`.

IMHO, having the average block commitment duration, it's not that hard to calculate the approximate amount of blocks which you can afford, and then round it up. For example, if in average every block is committed in 1 second, and we don't want the migration to run for longer than 2 hours, we should set the deadline height to (current height + 7200). Taking the time between obtaining the current height and sending the request for migration, we can round it up to 7500.

So, as a conclusion, I think it's better to have such an instrument, than to not have it.

One possible improvement here may be making `deadline_height` optional (meaning that if `deadline_height` is 0, migration will never be killed due to timeout), but it'll introduce more rules to the already not very easy interaction scheme.

services/supervisor/src/proto/schema.proto
```diff
@@ -33,21 +33,28 @@ message ErrorInfo {
...
     // Set only for `type == FAIL`.
     ErrorInfo error = 2;
 }
+
+// Migration state.
+message MigrationState {
+    AsyncEventState inner = 1;
+    uint64 request_origin_height = 2;
+    exonum.crypto.Hash expected_state_hash = 3;
+}
```
aleksuss: Nit: `\n` ?

popzxc: Yeah, current state of PR has a plenty of raw moments. Don't take them into account yet, I opened PR at this state for design discussion only :slightly_smiling_face:

services/supervisor/src/transactions.rs
```diff
@@ -531,11 +668,115 @@ impl Supervisor {
...
+    fn fail_migration(
+        &self,
+        mut context: CallContext<'_>,
+        request: MigrationRequest,
+        error: ExecutionError,
+        initiate_rollback: bool,
+    ) -> Result<(), ExecutionError> {
+        log::warn!(
+            "Migration for a request {:?} failed. Error: '{}'. \
+             This migration is going to be rolled back.",
```
slowli: Nit: This should be only output if `initiate_rollback` is `true`.

services/supervisor/src/transactions.rs
```diff
@@ -531,11 +668,115 @@ impl Supervisor {
...
+                // Hashes do not match, rollback the migration.
+                drop(schema); // Required for the context reborrow.
+                let initiate_rollback = true;
+                return self.fail_migration(context, request, error, initiate_rollback);
+            }
+        }
+
+        // Check if we have enough confirmations to confirm the migration.
+        let confirmations = schema.migration_confirmations.confirm(&request, author);
+        let validator_count = core_schema.consensus_config().validator_keys.len();
```
slowli: Can `consensus_config` change during the migration? If so, is it possible to rewrite this check to be sound?

services/supervisor/src/transactions.rs
```diff
@@ -531,11 +668,115 @@ impl Supervisor {
...
+            .migration_states
+            .get(&request)
+            .expect("BUG: Attempt to confirm a migration which does not have a stored state");
+
+        // Verify that state hash does match expected one.
+        match state.add_state_hash(state_hash) {
+            Ok(()) => {
+                // Hash is OK, process further.
+            }
+            Err(error) => {
```
slowli: Nit: Can be changed to `if let Err(error) = ...`

services/supervisor/src/lib.rs
```diff
@@ -55,6 +55,41 @@
...
+//! migration should be completed.
+//!
+//! The following pre-requirements should be satisfied in order to start a migration:
+//!
+//! - Target service instance should exist and be stopped.
+//! - End artifact for a migration should be a superior version of the artifact of target instance.
+//! - New (end) version of artifact should be deployed
+//!   (if migration will contain several steps, all the intermediate artifacts should
```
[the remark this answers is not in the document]
popzxc: Ah, OK. I didn't know that.

services/supervisor/src/transactions.rs
```diff
@@ -531,11 +668,115 @@ impl Supervisor {
...
+
+            // Migration is not pending anymore, remove it.
+            schema.pending_migrations.remove(&request);
+
+            drop(schema);
+
+            // Commit and flush the migration.
+            let mut supervisor_extensions = context.supervisor_extensions();
+            supervisor_extensions.commit_migration(request.service.as_ref(), state_hash)?;
+            supervisor_extensions.flush_migration(request.service.as_ref())?;
```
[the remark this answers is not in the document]
popzxc: Yeah, such an explanation will be really helpful. I've spent some time trying to understand why is this need to be 2 different operations, and didn't succeed.

I'll update the implementation.

services/supervisor/src/transactions.rs
popzxc: We can actualize the confirmations (as a set intersection of stored votes and current validator keys) every time we receive a confirmation, right before this check. Will this be enough?

slowli: I think yes. The other option is to somehow restrict validator changes when there are pending migrations, but this is not quite good from the UX standpoint.

services/supervisor/src/transactions.rs
```diff
@@ -413,6 +447,150 @@ impl SupervisorInterface<ExecutionContext<'_>> for Supervisor {
...
+
+    fn report_migration_result(
+        &self,
+        context: ExecutionContext<'_>,
+        result: MigrationResult,
+    ) -> Self::Output {
+        // Verifies that transaction author is validator.
+        let author = get_validator(&context)?;
+
+        result
```
slowli: Nit: Not sure I understand why this check is necessary here. Logically, the artifact in the `result.request` was checked when a migration was initiated (which is checked later).

services/supervisor/src/transactions.rs
```diff
@@ -413,6 +447,150 @@ impl SupervisorInterface<ExecutionContext<'_>> for Supervisor {
...
+            .migration_states
+            .get(&result.request)
+            .ok_or(MigrationError::MigrationRequestNotRegistered)?;
+
+        // Check if migration already failed.
+        if schema
+            .migration_states
+            .get(&result.request)
+            .map(|state| state.is_failed())
+            .unwrap_or_default()
```
slowli: Nit: IMO, `unwrap_or(false)` would be more readable. Also, isn't this whole thing redundant? You can bind a migration state from the previous check to a local variable (say, `state`) and simply check `if state.is_failed()` here.

services/supervisor/src/transactions.rs
```diff
@@ -413,6 +447,150 @@ impl SupervisorInterface<ExecutionContext<'_>> for Supervisor {
...
+        }
+
+        // Verify that we didn't reach deadline height.
+        if result.request.deadline_height < current_height {
+            return Err(SupervisorCommonError::DeadlineExceeded.into());
+        }
+
+        drop(schema);
+
+        match result.result.0 {
```
slowli: Nit: `result.result` duplication. As a suggestion, could it make sense to name the field `status` (given its type - `MigrationStatus`)?

services/supervisor/src/transactions.rs
```diff
@@ -534,11 +718,130 @@ impl Supervisor {
...
+
+        let mut state = schema
+            .migration_states
+            .get(&request)
+            .expect("BUG: Attempt to confirm a migration which does not have a stored state");
+
+        // Verify that state hash does match expected one.
+        if let Err(error) = state.add_state_hash(state_hash) {
+            // Hashes do not match, rollback the migration.
+            drop(schema); // Required for the context reborrow.
```
slowli: Nit: Is this `drop` still required, or is it a refactoring artifact? As far as I see, the context is not reborrowed within *this* method.

services/supervisor/src/api.rs
```diff
@@ -251,15 +266,12 @@ impl PrivateApi for ApiImpl<'_> {
     fn migration_status(
         &self,
         query: MigrationInfoQuery,
-    ) -> Result<ProcessStateResponse, Self::Error> {
+    ) -> Result<MigrationStateResponse, Self::Error> {
```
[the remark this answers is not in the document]
popzxc: Do you think so? Usually I poll the exonum API with `curl`, so when I see 404, I start looking for the mistake in URI. IMHO, it will cause more confusion (at least for me personally).

services/supervisor/tests/supervisor/migrations/migration_service.rs
```diff
@@ -113,6 +113,7 @@ pub(super) mod v01 {
         }
     }
 
+    #[allow(clippy::borrowed_box)] // We can't just convert `&Box<dyn Snapshot>` to `&dyn Snapshot`.
```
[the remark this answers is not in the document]
popzxc: I tried but encountered some issues with it. I'll try to investigate it once again I'll resolve the primary comments.

services/supervisor/src/lib.rs
```diff
@@ -69,18 +129,23 @@
...
-        DeployResult, ResumeService, ServiceConfig, StartService, StopService, SupervisorConfig,
+        DeployResult, MigrationRequest, MigrationResult, ResumeService, ServiceConfig,
+        StartService, StopService, SupervisorConfig,
     },
     schema::Schema,
     transactions::SupervisorInterface,
 };
 
+#[doc(hidden)] // Public for migration tests.
+pub use self::schema::SchemaImpl;
```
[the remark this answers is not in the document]
popzxc: Maybe, but not in this PR I think: it's too big already.

services/supervisor/src/lib.rs
```diff
@@ -432,8 +525,148 @@ impl Service for Supervisor {
...
+    /// This has to be done in the block other than one in which migration was committed,
+    /// so this method is invoked in `before_transactions` of the next block.
+    fn flush_completed_migrations(
+        &self,
+        context: &mut ExecutionContext<'_>,
+    ) -> Result<(), ExecutionError> {
+        let schema = SchemaImpl::new(context.service_data());
+
+        // Collect pending migration requests which are successfully completed.
+        let finished_migrations = schema.migrations_to_flush.iter().collect::<Vec<_>>();
```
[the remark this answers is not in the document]
popzxc: I was surprised as well.

services/supervisor/src/lib.rs
```diff
@@ -432,8 +525,148 @@ impl Service for Supervisor {
...
+                        .confirmed_by(&request, &service_key);
+
+                    // We are interested in requests that are both pending and not yet confirmed.
+                    // Despite the fact that `migration_confirmations` stores only successful
+                    // outcomes, receiving any failure report will immediately change the state
+                    // to `Failed`.
+                    // Thus if request is `pending` and there is no out signature in
+                    // `migration_confirmations` index, it means than we did not send the
+                    // result report, and should do it once core will provide this result.
+                    if state.state.is_pending() && !confirmed_by_us {
```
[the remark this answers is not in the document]
popzxc: Actually, it was `inner` in the previous iteration, but then I've renamed it because `inner` won't look adequate in the serialized form. I can just use `serde(rename = "state")` though.

services/supervisor/src/transactions.rs
popzxc: It is not required indeed, but it allows us to spawn a more appropriate error IMHO.

services/supervisor/src/api.rs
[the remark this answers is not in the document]
slowli: AFAIK, several APIs in other services work like that (e.g., the wallet endpoint in the cryptocurrency demos). One can argue that requesting a non-existing migration *is* wrong, since it should not happen with properly designed web front-end.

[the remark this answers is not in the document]
popzxc: Indeed. Well, we now have a pretty good API to build an error description, so it's a better alternative.

services/supervisor/src/transactions.rs
popzxc: `self.fail_migration(context, request, error, initiate_rollback)` fails without `drop`.

services/supervisor/tests/supervisor/migrations/migration_service.rs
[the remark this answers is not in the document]
popzxc: It worked. Dunno why it didn't work the first time...

services/supervisor/src/lib.rs
```diff
@@ -432,8 +525,148 @@ impl Service for Supervisor {
...
+            if state.state.is_pending() {
+                // If state is marked as pending, change it to failed as well.
+                state.fail(AsyncEventState::Timeout);
+                schema.migration_states.put(&request, state);
+
+                // Then, rollback the migration.
+                drop(schema);
+                context
+                    .supervisor_extensions()
+                    .rollback_migration(request.service.as_ref())?;
```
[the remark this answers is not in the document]
popzxc: I think this a minor improvement and can be done separately, if we'll decide that it worth it.

services/supervisor/tests/supervisor/migrations/migration_service.rs
```diff
@@ -0,0 +1,410 @@
...
+            InitMigrationError, LinearMigrations, MigrateData, MigrationContext, MigrationError,
+            MigrationScript,
+        },
+        versioning::Version,
+        ExecutionContext, ExecutionError, InstanceId,
+    },
+};
+use exonum_derive::*;
+use exonum_rust_runtime::{DefaultInstance, Service, ServiceFactory};
+// use rand::{seq::SliceRandom, thread_rng, Rng};
```
slowli: Nit: Commented code.

services/supervisor/tests/supervisor/migrations/migration_service.rs
```diff
@@ -0,0 +1,410 @@
...
+            balance: wallet.balance,
+        };
+        new_schema.wallets.put(&key, new_wallet);
+    }
+    Ok(())
+}
+
+/// Third migration script. Always fails.
+fn failing_migration(_ctx: &mut MigrationContext) -> Result<(), MigrationError> {
+    Err(MigrationError::Custom("This migration always fails".into()))
```
slowli: Nit: There is `MigrationError::new` method, which is more idiomatic in this case.

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+use crate::service_lifecycle::execute_transaction;
+
+use migration_service::{
+    FailingMigrationServiceV07, MigrationService, MigrationServiceV01_1, MigrationServiceV02,
+    MigrationServiceV05, MigrationServiceV05_1,
+};
+
+mod migration_service;
+
+/// Creates testkit with supervisor and three versions of migrating service.
```
slowli: Nit: There are more than three versions of the service now, aren't there? I'd avoid using specific numbers for future compatibility 🙃

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+    )
+    .expect("Stop service transaction should be processed");
+}
+
+fn obtain_expected_hash(testkit: &mut TestKit, request: &MigrationRequest) -> Hash {
+    for _ in 0..5 {
+        let snapshot = testkit.snapshot();
+        let prefixed = Prefixed::new(Supervisor::NAME, &snapshot);
+        let schema = SchemaImpl::new(prefixed);
+        let state = schema.migration_state_unchecked(&request);
```
slowli: Nit: Unnecessary reference; `request` is a reference already.

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+fn wait_for_migration_success(
+    testkit: &mut TestKit,
+    deadline_height: Height,
+    request: MigrationRequest,
+    version: Version,
+) {
+    let mut success = false;
+    while testkit.height() <= deadline_height.next() {
+        testkit.create_block();
+        let api = testkit.api();
```
slowli: Nit: Does `testkit.api()` needs to be called on each block? AFAIK, this call involves creating a new web server, so we should probably try to minimize the amount of invocations.

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+            // Migration is executed in the separate thread, so sleep a bit.
+            thread::sleep(Duration::from_millis(50));
+            // Then create a new block.
+            testkit.create_block();
+        }
+    }
+    panic!("Node didn't calculate the expected hash")
+}
+
+fn wait_for_migration_success(
```
slowli: Nit: It could make sense to extract common code for `wait_for_migration_success`, `wait_for_migration_fail` and `wait_for_migration_timeout`, which would wait until the migration request status differs from `Pending`.

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+        deadline_height,
+    };
+
+    let api = testkit.api();
+    let tx_hash = request_migration(&api, request.clone());
+    let block = testkit.create_block();
+
+    block[tx_hash]
+        .status()
+        .expect("Transaction should be executed successfully");
```
slowli: The code starting from `let api = ...` can be encapsulated into a function; it seems to be used often.

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+    );
+
+    let snapshot = testkit.snapshot();
+    let prefixed = Prefixed::new(MigrationService::INSTANCE_NAME, snapshot.as_ref());
+
+    migration_service::v02::verify_schema(prefixed);
+
+    // Request the same migration.
+    let deadline_height = Height(DEADLINE_HEIGHT.0 * 2);
+    let request = MigrationRequest {
```
slowli: Nit: To better illustrate the point, you could modify the deadline height in the previous `request`.

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+    let tx_hash = request_migration(&api, request.clone());
+    let block = testkit.create_block();
+
+    // Despite the fact that migration should fail, the transaction with request
+    // should be executed successfully.
+    block[tx_hash]
+        .status()
+        .expect("Transaction should be executed successfully");
+
+    let error = wait_for_migration_fail(&mut testkit, deadline_height, request);
```
slowli: Shouldn't the migration marked as failed right away? I think it could be beneficial to check that it is indeed so.

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,879 @@
...
+        request,
+        Version::new(0, 5, 1),
+    );
+
+    let snapshot = testkit.snapshot();
+    let prefixed = Prefixed::new(MigrationService::INSTANCE_NAME, snapshot.as_ref());
+
+    // Data should not change.
+    migration_service::v05::verify_schema(prefixed);
+}
```
slowli: Possible future work: It could make sense to test:

- Recovery after a failed migration (i.e., a migration fails, we restart it, it succeeds)
- Resuming a service after successful migration

exonum/src/helpers/mod.rs
```diff
@@ -60,3 +60,42 @@ pub trait ValidateInput: Sized {
 pub fn byzantine_quorum(total: usize) -> usize {
     total * 2 / 3 + 1
 }
+
+/// Module for serializing `Option<Hash>` to Protobuf.
```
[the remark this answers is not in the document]
slowli: Crap, the option should most probably be `ignore` instead of `no_run` (`no_run` compiles the code snippet, which requires the Protobuf source structure to exist).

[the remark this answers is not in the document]
popzxc: No problem, fixed.

services/supervisor/src/multisig.rs
```diff
@@ -81,6 +81,44 @@ where
...
         len
     }
+
+    /// Updates the stored confirmations to be an intersection of the set
+    /// with current confirmations, and set of the actual validator keys.
+    ///
+    /// This method is intended to be called before comparing the amount of
+    /// confirmations and amount of validators, so confirmations of nodes which
+    /// are not validators anymore won't be taken into account.
+    fn intersect(&mut self, id: &V, validator_keys: BTreeSet<PublicKey>) {
```
slowli: Nit: Can't a reference to a set be used here?

services/supervisor/src/multisig.rs
```diff
@@ -81,6 +81,44 @@ where
...
+        let mut confirmations = self.index.get(id).unwrap_or_default();
+
+        confirmations.0 = confirmations
+            .0
+            .intersection(&validator_keys)
+            .cloned()
```
slowli: Nit: Can't `copied` be used here since `PublicKey` implements `Copy`?

services/supervisor/src/multisig.rs
```diff
@@ -81,6 +81,44 @@ where
...
+        &mut self,
+        id: &V,
+        validator_keys: impl IntoIterator<Item = PublicKey>,
+    ) -> bool {
+        let validator_keys: BTreeSet<PublicKey> = validator_keys.into_iter().collect();
+        let validators_amount = validator_keys.len();
+
+        self.intersect(id, validator_keys);
+
+        self.confirmations(id) == validators_amount
```
slowli: Nit: If the length was returned from `intersect`, the additional access to DB could be avoided.

services/supervisor/src/transactions.rs
```diff
@@ -534,11 +705,124 @@ impl Supervisor {
...
+    /// and `fail_migration` method is invoked.
+    fn confirm_migration(
+        &self,
+        mut context: ExecutionContext<'_>,
+        request: MigrationRequest,
+        state_hash: Hash,
+        author: PublicKey,
+    ) -> Result<(), ExecutionError> {
+        let core_schema = context.data().for_core();
+
```
slowli: Micro-nit: having a newline every other line is a bit excessive, IMO.

services/supervisor/src/migration_state.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[protobuf_convert(with = "exonum::helpers::pb_version")]
+    pub version: Version,
+
+    /// Expected state hash. Equals to the first obtained local migration state hash.
+    /// For a good scenario, all the hashes should be equal between each other.
+    /// For the bad scenario, at least one node obtains the different hash and that's enough
+    /// to consider migration failed.
+    #[protobuf_convert(with = "exonum::helpers::pb_optional_hash")]
+    #[serde(skip)]
+    pub(crate) expected_state_hash: Option<Hash>,
```
slowli: Nit: maybe, call this field `reference_state_hash`? IMO, `expected` implies that this hash is *always* correct, which isn't the case.

services/supervisor/tests/supervisor/migrations/migration_service.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+impl<T: Access> Schema<T> {
+    pub fn new(access: T) -> Self {
+        Self::from_root(access).unwrap()
+    }
+}
+
+/// Initial service schema.
+pub(super) mod v01 {
```
slowli: Nit: IMO, `pub(super)`  qualifiers can be elided to `pub` given that it's test code. Also, since `ENTRY_VALUE` constants are seemingly used only in this module, you could use `pub(super)` for them.

services/supervisor/tests/supervisor/migrations/mod.rs
```diff
@@ -0,0 +1,784 @@
...
+    let builder = builder
+        .with_migrating_rust_service(MigrationServiceV05_1)
+        .with_artifact(MigrationServiceV05_1.artifact_id());
+
+    builder.create()
+}
+
+/// Same as `testkit_with_supervisor_and_service`, but services do not support migrations.
+fn testkit_with_supervisor_and_service_no_migrations(validator_count: u16) -> TestKit {
+    // Initialize builder;
```
slowli: Nit: Semicolon should be replaced with a period.

PR #1728, opened 2020-01-22
Add maintenance command to restart migration script [ECR-4095]

Kormick:


cli/src/lib.rs
```diff
@@ -47,6 +47,8 @@
 //!
 //! * `run-dev` command automatically generates network configuration with a single node and runs
 //! it. This command can be useful for fast testing of the services during development process.
+//! * `maintenance` command allows to clear node's consensus messages with `clear-cache`, and
```
popzxc: This entry conflicts with the next one.

exonum/src/runtime/dispatcher/helpers.rs
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+//! Helpers for a Dispatcher.
+
+use exonum_merkledb::Fork;
+
+use crate::runtime::dispatcher::Schema;
+
+/// Removes local nmigration result for specified service.
+pub fn remove_local_migration_result(fork: &Fork, service_name: &str) {
```
popzxc: I think it should be hidden under the `doc(hidden)`.

cli/src/command/maintenance.rs
```diff
@@ -70,6 +95,11 @@ impl ExonumCommand for Maintenance {
             Action::ClearCache => {
                 Action::clear_cache(self.node_config.clone(), self.db_path.clone())?
             }
+            Action::RestartMigration { ref service_name } => Action::restart_migration(
+                self.node_config.clone(),
+                self.db_path.clone(),
```
aleksuss: You can get rid of these clones if you'll change signature of the `restart_migration` function.

exonum/src/runtime/dispatcher/migration_tests.rs
```diff
@@ -173,6 +177,18 @@ fn complex_migration_part2(ctx: &mut MigrationContext) -> Result<(), MigrationEr
     Ok(())
 }
 
+fn not_good_then_ok(_ctx: &mut MigrationContext) -> Result<(), MigrationError> {
+    static SUCCESS: AtomicBool = AtomicBool::new(false);
```
slowli: Nit: `static`s and atomics could be avoided if `MigrationRuntime` would have a flag denoting whether to return a "good" script or a "bad" one for `"not-good-then-ok"` migration.

exonum/src/runtime/dispatcher/migration_tests.rs
```diff
@@ -203,6 +219,23 @@ impl Rig {
         }
     }
 
+    fn new_with_db(db: Arc<TemporaryDB>) -> Self {
```
slowli: Nit: `Rig::new()` can call `Self::new_with_db` in order to avoid code duplication. Also, `with_db` would be a more idiomatic name, IMO.

exonum/src/runtime/dispatcher/migration_tests.rs
```diff
@@ -964,25 +1000,108 @@ fn test_migration_commit_with_local_error(local_result: LocalResult) {
...
+        "0.5.2".parse().unwrap(),
+    );
+    let expected_status = InstanceStatus::migrating(InstanceMigration::from_raw_parts(
+        artifact,
+        Version::new(0, 5, 0),
+        Some(HashTag::empty_map_hash()),
+    ));
+    assert_eq!(state.status, Some(expected_status));
+
+    // Flush the migration.
```
slowli: Nit: This and the following lines do not test migration restart *per se* and thus could be removed.

exonum/src/runtime/dispatcher/schema.rs
```diff
@@ -135,7 +135,7 @@ impl<T: Access> Schema<T> {
         self.access.clone().get_map(PENDING_INSTANCES)
     }
 
-    fn local_migration_results(&self) -> MapIndex<T::Base, str, MigrationStatus> {
+    pub(crate) fn local_migration_results(&self) -> MapIndex<T::Base, str, MigrationStatus> {
```
slowli: Nit / subjective: IMO, `remove_local_migration_result` can be placed in this module; thus, creating a separate module for a single method would be avoided 🙃

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
```
slowli: Please reference the CLI command in `panic` messages in the `Dispatcher::block_on_migration` method; there's a FIXME there that should be removed in this PR.

CHANGELOG.md
```diff
@@ -113,6 +113,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `supervisor-mode` parameter has been added for `generate-template` subcommand.
   (#1598)
 
+- Added maintenance command `restart-migration` to restart migration script. (#1728)
```
slowli: This isn't a breaking change, is it? It should be placed into "New features" section.

cli/src/command/maintenance.rs
```diff
@@ -48,10 +49,16 @@ pub enum Action {
     /// Clear consensus messages cache.
     #[structopt(name = "clear-cache")]
     ClearCache,
+    /// Restart migration script.
+    #[structopt(name = "restart-migration")]
+    RestartMigration {
+        /// Name of the service to restart.
```
vitvakatu: nit: is it string representation of instance identifier? An example would be helpful for users, for sure.

exonum/src/runtime/dispatcher/migration_tests.rs
```diff
@@ -173,6 +177,18 @@ fn complex_migration_part2(ctx: &mut MigrationContext) -> Result<(), MigrationEr
     Ok(())
 }
 
+fn not_good_then_ok(_ctx: &mut MigrationContext) -> Result<(), MigrationError> {
+    static SUCCESS: AtomicBool = AtomicBool::new(false);
+
+    thread::sleep(DELAY);
+    if !SUCCESS.load(Ordering::SeqCst) {
+        SUCCESS.store(true, Ordering::SeqCst);
```
aleksuss: This can be replaces by `compare_and_swap` like this:
```rust 
if !SUCCESS.compare_and_swap(false, true, Ordering::SeqCst) {
```

cli/src/command/maintenance.rs
```diff
@@ -48,10 +49,16 @@ pub enum Action {
     /// Clear consensus messages cache.
     #[structopt(name = "clear-cache")]
     ClearCache,
+    /// Restart migration script.
+    #[structopt(name = "restart-migration")]
+    RestartMigration {
+        /// Name of the service to restart, e.g. explorer, time-oracle, my-service etc.
```
popzxc:
```suggestion
        /// Name of the service for migration restart, e.g. "explorer" or "my-service".
```

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -736,7 +736,8 @@ impl Dispatcher {
             // FIXME: Add a maintenance command for removing local migration result (ECR-4095).
             panic!(
                 "Migration for service `{}` is committed with migration hash {:?}, \
-                 but locally it has finished with an error: {}",
+                 but locally it has finished with an error: {}. You can remove local \
+                 migration result with CLI maintenance command `restart-migration`.",
```
slowli: Please add the same message to the `assert!(local_hash == global_hash)` below.

cli/src/command/maintenance.rs
```diff
@@ -62,13 +69,32 @@ impl Action {
...
+        service_name: &str,
+    ) -> Result<(), Error> {
+        let node_config: NodeConfig = load_config_file(node_config)?;
+        let db: Box<dyn Database> = Box::new(RocksDB::open(
+            db_path,
+            &node_config.private_config.database,
+        )?);
+        let fork = db.fork();
+
+        remove_local_migration_result(&fork, service_name);
```
slowli: I've understood just now (this is what you get when you rush to a release, I suppose 🤦‍♂): we need to remove migration data via `rollback_migration`, don't we? Otherwise, the restarted migration script (if it's written correctly) will do nothing and finish immediately with supposedly the same result. Ideally, this should be tested (for example, via a check in the script that no data is set in the migration).

exonum/src/runtime/dispatcher/migration_tests.rs
```diff
@@ -183,10 +201,18 @@ struct Rig {
 
 impl Rig {
     fn new() -> Self {
-        let blockchain = Blockchain::new(TemporaryDB::new(), gen_keypair(), ApiSender::closed());
+        Self::with_db_and_flag(Arc::new(TemporaryDB::new()), false)
+    }
+
+    fn with_db_and_flag(db: Arc<TemporaryDB>, flag: bool) -> Self {
+        let blockchain = Blockchain::new(
+            Arc::clone(&db) as Arc<dyn Database>,
```
slowli: Why is cloning required here?

cli/src/command/maintenance.rs
Kormick: So, in `remove_local_migration_result` we can call `Schema::add_migration_rollback` and this will remove data and local result, am I right?

PR #1730, opened 2020-01-22
Add more tests for "resume" operation

ivan-ochc:
Add more tests for "resume" operation; refactor existing tests.


aleksuss:
@ivan-ochc I'd propose to use `black` utility for the code formatting to achieve the same code formatting style.

test-suite/exonum-py-tests/suite/common.py
```diff
@@ -123,3 +134,27 @@ def wait_api_to_start(network: ExonumNetwork) -> None:
...
             except ConnectionError:
                 time.sleep(0.5)
+
+
+def generate_config(
+    network: ExonumNetwork,
+    deadline_height: int = 10000,
+    consensus: dict = None,
+    artifact_name: str = ARTIFACT_NAME,
+    instances: dict = {},
```
aleksuss: This is a bad practice to use a mutable value as a default value of the argument. Take a look [here](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html).

test-suite/exonum-py-tests/exonum_tests/deploy.py
```diff
@@ -307,32 +237,121 @@ def test_deploy_regular_stop_running_instance(self):
...
+            launcher.deploy_all()
+            launcher.wait_for_deploy()
+            launcher.start_all()
+            launcher.wait_for_start()
+
+        for validator_id in range(self.network.validators_count()):
+            host, public_port, private_port = self.network.api_address(validator_id)
+            client = ExonumClient(host, public_port, private_port)
+            available_services = client.public_api.available_services().json()
+            # crypto instance always first element in array
```
popzxc: IMHO it's a bit fragile. I'd suggest to create a helper function (e.g. `find_active_service(available_service, service_name)`), which finds required service by the name and returns its status.

test-suite/exonum-py-tests/exonum_tests/deploy.py
```diff
@@ -307,32 +238,119 @@ def test_deploy_regular_stop_running_instance(self):
...
```
popzxc: Nit: This line is not required anymore.

test-suite/exonum-py-tests/suite/common.py
```diff
@@ -123,3 +134,36 @@ def wait_api_to_start(network: ExonumNetwork) -> None:
...
+    }
+
+    return cryptocurrency_advanced_config_dict
+
+
+def find_service_status(available_service, service_name):
+    for service in available_service["services"]:
+        if service["spec"]["name"] == service_name:
+            return service["status"]["type"]
+    return "Service status has not been found"
```
popzxc: Nit: I think it's better to raise a `RuntimeError` here.

PR #1731, opened 2020-01-23
Clarify `InstanceState` protobuf message documentation [ECR-4053]

alekseysidorov:


exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -215,9 +215,25 @@ message InstanceMigration {
...
-  // Service instance activity status.
+  // Service instance activity status. 
+  //
+  // The `NONE` value means that it is initiated the addition of this instance to the blockchain
+  // and at the moment it is not yet completed. In this case, `pending_status` is `ACTIVE`.
+  //
+  // Note that status can be `NONE` only during the block execution.
   InstanceStatus status = 2;
-  // Pending status of instance if the value is not `NONE`.
+  // Pending status of instance.
```
aleksuss:
```suggestion
  // Pending status of the instance.
```

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -215,9 +215,25 @@ message InstanceMigration {
 message InstanceState {
   // Service instance specification.
   exonum.runtime.InstanceSpec spec = 1;
-  // Service instance activity status.
+  // Service instance activity status. 
+  //
+  // The `NONE` value means that it is initiated the addition of this instance to the blockchain
```
popzxc: I'd write it something like:

"This field can be `NONE` if instance was created, but activation routine for it is not yet completed. If this field
is set to `NONE`, the `pending_status` must have value `ACTIVE`."

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -215,9 +215,25 @@ message InstanceMigration {
...
+  // and at the moment it is not yet completed. In this case, `pending_status` is `ACTIVE`.
+  //
+  // Note that status can be `NONE` only during the block execution.
```
popzxc:
```suggestion
  // Status can be `NONE` only during the block execution, and this value can occur no more than
  // once in a service lifetime.
```

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -215,9 +215,25 @@ message InstanceMigration {
...
   InstanceStatus status = 2;
-  // Pending status of instance if the value is not `NONE`.
+  // Pending status of instance.
+  //
+  // If this state is not `NONE`, then this instance is in the process of initiating addition,
+  // resuming, or migration. After the block execution this status becomes an actual instance 
+  // status. 
+  //
+  // Note that this status can be not `NONE` only during the block execution.
```
popzxc:
```suggestion
  // Pending status can have value other than `NONE` only during the block execution.
```

exonum/src/proto/schema/exonum/runtime.proto
```diff
@@ -215,9 +215,25 @@ message InstanceMigration {
...
+  //
+  // The `NONE` value means that it is initiated the addition of this instance to the blockchain
+  // and at the moment it is not yet completed. In this case, `pending_status` is `ACTIVE`.
+  //
+  // Note that status can be `NONE` only during the block execution.
   InstanceStatus status = 2;
-  // Pending status of instance if the value is not `NONE`.
+  // Pending status of instance.
+  //
+  // If this state is not `NONE`, then this instance is in the process of initiating addition,
```
popzxc: I'd write it something like:

"Pending state can be not `NONE` iff core is in process of changing service status, e.g. service initialization, resuming or migration. If this field was set to value other than `NONE`, it always will be reset to `NONE` in the next block."

PR #1733, opened 2020-01-23
Prevent access to blockchain data after execution error [ECR-4033]

alekseysidorov:


exonum/src/runtime/execution_context.rs
```diff
@@ -46,6 +48,8 @@ pub struct ExecutionContext<'a> {
     dispatcher: &'a Dispatcher,
     /// Depth of the call stack.
     call_stack_depth: u64,
+    /// Flag indicates that there was an error occurred during the child call.
```
popzxc:
```suggestion
    /// Flag indicating an error occurred during the child call.
```

exonum/src/runtime/execution_context.rs
```diff
@@ -236,6 +255,11 @@ pub trait ExecutionContextUnstable {
     /// Invokes the interface method of the instance with the specified ID.
     ///
     /// See explanation about [`fallthrough_auth`](struct.ExecutionContext.html#child_context).
+    ///
+    /// # Panics
+    ///
+    /// If this method returns an error, do not access the blockchain data through this context
```
popzxc: I'd add a note that if child call errored, user is expected to propagate the error without any further processing.

exonum/src/runtime/execution_context.rs
```diff
@@ -236,6 +255,11 @@ pub trait ExecutionContextUnstable {
     /// Invokes the interface method of the instance with the specified ID.
     ///
     /// See explanation about [`fallthrough_auth`](struct.ExecutionContext.html#child_context).
+    ///
+    /// # Panics
```
slowli: Nit: I'd rather name this section "Return value", since the call does not panic *itself*. Also, I'd suggest to bubble errors up (e.g., via `?`) to avoid panics.

exonum/src/runtime/execution_context.rs
```diff
@@ -255,36 +279,44 @@ impl ExecutionContextUnstable for ExecutionContext<'_> {
...
-            .dispatcher
-            .get_service(called_instance)
-            .ok_or(CoreError::IncorrectInstanceId)?;
-
-        let (runtime_id, runtime) = self
-            .dispatcher
-            .runtime_for_service(descriptor.id)
-            .ok_or(CoreError::IncorrectRuntime)?;
+        // Move logic to closure to catch execution errors in uniform way.
+        let child_call = || {
```
slowli: Nit: I'd rather place this in a separate method (say, `make_child_call_inner`).

exonum/src/runtime/execution_context.rs
```diff
@@ -255,36 +279,44 @@ impl ExecutionContextUnstable for ExecutionContext<'_> {
...
+            if self.call_stack_depth + 1 >= Self::MAX_CALL_STACK_DEPTH {
+                let err = CoreError::stack_overflow(Self::MAX_CALL_STACK_DEPTH);
+                return Err(err);
```
slowli: Sorry for not discussing this before, but I wonder whether marking the fork as invalid here really makes sense. Logically, we want to mark the fork as failed if a service returns an error from a call expecting that changes during the call will be rolled back. There is no service call happening here, so (at least theoretically) not marking the fork as failed should be safe.

exonum/src/runtime/execution_context.rs
```diff
@@ -46,6 +52,8 @@ pub struct ExecutionContext<'a> {
     dispatcher: &'a Dispatcher,
     /// Depth of the call stack.
     call_stack_depth: u64,
+    /// Flag indicating an error occurred during the child call.
+    has_child_call_error: &'a mut bool,
```
popzxc: To be honest, a mutable reference to `bool` is a really weird thing. I'd suggest to encapsulate it into a separate type with meaningful method names.

test-suite/testkit/tests/interfaces/main.rs
```diff
@@ -389,3 +395,156 @@ fn test_any_call_panic_recursion_limit() {
...
+
+    assert_access_blockchain_data(res);
+    // Verify that the changes made by `execute_custom_call` have been reverted.
+    let snapshot = testkit.snapshot();
+    let schema = WalletSchema::new(
+        snapshot
+            .for_service(CustomCallServiceFactory::INSTANCE_NAME)
+            .unwrap(),
+    );
+    assert_eq!(schema.wallets.values().collect::<Vec<_>>(), vec![]);
```
slowli: Nit: You can use `count` instead of `collect` here.

test-suite/testkit/tests/interfaces/main.rs
```diff
@@ -389,3 +395,156 @@ fn test_any_call_panic_recursion_limit() {
...
+    execute_custom_call(|context| {
+        context.service_data();
+        Ok(())
+    })
+    .1
+    .unwrap();
+}
+
+#[test]
+fn custom_call_err_interface_call_access_panic() {
```
slowli: Nit: TBH, this name looks like word salad. I'd suggest something like `revert_after_error_in_child_call`.

test-suite/testkit/tests/interfaces/main.rs
```diff
@@ -389,3 +395,156 @@ fn test_any_call_panic_recursion_limit() {
...
+
+    let keypair = crypto::gen_keypair();
+    let res = execute_transaction(
+        &mut testkit,
+        keypair.custom_call(CustomCallServiceFactory::INSTANCE_ID, vec![]),
+    );
+    (testkit, res)
+}
+
+fn assert_access_blockchain_data(res: Result<(), ExecutionError>) {
```
slowli: Nit: `assert_access_error` would be a better name, IMO.

test-suite/testkit/tests/interfaces/main.rs
```diff
@@ -389,3 +395,156 @@ fn test_any_call_panic_recursion_limit() {
...
+    );
+}
+
+#[test]
+fn custom_call_ok() {
+    execute_custom_call(|context| {
+        context.service_data();
+        Ok(())
+    })
+    .1
```
slowli: Nit: I'd prefer destructuring assignment instead of accessing fields in a tuple.

test-suite/testkit/tests/interfaces/services.rs
```diff
@@ -222,3 +222,64 @@ impl DefaultInstance for AnyCallService {
...
     const INSTANCE_NAME: &'static str = "any-call";
 }
+
+#[exonum_interface(auto_ids)]
+pub trait CustomCallInterface<Ctx> {
+    type Output;
+    fn custom_call(&self, context: Ctx, arg: Vec<u8>) -> Self::Output;
+}
+
+type CustomCall = Box<dyn Fn(ExecutionContext<'_>) -> Result<(), ExecutionError> + Send + 'static>;
```
slowli: Nit: Do the custom call closures capture variables? (As far as I see, they don't.) If not, then it could be possible to use `fn(...)` pointers instead of closures, which would allow to simplify code substantially (`fn` pointers are `Copy + Send + 'static` by design).

exonum/src/runtime/execution_context.rs
slowli: As we've discussed with @alekseysidorov, it could be possible to encapsulate this flag (i.e., use `bool` instead of `&mut bool`) if a method is added to `ExecutionContext` to transform an execution result:

```rust
impl ExecutionContext<'_> {
    pub(crate) fn coerce_result(&self, res: Result<(), ExecutionError>) -> Result<(), ExecutionError> {
        if self.has_child_call_error {
            Err(CoreError::InvalidCall.into())
        } else {
            res
        }
    }
}
```

Would it be an improvement?

exonum/src/runtime/execution_context.rs
```diff
@@ -24,11 +24,17 @@ use crate::{
...
+
 /// Provides the current state of the blockchain and the caller information for the call
 /// which is being executed.
 ///
 /// The call can mean a transaction call, `before_transactions` / `after_transactions` hook,
 /// or the service constructor invocation.
+///
+/// # Safety
+///
+/// Errors that occur after making nested calls must bubble up to the upper level.
```
slowli: Nit: I'd add the following sentence:

> If an error has occurred in a nested call, but the returned result of the topmost call is `Ok(())`, the latter will be coerced to an error (`CoreError::InvalidCall`) and recorded as such in the blockchain. Accessing storage after an error in a nested call will result in a panic.

I'd also change "must bubble up" to "should be bubbled up" since the execution safety *does not* depend on whether the service logic is implemented according to this rule.

Finally, you should mention that making child calls is unstable functionality. Perhaps, it could make sense to move the docs to `ExecutionContextUnstable` because of this.

test-suite/testkit/tests/interfaces/services.rs
alekseysidorov: Yes, I can just use function pointer here instead of factory. But current approach is more flexible in future.

slowli: This is test code though; I don't think future compatibility should be the primary concern 🙃

alekseysidorov: Yes, but current code works fine and looks not so ugly.

exonum/src/runtime/execution_context.rs
```diff
@@ -226,16 +256,40 @@ impl<'a> ExecutionContext<'a> {
...
 }
 
 /// Collection of unstable execution context features.
+///
+/// # Safety
+///
+/// Errors that occur after making nested calls should be bubbled up to the upper level.
+///
+/// If an error has occurred in a nested call, but the returned result of the topmost
+/// call is `Ok(())`, the latter will be coerced to an error `(CoreError::InvalidCall)`
```
slowli: Nit: Unnecessary parentheses around `CoreError::InvalidCall`.

test-suite/testkit/tests/interfaces/services.rs
```diff
@@ -222,3 +222,55 @@ impl DefaultInstance for AnyCallService {
...
+    service_constructor = "Self::new_instance"
+)]
+#[service_dispatcher(implements("CustomCallInterface"))]
+pub struct CustomCallService {
+    handler: CustomCall,
+}
+
+impl CustomCallService {
+    pub fn new(handler: CustomCall) -> Self
+    {
```
slowli: Pretty sure the brace should be on the previous line.

exonum/src/runtime/execution_context.rs
```diff
@@ -269,10 +323,12 @@ impl ExecutionContextUnstable for ExecutionContext<'_> {
...
             .runtime_for_service(instance_id)
             .ok_or(CoreError::IncorrectRuntime)?;
+
         let context = self.child_context(interface_name, descriptor, fallthrough_auth);
         runtime
             .execute(context, method_id, arguments)
             .map_err(|mut err| {
+                self.should_rollback();
                 err.set_runtime_id(runtime_id).set_call_site(|| {
                     CallSite::new(
```
slowli: Nit: Wouldn't it make sense to use `CallSite::from_call_info` here?

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -523,21 +536,27 @@ impl Dispatcher {
...
-                CallSite::new(
-                    call_info.instance_id,
-                    CallType::Method {
-                        interface: String::new(),
-                        id: call_info.method_id,
-                    },
-                )
-            });
+            err.set_runtime_id(runtime_id)
+                .set_call_site(|| CallSite::from_call_info(call_info, String::new()));
```
slowli: Nit: Can't `""` be used instead of `String::new()` here? IMO, it would be better for readability.

CHANGELOG.md
```diff
@@ -119,6 +119,9 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Public structures and enums were made non-exhaustive. (#1710)
 
+- Blockchain data access from the `ExecutionContext` after interface call error
```
slowli: Nit: I'd rephrase "...after an error in a child call."

exonum/src/runtime/execution_context.rs
alekseysidorov: I am not have `CallInfo` instance here.

PR #1734, opened 2020-01-24
Document system API [ECR-4149]

popzxc:
This PR adds a documentation for public and private REST API.

Note that doctests are still sometimes failing due to some bug (which, as occurs, was not really fixed by the recent PR).

### DIFF
--- a/components/system-api/src/lib.rs
+++ b/components/system-api/src/lib.rs
@@ -14,6 +14,12 @@
 
 //! System API of an Exonum node, packaged as a node plugin.
 //!
+//! # HTTP API
+//!
+//! REST API of the service is documented in the corresponding modules:
+//! - [public API](public/index.html)
+//! - [private API](private/index.html)
+//!
 //! # Examples
 //!
 //! Use with the real node builder:
--- a/components/system-api/src/lib.rs
+++ b/components/system-api/src/lib.rs
@@ -17,6 +17,7 @@
 //! # HTTP API
 //!
 //! REST API of the service is documented in the corresponding modules:
+//!
 //! - [public API](public/index.html)
 //! - [private API](private/index.html)
 //!
--- a/components/system-api/src/private.rs
+++ b/components/system-api/src/private.rs
@@ -129,6 +129,14 @@
 //!
 //! Returns `true` if consensus is enabled on the node, and `false` otherwise.
 //!
+//! Note the difference between `consensus_enabled` and
+//! [public `healthcheck` endpoint](public/index.html#node-health-info):
+//!
+//! This endpoint only reports the setting value (should node participate in consensus
+//! or not), while `healthcheck` provides information about consensus status (since even
+//! with consensus setting turned on, node may not participate in consensus due to lack of
+//! peers).
+//!
 //! ```
 //! use exonum_system_api::SystemApiPlugin;
 //! use exonum_testkit::{ApiKind, TestKitBuilder};
@@ -183,7 +191,7 @@
 //! | Query type  | - |
 //! | Return type | - |
 //!
-//! Shutdowns the node.
+//! Shuts down the node.
 //!
 //! ```
 //! use exonum_system_api::SystemApiPlugin;
--- a/components/system-api/src/private.rs
+++ b/components/system-api/src/private.rs
@@ -130,9 +130,9 @@
 //! Returns `true` if consensus is enabled on the node, and `false` otherwise.
 //!
 //! Note the difference between `consensus_enabled` and
-//! [public `healthcheck` endpoint](public/index.html#node-health-info):
+//! [public `healthcheck` endpoint](../public/index.html#node-health-info):
 //!
-//! This endpoint only reports the setting value (should node participate in consensus
+//! This endpoint only reports the setting value (should the node participate in consensus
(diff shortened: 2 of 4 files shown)


components/system-api/src/public.rs
```diff
@@ -16,6 +16,120 @@
...
+//! | Property    | Value |
+//! |-------------|-------|
+//! | Path        | `/api/system/v1/services` |
+//! | Method      | GET   |
+//! | Query type  | - |
+//! | Return type | [`DispatcherInfo`] |
+//!
+//! Returns information about services available in the network.
+//!
+//! [`DispatcherInfo`]: struct.StatsInfo.html
```
slowli: Wrong link, should be `struct.DispatcherInfo.html`.

components/system-api/src/public.rs
```diff
@@ -16,6 +16,120 @@
...
+//! # User Agent
+//!
+//! | Property    | Value |
+//! |-------------|-------|
+//! | Path        | `/api/system/v1/user_agent` |
+//! | Method      | GET   |
+//! | Query type  | - |
+//! | Return type | `String` |
+//!
+//! Returns an user agent of the node.
```
slowli: Nit: Could be worth describing what user agent string consists of.

components/system-api/src/lib.rs
```diff
@@ -14,6 +14,12 @@
 
 //! System API of an Exonum node, packaged as a node plugin.
 //!
+//! # HTTP API
+//!
+//! REST API of the service is documented in the corresponding modules:
+//! - [public API](public/index.html)
```
slowli: Nit: A newline before the start of the list would be more idiomatic, IMO.

components/system-api/src/private.rs
```diff
@@ -16,6 +16,188 @@
...
+//! # Node Shutdown
+//!
+//! | Property    | Value |
+//! |-------------|-------|
+//! | Path        | `/api/system/v1/shutdown` |
+//! | Method      | POST   |
+//! | Query type  | - |
+//! | Return type | - |
+//!
+//! Shutdowns the node.
```
slowli: Nit / grammar: "Shuts down"?

components/system-api/src/private.rs
```diff
@@ -16,6 +16,188 @@
...
+//! # Check if Consensus is Enabled
+//!
+//! | Property    | Value |
+//! |-------------|-------|
+//! | Path        | `/api/system/v1/consensus_enabled` |
+//! | Method      | GET   |
+//! | Query type  | - |
+//! | Return type | `bool` |
+//!
+//! Returns `true` if consensus is enabled on the node, and `false` otherwise.
```
slowli: Nit / future work: It could be worth describing the difference with the consensus info returned by the `healthcheck` endpoint.

components/system-api/src/private.rs
```diff
@@ -16,6 +16,188 @@
...
+//! # Get Node Info
+//!
+//! | Property    | Value |
+//! |-------------|-------|
+//! | Path        | `/api/system/v1/network` |
+//! | Method      | GET   |
+//! | Query type  | - |
+//! | Return type | [`NodeInfo`] |
+//!
+//! Obtains information about node.
```
slowli: Nit / future work: I don't understand why this endpoint is called `network` or why it's not redundant given the user agent endpoint.

popzxc: I don't understand that either, but since the task is about documenting existing behavior, and not about changing the logic, I decided to go the straight way :)

components/system-api/src/public.rs
popzxc: I've added a short note and extended the example. Hope that'll be enough.

components/system-api/src/private.rs
slowli: Yeah, sure. I'll add a task to Jira to clarify APIs.

components/system-api/src/private.rs
```diff
@@ -129,6 +129,14 @@
 //!
 //! Returns `true` if consensus is enabled on the node, and `false` otherwise.
 //!
+//! Note the difference between `consensus_enabled` and
+//! [public `healthcheck` endpoint](public/index.html#node-health-info):
+//!
+//! This endpoint only reports the setting value (should node participate in consensus
```
slowli:
```suggestion
//! This endpoint only reports the setting value (should the node participate in consensus
```

components/system-api/src/public.rs
```diff
@@ -90,6 +90,8 @@
 //!
 //! Returns an user agent of the node.
 //!
+//! User agent includes Exonum, Rust compiler and OS version, separated by slashes.
```
slowli:
```suggestion
//! User agent includes versions of Exonum and the Rust compiler and the OS info,
//! all separated by slashes.
```

components/system-api/src/private.rs
```diff
@@ -129,6 +129,14 @@
 //!
 //! Returns `true` if consensus is enabled on the node, and `false` otherwise.
 //!
+//! Note the difference between `consensus_enabled` and
+//! [public `healthcheck` endpoint](public/index.html#node-health-info):
```
slowli:
```suggestion
//! [public `healthcheck` endpoint](../public/index.html#node-health-info):
```

PR #1735, opened 2020-01-24
Reexport types and move out tests for ProtobufConvert [ECR-4089, ECR-4090]

Kormick:
Reexported `failure` from `exonum-proto` crate, updated usage of `failure` in other crates.
Moved `ProtobufConvert` test in test-suite.


components/merkledb/src/indexes/proof_list/proof.rs
```diff
@@ -15,7 +15,7 @@
 pub use crate::ValidationError; // TODO Change for a type alias after EJB switching to rust > 1.36
 
 use exonum_crypto::Hash;
-use failure::Fail;
+use exonum_proto::failure::{self, Fail};
```
popzxc: Since `exonum_proto` dependency is optional, I don't think that importing `failure` from `exonum_proto` here is safe.

components/proto/src/lib.rs
```diff
@@ -39,6 +39,7 @@
 #[macro_use]
 extern crate serde_derive; // Required for Protobuf.
 
+pub use failure;
```
slowli: This isn't what "reexporting necessary types" means. Consider `exonum-merkledb`. Here are its reexports:

https://github.com/exonum/exonum/blob/869c61f42b47a6b56a44232f6dc05fdfab6c6e88/components/merkledb/src/lib.rs#L168-L173

...and here's the code that uses them in `exonum-derive`:

https://github.com/exonum/exonum/blob/869c61f42b47a6b56a44232f6dc05fdfab6c6e88/components/derive/src/db_traits.rs#L88-L93

Like in the case of `exonum_merkledb`, you need to create reexports in a hidden module (meaning that the existence of a module is an implementation detail) and use them in the derive macro.

components/crypto/src/proto/mod.rs
```diff
@@ -14,8 +14,10 @@
 
 pub use crate::proto::schema::*;
 
-use exonum_proto::ProtobufConvert;
-use failure::{ensure, format_err, Error};
+use exonum_proto::{
+    failure::{ensure, format_err, Error},
```
slowli: Relying on reexports that should logically be an implementation detail is a no-go, sorry.

PR #1736, opened 2020-01-24
Specify that ids are non-zero:

dmitry-timofeev:
According to `create_index_metadata`,
ids are non-zero:

> Identifier should be non-zero to translate to
a correct id in `ResolvedAddress`

## Overview

---

### Definition of Done

### DIFF
--- a/components/merkledb/src/views/metadata.rs
+++ b/components/merkledb/src/views/metadata.rs
@@ -164,7 +164,7 @@ impl Default for IndexType {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct IndexMetadata<V = Vec<u8>> {
     // A globally unique numeric index identifier. MerkleDB assigns a unique numeric ID for each
-    // fully-qualified index name.
+    // fully-qualified index name. Valid identifiers are non-zero.
     //
     // MerkleDB never re-uses the identifiers.
     identifier: u64,


components/merkledb/src/views/metadata.rs
```diff
@@ -164,7 +164,7 @@ impl Default for IndexType {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct IndexMetadata<V = Vec<u8>> {
     // A globally unique numeric index identifier. MerkleDB assigns a unique numeric ID for each
-    // fully-qualified index name.
+    // fully-qualified index name. Valid identifiers are non-zero.
```
popzxc: Not sure about this. This is not an invariant guaranteed by `IndexMetadata`, but rather an invariant of code that *use* `IndexMetadata`. That's a deep details and sort of verbalism though.

dmitry-timofeev: I see, without enforcing that invariant it is not reliable. We could postpone this doc update till this made a strict invariant, so that we can rely on it.

dmitry-timofeev:
> Also, will Java binding require a getter for the identifier from IndexMetadata? As far as I see, currently there's no way to get it.

Likely, yes, as we need an id to perform deduplication across various Fork-based accesses (which modify the name part, hence it can no longer be used as the id in the index pool): 
- https://wiki.bf.local/display/EJB/Java+Database+Accesses+Design#JavaDatabaseAccessesDesign-%D0%94%D0%B5%D0%B4%D1%83%D0%BF%D0%BB%D0%B8%D0%BA%D0%B0%D1%86%D0%B8%D1%8F
- https://github.com/exonum/exonum-java-binding/pull/1380

PR #1738, opened 2020-01-24
Clarify commit / flush split for migrations [ECR-4148]

slowli:
## Overview

This PR clarifies split between `commit_migration` and `flush_migration` operations in the core.

---

See: https://jira.bf.local/browse/ECR-4148

### DIFF
--- a/components/merkledb/src/migration.rs
+++ b/components/merkledb/src/migration.rs
@@ -624,7 +624,7 @@ impl AbortHandle {
 /// # Safety
 ///
 /// Flushing a migration needs to be performed on a `fork` which contains the final migration
-/// changes. Not doing so **may break the hash aggregation in the database.** A scenario when
+/// changes. Not doing so **may break the state aggregation in the database.** A scenario when
 /// this requirement would be violated is as follows:
 ///
 /// 1. Start a database migration in a separate thread, constructing a `MigrationHelper` around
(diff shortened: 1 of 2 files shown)


components/merkledb/src/migration.rs
```diff
@@ -620,6 +620,22 @@ impl AbortHandle {
 /// - Migrated indexes will be aggregated in the default namespace
 /// - Indexes marked with tombstones will be removed
 /// - Scratchpad associated with the migration will be cleared
+///
+/// # Safety
+///
+/// Flushing a migration needs to be performed on a `fork` which contains the final migration
```
popzxc:
```suggestion
/// Flushing a migration must be performed on a `fork` which contains the final migration
```

exonum/src/runtime/migrations.rs
```diff
@@ -76,11 +81,19 @@
 //! returns to the [`Stopped`] status. The local migration result is ignored; if the migration
 //! script has not completed locally, it is aborted.
 //!
+//! Note that commitment and flushing are separate operations and need to be performed in
```
popzxc:
```suggestion
//! Note that commitment and flushing are separate operations and must be performed in
```

exonum/src/runtime/migrations.rs
```diff
@@ -76,11 +81,19 @@
...
+//! different blocks. When a migration is flushed, the migrated data needs to have a definite
+//! state, which is ensured by an earlier commitment acting as a [Great Filter]. The requirement
```
popzxc: >acting as a Great Filter

Philosophy is good, but I think this analogy makes an already complex topic even harder to get :)

PR #1739, opened 2020-01-25
Remove lifetimes for storage accesses [ECR-4155]

aleksuss:
## Overview

This PR:
1. Removes lifetimes from structs: `Prefixed`, `InstanceDescriptor`, `Migration`, `Scratchpad` and `BlockchainData`. 
2. Adds a new trait `CopyAccessExt` to allow manipulate  indexes via structs with access `T: Access + Copy`.
3. Methods from `AccessExt` trait accept `self` by reference now.
3. Removes redundant clones.

---

See: https://jira.bf.local/browse/ECR-4155

### Definition of Done

### DIFF
--- a/components/merkledb/src/access/extensions.rs
+++ b/components/merkledb/src/access/extensions.rs
@@ -23,13 +23,14 @@ use crate::{
 };
 
 /// Extension trait allowing for easy access to indexes from any type implementing
-/// `Access`.
+/// [`Access`] + `Copy`.
 ///
 /// # Implementation details
 ///
 /// This trait is essentially a thin wrapper around [`FromAccess`]. Where `FromAccess` returns
 /// an access error, the methods of this trait will `unwrap()` the error and panic.
 ///
+/// [`Access`]: trait.Access.html
 /// [`FromAccess`]: trait.FromAccess.html
 ///
 /// # Examples
@@ -63,7 +64,7 @@ use crate::{
 /// let list = patch.get_list::<_, String>("list");
 /// assert_eq!(list.len(), 1);
 /// ```
-pub trait AccessExt: Access {
+pub trait AccessExt: Access + Copy {
     /// Returns a group of indexes. All indexes in the group have the same type.
     /// Indexes are initialized lazily; i.e., no initialization is performed when the group
     /// is created.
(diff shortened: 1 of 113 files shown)


exonum/src/runtime/mod.rs
```diff
@@ -660,32 +660,32 @@ impl<T: WellKnownRuntime> From<T> for RuntimeInstance {
 }
 
 /// Instance descriptor contains information to access the running service instance.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub struct InstanceDescriptor<'a> {
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct InstanceDescriptor {
```
slowli: I think we've agreed that `InstanceDescriptor` should remain as is 🙃 I'm not principally against this change though, it can simplify some code. Does it impact performance?

components/merkledb/src/access/extensions.rs
```diff
@@ -63,7 +64,7 @@ use crate::{
 /// let list = patch.get_list::<_, String>("list");
 /// assert_eq!(list.len(), 1);
 /// ```
-pub trait AccessExt: Access {
+pub trait AccessExt: Access + Copy {
```
slowli: Bikeshedding / opinions: I'd suggest to use the name `AccessExt` for a more generic extension trait (i.e., one which takes `&self` and doesn't require `Self: Copy`), and rename this trait to something like `CopyAccessExt`.

components/merkledb/src/access/extensions.rs
```diff
@@ -266,7 +267,227 @@ pub trait AccessExt: Access {
...
+/// [`Access`].
+///
+/// # Implementation details
+///
+/// This trait is essentially a thin wrapper around [`FromAccess`]. Where `FromAccess` returns
+/// an access error, the methods of this trait will `unwrap()` the error and panic.
+///
+/// [`Access`]: trait.Access.html
+/// [`FromAccess`]: trait.FromAccess.html
+///
```
slowli: Nit: redundant line.

components/merkledb/src/access/extensions.rs
```diff
@@ -23,13 +23,14 @@ use crate::{
 };
 
 /// Extension trait allowing for easy access to indexes from any type implementing
-/// `Access`.
+/// [`Access`] + `Copy`.
```
slowli: Explanation why both extension traits are necessary could be helpful. Something like

> This trait is helpful for *references* implementing `Access`, such as `&Fork` or `&dyn Snapshot` because Rust method resolution does not apply `AccessExt` to variables of corresponding types. For example, if `fork` has type `Fork`, then `fork.get_list("foo")` is not resolved as `AccessExt::get_list(..)`, only `(&fork).get_list("foo")` is.

(This assumes that the traits are renamed as mentioned in another comment.)

components/merkledb/tests/migration.rs
```diff
@@ -132,7 +132,7 @@ fn generate_action_with_rollbacks(namespaces: Strings) -> impl Strategy<Value =
 
 fn get_object_hash<S>(snapshot: S, name: &str, index_type: IndexType) -> Hash
```
slowli: Shouldn't `snapshot` be taken by reference if the `Copy` constraint is removed?

components/merkledb/tests/state_aggregation.rs
```diff
@@ -109,10 +109,12 @@ fn check_patch(patch: &Patch) -> TestCaseResult {
     let aggregator = system_schema.state_aggregator();
     prop_assert_eq!(state_hash, aggregator.object_hash());
 
-    let mut absent_lists: HashSet<_> = HashSet::from_iter(LIST_NAMES.iter().map(|s| s.to_string()));
-    let mut absent_maps: HashSet<_> = HashSet::from_iter(MAP_NAMES.iter().map(|s| s.to_string()));
+    let mut absent_lists: HashSet<_> =
+        HashSet::from_iter(LIST_NAMES.iter().map(|s| (*s).to_string()));
```
slowli: Is this some pedantic clippy lint? The change looks artificial, IMO.

runtimes/rust/src/service.rs
```diff
@@ -269,20 +269,20 @@ impl<'a> AfterCommitContext<'a> {
 
 // It is impossible to use `Cow` with `InstanceDescriptor` since it has a lifetime of its own.
 #[derive(Debug, Clone)]
-enum CowInstanceDescriptor<'a> {
-    Borrowed(InstanceDescriptor<'a>),
+enum CowInstanceDescriptor {
```
slowli: This enum is useless now and should be removed.

exonum/src/runtime/execution_context.rs
```diff
@@ -184,7 +184,7 @@ impl<'a> ExecutionContext<'a> {
     }
 
     /// Re-borrows an execution context with the given instance descriptor.
-    fn reborrow<'s>(&'s mut self, instance: InstanceDescriptor<'s>) -> ExecutionContext<'s> {
+    fn reborrow(&mut self, instance: InstanceDescriptor) -> ExecutionContext {
```
slowli: Use `-> ExecutionContext<'_>`; the child context still has lifetime dependent on `&mut self`.

components/merkledb/src/access/extensions.rs
popzxc: I'd also expect cross-references between traits in their descriptions, like 

```
For a version on `AccessExt` traits designed for `Copy` types (e.g. `&Fork` and
`&dyn Snapshot`) see [`CopyAccessExt` trait](trait.CopyAccessExt.html).
```

components/merkledb/src/db.rs
```diff
@@ -531,7 +531,7 @@ pub trait DatabaseExt: Database {
     /// and then applying backups in the reverse order:
     ///
     /// ```
-    /// # use exonum_merkledb::{access::{Access, AccessExt}, Database, DatabaseExt, TemporaryDB};
+    /// # use exonum_merkledb::{access::{Access, AccessExt, AccessRefExt}, Database, DatabaseExt, TemporaryDB};
```
popzxc: Nit: Are both traits required here?

components/merkledb/tests/state_aggregation.rs
popzxc: Yes, `clippy` in 1.40 complains (and this is not even `pedantic` lint):

```
Checking playground v0.0.1 (/playground)
warning: calling `to_string` on `&&str`
 --> src/main.rs:4:25
  |
4 |     list.iter().map(|s| s.to_string()).collect()
  |                         ^^^^^^^^^^^^^ help: try dereferencing the receiver: `(*s).to_string()`
  |
  = note: `#[warn(clippy::inefficient_to_string)]` on by default
  = help: `&str` implements `ToString` through a slower blanket impl, but `str` has a fast specialization of `ToString`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string
```

See [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=83ccef2d786d9ad6a7f209e89e1c7507).

This is a questionable topic though: on the one hand it's an almost free possibility to write more efficient code, but on the other hand it worsens the readability.

popzxc: And one more comment: you can compare the assembly generated for both versions:

- [`s.to_string()`](https://godbolt.org/z/jc87B8) - 534 lines;
- [`(*s).to_string()`](https://godbolt.org/z/5pkaG2) - 159 lines.

slowli: Would `|&s| s.to_string()` work in this case? IMO, it would be more readable. Using `Iterator::copied()` is another option, although if `|&s|` works, it would be optimal, IMO.

popzxc: Yep, seems that clippy is happy with `|&s|` and assembly looks identical to `(*s).to_string()`.

exonum/src/runtime/mod.rs
aleksuss: To be honest I haven't measured it. But I think it won't bring a significant regression.

runtimes/rust/src/service.rs
```diff
@@ -331,15 +303,15 @@ impl<'a> Broadcaster<'a> {
...
+    pub(super) fn instance(&self) -> &InstanceDescriptor {
+        &self.instance
     }
 
     /// Converts the broadcaster into the owned representation, which can be used to broadcast
     /// transactions asynchronously.
     pub fn into_owned(self) -> Broadcaster<'static> {
         Broadcaster {
-            instance: self.instance.into_owned(),
+            instance: self.instance.clone(),
```
slowli: Nit: Can't this be simplified to `self.instance`?

components/merkledb/src/access/mod.rs
```diff
@@ -162,7 +162,7 @@ impl<T: RawAccess> Access for T {
 /// # Examples
 ///
 /// ```
-/// use exonum_merkledb::{access::{AccessExt, AccessRefExt, Prefixed}, Database, TemporaryDB};
+/// use exonum_merkledb::{access::{CopyAccessExt, AccessExt, Prefixed}, Database, TemporaryDB};
```
slowli: Nit: Non-alphabetical ordering between `CopyAccessExt` and `AccessExt` (in some other places as well).

components/merkledb/src/indexes/group.rs
```diff
@@ -193,6 +193,7 @@ mod tests {
         // group.get(&3).push("quux".to_owned());
     }
 
+    #[allow(clippy::needless_pass_by_value)]
```
slowli: Nit: If we tolerate `prepare_key_iter(&&fork)`, this warning could be solved as intended (i.e., by using `fork: &A`).

components/merkledb/src/access/extensions.rs
```diff
@@ -29,14 +29,17 @@ use crate::{
 ///
 /// This trait is essentially a thin wrapper around [`FromAccess`]. Where `FromAccess` returns
 /// an access error, the methods of this trait will `unwrap()` the error and panic.
-///
+/// This trait is helpful for references implementing Access, such as `&Fork` or `&dyn Snapshot`
+/// because Rust method resolution does not apply `AccessExt` to variables of corresponding types.
+/// For example, if fork has type Fork, then `fork.get_list("foo")` is not resolved
+/// as `AccessExt::get_list(..)`, only `(&fork).get_list("foo")` is.
```
slowli: Nit: A newline after this paragraph could improve readability, IMO.

components/merkledb/src/access/extensions.rs
```diff
@@ -29,14 +29,17 @@ use crate::{
 ///
 /// This trait is essentially a thin wrapper around [`FromAccess`]. Where `FromAccess` returns
 /// an access error, the methods of this trait will `unwrap()` the error and panic.
-///
+/// This trait is helpful for references implementing Access, such as `&Fork` or `&dyn Snapshot`
```
slowli: Nit: `Access` should be code-formatted.

components/merkledb/src/access/extensions.rs
```diff
@@ -29,14 +29,17 @@ use crate::{
...
+/// because Rust method resolution does not apply `AccessExt` to variables of corresponding types.
```
slowli: Nit: `AccessExt` can be linked to.

PR #1740, opened 2020-01-26
Misc fixes

slowli:
## Overview

This PR fixes miscellaneous flaws in crate READMEs and throughout the codebase. Not meant to be exhaustive, just the surface-level stuff I've encountered lately.

### DIFF
--- a/components/merkledb/Cargo.toml
+++ b/components/merkledb/Cargo.toml
@@ -56,14 +56,21 @@ harness = false
 
 [features]
 default = ["rocksdb_snappy", "with-protobuf"]
-long_benchmarks = []
 with-protobuf = ["with-serde", "protobuf", "exonum-proto"]
 with-serde = []
+
+# Compression options passed to RocksDB backend.
 rocksdb_snappy = ["rocksdb/snappy"]
 rocksdb_lz4 = ["rocksdb/lz4"]
 rocksdb_zlib = ["rocksdb/zlib"]
 rocksdb_zstd = ["rocksdb/zstd"]
 rocksdb_bzip2 = ["rocksdb/bzip2"]
+
+# Enables long benchmarks; does not influence main crate code.
+long_benchmarks = []
+# Enables `unsafe` optimizations / experimental features.
+# Strongly discouraged for use in production.
+yolo = []
 
 [build-dependencies]
 exonum-build = { version = "0.13.0-rc.2", path = "../build" }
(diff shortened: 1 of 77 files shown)


components/merkledb/Cargo.toml
```diff
@@ -55,14 +56,21 @@ harness = false
...
 rocksdb_lz4 = ["rocksdb/lz4"]
 rocksdb_zlib = ["rocksdb/zlib"]
 rocksdb_zstd = ["rocksdb/zstd"]
 rocksdb_bzip2 = ["rocksdb/bzip2"]
 
+# Enables long benchmarks; does not influence main crate code.
+long_benchmarks = []
+# Enables `unsafe` optimizations / experimental features.
+# Strongly discouraged for use in production.
+yolo = []
```
popzxc: Should we create an even further restriction, like

```rust
#[cfg(all(not(debug_assertions), yolo))]
std::compile_error!("YOLO mode in release builds is not supported");
```

?

popzxc: Or, as an alternative, something like

```rust
#[cfg(all(not(debug_assertions),  yolo))]
mod yolo_warning {
    // Explicitly enable warning, so it cannot be suppressed with
    // `#![allow(dead_code)]`.
    #![warn(dead_code)]
    // Will spawn a compile-time warning about unused item.
    const YOLO_WARNING: &str = "Using YOLO mode in release builds \
                                is not recommended at all";
}
```

With that, one will be able to use this in the production, but will have to live with a non-avoidable warning.

slowli: Does the warning extend to the case when `exonum-merkledb` is a dependency of the crate being built? If so, will the warning be understood by users? I think it's enough that the user needs to perform some actions to enable the feature.

popzxc: Yes, I think that warning will be understood by users, if it will be phrased with a preventive explanation à la "You see this warning because your crate or one of your upstream dependencies enabled `yolo` feature of `exonum-merkledb` crate". IMHO you can never be overly defensive in protecting users from shooting in the foot :)

Currently this feature does not enable anything that scary though, so I think you're right and for now just an extra action is enough indeed.

test-suite/testkit/src/builder.rs
```diff
@@ -198,6 +160,9 @@ impl TestKitBuilder {
     }
 
     /// Adds a node plugin to the testkit.
+    ///
+    /// This method is only available if the crate is compiled with the `exonum-node` feature,
```
popzxc: It seems that since `exonum-node` feature is not enabled by default, this method will not appear on `docs.rs` (at least it does not in local `cargo doc` build).
Maybe we should make this feature a default one, and just specify `default-features = false` in packages that do not need it?

popzxc:
P.S. changelog is not updated yet :)

PR #1741, opened 2020-01-27
Document supervisor api [ECR-4149]

popzxc:
This PR adds a documentation for public and private REST API for supervisor.

Note that doctests are still sometimes failing due to some bug (which, as occurs, was not really fixed by the recent PR).

### DIFF
--- a/services/supervisor/src/api.rs
+++ b/services/supervisor/src/api.rs
@@ -78,12 +78,12 @@
 //! | Path        | `/api/supervisor/v1/config-proposal` |
 //! | Method      | GET   |
 //! | Query type  | - |
-//! | Return type | Option<[`ConfigProposalWithHash`]> |
+//! | Return type | `Option<[ConfigProposalWithHash]>` |
 //!
 //! Returns the configuration proposal which is currently pending. Returns `None` if there is no
 //! pending configuration at the moment.
 //!
-//! [`ConfigProposalWithHash`]: ../struct.ConfigProposalWithHash.html
+//! [ConfigProposalWithHash]: ../struct.ConfigProposalWithHash.html
 //!
 //! ```
 //! # use exonum_rust_runtime::ServiceFactory;
--- a/services/supervisor/src/api.rs
+++ b/services/supervisor/src/api.rs
@@ -153,7 +153,7 @@
 //! # impl Service for SomeService {}
 //! #
 //! # fn config_for_service() -> Vec<u8> {
-//! #     Vec::new()   
+//! #     Vec::new()
 //! # }
 //! #
 //! # fn main() -> Result<(), failure::Error> {
@@ -446,7 +446,7 @@
 //!
 //! Returns the current supervisor configuration, which includes the supervisor operating mode.
 //!
-//! [`SupervisorConfig`]: ../SupervisorConfig.html
+//! [`SupervisorConfig`]: ../struct.SupervisorConfig.html
 //!
 //! ```
 //! use exonum_rust_runtime::ServiceFactory;
@@ -482,7 +482,7 @@
 //! Returns the state of the deployment for a certain `ArtifactId`.
 //!
 //! [`DeployInfoQuery`]: struct.DeployInfoQuery.html
-//! [`AsyncEventState`]: ../struct.AsyncEventState.html
+//! [`AsyncEventState`]: ../enum.AsyncEventState.html
 //!
 //! ```
 //! # use exonum::{crypto::Hash, helpers::Height, merkledb::BinaryValue};
--- a/services/supervisor/src/api.rs
+++ b/services/supervisor/src/api.rs
@@ -34,7 +34,6 @@
 //!     - [Check the deployment status](#check-deployment-status)
 //!     - [Check the migration status](#check-migration-status)
 //!
-//!
 //! # Public API
 //!
 //! ## Obtaining Consensus Configuration
@@ -44,7 +43,7 @@
 //! | Path        | `/api/supervisor/v1/consensus-config` |
 //! | Method      | GET   |
 //! | Query type  | - |
-//! | Return type | [`BlockInfo`] |
+//! | Return type | [`ConsensusConfig`] |
 //!
 //! Returns the current consensus configuration.
 //!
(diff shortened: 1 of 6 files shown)


services/supervisor/src/api.rs
```diff
@@ -12,6 +12,578 @@
...
+//!
+//!     - [Request to deploy an artifact](#request-to-deploy-an-artifact)
+//!     - [Request service migration](#request-service-migration)
+//!     - [Request to accept new configuration](#request-to-accept-new-configuration)
+//!     - [Vote for configuration proposal](#vote-for-configuration-proposal)
+//!     - [Obtain current configuration number](#obtain-current-configuration-number)
+//!     - [Obtain current supervisor operating mode](#obtain-current-supervisor-operating-mode)
+//!     - [Check the deployment status](#check-deployment-status)
+//!     - [Check the migration status](#check-migration-status)
+//!
```
slowli: Nit: Unnecessary newline.

services/supervisor/src/api.rs
```diff
@@ -12,6 +12,578 @@
...
+//! # Public API
+//!
+//! ## Obtaining Consensus Configuration
+//!
+//! | Property    | Value |
+//! |-------------|-------|
+//! | Path        | `/api/supervisor/v1/consensus-config` |
+//! | Method      | GET   |
+//! | Query type  | - |
+//! | Return type | [`BlockInfo`] |
```
slowli: Shouldn't it be `ConsensusConfig` here?

services/supervisor/src/api.rs
```diff
@@ -12,6 +12,578 @@
...
+//! ```
+//!
+//! ## Obtaining Pending Configuration Proposal
+//!
+//! | Property    | Value |
+//! |-------------|-------|
+//! | Path        | `/api/supervisor/v1/config-proposal` |
+//! | Method      | GET   |
+//! | Query type  | - |
+//! | Return type | `Option<[ConfigProposalWithHash]>` |
```
slowli: The link here isn't parsed as a link; it probably needs to be written like

```text
`Option<`[`ConfigProposalWithHash`]`>`
```

services/supervisor/src/api.rs
```diff
@@ -12,6 +12,578 @@
...
+//!
+//! Returns the configuration proposal which is currently pending. Returns `None` if there is no
+//! pending configuration at the moment.
+//!
+//! [ConfigProposalWithHash]: ../struct.ConfigProposalWithHash.html
+//!
+//! ```
+//! use exonum_rust_runtime::ServiceFactory;
```
slowli: Nit: IMO, it would be safe to skip imports / `testkit` initialization if it coincides with the earlier tests:

```rust
let mut testkit = // same as in previous example
let pending_proposal: Option<ConfigProposalWithHash> = testkit
```

services/supervisor/src/api.rs
```diff
@@ -12,6 +12,578 @@
...
+//! Depending on the supervisor operating mode, it may be required to send such a request to
+//! majority of nodes (in "decentralized" mode), or one request will be enough (in "simple" mode).
+//!
+//! After receiving a deployment request, supervisor creates a corresponding transaction, signs it
+//! with node's keys and broadcasts the transaction within the network. The hash of broadcast
+//! transaction is returned from the endpoint.
+//!
+//! For more details on deploy requests, see [crate documentation](../index.html).
+//!
+//! **Warning:** `DeployRequest` structure should be serialized using corresponding protobuf message,
```
slowli: Off-topic: that `DeployRequest` is serialized into a hex string is unclear from the code sample below. IMO, a better way would be to have a wrapper type like

```rust
pub struct HexString<T>(pub T);

impl<T: BinaryValue> Serialize for HexString<T> { /* ... */ }

// used like
let response = testkit
    .api()
    .private(ApiKind::Service("supervisor"))
    .query(&HexString(deploy_request))
    .post("deploy-artifact")?;
```

This would have the added benefit that `DeployRequest` could have more "sane" serde serialization in other contexts (e.g., if it is returned from the API, or if somebody decides to serialize it in other format, such as `bincode`).

Also, how is a hex string serialized into query params? It could be provided as is for the request body, but as a query it would require the param name. If it *is* serialized as a request body, it makes sense to replace "query type" -> "body type" in the table above. (And maybe update docs in testkit.)

services/supervisor/src/api.rs
```diff
@@ -12,6 +12,578 @@
...
+//! #
+//! # fn main() -> Result<(), failure::Error> {
+//! let mut testkit = TestKitBuilder::validator()
+//!     .with_rust_service(Supervisor)
+//!     .with_artifact(Supervisor.artifact_id())
+//!     .with_instance(Supervisor::simple())
+//!     .with_rust_service(SomeService)
+//!     .create();
+//!
+//! // In this example, we will try to deploy supervisor service.
```
slowli: Isn't `SomeService` deployed, not a supervisor?

services/supervisor/src/api.rs
```diff
@@ -12,6 +12,578 @@
...
+//! };
+//!
+//! let tx_hash: Hash = testkit
+//!     .api()
+//!     .private(ApiKind::Service("supervisor"))
+//!     .query(&deploy_request)
+//!     .post("deploy-artifact")?;
+//!
+//! let block = testkit.create_block();
+//! let result = block[tx_hash].status();
```
slowli: Nit: Could this be shortened to

```rust
block[tx_hash].status()?;
```

services/supervisor/src/api.rs
```diff
@@ -12,6 +12,578 @@
...
+//! | Method      | POST   |
+//! | Query type  | [`DeployRequest`] |
+//! | Return type | [`Hash`] |
+//!
+//! Votes for a pending configuration with a certain hash.
+//!
+//! Depending on the supervisor operating mode, it may be required to vote by majority of
+//! nodes (in "decentralized" mode), or one vote will be enough (in "simple" mode).
+//!
+//! The first vote is sent automatically by the node which broadcast the proposal, there is
```
slowli: Nit: I'd rephrase "The node that broadcast the proposal is considered to have voted for it, there is no need..." because the node does not really *send* the vote, does it?

services/supervisor/src/api.rs
```diff
@@ -258,6 +830,7 @@ impl PublicApi for ApiImpl<'_> {
     }
 }
 
+/// Wires Supervisor API endpoints.
 pub fn wire(builder: &mut ServiceApiBuilder) {
```
slowli: Does this method need to be public? I'd prefer for it to be `pub(crate)`.

services/supervisor/src/lib.rs
```diff
@@ -157,9 +161,9 @@ use exonum_rust_runtime::{
 
 use crate::{configure::ConfigureMut, mode::Mode};
 
+pub mod api;
```
slowli: Nit: IMO, it makes sense not to re-export `api`-related types (i.e., `DeployInfoQuery` and `MigrationInfoQuery`) from the crate root.

services/supervisor/src/api.rs
popzxc: Yep, I forgot to fix the comment.

services/supervisor/src/api.rs
popzxc: I thought that with `assert` the example will be more demonstrative (in other places `?` is used to not-so-important determine places, where we don't want to `unwrap`, but here it's the key point of the example).

services/supervisor/src/api.rs
popzxc: I changed `query` to `body`, and added the notes to the `testkit` docs.

However, I did not introduce `HexString` here, since it's a docs-only PR, and I think it's better to make these changes separately.

services/supervisor/src/api.rs
popzxc: Is it OK now, or should I hide something else as well?

services/supervisor/src/api.rs
slowli: Sure, that's why I prefaced the comment as off-topic.

services/supervisor/src/api.rs
```diff
@@ -79,24 +78,25 @@
 //! | Path        | `/api/supervisor/v1/config-proposal` |
 //! | Method      | GET   |
 //! | Query type  | - |
-//! | Return type | `Option<[ConfigProposalWithHash]>` |
+//! | Return type | Option<[`ConfigProposalWithHash`]> |
```
slowli: Micro-nit: `Option` is not code-formatted now 🙃

PR #1742, opened 2020-01-27
Add public getter for IndexMetadata::identifier [ECR-4163]

vitvakatu:
## Overview

I decided to move documentation to the method, so it would be available on docs.rs

---

See: https://jira.bf.local/browse/ECR-4163

### Definition of Done


components/merkledb/src/views/metadata.rs
```diff
@@ -241,6 +237,16 @@ impl<V> IndexMetadata<V> {
     pub fn index_type(&self) -> IndexType {
         self.index_type
     }
+
+    /// Returns a globally unique numeric index identifier.
+    /// MerkleDB assigns a unique numeric ID for each fully-qualified index name.
+    ///
+    /// Valid identifiers are non-zero.
```
dmitry-timofeev: Shall we enforce it (in the constructor of this structure)? Or move the last two sentences to the thing that enforces these properties and link from here (@slowli , @popzxc)?

popzxc: I'd prefer the latter option.

vitvakatu: `IndexesPool::create_index_metadata` has a comment saying "Identifier should be non-zero to translate to a correct id in `ResolvedAddress`". `ResolveAddress::new` has a corresponding signature `(name, id: Option<NonZeroU64>)`. I'm not sure where should I put these sentences.

slowli: A provocative idea: Is it possible to change identifier to `NonZeroU64`?

vitvakatu: So the `1` would be the initial value?

slowli: I mean define `IndexMetadata.identifier` as `NonZeroU64` and return `NonZeroU64` from the corresponding public method.

components/merkledb/src/views/metadata.rs
```diff
@@ -340,7 +346,7 @@ impl<T: RawAccess> IndexesPool<T> {
     {
         let len = self.len();
         let metadata = IndexMetadata {
-            identifier: len + 1,
+            identifier: NonZeroU64::new(len + 1).unwrap(),
             // ^-- Identifier should be non-zero to translate to a correct id in `ResolvedAddress`
```
slowli: Nit / opinion: I feel that the comment is obsolete now since `NonZeroU64::new` is self-explanatory.

PR #1747, opened 2020-01-28
Generic access enumerations [ECR-4174]

slowli:
## Overview

This PR introduces access *enumerations* to MerkleDB, (hopefully) allowing to implement bindings more easily.

Discussion:

- Is this good enough for bindings?

---

See: https://jira.bf.local/browse/ECR-4174

### DIFF
--- a/components/merkledb/src/access/mod.rs
+++ b/components/merkledb/src/access/mod.rs
@@ -177,7 +177,13 @@ pub struct Prefixed<T> {
     prefix: String,
 }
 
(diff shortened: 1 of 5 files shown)


components/merkledb/src/access/mod.rs
```diff
@@ -177,6 +177,18 @@ pub struct Prefixed<T> {
     prefix: String,
 }
 
+// **NB.** Must not be made public! This would allow the caller to violate access restrictions
```
popzxc: I'd put it as a docstring to the both methods, just in case.

PR #1749, opened 2020-01-29
Implement Configure interface for timestamping example [ECR-4172]

Kormick:


examples/timestamping/backend/src/api.rs
```diff
@@ -84,6 +87,18 @@ impl PublicApi {
         })
     }
 
+    /// Endpoint for getting service configuration.
+    pub fn get_service_configuration(self, state: &ServiceApiState<'_>) -> api::Result<Config> {
+        let schema = Schema::new(state.service_data());
+        let config_entry = schema.config.get();
+
+        if let Some(config) = config_entry {
```
popzxc: Nit: You can use [`Option::ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else).

examples/timestamping/backend/src/lib.rs
```diff
@@ -77,3 +78,29 @@ impl Service for TimestampingService {
...
+
+impl Configure for TimestampingService {
+    type Params = Config;
+
+    fn verify_config(
+        &self,
+        _context: ExecutionContext<'_>,
+        params: Self::Params,
+    ) -> Result<(), ExecutionError> {
+        if params.time_service_name.is_empty() {
```
popzxc: I don't think this is a sufficient check. See [how it's checked above](https://github.com/exonum/exonum/blob/356b267d64ca2fe89e0cbd446fcdafd265fd7ec7/examples/timestamping/backend/src/lib.rs#L64).

examples/timestamping/backend/src/lib.rs
```diff
@@ -77,3 +78,29 @@ impl Service for TimestampingService {
         TimestampingApi.wire(builder);
     }
 }
+
+impl Configure for TimestampingService {
+    type Params = Config;
+
+    fn verify_config(
```
popzxc: Since `initialize` method and `Configure` interface for `TimestampingService` share the same configuration structure, you can add `impl Config` with verification and applying code to avoid code duplication.

examples/timestamping/backend/tests/config.rs
```diff
@@ -0,0 +1,104 @@
...
+        .snapshot()
+        .for_service(SERVICE_NAME)
+        .unwrap()
+        .get_entry("config")
+        .get()
+        .unwrap()
+}
+
+#[test]
+fn test_propose_configuration() {
```
popzxc: I'd expect this test to fail, because there is no service with such a name.

examples/timestamping/backend/src/lib.rs
```diff
@@ -39,35 +39,45 @@ pub use crate::{
...
+#[service_dispatcher(implements("TimestampingInterface", raw = "Configure<Params = Config>"))]
 #[service_factory(proto_sources = "proto")]
 pub struct TimestampingService;
 
+fn verify_config(context: &ExecutionContext<'_>, config: &Config) -> Result<(), ExecutionError> {
+    if !is_valid_index_name_component(&config.time_service_name) {
+        return Err(Error::InvalidConfig.into());
+    }
+
+    if context
```
slowli: I'd argue that this check is still not strong enough given that the `timestamp` method implementation uses `BlockchainData::service_schema`, which gives significantly higher guarantees:

- The service instance exists and is active
- The instance corresponds to a supported version of the time service artifact

These checks could be performed either by instantiating the time service schema as it's done in the `timestamp` handler, or by using `impl RequireArtifact for TimeSchema` directly.

There are marginal cases where the check may be insufficient (e.g., the time service is switched off in the same proposal) or too strict (the time service is switched on in the same proposal), but I'm not sure whether we should consider them.

examples/timestamping/backend/src/lib.rs
```diff
@@ -39,35 +39,45 @@ pub use crate::{
...
 use crate::{api::PublicApi as TimestampingApi, schema::Schema};
 
 #[derive(Debug, ServiceDispatcher, ServiceFactory)]
-#[service_dispatcher(implements("TimestampingInterface"))]
+#[service_dispatcher(implements("TimestampingInterface", raw = "Configure<Params = Config>"))]
 #[service_factory(proto_sources = "proto")]
 pub struct TimestampingService;
 
+fn verify_config(context: &ExecutionContext<'_>, config: &Config) -> Result<(), ExecutionError> {
+    if !is_valid_index_name_component(&config.time_service_name) {
```
slowli: Isn't this check redundant given that instance name sanitation is performed by the core? A service with an invalid name is guaranteed to never be instantiated.

examples/timestamping/backend/tests/config.rs
```diff
@@ -0,0 +1,137 @@
...
+                .with_constructor(Config {
+                    time_service_name: TIME_SERVICE_NAME.to_owned(),
+                }),
+        );
+
+    let testkit = if second_time_service {
+        let time_service = TimeServiceFactory::with_provider(mock_provider.clone());
+        let time_service_artifact = time_service.artifact_id();
+        testkit
+            .with_rust_service(time_service)
```
slowli: This is redundant: `TimeServiceFactory` is already added to the Rust runtime.

examples/timestamping/backend/tests/config.rs
```diff
@@ -0,0 +1,137 @@
...
+            .with_artifact(time_service_artifact.clone())
```
slowli: This also should be redundant: the time service artifact has already been added to the builder.

examples/timestamping/backend/tests/config.rs
```diff
@@ -0,0 +1,137 @@
...
+    let new_config = propose_configuration(&mut testkit, config);
+    // Check that configuration has not changed.
+    assert_eq!(new_config.time_service_name, TIME_SERVICE_NAME);
+
+    // Service name with illegal character.
+    let config = Config {
+        time_service_name: "illegal.illegal".to_string(),
+    };
+    // Propose configuration with not existing service name.
+    let new_config = propose_configuration(&mut testkit, config.clone());
```
slowli: Nit: Is `clone()` required here?

examples/timestamping/backend/tests/config.rs
```diff
@@ -0,0 +1,137 @@
...
+    // Check that configuration has not changed.
+    assert_eq!(new_config.time_service_name, TIME_SERVICE_NAME);
+
+    // Name of not existing service.
+    let config = Config {
+        time_service_name: "not_service".to_string(),
+    };
+    // Propose configuration with not existing service name.
+    let new_config = propose_configuration(&mut testkit, config.clone());
```
popzxc: Nit: You can create a test vector, like `let incorrect_names = vec!["", "illegal.illegal", "not_service"];` to avoid code duplication.

examples/timestamping/backend/tests/config.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    (testkit, mock_provider)
+}
+
+/// Creates block with `ConfigPropose` tx and returns current service configuration.
+fn propose_configuration(testkit: &mut TestKit, config: Config) -> Config {
+    let tx = ConfigPropose::immediate(0).service_config(SERVICE_ID, config);
+
+    let initiator_id = testkit.network().us().validator_id().unwrap();
+    let (pub_key, sec_key) = &testkit.validator(initiator_id).service_keypair();
```
slowli: Nit: `&` here looks non-idiomatic. Also, doesn't `testkit.network().us()` have `service_keypair()` method on its own?

examples/timestamping/backend/tests/config.rs
```diff
@@ -0,0 +1,137 @@
...
+    testkit.create_block_with_transaction(tx.sign_for_supervisor(*pub_key, sec_key));
+
+    testkit
+        .snapshot()
+        .for_service(SERVICE_NAME)
+        .unwrap()
+        .get_entry("config")
```
slowli: Nit: This looks fragile. I'd rather suggest to use the HTTP API of the service to retrieve the config. Also, to test configuration failures, it could be more appropriate to check the error of the corresponding proposal.

examples/timestamping/backend/tests/config.rs
```diff
@@ -0,0 +1,137 @@
...
+    assert_eq!(new_config.time_service_name, TIME_SERVICE_NAME);
+
+    // Name of not existing service.
+    let config = Config {
+        time_service_name: "not_service".to_string(),
+    };
+    // Propose configuration with not existing service name.
+    let new_config = propose_configuration(&mut testkit, config.clone());
+    // Check that configuration has not changed.
+    assert_eq!(new_config.time_service_name, TIME_SERVICE_NAME);
```
slowli: With the tightened checks I've proposed above, it would be possible to test that the service refuses a configuration with a service name that corresponds to a bogus artifact, for example, `SERVICE_NAME`.

examples/timestamping/backend/tests/config.rs
```diff
@@ -75,21 +73,24 @@ fn init_testkit(second_time_service: bool) -> (TestKit, MockTimeProvider) {
...
+    let block = testkit.create_block_with_transaction(tx.clone());
+
+    if let Err(e) = block.index(tx.object_hash()).status() {
+        return Err(e.clone());
+    }
+
+    let new_config = testkit
+        .api()
+        .public(ApiKind::Service(SERVICE_NAME))
+        .get::<Config>("v1/timestamps/config")
```
slowli: Nit: Turbofish is most probably redundant here; the type of `new_config` can be inferred from the return statement below.

examples/timestamping/backend/tests/config.rs
```diff
@@ -75,21 +73,24 @@ fn init_testkit(second_time_service: bool) -> (TestKit, MockTimeProvider) {
...
-        .for_service(SERVICE_NAME)
-        .unwrap()
-        .get_entry("config")
-        .get()
-        .unwrap()
+    let (pub_key, sec_key) = testkit.network().us().service_keypair();
+    let tx = tx.sign_for_supervisor(pub_key, &sec_key);
+    let block = testkit.create_block_with_transaction(tx.clone());
+
+    if let Err(e) = block.index(tx.object_hash()).status() {
```
slowli: Nit: `block[tx.object_hash()]` should work (or `block[0]` for that matter; `create_block_with_transaction` creates a block with exactly the specified transaction and nothing else). `block[0]` would also allow to avoid cloning `tx`.

examples/timestamping/backend/tests/config.rs
```diff
@@ -75,21 +73,24 @@ fn init_testkit(second_time_service: bool) -> (TestKit, MockTimeProvider) {
     (testkit, mock_provider)
 }
 
-/// Creates block with `ConfigPropose` tx and returns current service configuration.
-fn propose_configuration(testkit: &mut TestKit, config: Config) -> Config {
+/// Creates block with `ConfigPropose` tx and returns `Result` with new configuration or corresponding `ExecutionError`.
```
slowli: Micro-nit: This comment looks overly long. Also, does it make sense to *return* a `Config` if it's guaranteed to equal the input one? It could make more sense to return `Result<(), ExecutionError>` and possibly to check that the config is as expected *within* this method.

examples/timestamping/backend/tests/config.rs
```diff
@@ -73,24 +71,29 @@ fn init_testkit(second_time_service: bool) -> (TestKit, MockTimeProvider) {
...
         .api()
         .public(ApiKind::Service(SERVICE_NAME))
-        .get::<Config>("v1/timestamps/config")
+        .get("v1/timestamps/config")
         .expect("Failed to get service configuration");
-    Ok(new_config)
+    if config.time_service_name == new_config.time_service_name {
+        Ok(())
+    } else {
+        Err(Error::InvalidConfig.into())
```
slowli: Is this return reachable? If the transaction has completed successfully, the configuration should be applied, shouldn't it? If so, I'd use

```rust
assert_eq!(config.time_service_name, new_config.time_service_name);
```

examples/timestamping/backend/tests/config.rs
```diff
@@ -101,10 +104,11 @@ fn test_propose_configuration() {
...
 
     // Propose valid configuration.
-    let new_config = propose_configuration(&mut testkit, config.clone())
-        .expect("Configuration proposal failed.");
-
-    assert_eq!(new_config.time_service_name, config.time_service_name);
+    let res = propose_configuration(&mut testkit, config);
+    assert!(
+        res.is_ok(),
+        format!("Configuration proposal failed: {:?}", res)
```
slowli: This `assert` seems the same as calling `res.expect("Configuration proposal failed")`.

PR #1753, opened 2020-01-30
Add supervisor mode in README

aleksuss:
## Overview

---

### Definition of Done

### DIFF
--- a/examples/cryptocurrency-advanced/README.md
+++ b/examples/cryptocurrency-advanced/README.md
@@ -71,7 +71,7 @@ Generate template:
 ```sh
 mkdir example
 
-exonum-cryptocurrency-advanced generate-template example/common.toml --validators-count 4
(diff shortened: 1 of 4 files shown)


aleksuss:
Sadly, it is not. By the way, I though the simple mode is default as well.

popzxc:
@aleksuss maybe then make this behaviour the default in `cli`? :)

aleksuss:
I don't know if it really makes sense. It adds some ambiguity, doesn't it ? @slowli @alekseysidorov  what do you think ?

slowli:
@aleksuss I think adding a default value makes sense: the default value is prominently displayed in the command help, so there shouldn't be significant ambiguity. At the same time, it would make CLI commands less verbose.

aleksuss:
@slowli deal.

PR #1756, opened 2020-01-31
Structure Protobuf files [ECR-4187]

slowli:
## Overview

This PR structures exported Protobuf files (i.e., ones exported via `links = $fake_package` mechanism). Due to how the used code generator works, the same limitations still apply when generating Rust wrappers with `exonum-build` (all base names for Protobuf files should be different). However, this limitation is no longer pushed onto Protobuf declaration consumers.

Discussion:

- Should Protobuf files be structured elsewhere (e.g., in the node crate)?
- Should all files have the same value for `option java_package`?
- Is the chosen structure optimal? (E.g., it could be possible to move proofs from the core crate into `exonum.proof` package, together with list and map proofs, but I'm not sure what would be the best name of the file.)

---

See: https://jira.bf.local/browse/ECR-4187


dmitry-timofeev:
Regarding the question on the package structure, we can go into three routes to prevent an issue with split packages (esp. severe with Java Module System from Java 9) to occur downstream: 
1. Use the same package. It prevents us from distributing classes in separate modules(e.g., a light client might not need the dispatcher-specific messages, and use only "core"). 
2. Use the same _base_ package: "com.exonum.core.message" ->  "com.exonum.core.message.proofs", "....crypto". That allows both distributing them in one module "com.exonum.core.message" and, if we later decide to split into smaller modules, in separate ("com.exonum.core.message.crypto").
3. Use distinct base packages, e.g., "com.exonum.core.proofs[.message]", "....crypto[.message]". That requires putting them into separate modules from day one.

slowli:
@dmitry-timofeev What do you think the best option? Judging pure by your description, the second one looks most flexible, although I'm not sure about the specific package naming (more specifically, the "core" part in "com.exonum.core.message").

dmitry-timofeev:
I think the second is better (starting with a single module and splitting if necessary, e.g., in case the number of Exonum components/modules grows considerably). If "core" isn't needed, we can drop it.

PR #1758, opened 2020-01-31
FE cryptocurrency-advanced update to master

qvantor:


examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
```diff
@@ -4,200 +4,142 @@ import * as proto from '../../proto/stubs.js'
...
 const TX_WALLET_ID = 2
-const TABLE_INDEX = 0
 const Wallet = Exonum.newType(proto.exonum.examples.cryptocurrency_advanced.Wallet)
 
-function TransferTransaction(publicKey) {
-  return Exonum.newTransaction({
-    author: publicKey,
-    service_id: SERVICE_ID,
-    message_id: TX_TRANSFER_ID,
+function TransferTransaction () {
```
slowli: The transaction doesn't need to be created every time from scratch. Instead, you can define the type once in the module scope:

```javascript
const TransferTransaction = new Exonum.Transaction({
  serviceId: SERVICE_ID,
  methodId: TX_TRANSFER_ID,
  schema: proto.exonum.examples.cryptocurrency_advanced.Transfer
})
```

examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
```diff
@@ -4,200 +4,142 @@ import * as proto from '../../proto/stubs.js'
...
 }
 
-function getTransaction(transaction, publicKey) {
-  if (transaction.name) {
-    return new CreateTransaction(publicKey)
+function deserializeWalletTx (transaction) {
+  const txTypes = [TransferTransaction(), IssueTransaction()]
+  for (const tx of txTypes) {
+    const txData = tx.deserialize(Exonum.hexadecimalToUint8Array(transaction))
+    if (txData) return Object.assign({}, txData.payload, {
```
slowli: Nit: What's with this formatting? (Is this some new JS thing? 🙃) I'd suggest to use a block for `if` for it to read better.

examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
```diff
@@ -4,200 +4,142 @@ import * as proto from '../../proto/stubs.js'
...
+        const transferTx = TransferTransaction()
 
         // Transaction data
         const data = {
-          to: { data: Exonum.hexadecimalToUint8Array(receiver) },
+          to: { data: Exonum.hexadecimalToUint8Array(Exonum.publicKeyToAddress(receiver)) },
           amount: amountToTransfer,
           seed: seed
         }
+        console.log(data)
```
slowli: Please remove console output.

examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
```diff
@@ -4,200 +4,142 @@ import * as proto from '../../proto/stubs.js'
...
+        const transaction = transferTx.create(data, keyPair).serialize()
 
         // Send transaction into blockchain
-        return transaction.send(TRANSACTION_URL, data, keyPair.secretKey)
+        return Exonum.send(TRANSACTION_URL, transaction)
       },
 
-      getWallet(publicKey) {
-        return axios.get('/api/services/configuration/v1/configs/actual').then(response => {
+      getWallet (publicKey) {
```
slowli: Is `async` / `await` inapplicable here? Could significantly remove readability, IMO.

examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
```diff
@@ -4,200 +4,142 @@ import * as proto from '../../proto/stubs.js'
...
+            .then(({ block_proof, wallet_proof, wallet_history }) => {
+              if (Exonum.verifyBlock(block_proof, validators) !== undefined) throw new Error('Block is invalid')
+              const tableRootHash = Exonum.verifyTable(wallet_proof.to_table, block_proof.block.state_hash, 'crypto.wallets')
+              const walletProof = new Exonum.MapProof(wallet_proof.to_wallet, Exonum.MapProof.rawKey(Exonum.PublicKey), Wallet)
+              if (walletProof.merkleRoot !== tableRootHash) throw new Error('Wallet proof is corrupted')
+
+              const wallet = walletProof.entries.get(Exonum.publicKeyToAddress(publicKey))
+              if (typeof wallet === undefined) throw new Error('Wallet not found')
+
+              const transactions = wallet_history.transactions.map(deserializeWalletTx)
```
slowli: You've skipped some proof checks, e.g., that transaction hashes are appropriate. See exonum/exonum-doc#394 for how I see these checks.

examples/cryptocurrency-advanced/frontend/package.json
```diff
@@ -5,7 +5,7 @@
   "author": "Exonum Team <exonum@bitfury.com>",
   "license": "Apache-2.0",
   "scripts": {
-    "proto": "pbjs --keep-case -t static-module ../../../exonum/src/proto/schema/exonum/protocol.proto ../backend/src/proto/cryptocurrency.proto -o ./proto/stubs.js",
+    "proto": "pbjs --keep-case -t static-module node_modules/exonum-client/proto/**/*.proto ../backend/src/proto/service.proto -o ./proto/stubs.js",
```
slowli: I *think* you should use `--path` for includes (that is, `--path node_modules/exonum-client/proto`), at least this would be similar to `protoc` workflow. Can you try this approach?

examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
qvantor: The whole application on promises, for now, this why I stick this way. But in the future, this should be on async/await for the whole project - for sure.

examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
qvantor: I did :)

examples/cryptocurrency-advanced/frontend/package.json
qvantor: With `--path` it does not work as expected :(

ivan-ochc:
FE tests **'should transfer funds'** and **'should get wallet proof and verify it**' have been failed.

examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
```diff
@@ -4,201 +4,140 @@ import * as proto from '../../proto/stubs.js'
...
+              const verifiedTransactions = new Exonum.ListProof(wallet_history.proof, Exonum.Hash)
+              const hexHistoryHash = Exonum.uint8ArrayToHexadecimal(new Uint8Array(wallet.history_hash.data))
+              if (verifiedTransactions.merkleRoot !== hexHistoryHash) throw new Error('Transactions proof is corrupted')
+
+              const validIndexes = verifiedTransactions
+                .entries
+                .every(({ index }, i) => i === index)
+              if (!validIndexes) throw new Error('Invalid transaction indexes in the proof')
+
+              const transactions = wallet_history.transactions.map(deserializeWalletTx)
```
slowli: You still don't check that transactions have correct hashes 🙃 If I understand `deserializeWalletTx` correctly, the check should be something like

```javascript
const correctHashes = transactions.every(({ hash }, i) => verifiedTransactions[i] === hash)
if (!correctHashes) {
  throw new Error('Transaction hash mismatch')
}
```

examples/cryptocurrency-advanced/frontend/src/plugins/blockchain.js
```diff
@@ -4,201 +4,140 @@ import * as proto from '../../proto/stubs.js'
...
-                }
-
-                return {
-                  block: data.block_proof.block,
-                  wallet: wallet,
-                  transactions: transactions
-                }
-              })
+            .then(({ block_proof, wallet_proof, wallet_history }) => {
+              if (Exonum.verifyBlock(block_proof, validators) !== undefined) throw new Error('Block is invalid')
```
slowli: Nit / opinion: `if` with the body not separated into a block harms readability (especially after Rust, where blocks are mandatory).

Also, what's going on in this check? `verifyBlock` now throws on failure and returns `void` otherwise.

qvantor:
> FE tests **'should transfer funds'** and **'should get wallet proof and verify it**' have been failed.

fixed

PR #1760, opened 2020-01-31
Add a configuration to the cryptocurrency advanced service [ECR-4185]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-4185

### Definition of Done


examples/cryptocurrency-advanced/backend/src/config.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use super::proto;
+
+/// Cryptocurrency configuration parameters.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(ProtobufConvert, BinaryValue)]
+#[protobuf_convert(source = "proto::Config")]
+pub struct Config {
+    /// Initial balance of the wallet.
+    pub init_balance: i64,
```
slowli: Does using a signed integer only serve the purpose of getting an error during config verification? I'd suggest to use `u64` and check that `init_balance != 0`.

exonum/src/runtime/types.rs
```diff
@@ -275,6 +275,16 @@ impl Display for ArtifactId {
     }
 }
 
+impl Debug for ArtifactId {
```
slowli: Are this and the following `Debug` implementations motivated by the `non_exhaustive` marker? IMO, it's OK to output in `Debug`; it deals with the internal type structure anyway.

examples/cryptocurrency-advanced/backend/src/proto/service.proto
```diff
@@ -57,4 +57,7 @@ message Wallet {
 }
 
 // Cryptocurrency configuration.
-message Config {}
+message Config {
+  // Initial balance of the wallet
```
slowli: Nit: I'd rephrase to "Initial balance of newly created wallets."

examples/cryptocurrency-advanced/backend/src/config.rs
```diff
@@ -0,0 +1,47 @@
...
+            Ok(())
+        }
+    }
+}
+
+/// The enumeration represents errors that can occurred while validating config.
+#[derive(Debug, ExecutionFail)]
+pub enum ConfigError {
+    /// Initial balance is negative.
+    NegativeBalance = 1,
```
slowli: The code of this error will clash with the `Error::SenderNotFound` variant defined in the `transactions` module. It could make sense to start enumeration here from 16 or 32 (and, ideally, explain in a comment why such separation is beneficial for forward compatibility).

examples/cryptocurrency-advanced/backend/tests/api.rs
```diff
@@ -264,6 +268,55 @@ fn test_unknown_wallet_request() {
...
+
+    // Check that the user indeed is persisted by the service.
+    let wallet = api.get_wallet(tx.author()).unwrap();
+    assert_eq!(wallet.owner, author_address(&tx));
+    assert_eq!(wallet.name, ALICE_NAME);
+    assert_eq!(wallet.balance, 100);
+
+    // Change an initial balance to 200 via changing configuration.
+    let tx = api.change_config(&mut testkit, Config { init_balance: 200 });
+    testkit.create_block();
```
popzxc: Wouldn't it be simpler to use an existing API designed for this purpose:

```rust
let block = testkit.create_block();

assert!(block[tx.object_hash()].status().is_ok())
```

?

examples/cryptocurrency-advanced/backend/src/config.rs
popzxc: Note that such a separation is already introduced in [supervisor](https://github.com/exonum/exonum/blob/master/services/supervisor/src/errors.rs).

examples/cryptocurrency-advanced/backend/tests/api.rs
slowli: To be fair, these are HTTP API tests, and `assert_tx_status` method is already used throughout the code.

examples/cryptocurrency-advanced/backend/src/lib.rs
```diff
@@ -23,35 +23,62 @@ extern crate serde_derive; // Required for Protobuf.
...
     fn initialize(
         &self,
         context: ExecutionContext<'_>,
-        _params: Vec<u8>,
+        params: Vec<u8>,
     ) -> Result<(), ExecutionError> {
-        // Initialize indexes. Not doing this may lead to errors in HTTP API, since it relies on
-        // `wallets` indexes being initialized for returning corresponding proofs.
-        SchemaImpl::new(context.service_data());
+        let config = Config::from_bytes(params.into()).map_err(CommonError::malformed_arguments)?;
```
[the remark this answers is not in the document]
slowli: > AFAIK this `into` conversion will clone the vector data

Hm... Shouldn't the conversion consume `params` (i.e., simply wrap `Cow::Owned` variant around them)? Seems to be the case as per [this implementation](https://doc.rust-lang.org/src/alloc/vec.rs.html#2429-2433).

exonum/src/runtime/types.rs
aleksuss: But it appears in the log. This is a reason why i’ve reimplement Debug.

examples/cryptocurrency-advanced/backend/src/config.rs
aleksuss: Exactly

examples/cryptocurrency-advanced/backend/src/lib.rs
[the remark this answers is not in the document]
popzxc: Ah, indeed. Sorry, my bad.

examples/cryptocurrency-advanced/backend/tests/api.rs
popzxc: Then I think the approach should be unified, since suggested approach is already used in this PR: https://github.com/exonum/exonum/pull/1760/files#diff-480f4bde234f6ab105cb6acf68052b6aR315

exonum/src/runtime/types.rs
popzxc: I think the problem here is that log entry uses debug representation (which is misuse). Log entries should use `Display`.

slowli:
Another thing to note is that introducing config into the existing service will make the existing [advanced cryptocurrency tutorial](https://exonum.com/doc/version/latest/get-started/data-proofs/) clunky because it currently describes the initial wallet balance as a constant. I think the way around this would be place the version with config separately (I think it would be possible to reuse this newer version for the upcoming data migration tutorial). To minimize copy-paste, the reusable parts of the old service could be made public (with the comment explaining why) and would be used in the newer version.

examples/cryptocurrency-advanced/backend/src/schema.rs
```diff
@@ -36,6 +36,8 @@ pub(crate) struct SchemaImpl<T: Access> {
     pub public: Schema<T>,
     /// History for specific wallets.
     pub wallet_history: Group<T, Address, ProofListIndex<T::Base, Hash>>,
+    /// Configuration of the service.
+    pub config: Entry<T::Base, Config>,
```
slowli, line 40: Is config present after a data migration? Seemingly not, which may lead to panics during wallet creation.

aleksuss: The configuration is being available only after migration to 0.2.0 version. Simple version of the cryptocurrency doesn't have a configuration.

PR #1761, opened 2020-02-01
Use `KeyPair` in public APIs [ECR-4116]

slowli:
## Overview

Since we have time for breaking changes, here goes. This PR replaces uses of `(PublicKey, SecretKey)` with `KeyPair` (or `impl Into<KeyPair>`, which is backward-compatible) in public APIs. `KeyPair` now has constructors, making its use simpler.

Not done (can be done in separate PR):

- Replace usage of `crypto::gen_keypair()` with `KeyPair::random()` throughout the codebase.

---

See: https://jira.bf.local/browse/ECR-4116


components/crypto/src/lib.rs
```diff
@@ -461,13 +482,22 @@ implement_index_traits! {SecretKey}
 implement_index_traits! {Seed}
 implement_index_traits! {Signature}
 
+/// Pair of matching secret and public keys.
+///
+/// Prefer using this struct to `(PublicKey, SecretKey)`, since it asserts in compile time
```
popzxc: In compile time? How?

slowli: Perhaps not the best phrasing; what I meant is that public and secret keys in a `KeyPair` instance are guaranteed to match. Indeed, the `from_keys` constructor asserts this, and other constructors guarantee this by design.

popzxc: I'd remove "at compile time" part: code with an incorrect keypair will compile, but will panic in the runtime because of the assertion failure.

components/keys/src/lib.rs
```diff
@@ -38,30 +83,43 @@ fn validate_file_mode(mode: u32) -> Result<(), Error> {
...
+    /// by the crypto backend.
+    pub fn random() -> Self {
+        Self {
+            consensus: KeyPair::random(),
+            service: KeyPair::random(),
+            non_exhaustive: (),
+        }
+    }
+
+    /// Create validator keys from the provided keypairs.
```
aleksuss:
```suggestion
    /// Creates validator keys from the provided keypairs.
```

PR #1762, opened 2020-02-02
Unified iterator types [ECR-4192]

slowli:
## Overview
 
The season of last-minute breaking changes continues! This time, it's unification of iterator types for all MerkleDB indexes. This is (hopefully) better for bindings (they can now deal with a single iterator type extracted via a unified interface for all indexes), and simplifies support.

There are also some moderately unrelated changes:

- `KeySetIndex::insert` now takes the element by reference.
- `KeySetIndex` now supports unsized keys and does not have `K: Borrow<Q>` shenanigans going on, similar to `MapIndex` / `ProofMapIndex`
- Maximum height in `ProofListIndex` was fixed from the bogus value 58 to 56 (needs more detailed testing; I'm not sure this is fixed *everywhere*, and that there are no off-by-one errors)
- Bogus setting of the empty key was removed for `ListIndex`. (This code is seemingly an artifact from an era where indexes did not have metadata.)

Discussion:

- Does this really help bindings?
- Does iteration with `Key = ()`  need to be implemented for `Entry` / `ProofEntry` for uniformity?

------

See: https://jira.bf.local/browse/ECR-4192

### DIFF
--- a/components/merkledb/src/indexes/iter.rs
+++ b/components/merkledb/src/indexes/iter.rs
@@ -1,0 +1,109 @@
+// Copyright 2020 The Exonum Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Generic iterator types used by all indexes.
+
+use crate::{
+    views::{Iter, RawAccess, View},
+    BinaryKey, BinaryValue,
+};
+
+/// FIXME
+#[derive(Debug)]
+pub struct Entries<'a, K: ?Sized, V> {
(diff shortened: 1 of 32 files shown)


components/merkledb/src/indexes/value_set.rs
```diff
@@ -134,7 +107,8 @@ where
         self.base.contains(hash)
     }
 
-    /// Returns an iterator visiting all elements in arbitrary order. The iterator element type is V.
+    /// Returns an iterator over set elements and their hashes. The elements are ordered as per
+    /// lexicorgraphic ordering of their hashes (i.e., effectively randomly).
```
aleksuss:
```suggestion
    /// lexicographic ordering of their hashes (i.e., effectively randomly).
```

components/merkledb/src/indexes/iter.rs
```diff
@@ -0,0 +1,158 @@
...
+};
+
+/// Iterator over key-value pairs of an index.
+///
+/// This structure is returned by the [`IndexIterator`] trait and by inherent methods
+/// of some indexes.
+///
+/// [`IndexIterator`]: trait.IndexIterator.html
+#[derive(Debug)]
+pub struct Entries<'a, K: ?Sized, V> {
```
aleksuss: The plural is a little bit confusing here 🤔

slowli: But this is how iterators are usually named in the standard Rust library. Although, in this particular case, `Iter` is another option.

PR #1763, opened 2020-02-03
Readonly fork with static lifetime [ECR-4196]

slowli:
## Overview

This PR implements a version of `ReadonlyFork` with static lifetime. This version is then added to `GenericRawAccess` and used to implement `AsReadonly` for `GenericRawAccess`.

Discussion:

- Is `ReadonlyRcFork` a good name?

---

See: https://jira.bf.local/browse/ECR-4196

### DIFF
--- a/components/merkledb/src/db.rs
+++ b/components/merkledb/src/db.rs
@@ -920,6 +920,61 @@ impl<'a> RawAccess for ReadonlyFork<'a> {
     }
 }
 
+/// Version of `ReadonlyFork` with a static lifetime. Can be produced from an `Rc<Fork>` using
+/// the `AsReadonly` trait.
+///
+/// Beware that producing an instance increases the reference counter of the unrelying fork.
+/// If you need to obtain `Fork` from `Rc<Fork>` via [`Rc::try_unwrap`], make sure that all
+/// `ReadonlyRcFork` instances are dropped by this time.
+///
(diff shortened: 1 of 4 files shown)


components/merkledb/src/db.rs
```diff
@@ -920,6 +920,61 @@ impl<'a> RawAccess for ReadonlyFork<'a> {
...
+/// # use std::rc::Rc;
+/// let db = TemporaryDB::new();
+/// let fork = Rc::new(db.fork());
+/// fork.get_proof_list("list").extend(vec![1_u32, 2, 3]);
+/// let ro_fork: ReadonlyRcFork = fork.as_readonly();
+/// let list = ro_fork.get_proof_list::<_, u32>("list");
+/// assert_eq!(list.len(), 3);
+/// ```
+#[derive(Debug, Clone)]
+pub struct ReadonlyRcFork(Rc<Fork>);
```
popzxc: IMHO `Rc` in the middle looks a bit weird. I'd suggest `ReadonlyForkRc` (assuming that usually `Rc` is treated like a name, and not expanded mentally to "reference-counting").

slowli: Well, I arrived to `ReadonlyRcFork` by mentally expanding `Rc`: "readonly [version of] reference-counted fork" 🙃 Maybe, something like `OwnedReadonlyFork` would make sense?

popzxc: Yep, I think so.

PR #1765, opened 2020-02-03
Support instantiating built-in services in `exonum-cli` [ECR-4141]

alekseysidorov:


cli/src/lib.rs
```diff
@@ -158,14 +175,20 @@ impl NodeBuilder {
...
             let supervisor = Self::supervisor_service(&run_config);
-            let genesis_config = GenesisConfigBuilder::with_consensus_config(
-                run_config.node_config.public_config.consensus.clone(),
-            )
-            .with_artifact(Supervisor.artifact_id())
-            .with_instance(supervisor)
-            .with_artifact(ExplorerFactory.artifact_id())
-            .with_instance(ExplorerFactory.default_instance())
-            .build();
+            let genesis_config = {
```
popzxc: IMHO this block worth moving into a separate method.

examples/cryptocurrency-advanced/backend/src/main.rs
```diff
@@ -14,10 +14,18 @@
...
 fn main() -> Result<(), failure::Error> {
     exonum::helpers::init_logger().unwrap();
     NodeBuilder::new()
         .with_service(cryptocurrency::CryptocurrencyService)
+        // Starts cryptocurrency instance with given id and name
+        // immediately after genesis block creation.
+        .with_default_instance(
+            cryptocurrency::CryptocurrencyService
+                .artifact_id()
+                .into_default_instance(101, "cryptocurrency"),
```
popzxc: I'd suggest to create a constant in the library, which will be used here.

cli/src/lib.rs
```diff
@@ -149,6 +151,21 @@ impl NodeBuilder {
...
+    ///
+    /// Make sure that artifact for this instance is provided to this builder.
+    pub fn with_default_instance(mut self, instance: impl Into<InstanceInitParams>) -> Self {
+        self.default_instances.push(instance.into());
+        self
+    }
+
+    /// Adds a default Rust service instance that will be available immediately after creating a
+    /// genesis block.
+    pub fn with_default_service<S: DefaultInstance>(self, service: S) -> Self {
```
popzxc: Maybe `service: impl DefaultInstance`?

cli/src/lib.rs
```diff
@@ -149,23 +151,29 @@ impl NodeBuilder {
...
+    ///
+    /// Make sure that artifact for this instance is provided to this builder.
+    pub fn with_default_instance(mut self, instance: impl Into<InstanceInitParams>) -> Self {
+        self.default_instances.push(instance.into());
+        self
+    }
+
+    /// Adds a default Rust service instance that will be available immediately after creating a
+    /// genesis block.
+    pub fn with_default_service(self, service: impl DefaultInstance) -> Self {
```
slowli: I'd suggest to use `with_default_rust_service` for uniformity.

examples/cryptocurrency-advanced/backend/src/main.rs
```diff
@@ -19,5 +19,8 @@ fn main() -> Result<(), failure::Error> {
     exonum::helpers::init_logger().unwrap();
     NodeBuilder::new()
         .with_service(cryptocurrency::CryptocurrencyService)
```
slowli: Nit: Could make sense to import `CryptocurrencyService` directly.

cli/src/lib.rs
alekseysidorov: To uniformity we should also rename `with_service` to `with_rust_service` but this is breaking change.

cli/src/lib.rs
```diff
@@ -132,11 +133,12 @@ impl NodeBuilder {
                 .with_factory(Supervisor)
                 .with_factory(ExplorerFactory),
             external_runtimes: vec![],
+            default_instances: vec![],
```
slowli: I don't think `default_instances` is a good name; we call the services started in the genesis block *built-in* services. `DefaultInstance` does not necessarily mean that its parameters will be used in the genesis block.

cli/src/lib.rs
```diff
@@ -149,23 +151,29 @@ impl NodeBuilder {
         self
     }
 
+    /// Adds a service instance that will be available immediately after creating a genesis block.
+    ///
+    /// Make sure that artifact for this instance is provided to this builder.
+    pub fn with_default_instance(mut self, instance: impl Into<InstanceInitParams>) -> Self {
```
slowli: I'd call this method `with_builtin_instance` (or `with_instance`, to be consistent with `TestKitBuilder`); see the reasoning above.

cli/src/lib.rs
```diff
@@ -149,23 +151,29 @@ impl NodeBuilder {
         self
     }
 
+    /// Adds a service instance that will be available immediately after creating a genesis block.
+    ///
+    /// Make sure that artifact for this instance is provided to this builder.
```
slowli: Is this comment outdated? As far as I see, the artifact is added automatically in the `genesis_config` method. Or does the comment refer to the Rust service factory? In this case, I'd rephrase as

> For Rust services, the service factory needs to be supplied via `with_rust_service` separately.

cli/src/lib.rs
```diff
@@ -149,23 +151,29 @@ impl NodeBuilder {
...
+    pub fn with_default_instance(mut self, instance: impl Into<InstanceInitParams>) -> Self {
+        self.default_instances.push(instance.into());
+        self
+    }
+
+    /// Adds a default Rust service instance that will be available immediately after creating a
+    /// genesis block.
+    pub fn with_default_rust_service(self, service: impl DefaultInstance) -> Self {
```
slowli: I'd suggest to add a couple of tests to check that the builder performs decoupling as necessary; e.g., if `with_default_rust_service` and `with_default_instance` are called for the same service factory.

CHANGELOG.md
```diff
@@ -135,6 +135,8 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - `supervisor-mode` parameter has been added for `generate-template` subcommand.
   (#1598)
 
+- `with_service` has been renamed to the `with_rust_service`. (#1765)
```
slowli: Nit: I'd qualify `NodeBuilder::with_service`.

CHANGELOG.md
```diff
@@ -271,6 +273,9 @@ Indexes iterators names has been shortened to `Iter`, `Keys` and `Values`. (#162
 
 - Added maintenance command `restart-migration` to restart migration script. (#1728)
 
+- Added `with_default_instance` and `with_default_service` to start service
```
slowli: Nit: I'd qualify "Added `with_default_instance` and `with_default_service` to `NodeBuilder` in order to start service instances..."

PR #1768, opened 2020-02-05
Check index / height in `ProofListIndex` [ECR-4194]

slowli:
## Overview

This PR checks index / height "overflow" in `ListProof` and `ProofListIndex` methods. For example, `list.get(1 << 60)` previously panicked in the debug mode and returned garbage in the release mode; now it properly returns `None`.

Discussion:

- Is it appropriate to panic if the list length exceeds `1 << 56`? Is the error message output in this case good enough?

---

See: https://jira.bf.local/browse/ECR-4194


slowli:
@popzxc I agree that the triggering the assertion isn't realistic, but I'd argue that the message is harmless; if a user somehow triggers it, it could be useful.

PR #1769, opened 2020-02-05
Exclude itself from the connection list [ECR-4190]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-4190

### Definition of Done

### DIFF
--- a/cli/src/command/finalize.rs
+++ b/cli/src/command/finalize.rs
@@ -116,7 +116,7 @@ impl Finalize {
 
     fn create_connect_list_config(
         public_configs: &[NodePublicConfig],
-        skipped_key: &PublicKey,
+        key_to_skip: &PublicKey,
     ) -> ConnectListConfig {
         let peers = public_configs
             .iter()
@@ -124,7 +124,7 @@ impl Finalize {
                 let public_key = Self::get_consensus_key(config).unwrap();
                 // `skipped_key` is a consensus key of the current node. We don't need
                 // to include `ConnectInfo` with this key in the connect list.
-                if public_key != *skipped_key {
+                if public_key != *key_to_skip {
                     Some(ConnectInfo {
                         public_key,
                         address: config.address.clone().unwrap(),
--- a/cli/src/command/finalize.rs
+++ b/cli/src/command/finalize.rs
@@ -171,10 +171,8 @@ impl ExonumCommand for Finalize {
             .collect();
         let consensus = common.consensus.with_validator_keys(validator_keys);
 
-        let connect_list = Self::create_connect_list_config(
-            &public_configs,
-            &private_config.consensus_key.unwrap(),
-        );
+        let connect_list =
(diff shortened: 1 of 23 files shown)


cli/src/command/finalize.rs
```diff
@@ -170,6 +184,7 @@ impl ExonumCommand for Finalize {
                 ..private_config.api
             },
             connect_list,
+            consensus_key: None, // We use it in creation connect list. We don't need it anymore.
```
vitvakatu: Is there a good reason _not_ to store it in the config (i.e. get rid of `Option`)?

cli/src/command/finalize.rs
```diff
@@ -114,13 +114,24 @@ impl Finalize {
             .consensus_key)
     }
 
-    fn create_connect_list_config(public_configs: &[NodePublicConfig]) -> ConnectListConfig {
-        // FIXME: filter self (ECR-4190).
+    fn create_connect_list_config(
+        public_configs: &[NodePublicConfig],
+        skipped_key: &PublicKey,
```
vitvakatu: nit: `skip_key` is more a bit more logical variable name IMO

cli/src/command/finalize.rs
vitvakatu: Also you can consider renaming it into `consensus_public_key`, as it is not clear when reading configs.

aleksuss: The reason not to store is because of data duplication. We already have consensus key in *validator_keys* array.

vitvakatu: Is it bad? Right now we do not have a way to distinguish which private config corresponds to a public config.

cli/src/command/finalize.rs
slowli: @vitvakatu `skip_key` is a verbal phrase though; IMO, noun phrases are more appropriate for variable names.

cli/src/command/finalize.rs
slowli: I agree with @vitvakatu that `consensus_public_key` would be a better name. Regarding duplication, I don't see it as a significant problem; for example, a node may be an auditor, in which case `consensus_public_key` is not duplicated. (As far as I understand, `exonum-cli` does not have first-class auditor support during blockchain creation, but this is a subject of future improvements.)

cli/src/command/finalize.rs
vitvakatu: `skipped_key` means "the key that was skipped", while the actual meaning is "the key that is needed to be skipped". As `key_to_skip` is a bit long, I suggested to use `skip_key`, which is used nicely in a function call like `create_connect_list_config(..., skip_key: key)`. However, I'm not gonna to argue about variables names, it was a nit.

popzxc: IMHO `key_to_skip` is just fine.

PR #1770, opened 2020-02-06
Fix integration tests according to changes in protobuf structure

aleksuss:
## Overview

---

### Definition of Done

### DIFF
--- a/test-suite/testkit/server/package.json
+++ b/test-suite/testkit/server/package.json
@@ -25,7 +25,7 @@
   "homepage": "https://github.com/exonum/exonum-testkit/tree/master/examples/server#readme",
   "dependencies": {
     "chai": "^4.2.0",
-    "exonum-client": "^0.18.1",
+    "exonum-client": "https://github.com/exonum/exonum-client#master",
     "mocha": "^6.2.0",
     "node-fetch": "^2.6.0",
     "protobufjs": "^6.8.8",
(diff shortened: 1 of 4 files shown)


test-suite/testkit/server/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "private": true,
   "scripts": {
-    "proto": "pbjs -t static-module -r test ../../../examples/cryptocurrency/src/proto/service.proto ../../../components/crypto/src/proto/schema/types.proto -o ./test/stubs.js",
+    "proto": "pbjs -t static-module -r test ../../../examples/cryptocurrency/src/proto/service.proto ../../../components/crypto/src/proto/schema/exonum/crypto/types.proto -o ./test/stubs.js",
```
slowli: This is completely not your fault, but this command is arguably bogus. It should be changed to something like

```
pbjs -t static-module
  -r test
  -p node_modules/exonum-client/proto
  ../../../examples/cryptocurrency/src/proto/service.proto
  -o test/stubs.js
```

(Line breaks are inserted for readability.) The `-p` option specifies an include path (which, in this case, is a path to files included in the `exonum-client` package).

aleksuss: At least it works locally. I'll check the command line locally and change it in the release commit.

slowli: Yeah, it works, but the current command is more brittle. The version with an include path will continue working if `service.proto` requires more types from core, or if they are rearranged within the core.

aleksuss: But I'll fix it in the upcoming release commit to avoid one more CI flow.

PR #1775, opened 2020-02-09
Implementing BinaryValue by newtype

integrer:
## Overview

In this PR proposed attempt to implement BinaryValue trait across generic structure with
single member of this type.

Comments from  #1720 were taken to account.

---

### Definition of Done


components/merkledb/src/values.rs
```diff
@@ -261,6 +261,42 @@ impl BinaryValue for [u8; HASH_SIZE] {
     }
 }
 
+/// Concatenates slice of byte buffers prepending size of buffer before each one
+pub fn concat_buffers(buffers: &mut [Vec<u8>]) -> Vec<u8> {
+    use std::convert::TryFrom;
+
+    let mut res_buf = Vec::new();
+    for mut buf in buffers {
+        let size = u32::try_from(buf.len()).unwrap_or_else(|_| panic!("Too large buffer"));
```
aleksuss: Is it really a good idea to panic in this case ?

cli/src/command/generate_config.rs
```diff
@@ -95,7 +95,7 @@ impl GenerateConfig {
     }
 
     /// If no port is provided by user, uses `DEFAULT_EXONUM_LISTEN_PORT`.
-    fn parse_external_address(input: &str) -> Result<SocketAddr, Error> {
+    pub fn parse_external_address(input: &str) -> Result<SocketAddr, Error> {
```
aleksuss: Why `pub` ?

components/merkledb/src/macros.rs
```diff
@@ -152,3 +152,47 @@ macro_rules! impl_serde_hex_for_binary_value {
         }
     };
 }
+
+/// Implements `BinaryValue` traint for tuples cross single member generic tuple struct.
```
aleksuss: Typo: `trait`.

components/merkledb/src/macros.rs
```diff
@@ -152,3 +152,47 @@ macro_rules! impl_serde_hex_for_binary_value {
...
+/// If expression begins with for keyword it creates implementation for existend structure
```
aleksuss: Typo: `existed`

components/merkledb/src/values.rs
```diff
@@ -261,6 +261,42 @@ impl BinaryValue for [u8; HASH_SIZE] {
...
+    let mut pos = 0;
+    let mut result = Vec::with_capacity(qty);
+    for _ in 0..qty {
+        ensure!(
+            bytes.len() >= pos + size_of::<u32>(),
+            "Unexpected end of buffer"
+        );
+        let len = u32::from_bytes(bytes[pos..pos + size_of::<u32>()].into())? as usize;
+        pos += size_of::<u32>();
+        ensure!(bytes.len() >= pos + len, "Unexpected end of buffer");
```
aleksuss:
```suggestion
        ensure!(bytes.len() >= pos + len, "Unexpected end of the buffer");
```

components/merkledb/src/macros.rs
vitvakatu: Isn't it `existent`?

components/merkledb/src/macros.rs
```diff
@@ -152,3 +152,47 @@ macro_rules! impl_serde_hex_for_binary_value {
...
+/// supposing it meets requirements, mentioned above. Otherwise it creates new structure
+/// according to supplied meta attributes, visibility, name and type constraint for generic type.
+/// Target tuple types must be listed in curly braces.
```
vitvakatu: An example would be much better than a documentation. Also, consider the following fixes in grammar:

```
If an expression begins with `for` keyword, it creates an implementation for existent structure 
that meets the requirements mentioned above. Otherwise it creates a new structure 
according to supplied meta attributes, visibility, name and type constraints for a generic type.
```

components/merkledb/src/macros.rs
vitvakatu: What means `cross single member generic tuple struct`?

components/merkledb/src/macros.rs
```diff
@@ -152,3 +152,47 @@ macro_rules! impl_serde_hex_for_binary_value {
...
+#[macro_export]
+macro_rules! binary_value_tuple_impls {
+    () => ();
+    ($(#[$attr:meta])* $vis:vis $name:ident {$(($t1:ty$(, $t2:ty$(, $t3:ty$(, $t4:ty$(, $t5:ty$(, $t6:ty$(, $t7:ty$(, $t8:ty$(, $t9:ty$(, $t10:ty$(, $t11:ty$(, $t12:ty)?)?)?)?)?)?)?)?)?)?)?$(,)?)),+$(,)?} $($rest:tt)*) => {
```
vitvakatu: The docs doesn't mention that it supports only up to 12 elements

components/merkledb/src/values.rs
vitvakatu: I'd assert instead. Panic message isn't much self-describing

components/merkledb/src/values.rs
```diff
@@ -261,6 +261,42 @@ impl BinaryValue for [u8; HASH_SIZE] {
     }
 }
 
+/// Concatenates slice of byte buffers prepending size of buffer before each one
```
vitvakatu:
```suggestion
/// Concatenates buffer slices, prepends each slice with its size.
```

cli/src/command/generate_config.rs
integrer: I used this method in my own project for parsing command. This modification mistaken appears in this branch.

components/merkledb/src/values.rs
```diff
@@ -261,6 +261,42 @@ impl BinaryValue for [u8; HASH_SIZE] {
...
+        let size = u32::try_from(buf.len()).unwrap_or_else(|_| panic!("Too large buffer"));
+        res_buf.append(&mut size.to_bytes());
+        res_buf.append(&mut buf);
+    }
+    res_buf
+}
+
+/// Splits single byte buffer into vector of separate buffers according to its size
+pub fn split_buffer_into_sized_parts<'a>(
+    bytes: &'a Cow<[u8]>,
```
aleksuss: clippy suggests to pass `&[u8]` instead of `&'a Cow<[u8]>`

components/merkledb/src/macros.rs
```diff
@@ -152,3 +152,47 @@ macro_rules! impl_serde_hex_for_binary_value {
...
     };
 }
+
+/// Implements `BinaryValue` traint for tuples cross single member generic tuple struct.
+/// If expression begins with for keyword it creates implementation for existend structure
+/// supposing it meets requirements, mentioned above. Otherwise it creates new structure
+/// according to supplied meta attributes, visibility, name and type constraint for generic type.
+/// Target tuple types must be listed in curly braces.
+#[macro_export]
+macro_rules! binary_value_tuple_impls {
```
slowli: I don't understand why this macro needs to be public, or why it needs to be defined in the `merkledb` crate. See the overall comment for more details.

components/merkledb/src/values.rs
```diff
@@ -261,6 +261,42 @@ impl BinaryValue for [u8; HASH_SIZE] {
...
 
+/// Concatenates buffer slices, prepends each slice with its size.
+pub fn concat_buffers(buffers: &mut [Vec<u8>]) -> Vec<u8> {
+    use std::convert::TryFrom;
+
+    let mut res_buf = Vec::new();
+    for mut buf in buffers {
+        let size = u32::try_from(buf.len()).unwrap_or_else(|_| panic!("Too large buffer"));
+        res_buf.append(&mut size.to_bytes());
+        res_buf.append(&mut buf);
```
slowli: After doing some local benchmarks, using `append` appears ~2x slower than using `extend_from_slice`. Please use `extend_from_slice`. Also, it will allow to use `size.to_le_bytes()` above (which, unlike `to_bytes`, doesn't allocate on heap). Pre-allocating with `Vec::with_capacity` leads to ~30% performance boost, so I'd strongly suggest to use it as well.

components/merkledb/src/values.rs
```diff
@@ -261,6 +261,42 @@ impl BinaryValue for [u8; HASH_SIZE] {
...
+    }
+    res_buf
+}
+
+/// Splits single byte buffer into vector of separate buffers according to its size
+pub fn split_buffer_into_sized_parts<'a>(
+    bytes: &'a Cow<[u8]>,
+    qty: usize,
+) -> Result<Vec<Cow<'a, [u8]>>, failure::Error> {
```
slowli: Doesn't this method always return `Cow::Borrowed` slices? I'd suggest to change the return type to `Vec<&[u8]>`.

integrer:
@slowli I think it is good approach. Actually I misunderstood you last time. I will make effort for implement it. 
P.S. What do you think is it feature actually useful?

slowli:
@integrer 

> P.S. What do you think is it feature actually useful?

Yeah, I think so. The feature makes sense in the cases where the stored tuple cannot *a priori* evolve over time (e.g., due to semantics or during testing). If the stored type can evolve, Protobuf encoding could be a better alternative because it provides backward / forward compatibility in case of adding new fields.

components/merkledb/src/values.rs
```diff
@@ -386,4 +422,58 @@ mod tests {
         let values = [[1; HASH_SIZE]];
         assert_round_trip_eq(&values);
     }
+
+    binary_value_tuple_impls! {
+        #[derive(Debug, PartialEq)]
+        TupleContainer {
```
alekseysidorov: I suppose that we should keep rust syntax inside the macroses.

PR #1783, opened 2020-02-14
Added an utility for key generation from seed [ECR-4251]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-4251

### Definition of Done

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -146,8 +146,9 @@ jobs:
     - cargo run -p exonum-testkit --example timestamping
     - cargo run -p exonum-supervisor --example configuration_change
     - cargo run -p exonum-time --example simple_service
+    - cargo run -p exonum-keys --example key-generator -- -p passphrase -s a7839ea524f38d0e91a5ec96a723092719dc8a5b8a75f9131d9eb38f45e76344
 
-  # Integration tests.
+    # Integration tests.
   - name: integration-tests
     install:
     - python3 -m venv .venv
--- a/.travis.yml
+++ b/.travis.yml
@@ -148,7 +148,7 @@ jobs:
     - cargo run -p exonum-time --example simple_service
     - cargo run -p exonum-keys --example key-generator -- -p passphrase -s a7839ea524f38d0e91a5ec96a723092719dc8a5b8a75f9131d9eb38f45e76344
 
-    # Integration tests.
+  # Integration tests.
   - name: integration-tests
     install:
     - python3 -m venv .venv
--- a/components/keys/Cargo.toml
+++ b/components/keys/Cargo.toml
@@ -32,3 +32,4 @@ tempdir = "0.3.7"
 [[example]]
 name = "key-generator"
 path = "examples/key_generator.rs"
+test = true
--- a/components/keys/Cargo.toml
+++ b/components/keys/Cargo.toml
@@ -17,8 +17,10 @@ rand = "0.7"
 failure = "0.1.5"
 pwbox = "0.2.1"
 secret-tree = "0.2.0"
-serde = "1.0.10"
-serde_derive = "1.0.64"
+serde = "1.0"
+serde_json = "1.0"
+serde_derive = "1.0"
+structopt = "0.3"
 toml = "0.5.0"
 
 exonum-crypto = { version = "1.0.0-rc.1", path = "../crypto" }
@@ -26,3 +28,7 @@ exonum-crypto = { version = "1.0.0-rc.1", path = "../crypto" }
 [dev-dependencies]
 tempdir = "0.3.7"
 hex = "=0.4.0" # 0.4.1 is not compatible with Rust 1.36
+
+[[example]]
+name = "key-generator"
+path = "exmples/key_generator.rs"
--- a/components/keys/Cargo.toml
+++ b/components/keys/Cargo.toml
@@ -18,7 +18,6 @@ failure = "0.1.5"
 pwbox = "0.2.1"
 secret-tree = "0.2.0"
 serde = "1.0"
-serde_json = "1.0"
 serde_derive = "1.0"
 structopt = "0.3"
 toml = "0.5.0"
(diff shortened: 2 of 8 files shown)


components/keys/Cargo.toml
```diff
@@ -17,12 +17,18 @@ rand = "0.7"
...
+structopt = "0.3"
 toml = "0.5.0"
 
 exonum-crypto = { version = "1.0.0-rc.1", path = "../crypto" }
 
 [dev-dependencies]
 tempdir = "0.3.7"
 hex = "=0.4.0" # 0.4.1 is not compatible with Rust 1.36
+
+[[example]]
```
slowli: Nit: Aren't examples added automatically by Cargo?

aleksuss: Yes, they are. But I just prefer binary files with a dash as a splitter rather than an underscore 🙃

components/keys/exmples/key_generator.rs
```diff
@@ -0,0 +1,102 @@
...
+    let tempdir = TempDir::new("_prefix").map_err(|e| format_err!("{}", e))?;
+    let file_path = tempdir.path().join("master_key_file");
+    let keys = generate_keys_from_seed(&file_path, passphrase.as_bytes(), &seed)?;
+    let file_content =
+        read_to_string(file_path).expect("Couldn't read a content from the master key file");
+
+    let result = json!({
+        "consensus_pub_key": keys.consensus.public_key().to_hex(),
+        "service_pub_key": keys.service.public_key().to_hex(),
+        "master_key_file": file_content
```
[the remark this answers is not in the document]
aleksuss: It's just because I imagine it like a JSON object in which comma is unacceptable in such places. Ok. I'll fix it.

examples/timestamping/backend/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "exonum-timestamping"
-version = "0.0.0"
+version = "1.0.0-rc.1"
```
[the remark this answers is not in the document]
aleksuss: Because we use it as an artefact which has a version as well. There was an error when I've tried to deploy it with version `0.0.0`. By the way I even don't know if it's ok. Maybe it's a bug? What do you think ?

components/keys/Cargo.toml
aleksuss: And one more reason for adding this section. Without this section there is no possibility to run tests from the example. We should add  a `test=true` explicitly in this section to say to the cargo that we want to run tests from the example.

components/keys/exmples/key_generator.rs
```diff
@@ -0,0 +1,102 @@
...
+fn main() {
+    let args = Arguments::from_args();
+    let result = generate_json(&args.passphrase, &args.seed).unwrap();
+
+    println!("{}", result);
+}
+
+fn generate_json(passphrase: &str, seed: &str) -> Result<String, failure::Error> {
+    let seed = decode(seed).map_err(|e| format_err!("{}", e))?;
+    let tempdir = TempDir::new("_prefix").map_err(|e| format_err!("{}", e))?;
```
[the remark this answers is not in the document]
aleksuss: Yeah. Definitely makes sense. 👍🏻

components/keys/Cargo.toml
slowli: Ah, OK; these are good reasons to have the example declared explicitly.

examples/timestamping/backend/Cargo.toml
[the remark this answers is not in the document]
slowli: Artifact deployment is a good reason to have a "real" version, but not being able to deploy an artifact with version 0.0.0 looks strange and may be worth investigating.

.travis.yml
```diff
@@ -146,8 +146,9 @@ jobs:
     - cargo run -p exonum-testkit --example timestamping
     - cargo run -p exonum-supervisor --example configuration_change
     - cargo run -p exonum-time --example simple_service
+    - cargo run -p exonum-keys --example key-generator -- -p passphrase -s a7839ea524f38d0e91a5ec96a723092719dc8a5b8a75f9131d9eb38f45e76344
 
-  # Integration tests.
+    # Integration tests.
```
slowli: Nit: I don't think changing indent makes sense here; previously, the comment was aligned to the following list item mark (`- `).

aleksuss: Yeah. It happened accidentally.

components/keys/examples/key_generator.rs
```diff
@@ -0,0 +1,97 @@
...
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
+    name = "key-generator",
+    about = "An utility for keys generation from a seed."
+)]
+struct Arguments {
+    /// Seed for deriving keys.
+    #[structopt(short = "s", long)]
```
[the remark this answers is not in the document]
aleksuss: Hm. But the fields aren't optional in the struct. And the utility does nothing if you don't pass both arguments. So. If it so principle for you I can change it of cause. But in this case I suppose it is ok.

components/keys/src/lib.rs
```diff
@@ -205,11 +204,26 @@ impl EncryptedMasterKey {
...
+    let tree = SecretTree::from_seed(seed)
+        .ok_or_else(|| format_err!("Error creating SecretTree from seed"))?;
+    let encrypted_key = EncryptedMasterKey::encrypt(tree.seed(), passphrase)?;
+    let keys = generate_keys_from_master_password(&tree)
+        .ok_or_else(|| format_err!("Error deriving keys from master key."))?;
+
+    Ok((keys, encrypted_key))
+}
+
 fn generate_keys_from_master_password(tree: &SecretTree) -> Option<Keys> {
```
slowli: Off-topic: I doubt that this method should really return an `Option`; calls with `?` are seemingly infallible and could be replaced with `unwrap`s (or `Seed::new` constructor could be used).

components/keys/examples/key_generator.rs
```diff
@@ -0,0 +1,96 @@
...
+
+fn generate_json(passphrase: &str, seed: &str) -> Result<serde_json::Value, failure::Error> {
+    let seed = hex::decode(seed)?;
+    let (keys, encrypted_key) = generate_keys_from_seed(passphrase.as_bytes(), &seed)?;
+    let file_content = toml::to_string_pretty(&encrypted_key)?;
+
+    Ok(json!({
+        "consensus_pub_key": keys.consensus.public_key().to_hex(),
+        "service_pub_key": keys.service.public_key().to_hex(),
+        "master_key_file": file_content
```
slowli: Nit: You still didn't add a comma here 🙃

components/keys/src/lib.rs
aleksuss: Really good catch 👍🏻

PR #1784, opened 2020-02-17
Index proof verification [ECR-4249]

slowli:
## Overview

This PR introduces verification of `BlockProof`s and `IndexProof`s.

---

See: https://jira.bf.local/browse/ECR-4249

### DIFF
--- a/exonum/src/blockchain/block.rs
+++ b/exonum/src/blockchain/block.rs
@@ -252,8 +252,12 @@ impl BlockProof {
             }
             endorsements[validator_id] = true;
         }
+
+        // This assertion should always hold. Indeed, we've checked that there are +2/3 precommits
+        // and that there are no double endorsements; hence, the block should be approved
+        // by +2/3 validators.
         debug_assert!(
-            endorsements.iter().map(|&b| b as usize).sum::<usize>()
+            endorsements.iter().filter(|&&flag| flag).count()
                 >= byzantine_quorum(validator_keys.len())
         );
 
@@ -524,9 +528,8 @@ mod tests {
             .insert::<ProposerId>(ValidatorId(1));
 
         let precommits = keys.iter().enumerate().map(|(i, keypair)| {
-            let i = i as u16;
             let precommit = Precommit::new(
-                ValidatorId(i),
+                ValidatorId(i as u16),
                 Height(1),
                 Round(1),
                 Hash::zero(),
--- a/exonum/src/blockchain/block.rs
+++ b/exonum/src/blockchain/block.rs
@@ -14,7 +14,9 @@
 
 use exonum_crypto::{Hash, PublicKey};
 use exonum_derive::{BinaryValue, ObjectHash};
-use exonum_merkledb::{BinaryValue, MapProof, ObjectHash};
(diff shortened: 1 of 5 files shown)


exonum/src/blockchain/block.rs
```diff
@@ -386,4 +508,237 @@ mod tests {
...
+            additional_headers: AdditionalHeaders::default(),
+        };
+        block
+            .additional_headers
+            .insert::<ProposerId>(ValidatorId(1));
+
+        let precommits = keys.iter().enumerate().map(|(i, keypair)| {
+            let i = i as u16;
+            let precommit = Precommit::new(
+                ValidatorId(i),
```
aleksuss: Just out of curiosity. Why not just `i as u16` here ?

exonum/src/blockchain/block.rs
```diff
@@ -207,6 +210,103 @@ impl BlockProof {
...
+        }
+        if self.precommits.len() > validator_keys.len() {
+            return Err(ProofError::DoubleEndorsement);
+        }
+
+        let correct_heights = self
+            .precommits
+            .iter()
+            .all(|precommit| precommit.payload().height == self.block.height);
+        if !correct_heights {
```
aleksuss: Does it make sense to recognise which one of the precommits has a wrong height ?

exonum/src/blockchain/block.rs
```diff
@@ -207,6 +210,103 @@ impl BlockProof {
...
+            if expected_key != precommit.author() {
+                return Err(ProofError::ValidatorKeyMismatch);
+            }
+            if endorsements[validator_id] {
+                return Err(ProofError::DoubleEndorsement);
+            }
+            endorsements[validator_id] = true;
+        }
+        debug_assert!(
+            endorsements.iter().map(|&b| b as usize).sum::<usize>()
```
aleksuss: Isn't better ?
```suggestion
            endorsements.into_iter().filter(|e| *e == true).count()
```

exonum/src/blockchain/block.rs
```diff
@@ -207,6 +210,103 @@ impl BlockProof {
...
+                .ok_or(ProofError::IncorrectValidatorId)?;
+            if expected_key != precommit.author() {
+                return Err(ProofError::ValidatorKeyMismatch);
+            }
+            if endorsements[validator_id] {
+                return Err(ProofError::DoubleEndorsement);
+            }
+            endorsements[validator_id] = true;
+        }
+        debug_assert!(
```
aleksuss: Don't we really need it assert in the release code ?

exonum/src/blockchain/block.rs
slowli: I don't see a clear use case for this, tbh.

exonum/src/blockchain/block.rs
slowli: This assertion should always hold (otherwise, it would be a check, not an assertion). Indeed, we've checked that there are +2/3 precommits and that there are no double endorsements; hence, the block should be approved by +2/3 validators. The `debug_assert` is there to better illustrate the check logic.

PR #1786, opened 2020-02-18
Use fresh Rust version in travis [ECR-4211]

alekseysidorov:


components/merkledb/src/backends/temporarydb.rs
```diff
@@ -42,6 +42,7 @@ struct TemporarySnapshot {
 
 impl TemporaryDB {
     /// Creates a new, empty database.
+    #[must_use]
```
aleksuss: 🙃

aleksuss: What about to disable this lint ?

alekseysidorov: I think that this lint is useful in this component because it can warn about stupid mistakes.

slowli: I may be an old-fashioned guy, but I don't see utility of `#[must_use]` everywhere; I'd suggest to use it only when the non-use of the return value is non-obvious to a developer *and* can result in logical errors (cf. `must_use` for `Result`s and `Future` / `Stream` implementations). From [the relevant lint description](https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate):

> Why is this bad
>
> Not bad at all, this lint just shows places where you could add the attribute.

components/merkledb/src/views/metadata.rs
```diff
@@ -241,9 +242,9 @@ impl<V> IndexMetadata<V> {
     }
 
     /// Returns a globally unique numeric index identifier.
-    /// MerkleDB assigns a unique numeric ID for each fully-qualified index name.
+    /// `MerkleDB` assigns a unique numeric ID for each fully-qualified index name.
```
slowli: IMO, this change is confusing; I'm expecting code-formatted bits, which look like type names, to correspond to the type names.

PR #1787, opened 2020-02-18
Freeze support in supervisor [ECR-4226] [ECR-4227]

slowli:
## Overview

This PR introduces service freezing support in supervisor.

Discussion:

- Should this PR include logic to check if a runtime supports service freezing? This could be handy since the Java runtime will probably not support service freezing in the nearest future.

---

See: https://jira.bf.local/browse/ECR-4226

**Edit:** Now this PR covers a part of ECR-4227 as well.


exonum/src/runtime/dispatcher/migration_tests.rs
```diff
@@ -364,9 +375,11 @@ fn test_migration_workflow(freeze_service: bool) {
 
     // Now, the migration start should succeed.
     let fork = rig.blockchain.fork();
-    rig.dispatcher()
+    let ty = rig
```
aleksuss: `ty` 🙃 I've never seen such abbreviation before

exonum/src/runtime/migrations.rs
```diff
@@ -116,6 +137,18 @@ use crate::runtime::{CoreError, ExecutionError, ExecutionFail, InstanceSpec};
 
 type MigrationLogic = dyn FnOnce(&mut MigrationContext) -> Result<(), MigrationError> + Send;
 
+/// Denotes possible types of data migrations.
```
aleksuss: Maybe `Types of data migrations.` would be enough ?

services/supervisor/README.md
```diff
@@ -6,27 +6,29 @@
...
 Supervisor is an Exonum service capable of the following activities:
 
-- Service artifact deployment;
-- Service instances creation;
-- Changing consensus configuration;
-- Changing service instances configuration.
+- Deploying service artifacts
+- Instantiating services 
+- Changing configuration of instantiated services
+- Changing state of instantiated services: stopping, freezing, resuming,
```
aleksuss:
```suggestion
- Changing the state of instantiated services: stopping, freezing, resuming,
```

exonum/src/runtime/dispatcher/migration_tests.rs
slowli: It's used, for example, in [`syn`](https://docs.rs/syn/1.0.14/syn/struct.Field.html#structfield.ty) because `type` is a reserved keyword.

exonum/src/runtime/mod.rs
```diff
@@ -336,6 +371,19 @@ pub trait Runtime: Send + fmt::Debug + 'static {
...
+    /// Checks if the runtime supports an optional feature.
+    ///
+    /// This method can be called by the core before performing operations that might not
+    /// be implemented in a runtime, or by the supervisor service in order to check that a potential
+    /// service / artifact state transition can be handled by the runtime.
+    ///
+    /// An implementation should return `false` for all features the runtime does not recognize.
+    /// The default implementation always returns `false`, i.e., signals that the runtime supports
+    /// no optional features.
+    fn is_supported(&self, feature: &RuntimeFeature) -> bool {
```
slowli: @dmitry-timofeev Here's the new method in the `Runtime` interface, through which the runtime can signal that it supports service freezing (in the future, this could be extended to other features, such as artifact unloading).

dmitry-timofeev: I see, thanks! (cc @vitvakatu the things we don't support yet)

PR #1788, opened 2020-02-19
Remove future from the Runtime trait [ECR-4263]

alekseysidorov:


exonum/src/runtime/oneshot.rs
```diff
@@ -0,0 +1,72 @@
...
+use super::{ErrorKind, ExecutionError};
+
+/// Create a new one-shot channel for sending deployment status,
+/// returning the sender/receiver halves.
+pub fn channel() -> (Sender, Receiver) {
+    let (tx, rx) = mpsc::channel();
+    (Sender(tx), Receiver(rx))
+}
+
+/// The receiving half of [`channel`] type. This half can only be owned by one thread.
```
slowli: > This half can only be owned by one thread.

Nit: I don't understand what this comment achieves. Isn't this true for pretty much any data type?

exonum/src/runtime/oneshot.rs
```diff
@@ -0,0 +1,72 @@
...
+///
+/// [`channel`]: fn.channel.html
+#[derive(Debug)]
+pub struct Receiver(mpsc::Receiver<Result<(), ExecutionError>>);
+
+impl Receiver {
+    /// Attempts to wait for a value on this receiver, returning an error if the
+    /// corresponding channel has hung up.
+    pub(crate) fn wait(self) -> Result<(), ExecutionError> {
+        self.0.recv().map_err(|e| {
```
slowli: Nit: I don't think it makes sense to output an error; [`RecvError`](https://doc.rust-lang.org/std/sync/mpsc/struct.RecvError.html) is an empty struct, so it does not provide any details as to the error cause.

exonum/src/runtime/oneshot.rs
```diff
@@ -0,0 +1,72 @@
...
+                    "An error during waiting for deployment status occurred: {:?}",
+                    e
+                ),
+            )
+        })?
+    }
+}
+
+/// The sending-half of [`channel`] type. This half can only be
+/// owned by one thread, but it can be cloned to send to other threads.
```
slowli: > but it can be cloned

Unless I'm missing something, it can't be cloned right now since `#[derive(Clone)]` is missing.

exonum/src/runtime/oneshot.rs
```diff
@@ -0,0 +1,72 @@
...
+pub struct Sender(mpsc::Sender<Result<(), ExecutionError>>);
+
+impl Sender {
+    /// Sends a deployment status on this channel.
+    ///
+    /// This method will never block the current thread.
+    pub fn send(self, deployment_status: Result<(), ExecutionError>) {
+        if let Err(e) = self.0.send(deployment_status) {
+            log::warn!(
+                "Unable to send deployment status: an error occurred {:?}",
```
slowli: Nit: same as with `recv` errors, [`SendError`](https://doc.rust-lang.org/std/sync/mpsc/struct.SendError.html) doesn't provide info why an error has occurred, it just wraps the unsuccessfully sent value.

examples/sample_runtime/src/main.rs
```diff
@@ -95,12 +96,10 @@ impl SampleRuntime {
...
 
 impl Runtime for SampleRuntime {
-    fn deploy_artifact(
-        &mut self,
-        artifact: ArtifactId,
-        spec: Vec<u8>,
-    ) -> Box<dyn Future<Item = (), Error = ExecutionError>> {
-        Box::new(self.deploy_artifact(artifact, spec).into_future())
+    fn deploy_artifact(&mut self, artifact: ArtifactId, spec: Vec<u8>) -> oneshot::Receiver {
+        let (tx, rx) = oneshot::channel();
```
slowli: As I've suggested previously, it can make sense to introduce a `Receiver` constructor with an ready-made value:

```rust
impl Receiver {
    pub fn with_result(result: Result<(), ExecutionError>) -> Self {
        let (tx, this) = channel();
        tx.send(result);
        this
    }
}

// Usage:
Receiver::with_result(self.deploy_artifact(artifact, spec))
```

CHANGELOG.md
```diff
@@ -16,19 +24,15 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 - Switched off broadcasting of transactions which the node considers
   incorrect. (#1781)
 
+- Fixed incorrect invalidation of block proposals. (#1782)
+
 ### Breaking changes
```
slowli: I believe this "Breaking changes" section should be united with the above one.

exonum/src/runtime/dispatcher/mod.rs
```diff
@@ -482,13 +477,11 @@ impl Dispatcher {
...
         if !self.is_artifact_deployed(&artifact) {
-            self.deploy_artifact(artifact, payload)
-                .wait()
-                .unwrap_or_else(|e| {
-                    // In this case artifact deployment error is fatal because there are
-                    // confirmation that this node can deploy this artifact.
-                    panic!("Unable to deploy registered artifact. {}", e)
-                });
+            self.deploy_artifact(artifact, payload).unwrap_or_else(|e| {
+                // In this case artifact deployment error is fatal because there are
```
slowli: Nit / grammar: "...because the deploy was committed on the network level" would read better.

CHANGELOG.md
```diff
@@ -16,19 +16,21 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
```
slowli: Nit: In other release sections, "Breaking changes" comes before "Bug fixes".

CHANGELOG.md
```diff
@@ -16,19 +16,21 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
 
+#### exonum
+
+- `Runtime::deploy_artifact` longer returns `Box<dyn Future<...>>`. Instead a special
```
slowli: Also, this should probably say "`Runtime::deploy_artifact` **no** longer returns..."

PR #1790, opened 2020-02-20
Refine endpoints in system-api plugin [ECR-4154]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-4154

### Definition of Done


CHANGELOG.md
```diff
@@ -11,6 +11,21 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
...
+  been changed. (#1790) Current list of the endpoints:
+  - `v1/info` - obtains information about the node;
+  - `v1/stats` - obtains statistics of the node;
+  - `v1/peers` - adds a peer to the Exonum node;
+  - `v1/consensus_status` - enables or disables consensus on the node;
+  - `v1/shutdown` - shuts down the node.
+
+#### exonum-supervisor
+
+- `supervisor/services` has been added which obtains information about
```
slowli: Nit / grammar: `supervisor/services` **endpoint** has been added?

cli/tests/node.rs
```diff
@@ -106,7 +106,7 @@ fn node_basic_workflow() -> Result<(), failure::Error> {
     let client = reqwest::Client::new();
     // Check info returned by the system API plugin.
```
slowli: This comment should be updated.

components/system-api/src/private.rs
```diff
@@ -10,143 +10,108 @@
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
-// limitations under the License.
```
slowli: Why is this removed?

components/system-api/src/private.rs
```diff
@@ -10,143 +10,108 @@
...
+//! # Get Node Statistics
 //!
 //! | Property    | Value |
 //! |-------------|-------|
-//! | Path        | `/api/system/v1/network` |
+//! | Path        | `/api/system/v1/stats` |
 //! | Method      | GET   |
 //! | Query type  | - |
-//! | Return type | [`NodeInfo`] |
+//! | Return type | [`StatsInfo`] |
```
slowli: Nit: I'd call the type `NodeStats`.

components/system-api/src/private.rs
```diff
@@ -206,59 +171,80 @@
...
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-pub struct ServiceInfo {
-    /// Service name.
-    pub name: String,
-    /// Service identifier for the database schema and service messages.
-    pub id: InstanceId,
+/// Information about the current state of the node.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+pub struct StatsInfo {
+    /// Height of the last committed block.
```
slowli: Nit: I'd rephrase as "Height of the blockchain."

components/system-api/src/private.rs
```diff
@@ -206,59 +171,80 @@
...
+    pub height: u64,
+    /// Total number of uncommitted transactions stored in persistent pool.
+    pub tx_pool_size: u64,
+    /// Total number of transactions in the blockchain.
+    pub tx_count: u64,
+    /// Size of the transaction cache.
+    pub tx_cache_size: usize,
+    /// Work duration of the node.
```
slowli: In what units? (seconds? millis?)

components/system-api/src/private.rs
```diff
@@ -206,59 +171,80 @@
...
+    pub uptime: u64,
 }
 
-/// Short information about the current node.
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-pub struct NodeInfo {
-    /// Version of the `exonum` crate.
-    pub core_version: Option<String>,
+/// Information about whether it is possible to achieve the consensus between
```
slowli: Off-topic: Is this information garbage if the node is an auditor? Also, "Consensus enabled and the node has enough connected peers" does **not** mean that consensus can necessarily advance; for example, some of peers may have consensus disabled.

components/system-api/src/private.rs
```diff
@@ -206,59 +171,80 @@
...
-    }
+/// Info about connected peer.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct ConnectedPeerInfo {
+    /// Address of the peer.
+    pub address: String,
+    /// Consensus public key of the peer.
+    pub public_key: PublicKey,
+    /// Connect direction.
+    pub direction: String,
```
slowli: Why is this a string and not an enum with 2 variants?

components/system-api/src/private.rs
```diff
@@ -206,59 +171,80 @@
...
 }
 
-/// Short information about the current node.
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-pub struct NodeInfo {
-    /// Version of the `exonum` crate.
-    pub core_version: Option<String>,
+/// Information about whether it is possible to achieve the consensus between
+/// validators in the current state.
+#[derive(Deserialize, Serialize, Clone, Debug, PartialEq)]
```
slowli: We customarily use `#[serde(rename_all = "snake_case")]` for enums.

components/system-api/src/private.rs
```diff
@@ -206,59 +171,80 @@
...
+/// Short information about the current node.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct NodeInfo {
+    /// Consensus status.
+    pub consensus_status: ConsensusStatus,
+    /// List of connected peers.
+    pub connected_peers: Vec<ConnectedPeerInfo>,
+    /// Version of the `exonum` crate.
+    pub exonum_version: String,
+    /// Version of thr Rust.
```
slowli: Grammar: "Version of thr Rust" -> "Rust version"

components/system-api/src/private.rs
```diff
@@ -206,59 +171,80 @@
...
-    pub outgoing_connections: HashMap<SocketAddr, OutgoingConnection>,
+/// Short information about the current node.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct NodeInfo {
+    /// Consensus status.
+    pub consensus_status: ConsensusStatus,
+    /// List of connected peers.
+    pub connected_peers: Vec<ConnectedPeerInfo>,
+    /// Version of the `exonum` crate.
+    pub exonum_version: String,
```
slowli: Could this be made into a `semver::Version`?

exonum/src/helpers/user_agent.rs
```diff
@@ -32,10 +32,29 @@ pub fn user_agent() -> String {
...
+/// Returns OS info of host on which run the node.
+#[doc(hidden)]
+pub fn os_info() -> String {
+    os_info::get().to_string()
+}
+
+/// Returns a version of the exonum framework.
+#[doc(hidden)]
+pub fn exonum_version() -> Option<String> {
+    let versions: Vec<_> = USER_AGENT.split('/').collect();
```
slowli: Using `next()` instead of `collect()` would be more efficient.

exonum/src/helpers/user_agent.rs
```diff
@@ -32,10 +32,29 @@ pub fn user_agent() -> String {
...
+    versions.get(0).map(|v| v.trim().to_owned())
+}
+
+/// Returns a version of the rust compiler.
+#[doc(hidden)]
+pub fn rust_version() -> Option<String> {
+    let versions: Vec<_> = USER_AGENT.split('/').collect();
```
slowli: Using `nth(1)` instead of `collect()` would be more efficient.

components/system-api/src/private.rs
```diff
@@ -269,57 +255,85 @@ pub struct ConsensusEnabledQuery {
...
+                os_info: os_info(),
             })
         });
-        self
+        self_
     }
 
-    fn handle_peer_add(self, name: &'static str, api_scope: &mut ApiScope) -> Self {
+    fn handle_stats(self, name: &'static str, api_scope: &mut ApiScope) -> Self {
+        let self_ = self.clone();
```
slowli: Micro-nit / bikeshedding: IMO, moving `self_` into closure would be more idiomatic (you could also call the variable `this` if the underscore at the end looks bad).

components/system-api/src/private.rs
```diff
@@ -269,57 +255,85 @@ pub struct ConsensusEnabledQuery {
...
+                connected_peers.push(ConnectedPeerInfo::new(&connect_info, "incoming"));
             }
 
-            Ok(PeersInfo {
-                incoming_connections: shared_api_state.incoming_connections(),
-                outgoing_connections,
+            Ok(NodeInfo {
+                consensus_status: self.get_consensus_status(),
+                connected_peers,
+                exonum_version: exonum_version().unwrap_or_else(|| "unknown".to_owned()),
```
slowli: I think this and the following vars should be instantiated once and then cloned.

components/system-api/src/private.rs
```diff
@@ -269,57 +255,85 @@ pub struct ConsensusEnabledQuery {
...
+
+            for connect_info in shared_api_state.incoming_connections() {
+                connected_peers.push(ConnectedPeerInfo::new(&connect_info, "incoming"));
             }
 
-            Ok(PeersInfo {
-                incoming_connections: shared_api_state.incoming_connections(),
-                outgoing_connections,
+            Ok(NodeInfo {
+                consensus_status: self.get_consensus_status(),
```
slowli: Nit: `get_consensus_status()` can be made a free function or a static method accepting `&SharedNodeState` as an argument. In this way, you wouldn't need to clone `self` above.

exonum/src/helpers/user_agent.rs
```diff
@@ -48,4 +67,16 @@ mod tests {
             assert!(!val.is_empty());
         }
     }
+
+    #[test]
+    fn check_exonum_versions() {
+        let exonum_version = exonum_version().unwrap();
+        assert!(exonum_version.contains("exonum"));
```
slowli: What's the point of including "exonum" prefix into `exonum_version()`? Same question regarding the "rust" prefix for `rust_version`.

.travis.yml
```diff
@@ -132,6 +132,10 @@ jobs:
   - name: unit-test
     script:
     - cargo test --all --lib --tests --benches
+
+  # Doc tests.
+  - name: doc-test
```
slowli: Good idea, although inability to share build cache among this job and unit tests can increase build time for this job (but, hopefully, not the overall build time).

components/system-api/src/private.rs
aleksuss: Yeah. I'll rephrase this description.

components/system-api/src/private.rs
aleksuss: Potentially it could. But why ?

.travis.yml
aleksuss: Additionally, we use `sscache` for caching which stores its cache to `$HOME/.cache` and this directory shares between jobs. I suppose. By the way. I try to remove .cargo caching. I think in our case it works not optimal and consumes more time for archiving/unarchiving.

components/system-api/src/private.rs
aleksuss: But there is a case when we don't need this var at all. Redundant instantiation ?

components/system-api/src/private.rs
```diff
@@ -19,134 +19,101 @@
...
+//! # Get Node Statistics
 //!
 //! | Property    | Value |
 //! |-------------|-------|
-//! | Path        | `/api/system/v1/network` |
+//! | Path        | `/api/system/v1/stats` |
 //! | Method      | GET   |
 //! | Query type  | - |
-//! | Return type | [`NodeInfo`] |
+//! | Return type | [`StatsInfo`] |
```
slowli: Type name and link should be updated here to point to `NodeStats`. Weird that `deadlinks` didn't pick up this error. (`cargo clean --doc` should probably run before generating docs.)

exonum/build.rs
```diff
@@ -42,10 +42,9 @@ fn create_path_to_protobuf_schema_env() {
 }
 
 fn write_user_agent_file() {
-    let package_name = option_env!("CARGO_PKG_NAME").unwrap_or("exonum");
-    let package_version = option_env!("CARGO_PKG_VERSION").unwrap_or("?");
+    let exonum_version = option_env!("CARGO_PKG_VERSION").unwrap_or("?");
     let rust_version = rust_version().unwrap_or_else(|| "rust ?".to_string());
```
slowli: `rust ` prefix should be removed here.

exonum/build.rs
```diff
@@ -76,5 +75,7 @@ fn main() {
 fn rust_version() -> Option<String> {
     let rustc = option_env!("RUSTC").unwrap_or("rustc");
     let output = Command::new(rustc).arg("-V").output().ok()?.stdout;
-    String::from_utf8(output).ok()
+    std::str::from_utf8(&output)
+        .ok()
+        .map(|o| o.split_whitespace().nth(1).unwrap().to_owned())
```
slowli: Are you sure that `unwrap` is safe in this context? I'd return `None` in this case. Also, `?` could be used to simplify the code:

```rust
let rustc_output = std::str::from_utf8(&output).ok()?;
let version = rustc_output.split_whitespace().nth(1)?;
Some(version.to_owned())
```

PR #1792, opened 2020-02-21
Proofs for call status [ECR-4246]

slowli:
## Overview

This PR introduces ability to retrieve proofs for call statuses, both within Rust (using core schema) and using HTTP API of the explorer service. As a part of the PR, a bug concerning `ExecutionError` storage was fixed: `ExecutionError.description` was intended to be malleable, but this wasn't actually the case.

Not done yet:

- [x] Emulate (in the testkit?) different error descriptions with the same code; check that `error_hash`es are the same.

Discussion:

- Should error indexes be made public (with `#[doc(hidden)]` attribute) for the sake of bindings?
- Is implementing `Serialize` / `Deserialize` for `ExecutionError` a good idea? Technically, this makes it possible to spoof error location from the service code, but this may be outside the scope of our threat model.

---

See: https://jira.bf.local/browse/ECR-4246


slowli:
@dmitry-timofeev @vitvakatu Please comment on

> Should error indexes be made public (with `#[doc(hidden)]` attribute) for the sake of bindings?

...and on the PR in general.

dmitry-timofeev:
> Should error indexes be made public (with #[doc(hidden)] attribute) for the sake of bindings?

The public indexes are currently accessed by address directly from `CoreSchema`. Do we need to access the new private indexes at all?

exonum/src/blockchain/block.rs
```diff
@@ -349,18 +356,117 @@ impl IndexProof {
...
+/// from a non-existing service requires prior knowledge though.
+///
+/// `CallProof`s should not be mixed up with a proof of transaction commitment.
+/// To verify that a certain transaction was committed, use a proof from
+/// the `block_transactions` index of the [core schema].
+///
+/// # Verification
+///
+/// If you are implementing `CallProof` verification, bear in mind that `ExecutionError`
+/// need to be stripped off the non-hashed information before verifying `call_proof`.
```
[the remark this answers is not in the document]
slowli: > If the proof, on the other hand, comes only with the data to be hashed

Are we OK with not returning human-readable error description? IMO, the complexity introduced by error descriptions is manageable, but another option is to return error descriptions separately. Would it be OK with you?

slowli:
> Do we need to access the new private indexes at all?

What do you mean by private indexes? It's still possible to access error-related indexes by address, of course. The only possible problem is that `ExecutionErrorAux` (non-hashed part of `ExecutionError`) is treated as an implementation detail right now; if this approach is used, it should probably be exported.

exonum/src/blockchain/block.rs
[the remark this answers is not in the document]
dmitry-timofeev: I think the human-readable error description is valuable. Do you mean by returning separately returning in the same response, but in a different object? I would go with such approach, because it does not require special handling during verification.

dmitry-timofeev:
> What do you mean by private indexes?

That they are inaccessible through BlockchainSchema API. My question is whether the framework has to provide access to any of these indexes to service API: call_errors and call_errors_aux.

exonum/src/blockchain/block.rs
```diff
@@ -349,18 +356,122 @@ impl IndexProof {
...
+    /// The root hash of the proof must be equal to the `error_hash` mentioned in `block_proof`.
+    pub call_proof: MapProof<CallInBlock, ExecutionError>,
+
+    /// Human-readable description of an error if the call status is erroneous.
+    /// This description is not authenticated and thus should be used for diagnostic purposes only.
+    /// If the call is successful, the error description should be `None`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub error_description: Option<String>,
+
+    #[serde(skip, default)]
```
alekseysidorov: Can we use non exhaustive attribute for structures?

exonum/src/blockchain/block.rs
```diff
@@ -349,18 +356,122 @@ impl IndexProof {
...
+
+        let call_status = match maybe_status {
+            None => {
+                if self.error_description.is_some() {
+                    return Err(ProofError::MalformedStatus);
+                }
+                Ok(())
+            }
+            Some(e) => {
+                let mut full_error = e.to_owned();
```
alekseysidorov: Must full error always has description?

exonum/src/blockchain/block.rs
slowli: Sure, but this can be done in a separate PRs (e.g., #1797).

PR #1796, opened 2020-02-25
Port exonum to the actix-web 2.0 version.

alekseysidorov:


exonum-node/src/events/network.rs
```diff
@@ -382,7 +386,9 @@ impl NetworkHandler {
                                         "Couldn't take peer addr from socket = {}",
                                         e
                                     )))
-                                        as Box<dyn Future<Error = failure::Error, Item = ()>>;
+                                        as Box<
```
slowli: Nit: This formatting is terrible.

exonum/src/blockchain/api_sender.rs
```diff
@@ -45,12 +46,14 @@ impl<T: Send + 'static> ApiSender<T> {
     /// # Return value
     ///
     /// The failure means that the node is being shut down.
-    pub fn send_message(&self, message: T) -> impl Future<Item = (), Error = SendError> {
+    pub async fn send_message(&self, message: T) -> Result<(), SendError> {
```
slowli: Can be considered later: AFAIK, there is a subtle difference between

```rust
pub async fn send_message(&self, ...)
```

and

```rust
pub fn send_message(&self, ...) -> impl Future<Output = _>
```

The future returned by the first method has a lifetime dependent on `&self`, while the second method will return a future with a `'static` lifetime. Hence, if we reasonably expect that the users will expect a static-lifetime future, then we should use an `impl Future` approach.

runtimes/rust/src/api.rs
```diff
@@ -244,27 +251,32 @@ impl ServiceApiScope {
...
         let artifact = self.artifact.clone();
-        self.inner
-            .endpoint_mut(name, move |query: Q| -> FutureResult<I> {
+        self.inner.endpoint_mut(name, move |query: Q| {
+            let handler = handler.clone();
+            let blockchain = blockchain.clone();
+            let descriptor = descriptor.clone();
+            let artifact = artifact.clone();
+
+            async move {
```
slowli: I'd suggest to rework this to "`futures 0.1`-style" (i.e., minimized or absent `async {}`) in order to avoid cloning above. Can be done later.

runtimes/rust/src/api.rs
```diff
@@ -395,18 +419,20 @@ impl ServiceApiScope {
...
 ///     /// Simple handler without any parameters.
-///     pub fn ping(_state: &ServiceApiState, _query: ()) -> api::Result<()> {
+///     pub async fn ping(_state: ServiceApiState, _query: ()) -> api::Result<()> {
 ///         Ok(())
 ///     }
 ///
-///     /// You may also create asynchronous handlers for long requests.
-///     pub fn async_operation(
-///         _state: &ServiceApiState,
+///     /// You may also uses asynchronous tasks.
```
slowli: Grammar: "You may use"

runtimes/rust/src/api.rs
```diff
@@ -395,18 +419,20 @@ impl ServiceApiScope {
...
 ///         query: MyQuery,
-///     ) -> api::FutureResult<Option<Hash>> {
-///         Box::new(futures::lazy(move || {
-///             Ok(Some(query.block_height.object_hash()))
-///         }))
+///     ) -> api::Result<Option<Hash>> {
+///         # async fn long_async_task(query: MyQuery) -> Option<Hash> {
+///         #     Some(Hash::zero())
+///         # }
+///
```
slowli: Nit: This newline will look weird in generated docs.

runtimes/rust/src/runtime_api.rs
```diff
@@ -143,24 +143,29 @@ pub fn endpoints(runtime: &RustRuntime) -> impl IntoIterator<Item = (String, Api
         // This endpoint returns list of protobuf source files of the specified artifact,
         // otherwise it returns source files of Exonum itself.
         .endpoint("proto-sources", {
-            move |query: ProtoSourcesQuery| -> api::Result<Vec<ProtoSourceFile>> {
-                if let ProtoSourcesQuery::Artifact { name, version } = query {
-                    let artifact_id = ArtifactId::new(RuntimeIdentifier::Rust, name, version)
-                        .map_err(|e| {
+            move |query: ProtoSourcesQuery| {
```
slowli: Can't this approach be used to minimize cloning?

```rust
|query: ProtoSourcesQuery| async move {
    // ...
}
```

(Similarly, in some other places.)

services/explorer/src/api.rs
```diff
@@ -545,32 +544,32 @@ impl ExplorerApi {
     }
 
     /// Returns call status of `before_transactions` hook.
-    fn before_transactions_status(
-        schema: Schema<&dyn Snapshot>,
+    async fn before_transactions_status(
+        state: ServiceApiState,
```
slowli: Just to be clear: Did you change the signature because `async fn` automatically includes argument lifetimes into the output `Future`? You can still use an ordinary `fn` with an `impl Future` return value 🙃

services/explorer/src/api/websocket/mod.rs
```diff
@@ -221,6 +221,7 @@ pub(crate) struct SharedState {
     inner: Arc<Mutex<SharedStateInner>>,
 }
 
+// FIXME This code leads to crashes in api tests. [ECR-4268]
```
slowli: Is the referenced Jira task wrong here, or do you plan to fix this in this PR?

services/explorer/src/api/websocket/mod.rs
```diff
@@ -394,17 +402,20 @@ impl Server {
     fn handle_transaction(
         &self,
         message: Transaction,
-    ) -> impl Future<Item = TransactionResponse, Error = failure::Error> {
+    ) -> LocalBoxFuture<'static, Result<TransactionResponse, failure::Error>> {
```
slowli: Nit: I'd use `impl Future` here and move boxing closer to where it's necessary (i.e., `Handler<Transaction>` implementation).

services/explorer/tests/api.rs
```diff
@@ -485,28 +510,31 @@ fn test_explorer_transaction_statuses() {
...
-        let info: TransactionInfo = api
-            .public(ApiKind::Explorer)
-            .query(&TransactionQuery::new(*hash))
-            .get("v1/transactions")
-            .unwrap();
-        info.as_committed().unwrap().status().map_err(Clone::clone)
-    })
-    .collect();
+    let statuses = {
+        let mut output = Vec::new();
```
slowli: Nit: You can simply write `let mut statuses = Vec::new()` and avoid the additional block here.

services/explorer/tests/websocket.rs
```diff
@@ -87,6 +87,7 @@ fn assert_closure(mut client: Client<TcpStream>) {
 
 fn init_testkit() -> (TestKit, TestKitApi) {
     let mut testkit = TestKitBuilder::validator()
+        .with_logger()
```
slowli: Debug artifact?

services/supervisor/src/api.rs
```diff
@@ -561,15 +580,16 @@
...
+//! #     .await?;
 //! # testkit.create_block();
 //! let query = DeployInfoQuery::from(deploy_request);
 //!
 //! let deploy_state: AsyncEventState = testkit
 //!     .api()
 //!     .private(ApiKind::Service("supervisor"))
 //!     .query(&query)
-//!     .get("deploy-status")?;
+//!     .get("deploy-status").await?;
```
slowli: Nit: `await` should be on a separate line.

services/supervisor/src/lib.rs
```diff
@@ -531,9 +531,11 @@ impl Supervisor {
                 if let Some(tx_sender) = tx_sender {
```
slowli: IMO, we should consider making `start_deploy` accept futures; blocking in the handler looks weird.

test-suite/soak-tests/src/services.rs
```diff
@@ -59,14 +59,18 @@ impl Service for MainService {
...
-                log::error!(
-                    "[{}] Failed to broadcast transaction at height {}: {}",
-                    context.service_key(),
-                    context.height(),
-                    e
-                );
-            }
+            let height = context.height();
+            let service_key = context.service_key();
+            futures::executor::block_on(async move {
```
slowli: Maybe, it could make sense to introduce a sync version of broadcaster, at least as a temporary measure? Fragments similar to this one are repeated throughout the codebase.

exonum-node/src/events/mod.rs
```diff
@@ -27,7 +27,7 @@ use exonum::{
     helpers::{Height, Round},
     messages::{AnyTx, Verified},
 };
-use futures::{
+use futures_01::{
```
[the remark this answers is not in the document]
alekseysidorov: Yes, following task ECR-4269 just about removing old futures and tokio-compat.

services/supervisor/src/lib.rs
alekseysidorov: I thought about it, but anyway in current testkit implementation we should wait until this future will be finished.

services/explorer/src/api.rs
```diff
@@ -512,11 +511,11 @@ impl ExplorerApi {
             })
     }
 
-    fn transaction_status(
-        schema: Schema<&dyn Snapshot>,
+    async fn transaction_status(
```
slowli: Adding to previous comments: IMO, it makes little sense to make a function async just because you can (especially if the function's private). Doing this would be more transparent:

```rust
.endpoint("v1/call_status/transaction", |state, query| async {
    Self::transaction_status(state.data().for_core(), query)
})
```

alekseysidorov: But with async functions I can keep `wire_rest` boilerplate-less without any side effects.

runtimes/rust/src/runtime_api.rs
alekseysidorov: This approach will work if the endpoint handler accepts `FnOnce` instead of` Fn`.

services/explorer/src/api.rs
alekseysidorov: Imo in this case it doesn't make sense, because this handlers anyway take `ServiceApiState` by value. Moreover api state cloning is cheap operation: it just increments two atomic counters and clones two strings.

test-suite/soak-tests/src/services.rs
alekseysidorov: It make sense, but I don't know how to implement both of `GenericCall<(), Output = Future>` and  `GenericCall<(), Output = Result>` for the same context type.

services/explorer/Cargo.toml
```diff
@@ -22,10 +22,12 @@ exonum-derive = { version = "1.0.0-rc.1", path = "../../components/derive" }
...
 
-actix = "0.7.9"
-actix-web = { version = "0.7.18", default-features = false }
+actix = "0.9.0"
+actix-web = { version = "2.0.0", default-features = false }
+actix-web-actors = "2.0.0"
 failure = "0.1.5"
-futures = "0.1.25"
+futures = "0.3.4"
+futures-01 = { package = "futures", version = "0.1.24" }
```
slowli: This dependency seems to be unused.

components/api/src/manager.rs
```diff
@@ -40,6 +42,13 @@ impl WebServerConfig {
             allow_origin: None,
         }
     }
+
+    fn cors_factory(&self) -> CorsFactory {
+        match self.allow_origin.clone() {
+            Some(origin) => CorsFactory::from(origin),
+            None => Cors::new().finish(),
```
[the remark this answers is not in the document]
alekseysidorov: Unfortunately Cors Factory does not implement `Default` trait.

exonum-node/src/lib.rs
```diff
@@ -1157,8 +1149,20 @@ impl Node {
...
         let connect_message = self.state().our_connect_message().clone();
         let connect_list = self.state().connect_list();
-        let api_manager = ApiManager::new(self.api_manager_config, self.channel.endpoints.1);
-        SystemRuntime::start(api_manager).expect("Failed to start api_runtime.");
+
+        let api_manager = ApiManager::new(self.api_manager_config);
+        let endpoints = self.channel.endpoints.1;
+        thread::spawn(move || {
+            actix_rt::System::new("exonum-node").block_on(api_manager.run(
+                endpoints.compat().map_err(|_| {
```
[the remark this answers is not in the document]
alekseysidorov: Yes, and I believe that this should be fixed by the next ECR-4269 task.

components/api/src/manager.rs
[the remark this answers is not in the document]
slowli: I've meant [`Cors::default`](https://docs.rs/actix-cors/0.2.0/actix_cors/struct.Cors.html#method.default), sorry.

PR #1800, opened 2020-03-01
Refactor node builders [ECR-4298]

slowli:
## Overview

This PR refactors node builders in the testkit and the CLI crate, extracting reusable parts into a separate extendable interface.

---

See: https://jira.bf.local/browse/ECR-4298

### DIFF
--- a/runtimes/rust/src/spec.rs
+++ b/runtimes/rust/src/spec.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Specifications of Rust artifacts for use in deployment.
+//! Specifications of Rust and non-Rust artifacts for use in deployment.
 
 use exonum::{
     blockchain::config::{GenesisConfigBuilder, InstanceInitParams},
@@ -152,7 +152,8 @@ impl<T: ServiceFactory + MigrateData> Deploy for Spec<T, Migrating> {
     }
 }
 
-/// FIXME
+/// Deploy specification for a non-Rust artifact. The spec can include 0 or more
+/// instantiated services.
 #[derive(Debug)]
 pub struct ForeignSpec {
     artifact: ArtifactId,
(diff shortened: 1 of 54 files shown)


runtimes/rust/src/spec.rs
```diff
@@ -0,0 +1,247 @@
...
+///
+/// This is a sealed trait: it is not meant to be implemented by the external data types.
+pub trait Deploy: Sealed {
+    /// Modifies the genesis config and Rust runtime builders to deploy self.
+    fn deploy(self, genesis: &mut GenesisConfigBuilder, rt: &mut RustRuntimeBuilder);
+}
+
+/// Marker type for artifact deployment without migration support.
+#[derive(Debug)]
+pub struct Simple(());
```
alekseysidorov: Why not just `Service`?

runtimes/rust/src/spec.rs
```diff
@@ -0,0 +1,247 @@
...
+
+/// Marker type for artifact deployment with migration support.
+#[derive(Debug)]
+pub struct Migrating(());
+
+/// Deploy specification for a Rust artifact. The spec can include 0 or more instantiated services.
```
alekseysidorov: Imo zero instead of 0 is more readable.

runtimes/rust/src/spec.rs
slowli: I think `Simple` / `Migrating` dichotomy provides a better description than `Service` / `Migrating`. (For one, "simple" and "migrating" are both adjectives 🙃)

alekseysidorov: Oh, that is my mistake, I meant that we can just write `struct Simple` instead of `struct Simple(())`.

slowli: I wanted `Simple` / `Migrating` to be marker types (i.e., not instantiable).

PR #1801, opened 2020-03-01
Restore deploy tests [ECR-4299] [ECR-4182]

slowli:
## Overview

This PR restores deploy integration tests. To maximize reuse, restoring uses a new example binary from advanced cryptocurrency, which can also be used to test data migrations. Correspondingly, the advanced cryptocurrency is made an update of the simple cryptocurrency, and a data migration script is added to it together with tests.

Discussion:

- Advanced cryptocurrency was published for some reason. Where is it used? I've reworked the crate to be non-publishable.
- The launcher seems to lack functionality related to data migrations, so they cannot be tested in integration. Should we fix this?

---

See: https://jira.bf.local/browse/ECR-4299, https://jira.bf.local/browse/ECR-4182


examples/cryptocurrency-advanced/backend/src/migrations/tests.rs
```diff
@@ -0,0 +1,225 @@
...
+    let migration_req = MigrationRequest {
+        new_artifact: CryptocurrencyService.artifact_id(),
+        service: OldService::INSTANCE_NAME.to_owned(),
+        deadline_height: Height(100),
+    };
+    let migration = admin_keys.request_migration(SUPERVISOR_ID, migration_req.clone());
+    let block = testkit.create_block_with_transaction(migration);
+    assert!(block.errors.is_empty());
+
+    // Since migration is performed in a background, it may take a little while.
```
alekseysidorov: Interesting, can we use futures in this case?

examples/cryptocurrency-advanced/tutorial/frontend.md
```diff
@@ -1,320 +0,0 @@
-# Cryptocurrency frontend tutorial
```
alekseysidorov: What will happen to this guide in the future?

slowli: This guide duplicates [the one on exonum.com](https://exonum.com/doc/version/1.0.0-rc.1/get-started/light-client/), thus I think it can be safely removed.

examples/cryptocurrency-advanced/backend/src/migrations/tests.rs
slowli: Could you provide pseudocode how futures can be used? All other testkit code is synchronous, so I'm not sure how well the use of futures will fit into it.

alekseysidorov: Migration performs in the separate thread, isn't it? So in theory we can return the corresponding join handle or something like oneshot in the deploy_artifact method.

PR #1802, opened 2020-03-02
Brush up code [ECR-4268]

slowli:
## Overview

This PR brushes up some code after #1796. Tokio 0.2 support is **not** implemented in this PR.

When reviewing, one should deselect commits from the `master` branch. 

---

See: https://jira.bf.local/browse/ECR-4268


runtimes/rust/src/api.rs
```diff
@@ -220,18 +220,17 @@ impl ServiceApiScope {
...
-                handler(state, query)
-                    .await
-                    .map_err(|err| err.source(descriptor.to_string()))
-            }
+            let state = match maybe_state {
+                Ok(state) => state,
+                Err(err) => return future::err(err).left_future(),
+            };
+
+            let descriptor = descriptor.clone();
```
alekseysidorov: 👍

PR #1804, opened 2020-03-04
Use `tokio` 0.2 [ECR-4269]

slowli:
## Overview

This PR introduces support of `tokio` 0.2 everywhere it's possible.

Discussion:

- Does it make sense to replace `actix_rt` with `tokio` as a runtime?
- Would it make sense to have `!Send`-future returned from `Node::run()`? I don't like that right now `actix` is spawned onto a detached thread, with no way to stop it, but spawning it the sane way seemingly unavoidably makes `Node::run()` future non-`Send`.

---

See: https://jira.bf.local/browse/ECR-4269


exonum-node/src/events/mod.rs
```diff
@@ -210,10 +220,10 @@ pub struct EventsAggregator<S1, S2, S3, S4> {
 
 impl<S1, S2, S3, S4> EventsAggregator<S1, S2, S3, S4>
```
alekseysidorov: I suppose that we can replace this struct easier by utilising a new async/await syntax.

exonum-node/src/events/network.rs
```diff
@@ -261,245 +229,207 @@ impl NetworkHandler {
...
+        let mut handshake_params = handshake_params.clone();
+        handshake_params.set_remote_key(key);
+        let pool = self.pool.clone();
+        let network_tx = self.network_tx.clone();
 
-        if let Some(unresolved_address) = unresolved_address {
-            let action = {
-                let unresolved_address = unresolved_address.clone();
-                move || tokio_dns::TcpStream::connect(unresolved_address.as_str())
+        // TODO: use retries.
```
alekseysidorov: Do you plan to add retries in this PR or later?

exonum-node/src/lib.rs
```diff
@@ -935,10 +932,10 @@ impl NodeChannel {
     /// Returns the channel for sending timeouts, networks and API requests.
     fn node_sender(&self) -> NodeSender {
         NodeSender {
-            internal_requests: self.internal_requests.0.clone().wait(),
-            network_requests: self.network_requests.0.clone().wait(),
-            transactions: self.transactions.0.clone().wait(),
-            api_requests: self.api_requests.0.clone().wait(),
+            internal_requests: SyncSender::new(self.internal_requests.0.clone()),
```
alekseysidorov: Do we need to continue to use sync sender version in the node sender?
Now we can just spawn future in the any place of code.

exonum-node/src/events/network.rs
slowli: In this PR.

exonum-node/src/lib.rs
slowli: I'll try to encapsulate spawning within `SyncSender` so as not to break the existing usage. I guess this will require conditional compilation for sandbox tests or some other hack to get spawning work outside of `tokio` context 🙃

exonum-node/src/events/mod.rs
slowli: How though? I think the reasons `EventsAggregator` is implemented like now is to provide differing priorities for constituent streams, which doesn't seem like an easy task to do using only "high-level" `async` / `await`.

components/system-api/Cargo.toml
```diff
@@ -27,7 +27,11 @@ serde = "1.0"
...
-actix-rt = "1.0.0"
 exonum-testkit = { version = "1.0.0-rc.1", path = "../../test-suite/testkit", features = ["exonum-node"] }
-pretty_assertions = "0.6.1"
+
 failure = "0.1.5"
+pretty_assertions = "0.6.1"
+
+[dev-dependencies.tokio]
+version = "0.2.13"
+features = ["blocking", "dns", "io-util", "macros", "rt-threaded", "tcp", "time"]
```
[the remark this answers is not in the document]
slowli: Well yeah, it should. However, if I understand the inner workings of `cargo` correctly, a single version of the `tokio` crate (with a single feature set) is used throughout the workspace. Thus, we cannot easily test within CI or locally that a certain set of features is enough for a particular use case.

exonum-node/src/events/retries.rs
```diff
@@ -0,0 +1,58 @@
...
+// limitations under the License.
+
+use futures::{TryFuture, TryFutureExt};
+use tokio::time::delay_for;
+
+use std::time::Duration;
+
+/// Retries the specified fallible future with the delay strategy specified by the `timeouts`
+/// iterator.
+pub async fn retry_future<Fut>(
```
alekseysidorov: Why not just use `futures-retry` crate?

https://crates.io/crates/futures-retry

exonum-node/src/events/network.rs
```diff
@@ -105,135 +133,166 @@ pub struct NetworkPart {
...
+    fn contains(&self, address: &PublicKey) -> bool {
+        self.peers.get(address).is_some()
+    }
+
+    /// Drops the connection to a peer. The request can be optionally filtered by the connection ID
+    /// in order to avoid issuing obsolete requests.
+    ///
+    /// # Return value
+    ///
+    /// Returns `true` if the connection with the peer was dropped.
```
aleksuss: I'd add a description when we return `false`. If the connection with address doesn't exist we return `false`.

PR #1805, opened 2020-03-05
Update error crates [ECR-4314]

slowli:
## Overview

This PR updates error libraries used by the crates in the workspace, from `failure` to `anyhow` + `thiserror`. Since we use the external derive crate for `ProtobufConvert` which requires `failure::Error` internally, the integration of `anyhow` is *slightly* hacky.

---

See: https://jira.bf.local/browse/ECR-4314


cli/Cargo.toml
```diff
@@ -20,7 +20,7 @@ exonum-supervisor = { version = "1.0.0-rc.1", path = "../services/supervisor" }
 exonum-system-api = { version = "1.0.0-rc.1", path = "../components/system-api" }
 exonum-rust-runtime = { version = "1.0.0-rc.1", path = "../runtimes/rust" }
 
-failure = "0.1"
+failure = { package = "anyhow", version = "1.0" }
```
alekseysidorov: Nice hack 🥇 

But are you sure that it doesn't lead to misunderstandings in the end users code?

components/merkledb/src/indexes/proof_list/proof.rs
```diff
@@ -14,9 +14,10 @@
 
 pub use crate::ValidationError; // TODO Change for a type alias after EJB switching to rust > 1.36
 
+use anyhow as failure; // FIXME: remove once `ProtobufConvert` derive is improved (ECR-4314)
```
aleksuss: Is it really correct issue number ? Or you've meant to improve in a scope of this PR ?

components/merkledb/src/indexes/proof_map/proof.rs
```diff
@@ -102,15 +102,15 @@ impl<'de> serde::Deserialize<'de> for ProofPath {
 }
 
 /// An error returned when a map proof is invalid.
-#[derive(Debug, Fail)]
+#[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum MapProofError {
     /// Non-terminal node for a map consisting of a single node.
-    #[fail(display = "non-terminal node as a single key in proof")]
+    #[error("non-terminal node as a single key in proof")]
```
aleksuss: I'm not sure really, but I'd add a `the` article before `proof`.

components/merkledb/src/migration.rs
```diff
@@ -583,14 +584,14 @@ impl MigrationHelper {
 }
 
 /// Errors emitted by `MigrationHelper` methods.
-#[derive(Debug, Fail)]
+#[derive(Debug, Error)]
 pub enum MigrationError {
     /// Failed to merge migration changes to database.
-    #[fail(display = "Failed to merge migration changes to database: {}", _0)]
-    Merge(#[fail(cause)] crate::Error),
+    #[error("Failed to merge migration changes to database: {}", _0)]
```
aleksuss: The same here. `the` before database ?

components/merkledb/src/indexes/proof_list/proof.rs
slowli: I've meant for the Jira issue to correspond to this PR and another PR dealing with this problem, but I can create a new issue just for this FIXME.

PR #1809, opened 2020-03-08
Rework transaction pool [ECR-4317]

slowli:
## Overview

This PR resolves the issue with the transaction pool. Previously, the transactions in the pool were only flushed on block commit, which lead to explorer's `GET transactions` endpoint effectively broken for in-pool transactions. This is fixed by introducing flushing transactions to the DB on a scheduled interval. The interval can be changed in the node config.

Independently, this PR fixes another issue. Previously, nodes forgot all transactions in the executed blocks; such transactions were removed from the ephemeral transaction cache, but were not flushed to the DB or anywhere else. I'm pretty sure this could lead to consensus hang-up and/or panics in the consensus code.

Notes:

- The new soak test is functional, but the stats output by it are not accurate. I plan to fix this separately once `tokio` 0.2 support is merged.

Discussion:

- Any suggestions on API / doc improvement and reasonable defaults for the flushing frequency are appreciated.

---

See: https://jira.bf.local/browse/ECR-4317


exonum/src/blockchain/mod.rs
```diff
@@ -382,42 +480,28 @@ impl BlockchainMut {
...
-                if !schema.transactions().contains(&tx_hash) {
-                    schema.add_transaction_into_pool(tx);
-                }
-            }
-        }
-
         let patch = self.dispatcher.commit_block_and_notify_runtimes(fork);
         self.merge(patch)?;
+
+        // TODO: this makes `commit` non-atomic; can this be avoided?
```
alekseysidorov: ECR?

test-suite/soak-tests/README.md
```diff
@@ -6,9 +6,12 @@ This is an internal crate for subjecting Exonum networks to [soak tests].
 
 The crate exposes the following binaries:
 
-- [`toggle`](src/toggle.rs). Tests repeatedly switching a service on and off.
+- [`toggle`](src/bin/toggle.rs). Tests repeatedly switching a service on and off.
   The service generates transactions in the `after_commit` hook.
 
+- [`send_txs`](src/bin/send_txs.rs). Tests sending transactions to a service
```
alekseysidorov: I wonder can we use this test to approximately estimate TPS and then print it in summary?

exonum-node/src/events_impl.rs
```diff
@@ -115,29 +120,25 @@ impl NodeHandler {
...
+    pub(crate) fn flush_txs_into_pool(&mut self) {
         let tx_cache_size = self.state().tx_cache_len();
-
         if tx_cache_size == 0 {
-            //No need to do anything.
-            trace!("Transaction cache is empty.");
             return;
+        } else {
+            info!(
+                "Flushing {} transactions from pool to storage",
```
aleksuss: Like in place below `... from cache to persistent pool` is more appropriate, I suppose.

exonum-node/src/lib.rs
```diff
@@ -313,13 +315,44 @@ impl Default for EventsPoolCapacity {
...
+/// ```
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[non_exhaustive]
 pub struct MemoryPoolConfig {
     /// Sets the maximum number of messages that can be buffered on the event loop's
     /// notification channel before a send will fail.
-    events_pool_capacity: EventsPoolCapacity,
+    pub events_pool_capacity: EventsPoolCapacity,
+
+    /// Interval between flushing the transaction pool to the persistent storage.
```
aleksuss: I'd add that the interval in milliseconds.

exonum-node/src/requests.rs
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use exonum::{
-    blockchain::{get_transaction, Schema},
+    blockchain::{PersistentCache, Schema, TransactionCache},
```
aleksuss: `PersistentCache` looks a bit weird. Maybe `PersistentPool` would be better ?

exonum-node/src/state.rs
```diff
@@ -361,7 +362,7 @@ impl BlockState {
     }
 
     /// Returns block's transactions.
```
aleksuss:
```suggestion
    /// Returns block's transaction hashes.
```
???

exonum/src/blockchain/mod.rs
```diff
@@ -52,6 +53,92 @@ mod schema;
...
+    }
+}
+
+/// Cache that does not contain any transactions.
+impl TransactionCache for () {
+    fn get_transaction(&self, _hash: Hash) -> Option<Verified<AnyTx>> {
+        None
+    }
+
+    fn contains_transaction(&self, _hash: Hash) -> bool {
```
aleksuss: This implementation seems redundant ? And if you'd add a default implementation for the `get_transaction` method like:
```rust
fn get_transaction(&self, hash: Hash) -> Option<Verified<AnyTx>> {
    None
}
```
You could write just
```rust
impl TransactionCache for () { }
```

test-suite/soak-tests/README.md
slowli: Well yeah, but this is blocked by the transaction sending being hacky (which I plan to fix separately after transition to tokio 0.2).

PR #1820, opened 2020-03-17
Customize propose creation logic [ECR-4347] [ECR-4348]

slowli:
## Overview

This PR introduces customized generation of block `Propose`s.

- A node may send a propose that indicates skipping block creation for this consensus epoch (terminology TBD). Other than a flag, this propose does not differ from ordinary proposes. The consensus algorithm remains the same. In other words, during each epoch, the consensus algorithm makes a decision to accept a new block, or to skip a block.
- If a skip propose is committed, it is processed differently than ordinary blocks. Namely, no transactions are getting processed and no `before_transactions` / `after_transactions` / `after_commit` hooks are getting called. The blockchain height is not increased.
- The processed propose is saved as a block + `Precommit`s and may be sent to other nodes as a response to a  `BlockRequest`. Unlike ordinary blocks, such skip blocks are replaced if accepted sequentially.
- If a node is lagging behind, it can skip any amount of epochs if it receives an authenticated block with the next height or a pseudo-block with a greater epoch.

Scoping:

- Removing nodes from the pool is outside the scope of this PR.

Discussion:

- **Terminology.** As of 3b4b664, terms are used chaotically (e.g., skip block vs block skip vs pseudo-block).
- **Breaking changes.** There are some breaking changes, some of which could be avoided (e.g., saving skip blocks could theoretically be moved to node crate from the main crate; however, I believe this would complicate things in the perspective). Some changes seem unavoidable, such as `height` -> `epoch` renaming in `Precommit`.
- **Testing.** Any suggestions as to the test cases are appreciated. 
- Is `#[doc(hidden)]` enough to guard against the misuse? Or should we use conditional compilation etc.?
- There are most probably some issues with the implementation. As an example, `BlockOrEpoch` requests are never removed (at least explicitly).
- Is using `Height` type for the epoch appropriate? Should we introduce `Epoch(_)` for this purpose?

---

See: https://jira.bf.local/browse/ECR-4347, https://jira.bf.local/browse/ECR-4348


exonum-node/src/consensus.rs
```diff
@@ -1035,19 +1090,29 @@ impl NodeHandler {
...
                         peer,
-                        self.state.height(),
+                        self.state.epoch(),
                         round,
                         propose_hash,
                         self.state.known_prevotes(round, propose_hash),
                     ))
                     .into(),
+
                 RequestData::Block(height) => {
```
alekseysidorov: What is the main difference between `Block` and `BlockOrEpoch` request?

exonum-node/src/lib.rs
```diff
@@ -511,6 +515,7 @@ impl NodeRole {
 
 impl NodeHandler {
     /// Creates `NodeHandler` using specified `Configuration`.
+    #[allow(clippy::too_many_arguments)]
```
alekseysidorov: Imo the time for `NodeHandlerBuilder` is come.

exonum-node/src/consensus.rs
slowli: `Block` requests a block at the specified blockchain height; if the node does not have such a block, it ignores the request. `BlockOrEpoch`, on the other hand, requests a block at the specified blockchain height *or* a block skip with the specified or greater epoch.

- If the receiver has a block at the specified height, it returns this block.
- Otherwise, if the receiver has a block skip with the specified or greater epoch, it returns this block skip.
- Otherwise, the receiver does not reply.

I'll add this description to `BlockRequest` docs + other code comments.

exonum-node/src/lib.rs
slowli: `NodeHandler` is not public though, so IMO the priority of introducing the builder is not that high 🙃.

exonum-node/src/messages/types.rs
```diff
@@ -159,62 +148,57 @@ pub struct Propose {
     /// The validator id.
     pub validator: ValidatorId,
     /// The height to which the message is related.
```
aleksuss: Shouldn't the description be changed?

```suggestion
    /// The epoch to which the message is related.
```

exonum-node/src/messages/types.rs
```diff
@@ -232,7 +217,7 @@ pub struct Prevote {
...
```
aleksuss:
```suggestion
    /// The epoch to which the message is related.
```
???

exonum-node/src/messages/types.rs
slowli: Yeah, good catch 👍

PR #1822, opened 2020-03-19
Rework exonum_cli::RunDev [ECR-4354]

vitvakatu:
## Overview

- `artifacts_dir` param renamed to `blockchain_path`
- removed cleaning of database files
- configuration files now stored in `config` subdirectory.

---

See: https://jira.bf.local/browse/ECR-4354

### Definition of Done

### DIFF
--- a/cli/src/command/run_dev.rs
+++ b/cli/src/command/run_dev.rs
@@ -15,12 +15,12 @@
 //! Standard Exonum CLI command used to run the node with default parameters
 //! for developing purposes.
 
-use anyhow::Error;
+use anyhow::{Context, Error};
 use exonum_supervisor::mode::Mode as SupervisorMode;
 use serde_derive::{Deserialize, Serialize};
 use structopt::StructOpt;
 
-use std::{net::SocketAddr, path::PathBuf, str::FromStr};
+use std::{fs, net::SocketAddr, path::PathBuf, str::FromStr};
 
 use crate::command::{
     finalize::Finalize,
@@ -39,7 +39,7 @@ pub struct RunDev {
     /// Database is located in <blockchain_path>/db directory, node configuration files
     /// are located in <blockchain_path>/config directory. Existing files and directories are
     /// reused. To generate new node configuration and start a new blockchain, the user must
-    /// manually delete existing <blockchain_path> directory or specify a new one.
+    /// use --clean flag or specify an another directory.
     #[structopt(long)]
     pub blockchain_path: PathBuf,
     /// Listen address for node public API.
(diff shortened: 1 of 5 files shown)


cli/src/command/run_dev.rs
```diff
@@ -50,21 +55,6 @@ pub struct RunDev {
 }
 
 impl RunDev {
-    fn artifact_path(&self, artifact_name: &str) -> PathBuf {
-        let mut path = self.artifacts_dir.clone();
-        path.push(artifact_name);
-        path
-    }
-
-    fn cleanup(&self) -> Result<(), Error> {
```
slowli: I think `cleanup` should be called if `RunDev` is called with the corresponding flag (e.g., `--clean`). IMO, auto-clearing may make in some cases.

vitvakatu: Seems reasonable!

vitvakatu:
Is it okay that integration tests are failing? @aleksuss

aleksuss:
Yes. It's ok. Fixing in progress.

cli/src/command/run_dev.rs
```diff
@@ -34,9 +34,14 @@ use crate::command::{
...
-    /// The path where configuration and db files will be generated.
-    #[structopt(long, short = "a")]
-    pub artifacts_dir: PathBuf,
+    /// Path to a directory for blockchain database and configuration files.
+    ///
+    /// Database is located in <blockchain_path>/db directory, node configuration files
+    /// are located in <blockchain_path>/config directory. Existing files and directories are
+    /// reused. To generate new node configuration and start a new blockchain, the user must
+    /// use --clean flag or specify an another directory.
+    #[structopt(long)]
```
aleksuss: What do you think about to add a short name for this param? For example `-p` or something like this ? I suppose it'd be convenient since `run-dev` command uses often in manual mode.

cli/src/command/run_dev.rs
```diff
@@ -47,20 +52,17 @@ pub struct RunDev {
     /// Private API is used by node administrators for node monitoring and control.
     #[structopt(long, default_value = "127.0.0.1:8081")]
     pub private_api_address: SocketAddr,
+    /// Clean existing blockchain database and configuration files before run.
+    #[structopt(long)]
```
aleksuss: Here as well.

vitvakatu: I think `--clean` is short enough

cli/src/command/run_dev.rs
vitvakatu: Probably makes sense

PR #1826, opened 2020-03-23
Graceful shutdown of the http server

aleksuss:
## Overview

One more workaround for UB related to RocksDB. 

---

### Definition of Done

### DIFF
--- a/components/api/src/manager.rs
+++ b/components/api/src/manager.rs
@@ -26,6 +26,8 @@ use std::{collections::HashMap, io, net::SocketAddr, time::Duration};
 
 use crate::{AllowOrigin, ApiAccess, ApiAggregator, ApiBuilder};
 
+const SHUTDOWN_TIMEOUT: u64 = 2; // timeout before shut down in seconds
+
 /// Configuration parameters for a single web server.
 #[derive(Debug, Clone)]
 #[non_exhaustive]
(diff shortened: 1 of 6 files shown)


components/api/src/manager.rs
```diff
@@ -26,6 +26,8 @@ use std::{collections::HashMap, io, net::SocketAddr, time::Duration};
 
 use crate::{AllowOrigin, ApiAccess, ApiAggregator, ApiBuilder};
 
+const SHUTDOWN_TIMEOUT: u64 = 2; // timeout before shut down in seconds
```
slowli: Maybe, make this value configurable via `ApiManagerConfig`?

slowli: Also, I'm not quite sure that this can fix RocksDB UB. Is it the case that we hope that `rocksdb::DB` won't be destroyed because `Arc`s pointing to it are in the worker threads that are not terminated before the process exits? That is, this workaround only works if there are active connections to the HTTP server at the shutdown time?

aleksuss: Yeah. To make `SHUTDOWN_TIMEOUT` configurable is a really good idea. About active connections. If we terminate a node via http post request `/api/system/v1/shutdown` we would have active connections, wouldn't it ? An open question is what if we terminate a node via sending a termination signal like SIGTERM. In such case we potentially could haven't any active connection. By the way I think we should add an implementation for signal handling.

PR #1828, opened 2020-03-24
Work around server restart on Windows [ECR-3950]

slowli:
## Overview

This PR works around server restart not working on Windows. The work around is relatively "dirty" in the sense that it may lead to crashes or UB once the underlying problem is fixed in `actix-web` or its deps (`mio`?). This is clearly formulated in the safety notice, though.

---
See: https://jira.bf.local/browse/ECR-3950


components/api/src/manager.rs
```diff
@@ -121,12 +128,39 @@ async fn with_retries<T>(
...
+    async fn stop(self) {
+        self.inner.stop(false).await;
+    }
+
+    #[cfg(windows)]
+    #[allow(unsafe_code)]
+    async fn stop(self) {
+        self.inner.stop(false).await;
+
+        // SAFETY: The safety of `from_raw_socket` relies on the fact that the socket is not used
```
alekseysidorov: You may refer this issue https://github.com/actix/actix-web/issues/1249 in this comment to provide better explanation what it is actually happening.

PR #1834, opened 2020-03-26
Signal handling by Exonum nodes [ECR-4374]

slowli:
## Overview

This PR makes nodes process OS signals (namely, cross-platform "ctrl-c" corresponding to Unix's SIGINT; and SIGTERM on Unix).

Discussion:

- [x] ~Should the docs mention / link to [the caveats](https://docs.rs/tokio/0.2.13/tokio/signal/fn.ctrl_c.html#caveats) associated with signal processing using `tokio`?~ Added a note about signal handling to `Node` docs.
- [x] ~Should other signals be handled (e.g., SIGHUP)?~ SIGHUP is ignored, SIGQUIT gracefully shuts down the node.

---

See: https://jira.bf.local/browse/ECR-4374

### DIFF
--- a/components/api/src/manager.rs
+++ b/components/api/src/manager.rs
@@ -62,6 +62,7 @@ impl WebServerConfig {
 
 /// Configuration parameters for `ApiManager`.
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct ApiManagerConfig {
     /// Active API runtimes.
     pub servers: HashMap<ApiAccess, WebServerConfig>,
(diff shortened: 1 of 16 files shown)


components/api/src/manager.rs
```diff
@@ -72,6 +74,51 @@ pub struct ApiManagerConfig {
...
+
+    /// Sets the retry policy for this config.
+    pub fn with_retries(mut self, timeout: u64, max_retries: u16) -> Self {
+        self.server_restart_retry_timeout = timeout;
+        self.server_restart_max_retries = max_retries;
+        self
+    }
+
+    /// Disables signal handling.
+    pub fn disable_signals(mut self) -> Self {
```
aleksuss: Does it make sense to add the prefix `with_` here ?

slowli: I used the same naming [as used in `actix_web`](https://docs.rs/actix-web/2.0.0/actix_web/struct.HttpServer.html#method.disable_signals).

PR #1839, opened 2020-03-30
Change transactions pool layout to improve perfomance

alekseysidorov:


exonum/src/blockchain/mod.rs
```diff
@@ -154,13 +154,10 @@ where
...
-            let tx = self
-                .transactions
-                .get(&tx_hash)
-                .expect("Transaction in pool is lost");
-            (tx_hash, Cow::Owned(tx))
-        });
+        let pool_it = self
+            .transactions_pool
+            .iter()
+            .map(move |(tx_hash, tx)| (tx_hash, Cow::Owned(tx)));
```
slowli: Nit: is `move` necessary here? Seemingly, nothing gets moved into the closure.

exonum/src/blockchain/schema.rs
```diff
@@ -171,9 +171,9 @@ impl<T: Access> Schema<T> {
         pool.get().unwrap_or(0)
     }
 
-    /// Returns a table that represents a set of uncommitted transactions hashes.
-    pub fn transactions_pool(&self) -> KeySetIndex<T::Base, Hash> {
-        self.access.get_key_set(TRANSACTIONS_POOL)
+    /// Returns a table that represents a set of uncommitted transactions.
```
slowli: Nit: Maybe, rewrite as "Returns a table with uncommitted transactions, where the hash of the transaction is mapped to its contents"?

PR #1841, opened 2020-03-31
Update protobuf-convert version to remove "failure" dependency

alekseysidorov:


aleksuss:
Shouldn't we notify about this change in the CHANGELOG ?

PR #1847, opened 2020-04-07
Node tech debt [ECR-4118] [ECR-1818] [ECR-2257] [ECR-1635] [ECR-3788]

slowli:
## Overview

This PR resolves several small-scale issues with the node code.

---

See: https://jira.bf.local/browse/ECR-4118, https://jira.bf.local/browse/ECR-1818, https://jira.bf.local/browse/ECR-2257, https://jira.bf.local/browse/ECR-1635, https://jira.bf.local/browse/ECR-3788


exonum-node/src/events/mod.rs
```diff
@@ -57,21 +67,39 @@ impl<T: Send + 'static> SyncSender<T> {
...
+                    log::warn!(
+                        "Cannot send a {}; processing has shut down. {}",
+                        message_kind,
+                        Self::ERROR_MSG_SUFFIX
+                    );
+                }
             });
-        } else {
-            block_on(self.0.send(message)).log_error();
+        } else if block_on(self.inner.send(message)).is_err() {
```
aleksuss, line 81: Does it make sense to unpack an error message from `block_on` invoke and to add to the log message ? Or it doesn't have a cognitive load ?

exonum-node/src/state.rs
```diff
@@ -1097,21 +1096,29 @@ impl State {
...
             ProposeState {
                 propose: msg,
                 unknown_txs: HashSet::new(),
                 block_hash: None,
-                // TODO: For the moment it's true because this code gets called immediately after
-                // saving a propose to the cache. Think about making this approach less error-prone.
-                // (ECR-1635)
                 is_saved: true,
-                // We expect ourself not to produce invalid proposes.
+                // We expect our node not to produce invalid proposes.
```
aleksuss, line 1121:
```suggestion
                // We expect that our node not to produce invalid proposes.
```

slowli: I think "expect [A] to [B]" does not require "that"; with "that", it would look like "expect that [A] [B]s".

PR #1849, opened 2020-04-09
Brush up exonum-node logs [ECR-4045][ECR-1915]

alekseysidorov:


exonum/src/messages/signed.rs
```diff
@@ -237,6 +238,15 @@ where
     }
 }
 
+impl<T: Debug> Debug for Verified<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Verified")
+            .field("raw", &self.raw.to_string())
```
slowli, line 244: Could it make sense to output the message author instead of a complete `raw` message?

components/crypto/src/lib.rs
```diff
@@ -578,6 +586,14 @@ mod tests {
         assert_eq!(s, sh);
     }
 
+    #[test]
+    fn to_from_string_public_key() {
+        let (p, _s) = gen_keypair();
```
slowli, line 591: Nit: `_s` could be replaced with `_`. Also, I don't quite like one-char variable names 🙃

exonum/src/messages/signed.rs
alekseysidorov: Yes this is more useful information for debugging.

PR #1854, opened 2020-04-16
Document merge workflow [ECR-2046]

slowli:
## Overview

This PR documents intended merge workflow for `Database` and possible logical errors if this workflow is not followed.

---

See: https://jira.bf.local/browse/ECR-2046

### DIFF
--- a/components/merkledb/src/db.rs
+++ b/components/merkledb/src/db.rs
@@ -470,6 +470,63 @@ enum NextIterValue {
 /// db.merge(fork.into_patch()).unwrap();
 /// ```
 ///
+/// # Merge Workflow
+///
+/// The user of a `Database` is responsible to ensure that forks are either created and merged
+/// sequentially or do not contain overlapping changes. By sequential creation we mean the following
+/// workflow:
+///
+/// ```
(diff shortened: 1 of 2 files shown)


components/merkledb/src/db.rs
```diff
@@ -470,6 +470,63 @@ enum NextIterValue {
...
+/// let first_fork = db.fork();
+/// first_fork.get_list("list").extend(vec![1, 2, 3]);
+/// let second_fork = db.fork();
+/// second_fork.get_list("list").push(4);
+/// db.merge(first_fork.into_patch()).unwrap();
+/// db.merge(second_fork.into_patch()).unwrap();
+///
+/// let snapshot = db.snapshot();
+/// let list = snapshot.get_list::<_, i32>("list");
+/// assert_eq!(list.len(), 1);
```
aleksuss, line 521:
```suggestion
/// assert_eq!(list.len(), 3);
```
Shouldn't be 3 here ???

slowli: Nope, that's the point of the code snippet: the concurrent merge workflow leads to logical bugs. That's why the top of the snippet contains a SHOUTING WARNING 🙃

PR #1855, opened 2020-04-17
More verbose REST API error responses [ECR-2696]

alekseysidorov:
Use `JsonBody` in the error responses.
Provide mechanism to override default error handlers.

### DIFF
--- a/components/api/src/backends/actix.rs
+++ b/components/api/src/backends/actix.rs
@@ -18,16 +18,25 @@
 //! for HTTP API, based on the [Actix](https://github.com/actix/actix) framework.
 
 pub use actix_cors::{Cors, CorsFactory};
-pub use actix_web::{http::Method as HttpMethod, web::Payload, HttpRequest, HttpResponse};
+pub use actix_web::{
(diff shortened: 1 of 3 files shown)


components/api/src/backends/actix.rs
```diff
@@ -18,16 +18,25 @@
...
+    web::Payload,
+    HttpRequest, HttpResponse,
+};
 
 use actix_web::{
     body::Body,
+    dev::ServiceResponse,
     error::ResponseError,
     http::header,
+    middleware::errhandlers::{ErrorHandlerResponse, ErrorHandlers},
```
aleksuss, line 32: `errhandlers` 🙃

components/api/src/backends/actix.rs
```diff
@@ -268,6 +285,52 @@ impl From<AllowOrigin> for CorsFactory {
     }
 }
 
+trait ErrorHandlersEx {
```
slowli, line 288: Nit: This trait seems slightly redundant; OTOH, it does simplify the code.

PR #1869, opened 2020-05-14
Bump rust version up to 1.43.0

aleksuss:
## Overview

---

### Definition of Done


components/crypto/src/lib.rs
```diff
@@ -550,7 +550,11 @@ fn verify_keys_match(public_key: &PublicKey, secret_key: &SecretKey) -> bool {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
```
slowli, line 553: The relevant [clippy docs](https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_imports) state:

> `use super::*` is allowed in test modules. This is defined as any module with “test” in the name.

Is this not the case? IMO, `use super::*` for test modules does make sense.

components/keys/examples/key_generator.rs
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use exonum_keys::*;
```
slowli, line 15: Why is wildcard import not affected? 🙃 Intuitively, it looks more "suspicious" than ones in the test modules, for example.

exonum-node/src/events/internal.rs
```diff
@@ -93,9 +93,12 @@ mod tests {
         helpers::Height,
         messages::Verified,
     };
+    use futures::{SinkExt, StreamExt};
     use pretty_assertions::assert_eq;
 
-    use super::*;
+    use super::{
+        mpsc, BinaryValue, InternalEvent, InternalPart, InternalRequest, Message, SignedMessage,
```
slowli, line 100: Nit: Why are `mpsc` / `BinaryValue` re-imported from `super`, whereas `SinkExt` / `StreamExt` are imported directly from their definition location (`futures`)?

components/crypto/src/lib.rs
aleksuss: I followed by clippy suggestions only 🙃

components/keys/examples/key_generator.rs
aleksuss: Weird. But clippy skipped it.

PR #1870, opened 2020-05-15
Add possibility to use domain names along with IP addresses [ECR-4400]

aleksuss:
This PR add a possibility to use domain names of nodes on config generation stage and in the connect list as well.

## Overview

---

See: https://jira.bf.local/browse/ECR-4400

### Definition of Done


cli/src/command/generate_config.rs
```diff
@@ -104,30 +102,40 @@ impl GenerateConfig {
...
+        match hostname.to_socket_addrs() {
+            Ok(mut addrs) => addrs.next().ok_or_else(|| {
+                Error::msg(format!(
+                    "No one IP address is related to the domain: {}",
+                    hostname
+                ))
+            }),
+            Err(e) => {
+                if e.kind() == ErrorKind::InvalidInput {
+                    Self::resolve_peer_address(&format!(
```
slowli, line 115: This looks like a potential for an infinite recursion. I'd suggest to limit the recursion, e.g., by adding a Boolean function arg (have we appended a default port to the address?).

cli/src/command/generate_config.rs
```diff
@@ -104,30 +102,40 @@ impl GenerateConfig {
...
-            };
-            SocketAddr::new(ip_address, external_address.port())
-        }
+    /// Returns `provided` address or [`INADDR_ANY`](https://en.wikipedia.org/wiki/0.0.0.0) address
+    /// combined with the port number obtained from `peer_address`.
+    fn get_listen_address(provided: Option<SocketAddr>, peer_address: &str) -> SocketAddr {
+        provided.unwrap_or_else(|| match Self::resolve_peer_address(peer_address) {
+            Ok(address) => {
+                let ip_address = match address.ip() {
+                    IpAddr::V4(_) => "0.0.0.0".parse().unwrap(),
```
slowli, line 132: There is [`Ipv4Addr::UNSPECIFIED`](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#associatedconstant.UNSPECIFIED) for this purpose.

cli/src/command/generate_config.rs
```diff
@@ -104,30 +102,40 @@ impl GenerateConfig {
...
+                    IpAddr::V6(_) => "::".parse().unwrap(),
```
slowli, line 133: There is [`Ipv6Addr::UNSPECIFIED`](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#associatedconstant.UNSPECIFIED) for this purpose.

cli/src/command/generate_config.rs
```diff
@@ -104,30 +102,40 @@ impl GenerateConfig {
...
-            let ip_address = match external_address.ip() {
-                IpAddr::V4(_) => "0.0.0.0".parse().unwrap(),
-                IpAddr::V6(_) => "::".parse().unwrap(),
-            };
-            SocketAddr::new(ip_address, external_address.port())
-        }
+    /// Returns `provided` address or [`INADDR_ANY`](https://en.wikipedia.org/wiki/0.0.0.0) address
+    /// combined with the port number obtained from `peer_address`.
+    fn get_listen_address(provided: Option<SocketAddr>, peer_address: &str) -> SocketAddr {
+        provided.unwrap_or_else(|| match Self::resolve_peer_address(peer_address) {
```
slowli, line 129: Maybe, it could make more sense to just attempt to get *the port* from `peer_address` (which seems pretty easy - take the last ocurrence of `:`, parse the substring to the right as `u16`) instead of trying to resolve it? I'm not sure resolution will work in all cases, and the resolved IP address seems unused anyway.

CHANGELOG.md
```diff
@@ -23,6 +23,13 @@ The project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 - Backtrace is now included into `ExecutionError`s. (#1850)
 
+### New Features
+
+#### exonum-cli
+
+- Added a possibility to use domain names along with IP addresses on generation configs
+  stage and in the connect list as well. (#1870)
```
slowli, line 31: Nit / grammar: "as well" looks unnecessary.

aleksuss:
@slowli Thanks for review 👍🏻

cli/src/command/generate_config.rs
aleksuss: Yes. I've tried to do so. But:

> take the last occurrence of :

There can be a problem with IPv6 addresses with such approach. 

> the resolved IP address seems unused anyway.

But we can be sure we provide a correct domain name.

cli/src/command/generate_config.rs
aleksuss: Yeah. The reason I didn't do this is because it can occur only on config generation stage. And to be honest I couldn't think up a case when it happened.

cli/src/command/generate_config.rs
```diff
@@ -104,30 +102,50 @@ impl GenerateConfig {
...
     /// If no port is provided by user, uses `DEFAULT_EXONUM_LISTEN_PORT`.
-    fn parse_external_address(input: &str) -> Result<SocketAddr, Error> {
-        if let Ok(address) = input.parse() {
-            Ok(address)
-        } else {
-            let ip_address = input.parse()?;
-            Ok(SocketAddr::new(ip_address, DEFAULT_EXONUM_LISTEN_PORT))
+    fn resolve_peer_address(
+        hostname: &str,
+        is_host_with_port: &mut bool,
```
slowli, line 107: Nit: Why not pass this var by value, rather than by a mutable reference? A mutable reference could make sense if the var is read after the call, but this is seemingly not the case.

cli/src/command/generate_config.rs
```diff
@@ -104,30 +102,50 @@ impl GenerateConfig {
...
+    ) -> Result<SocketAddr, Error> {
+        match hostname.to_socket_addrs() {
+            Ok(mut addrs) => addrs.next().ok_or_else(|| {
+                Error::msg(format!(
+                    "No one IP address is related to the domain: {}",
+                    hostname
+                ))
+            }),
+            Err(e) => {
```
slowli, line 116: Nit: IMO, it could make sense to bring the condition here:

```rust
Err(e) if e.kind() == ErrorKind::InvalidInput && !*is_host_with_port => {
    // ...
}
Err(e) => Err(Error::from(e)),
```

AFAIK, the borrow checker no longer emits "by-ref & by-move" error, which would've prevented from doing this with old Rust versions.

cli/src/command/generate_config.rs
aleksuss: Influence of C 😄

aleksuss:
@slowli How can we define which listen address should be `Ipv4Addr :: UNSPECIFIED` or` Ipv6Addr :: UNSPECIFIED` if a user provides domain name and does not provide it explicitly ?

PR #1872, opened 2020-06-04
Re-work TemporaryDB for preventing UB in tests [ECR-4406]

aleksuss:
## Overview

---

See: https://jira.bf.local/browse/ECR-4406

### Definition of Done

### DIFF
--- a/components/merkledb/src/backends/rocksdb.rs
+++ b/components/merkledb/src/backends/rocksdb.rs
@@ -20,7 +20,8 @@ use crossbeam::sync::{ShardedLock, ShardedLockReadGuard};
 use rocksdb::{
     self, checkpoint::Checkpoint, ColumnFamily, DBIterator, Options as RocksDbOptions, WriteBatch,
 };
-use std::{fmt, iter::Peekable, mem, path::Path, ptr, sync::Arc};
+use smallvec::SmallVec;
+use std::{fmt, iter::Peekable, mem, path::Path, sync::Arc};
 
 use crate::{
     db::{check_database, Change},
@@ -154,7 +155,6 @@ impl RocksDB {
         Ok(())
     }
 
-    #[allow(unsafe_code)]
     fn do_merge(&self, patch: Patch, w_opts: &RocksDBWriteOptions) -> crate::Result<()> {
         let mut batch = WriteBatch::default();
         for (resolved, changes) in patch.into_changes() {
@@ -175,19 +175,15 @@ impl RocksDB {
 
                 // We assume that typical key sizes are less than `1_024 - ID_SIZE = 1_016` bytes,
                 // so that they fit into stack.
-                let mut buffer: [u8; 1_024] = [0; 1024];
-                unsafe {
-                    ptr::copy(id_bytes.as_ptr(), buffer.as_mut_ptr(), ID_SIZE);
-                }
+                let mut buffer: SmallVec<[u8; 1_024]> = SmallVec::new();
+                buffer.extend_from_slice(&id_bytes);
 
                 for (key, change) in changes.into_data() {
-                    let buffer_size = key.len() + ID_SIZE;
-                    unsafe {
-                        ptr::copy(key.as_ptr(), buffer.as_mut_ptr().add(ID_SIZE), key.len());
-                    }
+                    buffer.truncate(ID_SIZE);
+                    buffer.extend_from_slice(&key);
                     match change {
-                        Change::Put(ref value) => batch.put_cf(cf, &buffer[..buffer_size], value),
-                        Change::Delete => batch.delete_cf(cf, &buffer[..buffer_size]),
+                        Change::Put(ref value) => batch.put_cf(cf, &buffer, value),
+                        Change::Delete => batch.delete_cf(cf, &buffer),
                     }
                 }
             } else {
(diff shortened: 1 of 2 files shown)


components/merkledb/src/backends/temporarydb.rs
```diff
@@ -91,22 +84,137 @@ impl Database for TemporaryDB {
...
     }
 
     fn iter(&self, name: &ResolvedAddress, from: &[u8]) -> Iter<'_> {
-        self.snapshot.iter(name, from)
+        let collection = self
+            .snapshot
+            .get(&name.name)
+            .or_else(|| self.snapshot.get(DEFAULT_COLLECTION))
+            .unwrap();
+        let from = name.keyed(from).to_vec();
```
slowli, line 210: [`range` is parametric](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) by the key parameter; AFAIK, it's precisely to avoid unnecessary heap allocations, such as this one.

components/merkledb/src/backends/temporarydb.rs
```diff
@@ -14,74 +14,67 @@
...
+    backends::rocksdb::{next_id_bytes, ID_SIZE},
+    db::{check_database, Change, Iterator as DbIterator},
+    Database, Iter, Patch, ResolvedAddress, Result, Snapshot,
 };
 
-/// A wrapper over the `RocksDB` backend which stores data in the temporary directory
-/// using the `tempfile` crate.
-///
-/// This database is only used for testing and experimenting; is not designed to
+type MemoryDB = HashMap<String, BTreeMap<Vec<u8>, Vec<u8>>>;
```
slowli, line 31: Did you try this variant?

```rust
type MemoryDB = HashMap<ResolvedAddress, BTreeMap<Vec<u8>, Vec<u8>>>;
```

Intuitively, it could be more performant since we don't need iteration over multiple indexes in the same family.

components/merkledb/src/backends/temporarydb.rs
```diff
@@ -14,74 +14,67 @@
...
-    dir: Arc<TempDir>,
+    inner: Arc<ShardedLock<MemoryDB>>,
 }
 
-/// A wrapper over the `RocksDB` snapshot with the `TempDir` handle to prevent
-/// it from destroying until all the snapshots and database itself are dropped.
 struct TemporarySnapshot {
-    snapshot: RocksDBSnapshot,
-    _dir: Arc<TempDir>,
+    snapshot: MemoryDB,
```
slowli, line 43: Necessity to clone the entire database for each snapshot looks massively inefficient, and, AFAIK, we create a lot of snapshots during execution. The sketch for a corresponding solution would be as follows:

- Keep an auto-incrementing counter in the DB reflecting the number of merged patches.
- Deduplicate snapshots if the counter value for them is the same (e.g., by caching snapshots in a map `merge_counter -> Weak<MemoryDB>`). That is, a snapshot would wrap around `Arc<MemoryDB>`, not `MemoryDB` itself.

components/merkledb/benches/lib.rs
```diff
@@ -14,15 +14,12 @@
 
 use criterion::{criterion_group, criterion_main};
 
-use crate::{
+use crate::benchmarks::{
     encoding::bench_encoding, schema_patterns::bench_schema_patterns, storage::bench_storage,
     transactions::bench_transactions,
 };
 
-mod encoding;
```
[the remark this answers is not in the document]
aleksuss: I know it sounds odd, but I couldn't use `BenchDB` declared in `benches/lib.rs` in the modules. Possibly, it's because of some specific of files structure in benches.

components/merkledb/src/backends/rocksdb.rs
```diff
@@ -237,15 +175,19 @@ impl RocksDB {
 
                 // We assume that typical key sizes are less than `1_024 - ID_SIZE = 1_016` bytes,
                 // so that they fit into stack.
-                let mut buffer: SmallVec<[u8; 1_024]> = SmallVec::new();
-                buffer.extend_from_slice(&id_bytes);
+                let mut buffer: [u8; 1_024] = [0; 1024];
+                unsafe {
```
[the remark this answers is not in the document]
aleksuss: In some benchmarks it gives performance boost till 5%. I suppose  you're right. The unsafe and key limitation don't worth this performance boost.

components/merkledb/src/backends/temporarydb.rs
aleksuss: ... by the key parameter of the collection. A key of the collection has a type `Vec<u8>` and therefore we should use `Vec<u8>` in a `range`, shouldn't we ?

components/merkledb/src/backends/temporarydb.rs
aleksuss: Yeah. It works good. Thanks for advice 👍🏻

components/merkledb/src/backends/temporarydb.rs
```diff
@@ -91,22 +84,137 @@ impl Database for TemporaryDB {
...
+            if !inner.contains_key(&resolved.name) {
+                inner.insert(resolved.name.clone(), BTreeMap::new());
+            }
+
+            let collection: &mut BTreeMap<Vec<u8>, Vec<u8>> =
+                inner.get_mut(&resolved.name).unwrap();
+
+            if changes.is_cleared() {
+                if let Some(id_bytes) = resolved.id_to_bytes() {
+                    let next_bytes = next_id_bytes(id_bytes).to_vec();
```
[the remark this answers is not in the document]
aleksuss: Yeah. Makes sense.

PR #1958, opened 2021-03-18
Replace deprecated collections::Bound with ops::Bound

bstrie:


aleksuss:
Thanks for the PR. But as I can see `std::collections::Bound` is just [re-export](https://doc.rust-lang.org/src/std/collections/mod.rs.html#406) of `core::ops::Bound`. Does it really make sense to change it ?

aleksuss:
Oh. I see. `core::ops::Bound` is now deprecated. 🙃 Thanks 👍🏻

aleksuss (approved):

PR #2004, opened 2021-09-02
Bump RocksDB to 0.17.

tkintscher:
## Overview

Bring the RocksDb version to 0.17, and fix deprecation warnings with latest Rust 1.54.


aleksuss:
RocksDB was bumped in #2007

